PR #17, opened 2022-11-27
MIT license

edemaine:
Switch from 💖 license to explicit MIT license, to encourage open-source contributions.


STRd6:
End of an era.

PR #19, opened 2022-12-03
More consistent use of paths vs. file URIs

edemaine:
This fixes some lsp bugs on Windows, but sadly not all of them.

(It's still the case that updates to files don't trigger updates to the view. `getScriptSnapshot` seems to be returning the correct data (updated transpiled file), but as far as I can tell it's getting called *after* the TypeScript service returns results, which seems... wrong?)

Before merging, please test on other operating systems that this doesn't break anything.

### DIFF
--- a/lsp/source/lib/typescript-service.mts
+++ b/lsp/source/lib/typescript-service.mts
@@ -21,7 +21,7 @@ const {
 } = ts
 
 import { TextDocument } from "vscode-languageserver-textdocument"
-import { fileURLToPath } from "url"
+import { fileURLToPath, pathToFileURL } from "url"
 import { createRequire } from "module"
 // TODO: project Coffee version?
 import { compile as coffeeCompile } from "coffeescript"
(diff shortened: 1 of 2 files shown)


STRd6:
I'll try and summarize how the snapshots work generally from what I remember.

`addOrUpdateDocument` is called every time a document changes. Inside that function it checks for a transpiler for the extension (a transpiler being something that will translate whatever the source file is like .civet, .hera, .coffee to .ts/.js). If a transpiler is found it initializes a virtual transpiledDoc at the infered transpilation path by appending the transpilation target suffix to the path (ie `somefile.civet.ts`). It then clears any snapshot for the transpilation path so when a TSServer command is run the file will be recompiled and resnapshotted on demand.

`getOrCreatePathSnapshot` is called from TSServer through `getScriptSnapshot` (kept up to date by incrementing `projectVersion` and clearing any cached snapshot. This way TSServer only re-transpiles when the source document has changed and we're asking TSServer for data (by calling `service.getSyntacticDiagnostics`, `getNavigationTree`, `getDefinitionAtPosition`, etc.)

I'll try and review this PR further soon, just wanted to brain dump while I was reviewing the LSP source.

edemaine:
That all matches my understanding. From console logging, the virtual documents all look OK (they have the updated content), but TypeScript seems to be loading them / asking for them (asynchronously?) after we ask for information from the TypeScript server, and returning nothing meanwhile.

Does the TS server have an async mode or something? Or maybe it's just not understanding the document we give it in update, only in add; I guess I should compare the inputs in both cases.

PR #21, opened 2022-12-07
Support for unbraced `export x, y`

edemaine:
This PR adds support for `export x, y` shorthand for `export {x, y}`.  Also supports `export x as foo, y as bar`.

It's meant as a step toward `export x = 5` from CoffeeScript, but I haven't gotten there yet.  It seems useful by itself.  It feels like a natural extension to CoffeeScript's implicit object literals.

Also not yet supported:
* `export x, y from 'foo'`.  This seems like it should be supported by the grammar as I wrote it, but it won't parse. Any ideas?
* Multi-line implicit objects: (I couldn't quite figure out how this works with `ImplicitObjectLiteral`)
  ```
  export x,
    y
    z
  ```
  
I'm also a little uncomfortable with why I needed to re-order `Declaration` and `NamedExports`. I thought both branches would try to parse...?

### DIFF
--- a/README.md
+++ b/README.md
@@ -198,7 +198,8 @@ Things Added that CoffeeScript didn't
 - Convenience for ES6+ Features
   - Const assignment shorthand `a := b` -> `const a = b`; `{a, b} := c` -> `const {a, b} = c`
   - `@#id` -> `this.#id` shorthand for private identifiers
-  - `import` shorthand `x from ./x` -> `import x from "./x"`
+  - `import` shorthand: `x from ./x` -> `import x from "./x"`
+  - `export` shorthand: `export x, y` -> `export {x, y}`
   - Triple backtick Template Strings remove leading indentation for clarity
   - Class constructor shorthand `@( ... )`
   - ClassStaticBlock `@ { ... }`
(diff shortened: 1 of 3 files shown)


STRd6:
@edemaine PEG `/` is ordered choice so the order of choices affects the parse. Adding a note about the ordering is good and correct.

I like this feature. The first time I learned about ES module exports I was confused why things like `export x` wouldn't work.

Thanks for the PR!

edemaine:
Ah, that makes sense. Do you understand why `export x from 'foo'` doesn't work?

STRd6:
It was because `from` isn't in the reserved words list since it is only reserved in some contexts. Here's what I ended up adding: https://github.com/DanielXMoore/Civet/commit/1a1b8103e1a690ceb66e1180f4c00b265d661d63

For debugging the parser you can add the `Debugger` rule into the production you want to inspect and run the tests with `yarn test --inspect-brk` and step through with Chrome dev tools. It's a little wild with all the generated code in the parser but it can help once you get your bearings and get comfortable stepping up and down through the production rules. I'd really like to work on some improved parser debugging tools at some point as well.

edemaine:
Oh, right — nice! Thanks for fixing this and for the tips!

Issue #20, opened 2022-12-07
Indentation based jsx syntax


lorefnon:
Hello, thanks a lot for creating Civet - This looks great. I spent some time playing around with it yesterday and loved the DX.

Since this language is quite early, I was wondering if we could eliminate the need for closing xml tags in jsx, and adopt an indentation based syntax similar to [imba](https://imba.io/). I think this will make it more coherent with the rest of the language as well. 

Imba is tied to its own dom/rendering impl. where as I am trying to use civet with solidjs which has its own jsx preprocessor.

edemaine:
I agree 100%. We had some discussions about this in the Discord, and I was planning to take a stab at implementing it (as Daniel doesn't use JSX himself).

I think the right way to do this is to support both syntaxes, via a parser flag in the first line (the existing `"civet ..."` directive). That way we can still offer CoffeeScript compatibility. For example:

* `"civet coffeeCompat"` - closing tags, not indentation based
* `"civet coffeeCompat indentJSX"` - CoffeeScript but with indentation JSX
* `"civet closeJSX"` - Civet as it is now, requiring closing tags
* default - indentation JSX

@STRd6 I haven't looked at how JSX is handled, but I assume there's significant support in the lexer. Will it be feasible to support both modes?

STRd6:
It should be possible to support both, and it may even be possible to support optional closing tags by default. With the caching implemented in the parser there is much less cost to distant lookaheads. Since optional closing tags is a superset of explicit closing tags then it could make a very compatible default that people could opt into by usage. Perhaps the only necessary directive would be `"civet closeJSX"` to opt into mandatory explicit JSX closing.

@edemaine take a look at the `NestedBlockStatements` rule for an example of how the parser currently handles nesting. You can add a production to `JSXElement` that uses nesting and inserts a synthetic closing tag (similar to how it checks for mismatched tags). Let me know if you have any other questions, thanks!

lorefnon:
So the indentation based syntax in imba has a few more nuances other than optional closing tag. I am not suggesting we adopt them all, but one thing of interest is that because strings need to be explicitly quoted, we can avoid the need to wrap expressions in braces as in JSX.

This enables us to write code like: 

```imba
<div.tiles> for tile,i in game.tiles
	<Tile data=game nr=i @click=game.place(i)> tile
```

which is more succinct compared to: 


```imba
<div.tiles>{for tile,i in game.tiles
[...]

PR #25, opened 2022-12-09
Indentation-based JSX

edemaine:
This is an attempt at implementing basic auto-closing JSX tags via indentation, while co-existing with explicitly closed tags, as discussed in #20.

Here's an example:

```coffee
return
  <div>
    <b>
      Bold
    <i>
      Italic
console.log 'hi'
```

But it currently compiles to:

```js
return
  <div>
    <b>
      Bold
</b>
    <i>
      Italic
</i>
</div>
console.log('hi')
```

The issue is that the added closing tags aren't indented properly. @STRd6 Could you explain how I'm using `IndentIndent` incorrectly? I haven't been able to figure it out, despite matching the `Nested` (as far as I can tell). Is the function associated with the `JSXElement` rule just called too late, and I need to add an `InsertClosingTag` rule? If so, how would I give it the opening tag?

### DIFF
--- a/source/parser.hera
+++ b/source/parser.hera
@@ -3465,12 +3465,14 @@ Yield
 # https://facebook.github.io/jsx/#prod-JSXElement
 JSXElement
   JSXSelfClosingElement
-  JSXOpeningElement JSXChildren* __ JSXClosingElement ->
+  JSXOpeningElement JSXChildren? __ JSXClosingElement ->
     // Check that tags match
     if ($1[1] !== $4[2]) {
       throw new Error(`mismatched closing tags at ${JSON.stringify($loc)}`)
     }
     return $0
+  JSXOpeningElement:open JSXNestedChildren InsertNewline InsertIndent ->
+    return [...$0, ["</", open[1], ">"]]
 
 # https://facebook.github.io/jsx/#prod-JSXSelfClosingElement
 JSXSelfClosingElement
@@ -3536,6 +3538,9 @@ JSXAttributeValue
 JSXChildren
   ( __ JSXChild )*
 
+JSXNestedChildren
+  PushIndent ( Nested JSXChild )+ PopIndent
+
 # https://facebook.github.io/jsx/#prod-JSXChild
 JSXChild
   JSXText
@@ -3546,7 +3551,9 @@ JSXChild
 # https://facebook.github.io/jsx/#prod-JSXText
 JSXText
   # NOTE: not currently excluding https://facebook.github.io/jsx/#prod-HTMLCharacterReference
-  [^{}<>]+
+  # NOTE: Additionally forbidding leading whitespace so it can be used for
+  # indentation (by JSXNestedChildren) or ignored by JSXChildren.
+  [^{}<>\s][^{}<>\n\r]*
 
 # https://facebook.github.io/jsx/#prod-JSXChildExpression
 JSXChildExpression
--- a/source/parser.hera
+++ b/source/parser.hera
@@ -3940,8 +3940,7 @@ Reset
       })
     }
 
-    module.verbose = false
-    module.config = {
+    module.config = parse.config = {
       autoVar: false,
       coffeeBinaryExistential: false,
       coffeeBooleans: false,
@@ -3956,6 +3955,7 @@ Reset
       coffeeNot: false,
       coffeeOf: false,
       implicitReturns: true,
+      verbose: false,
     }
 
     let indexOfRef, hasPropRef, spliceRef
@@ -5084,7 +5084,7 @@ TrackIndented
     if (level <= module.currentIndent.level) {
       return $skip
     }
-    if (module.verbose) {
+    if (module.config.verbose) {
       console.log("pushing indent", indent)
     }
 
@@ -5117,7 +5117,7 @@ PushIndent
 
 PopIndent
   "" ->
-    if (module.verbose) {
+    if (module.config.verbose) {
       console.log("popping indent", module.indentLevels[module.indentLevels.length-1], "->", module.indentLevels[module.indentLevels.length-2])
     }
     module.indentLevels./**/pop()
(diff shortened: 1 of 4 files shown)


source/parser.hera
```diff
@@ -3536,6 +3538,9 @@ JSXAttributeValue
 JSXChildren
   ( __ JSXChild )*
 
+JSXNestedChildren
+  PushIndent ( Nested JSXChild )+ PopIndent
```
STRd6, line 3542: One quirk of tracking indentation state is that `PopIndent` must always be called if `PushIndent` succeeds. This means that you need to use `*` and `return $skip` in the handler if the length is zero to make sure the indentation state stack resets.

source/parser.hera
```diff
@@ -3465,12 +3465,14 @@ Yield
...
 JSXElement
   JSXSelfClosingElement
-  JSXOpeningElement JSXChildren* __ JSXClosingElement ->
+  JSXOpeningElement JSXChildren? __ JSXClosingElement ->
     // Check that tags match
     if ($1[1] !== $4[2]) {
       throw new Error(`mismatched closing tags at ${JSON.stringify($loc)}`)
     }
     return $0
+  JSXOpeningElement:open JSXNestedChildren InsertNewline InsertIndent ->
```
STRd6, line 3474: I would probably move this rule above the non-indented rule. It shouldn't really matter but with caching and the parser state tracking sometimes it does.

STRd6:
Overall it looks like you're on the right track, let me know if those suggestions make a difference.

STRd6:
As for `InsertIndent` it looks like that should work the way you are using it here. It looks the same as `ImplicitNestedBlock` which works.

source/parser.hera
edemaine: Thanks; I wondered why that code was there.

On another point, I found that replacing `( Nested JSXChild )` with `JSXNestedChild` and a separate rule:
```hera
NestedJSXChild
  Nested JSXChild
```
broke the parsing.  Any ideas why this would occur?

source/parser.hera
edemaine: Oddly, this fixed the `InsertIndent` problem. Now I get:
```js
return
  <div>
    <b>
      Bold
    </b>
    <i>
      Italic
    </i>
</div>
console.log('hi')
```
which seems about right.

I had in mind that it's good to check for an explicit closing tag first, but I'm not sure what's best from a performance perspective. What's weird is that it affects correctness of `InsertIndent`...

edemaine: This change also breaks regular JSX, specifically the "nested tags" example.
```jsx
    <p>
      <a href="https://www.reactjs.org"> link </a>
      <img src={user.avatarUrl}></img>
    </p>
```
I guess indentation-based and explicitly closed tags aren't so easy to make co-exist...

source/parser.hera
STRd6: I think it might have to do with caching in the parser. In `source/main` the exit handler there is a big list of rules that shouldn't be cached. Since `JSXChild` includes `__` if it is cached it might override indentation after backtracking. Let me know if that is making sense.

source/parser.hera
```diff
@@ -1693,7 +1693,10 @@ BinaryOpSymbol
   "<="
   ">="
   "<<"
-  "<"
+  # NOTE: Avoid matching JSX opening tag by requiring non-identifier character
+  # (e.g. whitespace) after "<", or no whitespace before "<" (for e.g. "1<2").
+  /<(?!\p{ID_Start}|[_$])|(?<=\S)</ ->
```
STRd6, line 1698: Lookbehinds don't work in Safari so we may need to find another way.

test/jsx/indent.civet
```diff
@@ -1,6 +1,18 @@
...
 describe "Indentation-based JSX", ->
+  testCase """
+    no children
+    ---
+    return
+      <div>
+    console.log 'unreachable'
+    ---
+    return
+      <div />
```
STRd6, line 12: We should wrap nested return expressions in parens because JS/TS treats `return\n` as a statement end.

Don't let this block the feature though, it's likely a more general problem in the parser currently.

edemaine: Ah, good point. I noticed CoffeeScript has an explicit "RETURN INDENT ... DEDENT" rule. Maybe we want to add one to Civet. (This would also improve the indentation of the added closing tags.) Then we could easily add parens in that rule.

source/parser.hera
edemaine: Are we OK forbidding `1<2` and `x<y`?

Alternatively, inside the `->` can we get the text right before the match, and return `$skip` if appropriate?

I was thinking about this but wasn't sure we needed the Civet compiler to run in Safari. Might be nice though.

STRd6: We can require mandatory whitespace between binary operators.

The compiler should be able to run in Safari (Civet Playground etc.)

test/jsx/indent.civet
STRd6: I sort of already have this in `MaybeNestedExpression` (probably needs a better name). Essentially if there is a `&EOS ExtendedExpression` we should make sure to wrap in parens.

edemaine:
This could no doubt use more testing, but it's working reasonably well now, so I think it might be ready to merge and play with.

source/parser.hera
```diff
@@ -3565,7 +3590,10 @@ JSXChild
 # https://facebook.github.io/jsx/#prod-JSXText
 JSXText
   # NOTE: not currently excluding https://facebook.github.io/jsx/#prod-HTMLCharacterReference
-  [^{}<>]+
+  # NOTE: Additionally forbidding leading whitespace, so it can be used for
+  # indentation (by JSXNestedChildren) or ignored by JSXChildren,
+  # and newlines which we leave for the next indentation.
+  [^{}<>\s][^{}<>\n\r]*
```
STRd6, line 3596: This might not be necessary since the `Nested JSXChild` rule should consume and hold the space before getting to here.

edemaine: I had a ton of extremely difficult to debug bugs before adding this; feel free to try, but I'm pretty sure it's necessary. The issue is that `<foo>` can also be parsed as `< foo >` given certain stuff on the preceding and following lines.

STRd6: It looks like stopping at EOL was the key ingredient: https://github.com/DanielXMoore/Civet/commit/fe92ceda94883f4e35c638ed3ce5d9a376db7e2b

PR #26, opened 2022-12-10
Indentation flexibility

edemaine:
1. [Allow arbitrary spaces in indentation](https://github.com/DanielXMoore/Civet/commit/c6633644dc50c3bf5e2f3f5cac392c775bcdb086), instead of an even number of spaces. In particular allows for 3-space indent, or mixtures.
2. [Custom unification of tabs and spaces](https://github.com/DanielXMoore/Civet/commit/7e0619b10a18ed09add88a5d42489fc049c74973) via `"civet tab=NNN"` option.

### DIFF
--- a/source/parser.hera
+++ b/source/parser.hera
@@ -3943,6 +3943,7 @@ Reset
       coffeeNot: false,
       coffeeOf: false,
       implicitReturns: true,
+      tab: undefined, // default behavior = same as space
     }
 
     let indexOfRef, hasPropRef, spliceRef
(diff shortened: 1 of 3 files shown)


source/parser.hera
```diff
@@ -5047,8 +5054,20 @@ Init
 # EOS/Nested rules are used carefully and if we only compare to the
 # pushed value.
 Indent
-  ( "  " / "\t" )* ->
-    const level = $1.length
+  ( " " / "\t" )* ->
```
STRd6, line 5057: Here you can use `[ \t]*` for the rule. This will allow the parser to spend more time in optimized regex land. Then in the handler `$0` will be the entire string match so `$0.split('').forEach` or similar.

edemaine: Done.

PR #29, opened 2022-12-11
LSP: Update other documents asynchronously when content gets updated

lorefnon:
Currently civet lsp client does not update diagnostics for other files when a file is updated. This often results in inconsistent diagnostics when for example file foo.civet imports a non-existent function from bar.civet. Later if I update bar.civet adding this function, foo.civet continues to show an error until I edit foo.civet itself which is not ideal. 

This PR makes the assumption that a user will usually not have hundreds of files open and just updates diaganostics for all open files rather than hooking into dependency analysis. Other extensions like volar also make similar assumption.

### DIFF
--- a/lsp/source/server.mts
+++ b/lsp/source/server.mts
@@ -9,7 +9,6 @@ import {
   InitializeResult,
   MarkupKind,
   TextDocumentIdentifier,
-  HandlerResult,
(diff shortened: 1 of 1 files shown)


STRd6:
Looks good, thanks!

PR #32, opened 2022-12-11
JSX object shorthand

edemaine:
Support any object literal as a JSX attribute:
* `{foo}` as shorthand for `foo={foo}`
* `{foo, bar}` is shorthand for `foo={foo} bar={bar}`
* `{foo: bar}` is shorthand for `foo={bar}`
* Multiple `...rest` supported
* Methods get converted into functions
* Computed property names and `get`/`set` methods supported via `{...rest}` JSX notation

~~Oops, `[foo]: bar` isn't supported yet.~~

### DIFF
--- a/source/parser.hera
+++ b/source/parser.hera
@@ -1571,7 +1571,11 @@ PropertyName
   IdentifierName
   # https://262.ecma-international.org/#prod-ComputedPropertyName
   # NOTE: Using ExtendedExpression to allow If/Switch expressions
-  OpenBracket ExtendedExpression __ CloseBracket
+  OpenBracket ExtendedExpression __ CloseBracket ->
+    return {
+      type: "ComputedPropertyName",
+      children: $0,
+    }
 
 Decorator
   AtAt IdentifierReference Arguments?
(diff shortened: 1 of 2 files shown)


STRd6:
Looks good to me. At first I wasn't convinced about the `{foo: bar}` aliasing example (it's not really any simpler than the regular JSX) but seeing how the implementation works and the tests that show it in situations where there are many keys it makes a lot of sense.

I assume passing getter/setters allows you to declare inline objects to pass to a complex component or something along those lines.

edemaine:
Thanks!

Agreed that `{foo: bar}` isn't really better than `foo={bar}` but object literals seemed like the natural generalization for "what can go in the braces" given both `{foo}` and `{...foo}`, and it probably makes a difference when you have a big multiline object literal. And I think methods could be handy for passing callbacks in.

Getters (and sometimes setters) are used a lot in SolidJS to simulate what look like simple props but are actually reactive. In the Solid Discord I've seen the suggestion to use `<Component {...{get foo() { ... }}}>` in order to pass in custom getters. Normally getters are created automatically for you; see, for example, the Output tab of [this playground](https://playground.solidjs.com/anonymous/4b5207c9-cb6e-41ba-9246-68c5c23ab5a1). But it's nice to be able to make your own sometimes, and this PR definitely cuts down on the notational overhead.

PR #33, opened 2022-12-11
Indentation in middle of arrays

### LINKED ISSUE #27: Indentations in the middle
CoffeeScript allows for indentation in the middle of e.g. an array literal:

```coffee
[1
 2
 3]
```

Civet won't parse this, even after #26.

More generally, I think whenever we have indented forms, we often need to support switching from regular form to indented form, at least once. For example:
```coffee
[1, 2, 3   # regular form
 4, 5, 6   # switch to indented
 7, 8, 9]
```

Oddly CoffeeScript (see `ArgElisionList` in `grammar.coffee`) supports additional indentation blocks in the middle of an array literal:
```coffee
[1
 2
   3
]
```
This is probably less important to support. But one indentation level is pretty common I think.

edemaine:
Support for indentation after the first line of items in an array, e.g.:

```coffee
[1, 2, 3
 4, 5, 6
 7, 8, 9]
```

Fixes #27 (at least for arrays — maybe there are other cases of indentation-in-the-middle that we need to find).

This is still a lot more strict than CoffeeScript, which allows some weird things:

```coffee
[x
   y
     z
] → [x, y, z]

[x
    y
 z
   w
] → [x, y, z, w]

[x
   y: z
] → [x, {y: z}] (!)

[x
   y
     z: w
] → [x, y({z: w})]
```

### DIFF
--- a/source/parser.hera
+++ b/source/parser.hera
@@ -1312,7 +1312,14 @@ RangeExpression
 ArrayLiteralContent
   RangeExpression
   NestedElementList
-  ElementList
+  ElementList NestedElementList? ->
+    if ($2) {
+      // ElementList never ends in comma; otherwise, the next line of elements
+      // would be absorbed into the ElementList.
+      return [...$1, ",", ...$2]
+    } else {
+      return $1
+    }
 
 NestedElementList
   PushIndent NestedElement*:elements PopIndent ->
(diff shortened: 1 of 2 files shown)


STRd6:
Nice!

I'm fine with being more strict about indentation than CoffeeScript in some places, especially when they are almost never used in practice.

PR #36, opened 2022-12-12
Document JSX object literal enhancement

edemaine:


README.md
```diff
@@ -97,7 +97,7 @@ Things Kept from CoffeeScript
 - RestElement/RestParameter in any position `(first, ...midle, last) ->` -> `function(first, ...middle) { let [last] = middle.splice(-1)}`
 - `///` Heregexp
   - With some [changes](#things-changed-from-coffeescript).
-- JSX 😿
```
STRd6, line 100: I suppose with these enhancements the cat is no longer crying 😸

edemaine: That's the idea! Given that we're adding cool JSX enhancements, I don't think it makes sense to discourage its use (if it ever did).

PR #38, opened 2022-12-12
Support for many unbraced JSX attribute values

edemaine:
The list of allowed types matches [this list](https://github.com/LXSMNSYC/ideas/discussions/11) pretty well.

We could consider adding more of `MemberExpression`, but I fear this may not leave as much room for future extension. For example, `foo=array[i]` might conflict (especially if we allow whitespace before `[`) with a future meaning for `[key]`, namely, set an attribute with key equal to the *variable* `key` (as opposed to the name `"key"`). Similarly, `foo=obj.name` might conflict with an attribute starting with `.` (though I'm not yet sure what this would mean... perhaps an automatic `class`).

### DIFF
--- a/source/parser.hera
+++ b/source/parser.hera
@@ -3674,6 +3674,25 @@ JSXAttributeValue
   OpenBrace ExtendedExpression __ CloseBrace
   JSXElement
   JSXFragment
+  InsertInlineOpenBrace InlineJSXAttributeValue InsertCloseBrace
+
+# JSX shorthand to avoid explicit braces when unnecessary.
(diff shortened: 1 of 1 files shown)


STRd6:
I think the only thing that wouldn't work would be function application with implicit parens since that requires a space to start. Having member access and function application with explicit parens should work fine. Civet is already a little bit of a stickler about adding extra whitespace between some operators so this wouldn't be an excessive requirement.

edemaine:
Cool, I can work on these cases in another PR.

PR #39, opened 2022-12-12
Support computed JSX attribute names

edemaine:
`[name]=value` translates into `{...{[name]: value}}`

This PR is relative to #38, because it also touches the file `test/jsx/attr.civet`; I can rebase once #38 is merged.


STRd6:
Looks like another good addition!

edemaine:
Rebased!

PR #41, opened 2022-12-13
Omit type parameters in js output

lorefnon:


source/parser.hera
```diff
@@ -360,7 +360,7 @@ ExtendsToken
     return { $loc, token: $1 }
 
 ExtendsTarget
-  LeftHandSideExpression TypeArguments?
```
STRd6, line 363: I think this one should stay `TypeArguments` and `TypeArguments` should have `{ts: true, children: $0}` added. The `TypeParameters` rule also probably needs to be updated because it should only allow type identifiers and not any type.

The thinking is that TypeParameters are like function parameters in that they name a slot for a type to be received and  that TypeArguments are the actual types being slotted in.

lorefnon: Ah ok, so in following example: 

```ts
type Foo<T> = { x: T }
                ^
                 Type parameter

type FooStr = Foo<String>
                               ^
                                Type arguments
```

Then type parameters will need to be restricted to identifiers (and optionally constraints) where as Type args will be concrete types.

source/parser.hera
```diff
@@ -3871,14 +3871,15 @@ FunctionType
...
+    return { ts: true, children: $0 }
 
 TypeParameters
-  __ "<" __ Type (__ Comma __ Type)* (__ Comma)? __ ">" ->
+  __ "<" __ TypeParameter (__ Comma __ TypeParameter)* (__ Comma)? __ ">" ->
     return { ts: true, children: $0 }
 
 TypeParameter
-  __ Identifier TypeConstraint? TypeParameterDelimiter
+  __ Identifier TypeConstraint? # TypeParameterDelimiter
```
lorefnon, line 3882: @STRd6 I am not quite sure what was the purpose of the last part ...

lorefnon: As in ts doesn't seem to care about trailing commas, so do we have a need to ?

STRd6: It's so we can do:

```
TypeParameters
  __ "<" TypeParameter* __ ">" ->
    return { ts: true, children: $0 }
```

It looks like this was partially implemented, just not hooked up.

TypeParameterDelimiter could also be simplified by removing the `&( _* ">" )` assertion and changing `&( __ ">" )` to not insert a comma.

lorefnon: Ok, updated. The `  &EOS InsertComma -> $2` doesn't seem to be working for me though and I can't quite figure why

STRd6: I took a look and source/parser.hera:345 should be `TypeParameters` instead of `TypeArguments`.

Looks great overall!

lorefnon: facepalm, updated.

PR #43, opened 2022-12-14
Add autolet

IamIpanda:
All effects are shown in the test.

NOTES: 
+ Assignment children process missed. I can’t construct a legal case.
+ How to handle type?

### DIFF
--- a/source/parser.hera
+++ b/source/parser.hera
@@ -5202,7 +5202,7 @@ Init
       }
 
       function insertBeforeAssignment(assignment, content) {
-        if (assignment.children) {
+        if (assignment.children && assignment.children.length) {
           let indent = assignment.children[0][0][0]
           if (Array.isArray(indent)) indent = indent[indent.length - 1]
           assignment.children.unshift([indent, ...content])
@@ -5220,6 +5220,7 @@ Init
       for (const node of nodes) {
         if (node.type == "AssignmentExpression") {
           let undeclaredIdentifiers = node.names.filter(name => !hasDec(name))
+          if (node.children.length) createLetDecs(node.children, scopes)
           undeclaredIdentifiers.forEach((name) => currentScope.add(name))
           if (node.names.length === undeclaredIdentifiers.length) { 
             insertBeforeAssignment(node, ["let "])
--- a/source/parser.hera
+++ b/source/parser.hera
@@ -4068,6 +4068,7 @@ Reset
 
     module.config = parse.config = {
       autoVar: false,
+      autoLet: false,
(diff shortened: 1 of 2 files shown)


source/parser.hera
```diff
@@ -4147,6 +4148,7 @@ Reset
       set(b) {
         if (b) {
           this.autoVar = true
+          this.autoLet = false
```
STRd6, line 4151: This isn't necessary here or in the else branch below since `autoLet` is unrelated to the `coffeeCompat` flags.

edemaine: `coffeeCompat` implies `autoVar`, right?  How do `autoVar` and `autoLet` interact?  Presumably one should disable the other, maybe whichever is specified last.  So you could `"civet coffeeCompat autoLet"` and `autoLet` would win.

source/parser.hera
```diff
@@ -4161,6 +4162,7 @@ Reset
           this.coffeeOf = true
         } else {
           this.autoVar = false
+          this.autoLet = false
```
STRd6, line 4165: This one can be removed too.

source/parser.hera
STRd6: To resolve that we should check and process `autoLet` before `autoVar` inside `processProgram`.

IamIpanda:
Setting `autolet` is removed for `coffeeCompat`.

Add a recursion inner AssignmentExpression to resolve NOTE 1.

STRd6:
@IamIpanda Thanks for the PR, looks great!

I think for `autoLet` types should be handled by TypeScript's type inference. In cases where you need to specify a particular type use an explicit declaration.

This looks good enough to merge in. It's ok for features flagged behind `"civet ..."` directives to be somewhat experimental and open to future improvements.

@edemaine any other thoughts?

edemaine:
I agree!  I'd like to see a short description of what `autoLet` does in the README, though, ideally before merging.

I also had one question: Does `autoLet` still prevent shadowing across functions, as in:
```coffee
x = 5
function f()
  x = 10  # is this the global x?
```

I assume so, like `autoVar`.  I think that's fine, but I wanted to be clear which one it does. (And maybe it's worth adding a test to this effect.)

Assuming yes, in the future, I'd like to make a `shadow` flag where `autoVar` and `autoLet` both don't do this, and limit their automatic scoping to the current function, not parent functions, except when overridden by `global`/`nonlocal` (or `outer`?) directives, like Python. This can lead to fewer nonshadowing bugs (which I've run into maaany times), though it's probably less common with TypeScript.

PR #45, opened 2022-12-14
Support whitespace-free JSX attributes

edemaine:
* Fix output of `""` in `coffeeCompat` mode (previously was ``` `` ```)
* Oops, I never added `test/jsx/attr.civet` which had been testing past features
* Add support for whitespace-free JSX attributes, including unary, binary, call, attribute expressions. This involved refactoring unary processing to avoid duplicating that code.
* Fix support for strings in JSX attributes (previously, `coffeeInterpolation` would have broken things)

### DIFF
--- a/source/parser.hera
+++ b/source/parser.hera
@@ -44,7 +44,8 @@ Expression
 
 # https://262.ecma-international.org/#prod-Arguments
 Arguments
-  OpenParen ArgumentList? ( __ Comma )? __ CloseParen
+  ExplicitArguments ->
+    return $1
   # NOTE: Added spacing based implicit function application
   # Trailing (__ MemberExpressionRest)* is to capture trailing .someMethod and bind them at the right place
 
@@ -55,6 +56,9 @@ Arguments
   # x y z => x(y(z))
   ApplicationStart InsertOpenParen:open _*:ws ArgumentList:args InsertCloseParen:close ->
     return [open, module.insertTrimmingSpace(ws, ""), args, close]
+
+ExplicitArguments
+  OpenParen ArgumentList? ( __ Comma )? __ CloseParen
 
 # Start of function application, inserts an open parenthesis, maintains spacing and comments when possible
 ApplicationStart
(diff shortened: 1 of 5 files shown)


source/parser.hera
```diff
@@ -44,7 +44,8 @@ Expression
 
 # https://262.ecma-international.org/#prod-Arguments
 Arguments
-  OpenParen ArgumentList? ( __ Comma )? __ CloseParen
+  ExplicitArguments ->
```
STRd6, line 47: If the rule is only a single expression it is automatically returned so you don't need to add

```
->
  return $1
```

edemaine: Oh, I didn't know that — thanks!

source/parser.hera
```diff
@@ -3719,32 +3680,93 @@ JSXAttributeInitializer
...
+  InlineJSXCallExpression UpdateExpressionSymbol? ->
+    if ($2) return $0
+    return $1
+
+# CallExpression, with only explicit function calls.
+# This also acts as a replacement for LeftHandSideExpression
+# (we don't allow New because that has whitespace).
+InlineJSXCallExpression
+  "super" ExplicitArguments
+  "import" OpenParen ExtendedExpression __ CloseParen
```
STRd6, line 3734: You'll want to remove the `__` before `CloseParen` here.

edemaine: Why? Whitespace within parens seems fine. And I assume it's useful given that it's in `CallExpression` (but maybe I'm wrong).

STRd6: Actually it's fine, once inside the `(` whitespace is allowed.

source/parser.hera
```diff
@@ -3719,32 +3680,93 @@ JSXAttributeInitializer
...
+
+# CallExpressionRest, with only explicit function calls.
+InlineJSXCallExpressionRest
+  MemberExpressionRest
+  TemplateLiteral
+  ( OptionalShorthand / NonNullAssertion )? ExplicitArguments
+
+# MemberExpression, with PrimaryExpression -> InlineJSXPrimaryExpression
+InlineJSXMemberExpression
+  InlineJSXPrimaryExpression MemberExpressionRest* ->
```
STRd6, line 3747: There is an indented construction inside `MemberExpressionRest` to handle things like:

```
f a
  .b
f a
.b
---
f(a
.b)
f(a)
.b
```

We probably don't want to allow nested indentation accese here.

PR #56, opened 2022-12-15
support typescript `satisfies`

### LINKED ISSUE #42: TS: Satisfies
We need to keep up with additions to TS like satisfies. My initial feeling is that this will fit in similar to `as` which we already support.

https://devblogs.microsoft.com/typescript/announcing-typescript-4-9/#satisfies

samualtnorman:
fixes #42


STRd6:
Looks good, thanks!

PR #57, opened 2022-12-16
Default to tsx extension for transpiled files

lorefnon:


source/esm.civet
```diff
@@ -76,7 +76,7 @@ export async function load(url: string, context: any, next: any)
       filename: path
       sourceMap: true
 
-    transpiledPath := url + ".ts"
```
STRd6, line 79: I think this should stay as `.ts`.

register.mjs
```diff
@@ -31,7 +31,7 @@ export async function load(url, context, next) {
 
     // NOTE: Assuming ts-node hook follows load hook
     // NOTE: This causes .civet files to show up as .ts in ts-node error reporting (TODO: May be able to add a sourcemapping)
-    const result = await next(url.replace(extensionsRegex, ".ts"), {
```
STRd6, line 34: Also this should stay `.ts` as well.

STRd6:
Some more clarifying info:

I think only the the `.ts` -> `.tsx` in `typescript-service.mts` is necessary. That handles all the LSP stuff. `register` and `esm` affect running the files directly in Node or through `ts-node` (the change to transpiled path in `esm.civet` is what is breaking the tests).

Essentially any `.civet` file has three divergent paths that it may take:

- Civet LSP -> TSServer -> VSCode UI ( typescript-service.mts )
- ts-node -> Node.js runtime ( register.mjs / esm.civet )
- esbuild, vite, etc. -> built package ( esbuild-plugin.civet, other future build plugins )

The challenge is keeping the TSServer errors accurate for no matter where the file ends up. It looks like this PR will fix that (with the change to typescript-service.mts).

Hopefully that explains things a little better.

Also, thanks for the PR! Defaulting `.civet` files to be `.civet.tsx` looks like it should work great.

lorefnon:
But then if someone is using esm loader to server render react components etc. it wont work ?

Ideally we'd want to replicate the tsconfig resolution that ts-node does, and switch to tsx extension if jsx is configured in tsconfig ?

STRd6:
Yeah, there is more work that should be done later with the esm loader for server side JSX rendering. We can have the parser output additional info such as if any JSX was found in the file and auto-pick the proper extension. It may require a tweak to ts-node config or an additional TSX/JSX loader inserted in the chain as well.

STRd6:
Looks like the `tsconfig` fixes the build so I'm fine defaulting to `tsx` output. Might need to add a note in the README later that `jsxPreserve` is necessary to run `.civet` files with ts-node.

PR #59, opened 2022-12-16
support typescript parameter properties

samualtnorman:
okay it works but where should I go from here?

### DIFF
--- a/source/parser.hera
+++ b/source/parser.hera
@@ -346,6 +346,7 @@ NestedClassElement
 
 # https://262.ecma-international.org/#prod-ClassElement
 ClassElement
+  Constructor
   MethodDefinition
   # NOTE: Combined optional static and Method/Field definition
   ( Static TrailingComment* )? ( MethodDefinition / FieldDefinition )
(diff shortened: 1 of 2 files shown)


source/parser.hera
```diff
@@ -1594,6 +1595,80 @@ MethodDefinition
...
+      name: signature.name,
+      decorators,
+      signature,
+      block,
+      parameters: signature.parameters,
+    }
+
+ConstructorSignature
+  # NOTE: If this node layout changes, be sure to update `convertMethodTOFunction`
+  MethodModifier? "constructor" _* ConstructorParameters:parameters ReturnTypeSuffix?:suffix ->
```
STRd6, line 1614: `ConstructorSignature` can't have `MethodModifier` or `ReturnTypeSuffix` so those can be removed here.

Also there is a `@` constructor shorthand that should be added as an alternative to `"constructor"`.

STRd6:
Looks like a good start!

To add statements into the body of the constructor add them into `blockPrefix`. Later when the parser processes functions/methods/constructors those get added at the start of the block.

Let me know if you have any other specific questions.

STRd6:
Resolved in #80

PR #63, opened 2022-12-17
Refactor ref making

edemaine:
I'm planning some future changes that make even more global refs. This is a refactor to make this very easy.

### DIFF
--- a/source/parser.hera
+++ b/source/parser.hera
@@ -4220,7 +4220,7 @@ Reset
     }
     module.prelude = []
 
-    const refMakers = {
+    const declareRef = {
       indexOf(indexOfRef) {
         const typeSuffix = {
           ts: true,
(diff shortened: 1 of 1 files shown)


source/parser.hera
```diff
@@ -4214,62 +4214,57 @@ Reset
...
-      if (spliceRef) return spliceRef
-
-      spliceRef = {
-        type: "Ref",
-        base: "splice",
-        id: "splice",
-      }
-
-      return spliceRef
+    const refMakers = {
```
STRd6, line 4223: Maybe name this `addRefDeclaration` instead of `refMakers`?

STRd6:
Looks great overall, just added a minor naming suggestion.

source/parser.hera
edemaine: Good idea. How about `declareRef`?

PR #64, opened 2022-12-17
autoWithType / Implicit Let

IamIpanda:
# Warning
+ This flag will influence coffeescript compability.
+ This flag is not property named, may be named like `Implicit let/var`

# Summary
When `autoLet` or `autoVar` is enabled, and `autoWithType` enabled, the sentence
```
a: number = 3
```
will compile to 
```
let a: number = 3
```

While in coffeescript, it compile to
```
var number;

({
  a: number = 3
})
```
in civet, it compile to
```
{a:number = 3}
```
with coffeeCompat it compile to
```
var number
{a:number = 3}
```

This let/var only trigger when there is an assignment with strict `=`, the pattern `a: number` still compile to `{a: number}`, the pattern `a: (number = 3)` still compile to `a: {let number = 3}`.

Breaking down example, this actually is from difference between coffee and civet:
```
"civet autoLet autoWithType"
foo := () =>
    number = 3
    v = v0: number
    v1: number = 3
    v2: (number = 3)
    v3: (number = 3)
```
compile to 
```
const foo = () => {
    let number = 3
    return let v = {v0: number,
    v1:number = 3,
    v2: (number = 3),
    v3: (number = 3)}
}
```
In Coffeescript
```
(() => {
  var number, v;
  number = 3;
  v = {
    v0: number
  };
  return {
    v1: number = 3,
    v2: (number = 3),
    v3: (number = 3)
  };
});
```

### DIFF
--- a/source/parser.hera
+++ b/source/parser.hera
@@ -2941,6 +2941,28 @@ LexicalDeclaration
       children,
     }
 
+  InsertAutoWithType:l (BindingPattern / BindingIdentifier):id TypeSuffix:suffix __:ws Equals:eq ExtendedExpression:e ->
+    l = {
+      ...l,
+      $loc: {
+        pos: eq.$loc.pos - 1,
+        length: eq.$loc.length + 1,
+      }
+    }
+
+    let [splices, thisAssignments] = module.gatherBindingCode(id)
+
+    splices = splices.map(s => [", ", s])
+    thisAssignments = thisAssignments.map(a => [";", a])
+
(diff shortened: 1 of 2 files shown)


IamIpanda:
As discussed, this is a workaround for Coffeescript with type to fix conflicting character `:`.    
In Civet, we use `let`, `var`, `:=`, `.=` for type definition without ambiguity.    
So that proposal come to not such interesting but become dangerous in sensitive fields.    
So, close this PR before more people be interested in declaring without prefix or special `=`.

PR #65, opened 2022-12-17
Fix singleton TypeParameters compatibility with TSX

edemaine:
As discussed in discord, `<T>(x) => x` is not valid TSX (but is valid TS).  Convert to `<T,>(x) => x` for safety.

### DIFF
--- a/source/parser.hera
+++ b/source/parser.hera
@@ -564,7 +564,7 @@ Parameters
   "" ->
     return {
       type: "Parameters",
-      children:[{$loc, token: "()"}],
+      children: [{$loc, token: "()"}],
       names: [],
     }
 
@@ -601,6 +601,7 @@ NonEmptyParameters
           {...rest, children: rest.children.slice(0, -1)},
           close,
         ],
+        tp,
         names,
         blockPrefix,
       }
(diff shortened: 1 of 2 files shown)


source/parser.hera
```diff
@@ -4914,6 +4922,16 @@ Init
       return [rest, after]
     }
 
+    // Ensure given Parameters structure isn't parsed as JSX by TypeScript,
+    // by replacing singleton TypeParameters <Foo> with <Foo,>.
+    // Modifies `parameters` in-place.
+    module.parametersNotJSX = function(parameters) {
+      const [tp] = parameters.children
+      if (tp && tp.type === "TypeParameters" && tp.parameters.length === 1) {
+        tp.parameters.push(",")
```
STRd6, line 4931: Since this mutates in place it may cause a problem with some caching/backtracking situations. I'd recommend either moving the mutation to after the entire parse completes (no more backtracking so we're free to mutate) or returning a new `parameters` structure and re-assigning at the calling location.

edemaine: I'm new to the mutation after the entire parse completes. Could I just add some code to `processParams` say?

STRd6: Yes. Basically anything called inside `processProgram` (and so forth) is free to mutate the tree and nodes. Anything called in a handler shouldn't mutate anything that is not created from inside the handler itself since it may receive the same cached results multiple times.

PR #66, opened 2022-12-17
Support type parameters in JSX tag names

edemaine:
https://twitter.com/srsholmes/status/1593891832307634182

### DIFF
--- a/source/parser.hera
+++ b/source/parser.hera
@@ -3566,11 +3566,11 @@ JSXElement
 
 # https://facebook.github.io/jsx/#prod-JSXSelfClosingElement
 JSXSelfClosingElement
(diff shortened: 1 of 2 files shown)


source/parser.hera
```diff
@@ -3994,6 +3994,11 @@ TypeParameters
   __ "<" TypeParameter+ __ ">" ->
     return { ts: true, children: $0 }
 
+# TypeParameters without the leading whitespace
+CompactTypeParameters
```
STRd6, line 3998: I think this should be `CompactTypeArguments` and `TypeArgument+` since it is part of the application.

Parameters for declaration, arguments for application.

PR #68, opened 2022-12-18
Typed JSX elements/fragments for Solid

edemaine:


STRd6:
It's very cool that you're able to use `"civet"` directives to customize the emitted types to match Solid. It looks like a good approach that can be expanded to other frameworks later.

Looks like maybe the tests files didn't get added yet.

edemaine:
Yes, the hope is that we could support other JSX backends where that makes sense (`JSX.Element` doesn't suffice), and maybe re-use `client` and `server` for other things. Meanwhile I hope to add more features to the `solid` directive where Solid and TypeScript get in each others' way (e.g. `ref` and `use:___` directives).

> Looks like maybe the tests files didn't get added yet.

Oops! I often forget to `git add`; thanks.

STRd6:
Can you add a test for the situation where there is already a `type JSX` in scope? I'm not sure the parser tracks type refs the same as identifier refs (and I think they exist in separate namespaces in TS). Also maybe a test for if there is a userspace `class JSX` in scope since that would should create a type and and identifier.

STRd6:
Looks good!

PR #82, opened 2022-12-24
Class enhancements: Support abstract & readonly

lorefnon:


source/parser.hera
```diff
@@ -347,32 +352,40 @@ NestedClassElement
 # https://262.ecma-international.org/#prod-ClassElement
 ClassElement
   # NOTE: Combined optional static and Method/Field definition
-  Decorators? AccessModifier? ( Static TrailingComment* )? ClassElementDefinition
+  Decorators? ( AccessModifier __ )? ( Static TrailingComment* )? ClassElementDefinition
```
STRd6, line 355: This `__` is redundant because `AccessModifier` already ends with `__`.

source/parser.hera
```diff
@@ -347,32 +352,40 @@ NestedClassElement
...
+ProtectedLiteral
+  "protected" NonIdContinue ->
     return { $loc, token: $1 }
 
+ReadonlyLiteral
+  "readonly" NonIdContinue ->
+    return { $loc, token: $1, ts: true }
+
+AbstractLiteral
+  "abstract" NonIdContinue ->
```
STRd6, line 386: You can avoid the whitespace checking/trimming by adding `" "?` after `NonIdContinue` here and wrapping the whole rule in `$( )`. That way the `AbstractLiteral` will include a trailing space if present. See how `If` does it.

For the naming of these `Abstract` is more in line with the naming conventions than `AbstractLiteral`. They could also be moved into the `## Tokens` section and alphabetized.

lorefnon: Sorry what does `$()` do ? Not able to find the relevant usage in if clause ...

lorefnon: Ok, works now. What does `$( )` do exactly ? Prevent $1 from becoming an array ?

STRd6: `$("abstract" NonIdContinue " "?)` will replace the content inside of the parentheses with a string of the matching source text. So instead of ["abstract", undefined, " " | undefined] the result will be `"abstract"` or `"abstract "`.

lorefnon: ok, thank you.

STRd6:
Looks great, thanks!

PR #83, opened 2022-12-24
Add support for |> pipelines

lorefnon:
This PR conservatively implements the Pipe operator as described in https://github.com/tc39/proposal-pipeline-operator. 

Advanced features like placeholders or other features from hack are still in flux and likely to cause compatibility problems in future.

### DIFF
--- a/source/parser.hera
+++ b/source/parser.hera
@@ -15,6 +15,14 @@ TopLevelStatement
 
 # Expressions with If and Switch
 ExtendedExpression
+  NonAssignmentExtendedExpression
+  AssignmentExpression
+
+NonPipelineExtendedExpression
+  NonAssignmentExtendedExpression
+  NonPipelineAssignmentExpression
+
+NonAssignmentExtendedExpression
   # Check for nested expressionized statements first
   &EOS PushIndent ( Nested ExpressionizedStatement )?:expression PopIndent ->
     if (expression) return expression
@@ -23,7 +31,6 @@ ExtendedExpression
     return {...$2,
       children: [...$1, ...$2.children]
     }
-  AssignmentExpression
 
 ExpressionizedStatement
   DebuggerExpression
@@ -151,6 +158,12 @@ UpdateExpressionSymbol
 
 # https://262.ecma-international.org/#prod-AssignmentExpression
 AssignmentExpression
+  # It is important for pipeline to have higher precedence than
+  # usual binary operators, so that x |> & + 2 |> & * 3
+  # is equivalent to x |> (& + 2) |> (& * 3)
+  PipelineExpression
+  # TODO If NonPipelineAssignmentExpression or SingleLineAssignmentExpression is used here then behavior changes
+
   # NOTE: Try to match a single line assignment expression before matching newline then assignment expression
   TrailingComment*:ws AssignmentExpressionTail:tail ->
     if (ws.length) {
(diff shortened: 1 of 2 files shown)


source/parser.hera
```diff
@@ -151,6 +158,9 @@ UpdateExpressionSymbol
 
 # https://262.ecma-international.org/#prod-AssignmentExpression
 AssignmentExpression
+  PipelineExpression
+  # TODO If NonPipelineAssignmentExpression or SingleLineAssignmentExpression is used here then behavior changes
```
lorefnon, line 162: @STRd6 I am not entirely sure why I can't use the SingleLineAssignmentExpression rule here to remove the duplication. Somehow how the parser parses call expressions changes if I do that.

test/binary-op.civet
```diff
@@ -146,3 +146,35 @@ describe "binary operations", ->
     ---
     a instanceof b
   """
+
+  testCase """
+    pipe expression
+    ---
+    a |> fn
+    ---
+    (fn)(a)
```
STRd6, line 155: We could check for common cases where the fn is an identifier or already a parenthesized expression so we won't wrap in excess parens.

Ideally this would generate `fn(a)`.

test/binary-op.civet
```diff
@@ -146,3 +146,35 @@ describe "binary operations", ->
...
+  """
+
+  testCase """
+    chained pipe expression
+    ---
+    a |> foo |> bar
+    ---
+    (bar)((foo)(a))
```
STRd6, line 163: Similarly here should ideally be `bar(foo(a))`, only wrapping with parens if the expression is more complex.

source/parser.hera
STRd6: I've run into similar trouble trying to simplify this as well. It's not necessary to have it for this PR, but hopefully one of us can figure it out at some point.

STRd6:
I agree with skipping placeholders and the more complex piping syntax. The goal of Civet is for features to be simple and compose well so having the simplest pipe makes sense especially since we can compose with block function syntax (i.e. `&+1`) to cover the majority of cases.

Looks great overall, my only recommendation is to reduce excess paren wrapping if possible.

edemaine:
This looks very cool! I especially like the new `yield` example. Could/does `return` behave the same way?

I feel like this feature really highlights the lack of a different feature: the ability to define functions with placeholders. For example, I imagine it's really common to want to pass a value in as a first argument of a function, with other arguments given explicitly. What we want is shorthand for `($) => func($, foo, bar)`.  One proposal on Discord was `func((&), foo, bar)`, where each extra layer of parens around `&` pushes the function wrapper up a level.  Another option might be `func(&&, foo, bar)`, where each additional `&` pushes the function wrapper up a level. The TC39 proposal suggests `func(%, foo, bar)`.  Not sure what's best, but it kind of highlights the limitation of the current `&` notation... yet it feels like we're close.

Another optimization I'd like to see is that implicit functions don't actually get created.  For example, I'd like `x |> &+1` to compile to `x+1`, not `(($) => $+1)(x)`.  I'm not sure how far to go in this direction, but a little bit would go a long way I think.  Specifically, I think we should aim to optimize away a top-level double-arrow function (such as the one resulting from `&` notation).

Anyway, none of these are stopping the current PR, and I think it's fine to merge as is, but maybe we should raise these as separate issues to deal with next.

STRd6:
Thanks for the great PR!

Live in v0.5.11

lorefnon:
Thanks to both of you. 

> implicit functions don't actually get created

I think this is a low hanging fruit here. Esp. when `&` is used exactly once. 

Before we do the others, I think we need a better way to hygenically inject a temporary variable without potential of conflict. One common approach seems to be to have a registry of user defined/imported variables, which the parser can lookup and augment with generated temporary variables. Babel has a [nice utility](http://hzoo.github.io/babel.github.io/docs/advanced/plugins/scope/#generating-a-uid) that is quite handy. This may need the compiler to become multipass (or atleast be more scope aware) though which was beyond what I was comfortable attempting at this point.

edemaine:
I believe the "ref" mechanism could serve to create temporary variables. At least, that's what it does in CoffeeScript, so I think it's one of its goals. (Not sure it's been used for that yet though. Maybe in `for` loops?)

PR #84, opened 2022-12-26
Fix function application with implicit object

lorefnon:
Passing implicit (not brace delimited) object to a function as first parameter has some oddities currently. 

Something like this appears to work in basic example: 

```
foo
 a: 10
```
```js
foo(
 {a: 10},)
```

But if one more property is added things get weird: 

```
foo
 a: 10
 b: 20
```
```js
foo(
 {a: 10(
 {b: 20},)},)
```

Making the function invocation explicit through parens doesn't help: 

```
foo(
 a: 10
 b: 20
)
```
```js
foo(
 {a: 10(
 {b: 20})}
)
```

But interestingly if I add a space, then it is parsed as an implicit function application whose argument is a parenthesized implicit object expression, which works: 

```
foo (
 a: 10
 b: 20
) 
```
```js
foo(( {
 a: 10,
 b: 20,
}
))
```

This PR adds an additional rule so the first two scenarios also behave as expected.

### DIFF
--- a/source/parser.hera
+++ b/source/parser.hera
@@ -85,6 +85,8 @@ ArgumentsWithTrailingCallExpressions
 ArgumentList
   ArgumentPart ( __ Comma ( NestedImplicitObjectLiteral / NestedArgumentList ) )+
   # NOTE: Added nested arguments on separate new lines
+  InsertNewline InsertIndent NestedImplicitObjectLiteral ->
+    return [$1, $2, module.insertTrimmingSpace($3, '')]
   NestedArgumentList
   InlineArgumentExpressions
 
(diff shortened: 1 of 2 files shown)


source/parser.hera
```diff
@@ -85,6 +85,8 @@ ArgumentsWithTrailingCallExpressions
 ArgumentList
   ArgumentPart ( __ Comma ( NestedImplicitObjectLiteral / NestedArgumentList ) )+
   # NOTE: Added nested arguments on separate new lines
+  InsertNewline InsertIndent NestedImplicitObjectLiteral ->
```
STRd6, line 88: I would leave out the `InsertNewline` and `InsertIndent`

STRd6:
I made some minor changes here: https://github.com/DanielXMoore/Civet/commit/8a5e632f47ecb985372701875c10cff2c49bc77b

lorefnon:
@STRd6 Thank you - that looks better.

PR #86, opened 2022-12-26
Implicit dynamic import()

edemaine:
With this PR, I believe `import foo` works as well as `import(foo)` currently does.

However, `import(foo).then(...)` doesn't currently work, and still doesn't work with this PR. Any ideas on how to fix? It doesn't seem to be a caching issue...

### DIFF
--- a/source/parser.hera
+++ b/source/parser.hera
@@ -60,6 +60,9 @@ Arguments
   # a b => a(b)
   # a b, c, d => a(b, c, d)
   # x y z => x(y(z))
+  ImplicitArguments
+
+ImplicitArguments
   ApplicationStart InsertOpenParen:open _*:ws ArgumentList:args InsertCloseParen:close ->
     return [open, module.insertTrimmingSpace(ws, ""), args, close]
 
@@ -80,6 +83,7 @@ ArgumentsWithTrailingCallExpressions
   # Since this is recursive Arguments must consume input to avoid infinite recursion
   # NOTE: Do not allow trailing template literals to match
   Arguments ( Samedent !Backtick CallExpressionRest )*
+  ExplicitArguments ( !Backtick CallExpressionRest )*
 
 # https://262.ecma-international.org/#prod-ArgumentList
 ArgumentList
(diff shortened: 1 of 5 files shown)


source/parser.hera
```diff
@@ -80,6 +83,7 @@ ArgumentsWithTrailingCallExpressions
   # Since this is recursive Arguments must consume input to avoid infinite recursion
   # NOTE: Do not allow trailing template literals to match
   Arguments ( Samedent !Backtick CallExpressionRest )*
+  ExplicitArguments ( !Backtick CallExpressionRest )*
```
STRd6, line 86: You can remove this

source/parser.hera
```diff
@@ -489,8 +493,10 @@ LeftHandSideExpression
 CallExpression
   # NOTE: Tracking trailing member expressions based on indentation level for implicit arguments
   "super" ArgumentsWithTrailingCallExpressions
+  # Dynamic import(), with optional parentheses when not used at top level.
+  # (At top level, ImportDeclaration will match first.)
   # NOTE: Using ExtendedExpression to allow for If/Switch expressions
-  "import" __ OpenParen ExtendedExpression __ CloseParen
+  "import" ArgumentsWithTrailingCallExpressions
```
STRd6, line 499: and add `CallExpressionRest*` at the end of this line.

source/parser.hera
```diff
@@ -489,8 +493,10 @@ LeftHandSideExpression
 CallExpression
   # NOTE: Tracking trailing member expressions based on indentation level for implicit arguments
   "super" ArgumentsWithTrailingCallExpressions
```
STRd6, line 495: This line probably also needs `CallExpressionRest*` at the end as well.

source/parser.hera
edemaine: `Samedent` would prevent the first rule from matching, right?

STRd6: Since `Samedent` is inside the `*` it will match with zero items in the array if no `Samedent` was present, then that match would shadow this second production.

edemaine:
Thanks @STRd6 for the help! It all makes sense now. Thinking parser is difficult...

PR #87, opened 2022-12-28
Don't deploy civet.dev website from forks

edemaine:
I can disable the action on my fork, but it seems more polite to new contributors to not have this action run whenever they sync their repo with upstream.

I haven't tested this, but it seems correct from https://docs.github.com/en/actions/using-jobs/using-conditions-to-control-job-execution


edemaine:
Looks like it works! The deploy job gets skipped on my fork (after syncing), but seems to have run here.

PR #95, opened 2022-12-29
JSX #id and .class shorthands

edemaine:
* JSX `.class` shorthand, which stack together and with `class`/`className`
* JSX `#id` shorthand
* Improve JSX attribute space generation (fewer double spaces)
* Forbid JS comments in JSX (outside braces) — this never should have been allowed


STRd6:
Looks good!

PR #99, opened 2022-12-30
New CLI including REPL

edemaine:
* `-c`/`--compile` to compile specified `.civet` files to `.civet.tsx` or `.civet.jsx` files (according to `--js` flag)
* `-o`/`--output` lets you override the output directory/extension/filename
* Without `-c`, runs files directly (like `coffee`)
* `-` to run or compile stdin; `-o -` to output to stdout
* `civet <input >output` still works
* `civet` alone runs simple REPL with cute prompt: `🐱> `

I also made the main `compile` function turn on `coffeeCompat` by default (when no other `civet` directive) for filenames ending with `.coffee`. This currently messes up line numbers by 1 though... Ideally we'd have a way to send information to the parser, but I don't know how.

Builds on #96

TODO: `require` registration, so `require`s work (this is generally missing in Civet I think)

TODO: `civet foo.civet` in ESM mode?

### DIFF
--- a/build/build.sh
+++ b/build/build.sh
@@ -11,10 +11,10 @@ chmod +x "$BIN"
 rm dist/cli.js
 
 # esbuild-plugin
-./dist/civet --js -c source/esbuild-plugin.civet -o dist/.js
+./dist/civet --js -c source/esbuild-plugin.civet -o dist/esbuild-plugin.js
 
 # esm loader
-./dist/civet --js -c source/esm.civet -o dist/.mjs
+./dist/civet --js -c source/esm.civet -o dist/esm.mjs
 
 # types
 cp types/types.d.ts dist/types.d.ts
--- a/build/build.sh
+++ b/build/build.sh
@@ -11,10 +11,10 @@ chmod +x "$BIN"
 rm dist/cli.js
 
 # esbuild-plugin
-./dist/civet --js < source/esbuild-plugin.civet > dist/esbuild-plugin.js
+./dist/civet --js -c source/esbuild-plugin.civet -o dist/.js
 
 # esm loader
-./dist/civet --js < source/esm.civet > dist/esm.mjs
+./dist/civet --js -c source/esm.civet -o dist/.mjs
 
 # types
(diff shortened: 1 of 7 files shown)


build/build.sh
```diff
@@ -11,10 +11,10 @@ chmod +x "$BIN"
 rm dist/cli.js
 
 # esbuild-plugin
-./dist/civet --js < source/esbuild-plugin.civet > dist/esbuild-plugin.js
+./dist/civet --js -c source/esbuild-plugin.civet -o dist/.js
```
STRd6, line 14: Why not use the piping here?

source/main.coffee
```diff
@@ -36,6 +36,14 @@ module.exports =
   compile: (src, options=defaultOptions) ->
     filename = options.filename or "unknown"
 
+    # TODO: This incorrectly shifts line numbers by 1
+    if filename.endsWith('.coffee') and not
+       /^(#![^\r\n]*(\r\n|\n|\r))?\s*['"]civet/.test src
+      src = """
+        "civet coffeeCompat"
```
STRd6, line 43: It may be possible to update the CivetPrologue rule to not be as strict about EOL and then to have this as
```
"civet coffeeCompat";#{src}
```

build/build.sh
edemaine: The piping still works, but I figured that I built the CLI to make this more convenient, so why not use it? (It serves as an example of how to use the CLI.)

STRd6: I'd probably go with the long names or the `<` `>` in build scripts with a preference towards `< >` because it is an ancient and venerable bash tradition :)

edemaine:
Rebased and added a commit that does `"civet coffeeCompat";#{src}`

build/build.sh
edemaine: One more argument for not using `<`/`>`: on all operating systems (not just those with `/dev/tty` links), Civet will correctly know the input filename, in case there are bugs/errors during compilation.

I've switched to full filenames given your preference. I can switch fully back to `<`/`>` if you don't buy the argument above.

PR #100, opened 2022-12-30
Filter combined JSX classes

edemaine:
Currently, JSX attributes `.class1 .class2 class={foo()}` compile to ``class={`class1 class2 ${foo()}`}``. This isn't what people expect when `foo()` returns `undefined` or `false`.

This PR instead compiles `.class1 .class2 class={foo()}` to `class={["class1 class2", foo()].filter(Boolean).join(" ")}`.

### DIFF
--- a/test/jsx/attr.civet
+++ b/test/jsx/attr.civet
@@ -246,7 +246,7 @@ describe "JSX class shorthand", ->
     ---
     <div .foo class={bar()} />
     ---
-    <div class={`foo ${bar()}`} />
+    <div class={["foo", bar()].filter(Boolean).join(" ")} />
   """
 
   testCase """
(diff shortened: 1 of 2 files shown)


test/jsx/attr.civet
```diff
@@ -279,5 +279,5 @@ describe "JSX class shorthand", ->
     ---
     <div#id.class1.class2.t-[5px].{myClass()} class={anotherClass}/>
     ---
-    <div class={`class1 class2 t-[5px] ${myClass()} ${anotherClass}`} id="id"/>
+    <div class={["class1 class2 t-[5px]", myClass(), anotherClass].filter(Boolean).join(" ")} id="id"/>
   """
```
STRd6, line 283: Remember to add 'Files: Insert Final Newline' to your VSCode config.

PR #101, opened 2022-12-30
Insert Final Newline setting for VSCode

edemaine:
https://github.com/DanielXMoore/Civet/pull/100/files/51f4cb99b904b1b46cc1d30d7957f9a6907970ed#r1059542457


.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
+  "files.insertFinalNewline": true
+}
```
STRd6, line 3: 👻

edemaine: In case it's not obvious, this was intentional: by making the setting a workspace setting, you'll never have to tell anyone the same comment. 😅

edemaine: If you're complaining that the file has no terminating newline... That's how VSCode does it. 😅

STRd6: It just doesn't include the final newline in the settings file yet. :)

STRd6: VSCode!!!🙀

PR #102, opened 2022-12-31
XML comment support within JSX

edemaine:
Mentioned as idea 7 on https://github.com/lxsmnsyc/ideas/discussions/11
I've wanted this myself in the past, though partly because CoffeeScript's equivalent (`{### ... ###}`) is so verbose.

One use-case for `<!-- ... -->` comments is that they let you comment out a block of JSX that itself has `/*...*/` comments (similar to single vs. double quotes and their tripled forms).

(Of course this only works one level deep... you can't use `<!-- ... -->` to comment out a block that has other `<!-- ... -->` comments.)

### DIFF
--- a/README.md
+++ b/README.md
@@ -239,6 +239,7 @@ Things Added that CoffeeScript didn't
     `"civet solid server"` for server-only code (SSR only), or
     `"civet solid client server"` for isomorphic code that runs on
     client and server (SSR + hydration).
+  - XML comments: `<!-- ... -->` → `{/* ... */}`
 - CoffeeScript improvements
   - Postfix loop `run() loop` → `while(true) run()`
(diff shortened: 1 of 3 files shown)


STRd6:
It would be possible to support nested JSX comments as well by slightly changing the JSXComment rule. I've also considered allowing nested `/* */` comments in Civet since it seems like a good convenience.

edemaine:
XML comments can't normally nest inside each other, I don't think, so it feels a little weird to support... But same goes for C comments, I guess. So I'm not sure what's best; I agree it'd be useful to allow `/*`/`*/` to nest, but perhaps better for another PR?

I came up with a "better" justification for this PR: JSX encourages thinking in XML mode. It feels natural to write XML comments in that context.

But it's not a big deal either way...

STRd6:
I agree that we can consider nested comments as a separate PR.

PR #103, opened 2023-01-02
Add civet config

IamIpanda:
This pull request try to add a civet-config.json reader.

```
{
    "compilerOptions": {
        “coffeeCompat”: true
    }
}
```

Now it works by:
+ For first file passed to civet, search its directory for a `package.json`, if it not exists, then its upper directory.
+ Try to load `civet-config.json`(or other 2 alias) under that directory.

Limits:
+ Reading file from stdin, cannot locate where file from, so cannot load config.
+ REPL don’t have a filename, cannot load config.
 
Tested on Mac, Not tested for Windows. 
   
Is there some way to test cli features?

### DIFF
--- a/source/parser.hera
+++ b/source/parser.hera
@@ -4603,7 +4603,7 @@ Reset
       })
     }
 
-    module.config = parse.config = {
+    module.config = {
       autoVar: false,
       autoLet: false,
       coffeeBinaryExistential: false,
@@ -4804,6 +4804,11 @@ Reset
         }
       }
     })
+
+    if (parse.config != null) {
+      Object.assign(module.config, parse.config)
+    }
+    parse.config = module.config
 
     // TODO: this is only here because directive prologues depend on it
     // eventually all these `module.*` variables should be handled better in Hera
--- a/build/esbuild.coffee
+++ b/build/esbuild.coffee
@@ -33,6 +33,23 @@ extensionResolverPlugin = (extensions) ->
 resolveExtensions = extensionResolverPlugin(["hera", "coffee"])
 
 esbuild.build({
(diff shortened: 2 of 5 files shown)


Argeento:
[cosmiconfig](https://github.com/davidtheclark/cosmiconfig) is a good, tested way to find and parse a config file. Maybe there is a space for this tool here?

IamIpanda:
I add a commit for:
+ `--config` option
+ use cwd when there is no one input file or from stdin.

2 questions left:
+ Do we sure to add a dependency to search config? use [cosmicconfig](https://github.com/davidtheclark/cosmiconfig) or [lilconfig](https://github.com/antonk52/lilconfig)?
+ I should move config code out to a separate file, but failed to build it: `cli.coffee` compiles in esbuild with `bundle = false`, because the only dependency `main` will compile later. But now come a new file which shouldn't compile itself, so I have to set `bundle = true` (I'm not familiar with esbuild. Is there way to bundle one file in but one file out in esbuild?). That will produce usable binary, but cannot pass [Github Action](https://github.com/DanielXMoore/Civet/actions/runs/3847293071/jobs/6553570010)

STRd6:
We do not need to add a dependency for config right now.

For building you can add another esbuild step to build the config file then require it from the cli file similar to how it requires `./main`. They should all end up in the `dist` folder as `.js`.

IamIpanda:
move config.coffee out.

source/parser.hera
```diff
@@ -4603,7 +4603,7 @@ Reset
       })
     }
 
-    module.config = parse.config = {
```
STRd6, line 4606: Won't this get overridden with every parse?

PR #105, opened 2023-01-03
Auto let Fix

IamIpanda:
Fix missing let decs inner declaration.    
I'm not sure if it's needed to fix that breaking `return let` which won't happen on `auto Var` as autovar move all variables to top.

### DIFF
--- a/source/parser.hera
+++ b/source/parser.hera
@@ -5851,6 +5851,7 @@ Init
           }
         }
         else if (node.type == "Declaration") {
+            if (node.children && node.children.length) createLetDecs(node.children, scopes)
             node.names.forEach((name) => currentScope.add(name))
         }
         else {
--- a/test/autolet.civet
+++ b/test/autolet.civet
@@ -31,6 +31,7 @@ describe "auto let", ->
     """
 
 describe "auto let with function", ->
+    // TODO: We should fix the return let here, this is invalid.
     testCase """
         auto let with function
         ---
@@ -41,6 +42,9 @@ describe "auto let with function", ->
             var b = 2
         }
         a = 2
+        let c = (b) => {
+            c = 1
+        }
         ---
         let b = 1
         function b(a) {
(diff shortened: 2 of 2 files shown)


edemaine:
If we want `autoLet` to be a viable option, I think we should fix the `return let`...

IamIpanda:
> If we want `autoLet` to be a viable option, I think we should fix the `return let`...

Return let can be fixed, but it would be only the start as `let` don’t return value, that means any assignment as an expression should be checked, and hoist to the a proper position (Should be nearest position where can put a `let`, or maybe just block top to make it possible.) I’m not sure if now AST support to detect the `AssignmentExpression`’s return value is “used”

I’m checking the PR similar to that for coffeescript: https://github.com/jashkenas/coffeescript/pull/5395

STRd6:
Here's an idea:

When a declaration is in return position prepend it with a simple `let ref`, then assign the final declaration to the ref, then return the ref.

```
->
  let a = 1, b = 2
---
function() {
  let ref
  let a, b = ref = 2
  return ref
}
```

This may be useful for Civet in general to have declarations behave more like expressions.

edemaine:
Nice, yes, maybe we can use this to allow `let` declarations within `if..else` blocks (along with `return`). But it doesn't work with `const`... An alternative is to wrap in IFFEs, like `for` and `do`. This seems more universal.  Hmm, except then it's hard to `return` out of the block (e.g. `if x then return let y = 5`).

IamIpanda:
I think it would be quite normal for following code before how it interact with implicit return
```
a = b = 3
c = (d = 3)
export default MyReactFC = () => <></>
```

edemaine:
Anyway, this PR is probably worth merging already, as it fixes some bugs in `autoLet`. Perhaps should add a comment to the `return let` example like `// TODO This is currently invalid`. And we can continue to improve inner declarations in future PRs.

IamIpanda:
Comment added.

STRd6:
`autoVar` does hoisting to the enclosing function blocks so a similar mechanism could be used here as well.

PR #106, opened 2023-01-03
Support `|> return` and document pipes

edemaine:


README.md
```diff
@@ -209,6 +209,11 @@ Things Added that CoffeeScript didn't
   - Function call: `x.map &.callback a, b` → `x.map($ => $.callback(a, b))`
   - Unary operators: `x.map !!&` → `x.map($ => !!$)`
   - Binary operators: `x.map &+1` → `x.map($ => $+1)`
+- Pipe operator (based on [Hack pipes](https://docs.hhvm.com/hack/expressions-and-operators/pipe) and the [TC39 proposal](https://github.com/tc39/proposal-pipeline-operator))
```
STRd6, line 212: I think this is more closely related to F# pipes. Hack pipes use `$$` as the topic reference and require using the topic reference in the right hand side.

STRd6: It's probably worth linking to both as long as it's clear that it's the unary function application implementation that exists in Civet. If TC39 adopts Hack pipes we'll need to adjust to support it at that time but I don't foresee them ever adopting the pipe proposal.

PR #109, opened 2023-01-03
Avoid implicit return via trailing semicolon

edemaine:
As suggested by @orenelbaum on Discord, this PR disables `implicitReturns` in functions that have a trailing semicolon.

I may not have dealt with all cases; I'm not sure what a trailing semicolon at the end of a `for` loop should mean, for example. But hopefully a useful enough start!

### DIFF
--- a/source/parser.hera
+++ b/source/parser.hera
@@ -1166,7 +1166,7 @@ BracedBlock
     }
 
 SingleLineStatements
-  ( TrailingComment* Statement ):first ( ( TrailingComment* Semicolon TrailingComment* ) Statement? )*:rest ->
+  ( TrailingComment* Statement ):first ( SemicolonDelimiter Statement? )*:rest ->
     if (rest.length) {
       return [first, ...rest]
     }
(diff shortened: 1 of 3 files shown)


source/parser.hera
```diff
@@ -1166,7 +1166,7 @@ BracedBlock
     }
 
 SingleLineStatements
-  ( TrailingComment* Statement ):first ( ( TrailingComment* Semicolon TrailingComment* ) Statement )*:rest ->
+  ( TrailingComment* Statement ):first ( ( TrailingComment* Semicolon TrailingComment* ) Statement? )*:rest ->
```
STRd6, line 1169: `( TrailingComment* Semicolon TrailingComment* )` could be updated to `StatementDelimiter`

edemaine: `StatementDelimiter` includes `EOS` as a special case, which we probably don't want for `SingleLineStatements`. I could add an `ExplicitStatementDelimiter` perhaps?

STRd6: Good point, `ExplicitStatementDelimiter` or `SemicolonDelimiter` would work for the name of that common part.

PR #122, opened 2023-01-05
CLI support for running script with arguments

edemaine:
Previously, `civet foo bar` would try to run two scripts, `foo` and `bar`. This is not how `node` or `coffee` work.

With this PR, `civet foo bar` runs script `foo` with argument `bar` (via `process.argv`). This is exactly how `node` and `coffee` work.

Note that `civet foo.civet -c` will no longer compile `foo.civet` (it will run it with an argument of `-c`).  Now `-c` needs to come first, as in `civet -c foo.civet`.  I think this is a worthwhile tradeoff.

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,6 +28,8 @@ npm install -g @danielx/civet
 civet
 # Compile civet source file to typescript
 civet < source.civet > output.ts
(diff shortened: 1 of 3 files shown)


STRd6:
I like it.

PR #123, opened 2023-01-05
Document void return to avoid implicit return

edemaine:
Let me know if I should just commit small things like this directly to `master`.


STRd6:
> Let me know if I should just commit small things like this directly to `master`.

Please do!

PR #124, opened 2023-01-05
New operators: is not, is in, is not in, not in

edemaine:
* `is not` &rarr; `!==`
* `a not in b` &rarr; `!(a in b)` (like CoffeeScript, but with `in` instead of `of`; disabled when `coffeeOf` is on)
* `is in` and `is not in` are like CoffeeScript's `in` and `not in` but for general use

Open questions:
* Should we allow implicit arrays like `role is in 'admin', 'editor'` and `char is in 'a'..'f'`?
* Or float ranges like `0..1`?
* Should we use `includes` instead of `indexOf` with the new Civet `is [not] in` operators? (Daniel says CoffeeScript operators should probably stay as-is for compatibility. I'm not totally convinced...)

### DIFF
--- a/README.md
+++ b/README.md
@@ -139,14 +139,14 @@ Things Changed from CoffeeScript
 
 - `==` → `==` rather than `===` (unless you specify `"civet coffeeCompat"` or `"civet coffeeEq"`)
 - `!=` → `!=` rather than `!==` (unless you specify `"civet coffeeCompat"` or `"civet coffeeEq"`)
-- `is not` → `===`
+- `is not` → `!==`
   (unless you specify `"civet coffeeCompat"` or `"civet coffeeNot"`),
   instead of `isnt`
   (unless you specify `"civet coffeeCompat"` or `"civet coffeeIsnt"`)
 - `for in` and `for of` are no longer swapped and become their JS equivalents (unless you specify `"civet coffeeCompat"` or `"civet CoffeeOf"`)
 - `a is in b` → `b.indexOf(a) >= 0` and
   `a is not in b` → `b.indexOf(a) < 0` instead of `a in b` and `a not in b`;
-  `a in b` remains `a in b` as in JS
+  `a in b` remains `a in b` as in JS, and `a not in b` → `!(a in b)`
   (unless you specify `"civet coffeeCompat"` or `"civet coffeeOf"`)
 - `x?.y` now compiles to `x?.y` rather than the `if typeof x !== 'undefined' && x !== null` if check
 - Existential `x?` → `(x != null)` no longer checks for undeclared variables.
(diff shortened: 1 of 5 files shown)


source/parser.hera
```diff
@@ -1870,18 +1869,22 @@ BinaryOpSymbol
...
     }
-  CoffeeOfEnabled "not" NonIdContinue __ "in" NonIdContinue ->
+  ( "is" NonIdContinue __ "not" NonIdContinue __ "in" NonIdContinue ) / ( CoffeeOfEnabled "not" NonIdContinue __ "in" NonIdContinue ) ->
     return {
       ref: module.getRef("indexOf"),
       suffix: " < 0",
       special: true,
     }
+  # NOTE: "is not" must come after "is not in"
+  "is" NonIdContinue __ "not" NonIdContinue -> "!=="
```
STRd6, line 1885: We probably want to use `_*` here because if the following line starts with an indented `not` it will get eaten up by `__ "not"`.

edemaine: Isn't that intentional? What would `is` at the end of a line mean? I was just matching the existing style from CoffeeScript's `not in` above. See also https://coffeescript.org/#try:5%20not%0A%0Ain%2010

edemaine: This does raise an ambiguity in `coffeeCompat` mode: `x is not y` already means `x !== not y`. This PR changes that behavior to `x !== y`... 🙁

STRd6: > Isn't that intentional? What would `is` at the end of a line mean? I was just matching the existing style from CoffeeScript's `not in` above. See also https://coffeescript.org/#try:5%20not%0A%0Ain%2010

That's probably correct. The difficulty of whitespace sensitivity is you always have to consider it when writing the parser :)

edemaine:
I disabled `is not` when `coffeeNot` is enabled.

I tried to also add a `not in` operator in non-Coffee mode, but for some reason it's not working... If you see my mistake, please let me know!

README.md
```diff
@@ -140,8 +139,15 @@ Things Changed from CoffeeScript
 
 - `==` → `==` rather than `===` (unless you specify `"civet coffeeCompat"` or `"civet coffeeEq"`)
 - `!=` → `!=` rather than `!==` (unless you specify `"civet coffeeCompat"` or `"civet coffeeEq"`)
+- `is not` → `===`
```
STRd6, line 142: Looks like this should be `!==`

edemaine:
I fixed `not in` by correctly augmenting `ReservedWord`. At the same time I fixed `not` being assignable when `coffeeNot` is on, which seems wrong. (CoffeeScript doesn't allow it, and in Civet with `coffeeNot` on, you can't ever read the value of `not`.)

STRd6:
I think we can merge this in as it is now and split out the open questions into follow on PRs as needed, @edemaine thoughts?

edemaine:
Yep, I agree!

PR #125, opened 2023-01-06
New Playground tag in docs

Argeento:
Add snippets compiler

TODO:
  - [ ] Fix missing code after 2 new lines in the Playground slot
  - [ ] Add editable textarea in the Playground
  - [x] Import civet compiler from `dist`, instead of exec `node dist/civet -c tmp.civet`


build/build.sh
```diff
@@ -18,3 +18,6 @@ rm dist/cli.js
 
 # types
 cp types/types.d.ts dist/types.d.ts
+
+# create browser build for docs
+npx terser dist/browser.js --compress --mangle --ecma 2015 --output civet.dev/public/__civet.js
```
STRd6, line 23: If we're going to use `terser` we should add it as a dev dependency instead of using `npx` to lock the version.

STRd6: I should have read further, it looks like it was already added so we should be able to just call it directly as `terser dist/browser.js --compress --mangle --ecma 2015 --output civet.dev/public/__civet.js`

Argeento: You are right. Fixed

edemaine:
I looked some at the `import` issue.

In `civet.dev/.vitepress/compileCivet.ts`, I cannot `import {compile} from '../../dist/main.js';` because then the `import` gets attempted in the browser (at least in Vite dev mode, which `import`s modules), and `dist/main.js` sets `module.exports` (CJS style) instead of `export`ing (ESM).

@STRd6 Are we OK with rewriting everything to be ESM? If we still want a CJS build (which I think is necessary for using the compiler in CJS contexts), we can use e.g. [NPM package esm-cjs](https://www.npmjs.com/package/esm-cjs) to rewrite `export`s back into `module.exports`. (The reverse is harder.) I did such conversion in [this repo](https://github.com/edemaine/preact-render-to-dom) for example.

Argeento:
> Another related problem is I think Prettier and Civet are both included in the main bundle and then included again (via unpkg) in the worker. So we're loading two copies of Prettier/Civet, which is not good.

Vite treeshakes imported Civet and Prettier in `compileCivet.ts` -> This packages are not included on client side. See the client build size in devtools network.

> I'm less clear on how to import Civet from the worker, civet.dev/public/playground.worker.js. Should we load from unpkg.com, or copy the built main.js into public?

By copying the build to the vitepress public folder, we always have the newest version of compiler.
// TODO: Add version of civet to the `__civet.js`. Browser should request this file like this: `/__civet,js?v=0.54.0` to avoid service worker caching.

> Another thing: it seems like the site transpiles and rerenders all code samples on load / hydration. This seems wasteful, and less robust. (In particular, when things aren't working, all the code samples disappear.) I'd rather see code transpilation happen on when changing code. Is this possible?

Transpile and render code samples are done during the build process. You can disable JS and refresh the site to make sure of this.

This is the code to rerender on clinet side: https://github.com/DanielXMoore/Civet/pull/125/files#diff-5009f1b4a3d4e037a5e0781741f57f95af657682b868186fad68245e8949227fR52-R56

PR #131, opened 2023-01-06
Avoid object literal calls with class extends

### LINKED ISSUE #129: Implicit object literal calls vs class definitions
@IamIpanda posted this example:

```coffee
class A
  a: number
  
class B extends A
  b: number
```

https://civet-web.vercel.app/#Y2xhc3MgQQogIGE6IG51bWJlcgogIApjbGFzcyBCIGV4dGVuZHMgQQogIGI6IG51bWJlcg==

Currently, this looks like and gets treated as a function call `A({b: number})`. But this is almost certainly not what the user meant. This is a special feature of classes where we can have bare type declarations like `b: number` (`FieldDefinition` within `ClassElementDefinition` within `ClassElement`).

@STRd6 What's the best way to communicate that `ExtendsClause` shouldn't, at the top level, use the implicit-function-call-with-implicit-object rule? Do we need to duplicate the whole rule tree from `LeftHandSideExpression`? Or set some state and disable caching? Pretty annoying...

It's not strictly necessary, but we ideally would still allow implicit function calls *not* at the top level, e.g.:

```coffee
class B extends (A
  b: c
)
```

Update: Oh, I notice `IndentedApplicationAllowed` is exactly for this purpose. Just need to figure out which cache rules to invalidate...

edemaine:
Fixes #129


edemaine:
I tried replacing `LeftHandSideExpressionWithIndentedApplicationSuppressed` with `ExpressionWithIndentedApplicationSuppressed` but I find that

```coffee
class A extends if x then B else C
  x: number
```

parses as

```js
class A extends (x)? B : C({
  x: number,
}) {
}
```

Any idea why? I tried disabling caching.

STRd6:
I think we probably need to use `ExpressionWithIndentedApplicationSuppressed` in the one line `else` clause as well since it is `C` that is doing the application.

STRd6:
Look inside `ElseExpressionBlock`

STRd6:
When testing out the if expressions inside of `extends` it looks like JS doesn't like them. We'd probably need to use a ref and hoist the conditional to before the class declaration.

It's probably simpler for now to stick with the LHS with suppressed indented application. Sorry for leading you astray!

edemaine:
I did test this in Chrome:

image

Also works in Node:

image

But you're right that it needs wrapping parentheses. Is there an easy way to add this only when needed?

STRd6:
What I've done previously is:

```
switch(exp.type) {
  case "Identifier":
  case "Literal":
  case "ParenthesizedExpression":
    return exp;
    break;
  default:
    return {
      type: "ParenthesizedExpression",
      children: ["(", exp, ")"]
    }
```

... which should probably be extracted to a module level helper function.

STRd6:
Thanks, great work!

PR #134, opened 2023-01-06
ESM build and refactoring

edemaine:
* Add support for ESM `.coffee` files in `coffee-esm.mjs` (autodetection)
* Build new ESM version `dist/main.mjs`, which Vite seems to need in #125 (I've tested this in code I haven't yet pushed). (If I just told esbuild to convert the CJS code to ESM, you wouldn't be able to do `import {compile} from '@danielx/civet'`; only the default export gets preserved.)
* CoffeeScript code refactored to use ESM (but some things like Hera parser are still CJS)
* Note that default exports no longer have everything; need to `import *`

### DIFF
--- a/source/main.coffee
+++ b/source/main.coffee
@@ -1,8 +1,11 @@
 "civet coffeeCompat"
 
-{ parse } = require "./parser"
-{ prune } = gen = require "./generate"
-{ SourceMap, base64Encode } = util = require "./util.coffee"
+import parser from "./parser.hera"
+{ parse } = parser
+import generate, { prune } from "./generate.coffee"
+import * as util from "./util.coffee"
+{ SourceMap, base64Encode } = util
+export { parse, generate, util }
 
 defaultOptions = {}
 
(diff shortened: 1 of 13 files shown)


edemaine:
Oh, in `package.json`, I'm not sure whether `"exports"`/`"."` should point to the `.js` or `.mjs` file... Vite was unhappy with the former, but it was fine when I explicitly imported `".../main.mjs"`.

STRd6:
I think using `import` and `require` conditions is a good idea. My main goal is to have Civet "just work" whatever type of project people `npm install` into. And that probably means adding both ESM and CJS builds to the default dist with the proper `package.json` config.

source/main.coffee
```diff
@@ -34,45 +37,41 @@ uncacheable = new Set [
...
+      srcMapJSON = sm.json(filename, "")
+      # NOTE: separate comment to prevent this string getting picked up as actual sourceMappingURL in tools
+      return "#{code}\n#{"//#"} sourceMappingURL=data:application/json;base64,#{base64Encode JSON.stringify(srcMapJSON)}\n"
+    else
+      return {
+        code,
+        sourceMap: sm
+      }
+
+  generate ast, options
```
STRd6, line 75: We may also want add a default export object that holds all of the named exports as a convenience to the importer.

PR #135, opened 2023-01-06
JSX +- toggles like LiveScript object literals

edemaine:
Inspired by https://github.com/facebook/jsx/issues/134 but with both a true and false version matching LiveScript

### DIFF
--- a/civet.dev/cheatsheet.md
+++ b/civet.dev/cheatsheet.md
@@ -357,6 +357,7 @@ for item of iterable
 ## JSX
 
 Enhancements, inspired by [solid-dsl discussions](https://github.com/solidjs-community/solid-dsl/discussions)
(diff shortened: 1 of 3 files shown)


STRd6:
We might want to add `!` as a synonym for `-` in both JSX and object literal contexts.

edemaine:
I did notice an issue with `+` and `-`: because `-` is a valid part of an identifier, you can't write `+foo-bar` (it's treated as a single `+`).  Still, I like the symmetry of `+` and `-`... Should we leave the possibility of `!` for a future PR?

STRd6:
I think it is correct that `+foo-bar` is a single attribute. The correct way for someone to work around it would be to add a space. We already have several other places where whitespace around operators is significant.

PR #136, opened 2023-01-07
Expanded LiveScript toggles to include ! and added ! to JSX attribute toggle shorthand

STRd6:


edemaine (commented):
This seems fine to me, aside from lack of documentation. Just want to point out:

* No existing language does this (but JSX open issue does)
* Redundant with `-`
* Asymmetric with `+`
* `{!foo}` (which means `{foo: false}`) is quite asymmetric/different from `{foo}` (which means `{foo: foo}`)

edemaine:
Hmm, maybe I'd be happier with the symmetry if we also added `!!` for the true version??

STRd6:
> Hmm, maybe I'd be happier with the symmetry if we also added `!!` for the true version??

A possibility for sure.

`{¡foo}` -> `{foo: true}` 🙃

civet.dev/cheatsheet.md
```diff
@@ -41,11 +41,21 @@ c .= 10
...
+</Playground>
+
+You can use a flagging shorthand inspired by [LiveScript](https://livescript.net/#literals-objects).
+
+<Playground>
+config := {
+  +debug
+  -live
+  !verbose
+}
```
edemaine, line 58: This actually duplicates an example below; could you remove that one?

PR #139, opened 2023-01-07
JSX implicit fragments

edemaine:


STRd6:
I think this is good overall but I think keeping array items as separate would be best.

```
[
  <a>
  <a>
  <a>
]
---
[
  <a />,
  <a />,
  <a />,
]
```

We can merge it as is now with plans to follow up on the array stuff.

edemaine:
Oren's argument in Discord for this behavior was "that's how object literals work". Do you buy this or still have the same preference?

A more practical issue is that I don't know how to implement it, except as a flag like the no-implicit-function-call-with-object-literal flag. Which I guess is a reasonable way to do it.

STRd6:
> Oren's argument in Discord for this behavior was "that's how object literals work". Do you buy this or still have the same preference?

I haven't been doing much JSX recently so I'll leave it up to your judgement about how often people would use arrays of JSX elements. I know I used them at least occasionally when I was using React a few years ago.

> A more practical issue is that I don't know how to implement it, except as a flag like the no-implicit-function-call-with-object-literal flag. Which I guess is a reasonable way to do it.

I think that is basically the way to do it. An additional note is that it should ideally be a stack rather than a flag so that we can push and pop when backtracking without losing data but we won't have to implement it that way for the first version.

edemaine:
After building the [whole stack thing](https://github.com/DanielXMoore/Civet/pull/139/commits/18905e80afc9e105c017fc996f758291152b7de9) I realized a simpler workaround: check for a matching `JSXTag` as array content before trying a full `ExtendedExpression`. So much easier!

While playing with caching I noticed [some more cache disablings that changed test results](https://github.com/DanielXMoore/Civet/pull/139/commits/6be913fec461b22396ee8e9534b12d9b675202f6). So I've added those too.

STRd6:
> While playing with caching I noticed [some more cache disablings that changed test results](https://github.com/DanielXMoore/Civet/pull/139/commits/6be913fec461b22396ee8e9534b12d9b675202f6). So I've added those too.

A feature I'd like to add to Hera is to annotate rules like:

```
@no-cache
Nested
  ...
```

Then process the dependency tree of rules to propagate the `@no-cache` annotation to all the parent rules and use that to populate the Set. Something for the future!

The stack work will still be useful for updating `ExpressionWithIndentedApplicationSuppressed` eventually.

STRd6:
Looks great!

PR #151, opened 2023-01-07
Support `import`/`export` `type`/`{type}`

### LINKED ISSUE #143: Support import named type
```
import A, {B, type C} from 'library'
```

Should compile to:

```
import A, {B, type C} from 'library'
```

and should compile to:

```
import A, {B} from 'library'
```

when passed compiler option `js: true`.

edemaine:
Fixes #143


STRd6:
Thanks, very good!

PR #152, opened 2023-01-07
Improve extends '<' shorthand

STRd6:


source/parser.hera
```diff
@@ -381,14 +381,26 @@ ClassHeritage
...
-  __ ExtendsToken __ ExtendsTarget
+  ExtendsToken __ ExtendsTarget
 
 ExtendsToken
-  # NOTE: Added "<" extends shorthand
-  "<" ->
-    return { $loc, token: "extends" }
-  "extends" NonIdContinue ->
-    return { $loc, token: $1 }
+  # NOTE: Added "<" extends shorthand which must be surrounded by spaces
```
edemaine, line 387: This comment feels off. The shorthand *doesn't* need to be surrounded by spaces (as the tests show), but the expansion does. Otherwise looks great!

STRd6: You're right, that was from an intermediate version. I've updated the comment.

PR #159, opened 2023-01-08
Function declarations without parentheses

edemaine:
Remove need for `()` in defining argumentless functions with `function`:

```coffee
function name
  data?.part[0]?.name
function abort: void
  process.exit 1
```

One catch is that I removed the ability to have newlines between the function name and the opening parenthesis of the arguments.  Otherwise we have issues with tracking indentation but also the following ambiguity:

```coffee
function f
  (x)
```

I'd say this is an argumentless function `f()` that returns `x`.  But currently it parses as the TS function declaration `function f(x)`.

Note that JS does allow

```js
function f
  (x)
{}
```

so this is technically an incompatibility with JS... Perhaps there should be a toggle?

### DIFF
--- a/source/parser.hera
+++ b/source/parser.hera
@@ -687,7 +687,8 @@ Parameters
     return {
       type: "Parameters",
       children: [{$loc, token: "()"}],
-      names: []
+      names: [],
+      implicit: true,
     }
 
 NonEmptyParameters
@@ -1037,15 +1038,17 @@ FunctionDeclaration
 
 FunctionSignature
   # NOTE: Merged in async and generator with optionals
-  ( Async __ )? Function Star? BindingIdentifier?:id __ NonEmptyParameters:parameters ReturnTypeSuffix?:suffix ->
+  ( Async __ )?:async Function:func ( __ Star )?:star ( __ BindingIdentifier )?:wid _?:w Parameters:parameters ReturnTypeSuffix?:suffix ->
     return {
       type: "FunctionSignature",
-      id,
+      id: wid?.[1],
       parameters,
       returnType: suffix?.children?.[1]?.[0]?.[1]?.token,
       ts: false,
       block: null,
-      children: $0,
+      children: !parameters.implicit ? $0 :
+        [ async, func, star, wid, parameters, w, suffix ],
+        // move whitespace w to after implicit () in parameters
     }
 
 # https://262.ecma-international.org/#prod-FunctionExpression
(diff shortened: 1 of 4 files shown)


source/parser.hera
```diff
@@ -1037,15 +1038,17 @@ FunctionDeclaration
 
 FunctionSignature
   # NOTE: Merged in async and generator with optionals
-  ( Async __ )? Function Star? BindingIdentifier?:id __ NonEmptyParameters:parameters ReturnTypeSuffix?:suffix ->
+  ( Async __ )?:async Function:func ( __ Star )?:star ( __ BindingIdentifier )?:wid _?:w Parameters:parameters ReturnTypeSuffix?:suffix ->
```
STRd6, line 1041: `BindingIdentifier` already includes `__` inside its productions.

edemaine: Ah, indeed. But we actually need a `_?` version, to support anonymous functions.

It'd be nice to have consistent names for rules that have or don't have whitespace at the start... The current grammar is a bit of a mix.  Maybe a leading `W` or something?

STRd6:
With this change both of these are equivalent:

```
function f 5
f := -> 5
```

I think that's fine and some people may prefer one to the other. `function f 5` looks a little weird to me but I understand how it makes sense and follows logically from optional empty parameters. The ones with nested blocks look good though and it wouldn't make sense to arbitrarily disallow the inline version.

Will someone prefer this style to the other? Probably. Will someone go nuts and annoy their teammates? Also probably, but that is true for many Civet features :)

source/parser.hera
STRd6: Definitely worth exploring some conventions now that the parser is getting large.

edemaine:
Yeah, I had similar thoughts. I was going for the indented block case, where the parens really seemed unnecessary. The one-line tests were just seeing what it could do, and maybe shouldn't be encouraged e.g. via docs. It's particularly weird that `function 5 f` is equivalent to `-> 5(f)` while `function f 5` is equivalent to `function f { 5 }`.

I wonder of we should introduce `then` for the one-line case, like `if` and such, so it's clear when you enter the body. Should we require it? Less clear.

STRd6:
I think `5 f` -> `5(f)` is a separate bug which might be fixed by making spaced application `$skip` on literals in the LHS.

`then` but perhaps spelled as `returns` might make the one-liner look more readable.

PR #164, opened 2023-01-08
Create directory in `-o` option if it doesn't exist

Milo123459:
For example, if `civet ... -o dist/.ts` was used and the `dist` directory didn't exist, this PR makes it so that Civet automatically creates it.


edemaine:
Your code had a few typos, and I generalized it to support things like `civet -o dist/`. Assuming you're happy with it, I'll merge.

Milo123459:
yup, thank you!

PR #167, opened 2023-01-08
Function with `then` body

edemaine:
Following up on https://github.com/DanielXMoore/Civet/pull/159#issuecomment-1374862084, this PR extends `BracedBlock` to allow for a `ThenClause`. This seems symmetric and removes ambiguity for some one-line `function` definitions: `function f` looks like a named function, while `function then f` is clearly an anonymous function that evaluates to `f`.

I also added `then` to the list of keywords. Otherwise `function then` looked like a named function too!

As @STRd6 suggested, I could add a special `returns` but it might be overkill... And what if `returns` is an identifier and you're intending to call it?  Like `function f returns 5` could be construed as `function f { returns(5) }`.  Whereas `then` is consistent with `if`, `when`, etc., and I believe should already be treated as a keyword. (CoffeeScript does.)

I believe this change also allows `do then` and `try then` but that seems fine to (optionally) allow — arguably more symmetric?

### DIFF
--- a/source/parser.hera
+++ b/source/parser.hera
@@ -1148,7 +1148,7 @@ Block
     }
 
 ThenClause
-  Then TrailingComment*:ws Statement:s  ->
+  Then TrailingComment*:ws Statement:s ->
     const expressions = [[ws, s]]
     return {
       type: "BlockStatement",
@@ -1203,6 +1203,8 @@ BracedBlock
       // Remove !EOS assertion
       children: [o, s, ws, c],
     }
+  InsertOpenBrace:o ThenClause:then InsertSpace:ws InsertCloseBrace:c ->
+    return {...then, children: [o, ...then.children, ws, c]}
 
 SingleLineStatements
   ( TrailingComment* Statement ):first ( SemicolonDelimiter Statement? )*:rest ->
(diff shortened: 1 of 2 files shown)


STRd6:
Maybe only allow `ReturnStatement` for the inline function body? That way it will enforce the use of `return` keyword and we won't need to expand the keyword list to include `then`.

If we do want to keep `then` we can adjust the parse order to check for ThenClause in the function context without necessarily adding a keyword.

I'm not sure about expanding `BracedBlock` to include `ThenClause`. `ThenClause` was primarily for CoffeeScript compatibility but I'm also not opposed to expanding and making it more consistent.

edemaine:
`return` is cool for functions that return a value, but less good if you wanted a void return... I like the idea of using existing keywords. 

`do` also comes to mind, once we have `do` expressions:

```coffee
function f do console.log 'hello';
```

(the `;` avoids returning a value)

Or

```coffee
function f: void do console.log 'hello'
```

`do` sounds descriptive: it's what the function does :) And top-level do is equivalent to no do...

STRd6:
`do` might also be a good generic block syntax and possible replacement for `then`...

```
if x do y else z

do
  x := 3

x = do
  // more block stuff ...

function f do console.log x
```

They all have a block-like nature to them. Then when we add full `BlockExpressions` we can handle assignment with `do` blocks on an rhs.

Mainly just thinking out loud here.

STRd6:
I don't want to merge this if it adds `then` as a reserved word.

Using `do` for this is 👍 for me.

```
function f do console.log x
```

edemaine:
It turns out that the recently added `do` expressions magically added support for one-line `function`s (though it adds a second layer of braces):

```coffee
function f do console.log x
---
function f()  {{ return console.log(x) } }
```

I'll just add some documentation and tests around this, but this PR can be decommissioned.

PR #168, opened 2023-01-08
Fix array parsing bug

### LINKED ISSUE #147: Array parsing bug
```coffee
[
  1
  y: 2
  z: 1
  w: 2
]
```

currently compiles to

```js
[
  1,
  {y: 2({
  z: 1,
  w: 2,
})}
]
```

[Playground](https://civet-web.vercel.app/#WwogIDEKICB5OiAyCiAgejogMQogIHc6IDIKXQ==)

Reported by Oren

edemaine:
Fixes #147 by adding a no-cache rule.

I'm not totally sure why this works (or rather why it wasn't working) but it does. 😅


STRd6:
Good fix! I'm not 100% sure either but it may be that when it recurses into `ExtendedExpression` the following property gets cached with an incorrect indent level.

I'll need to get those Hera updates so we can automate all the caching!

PR #171, opened 2023-01-08
Fix inline ifs

### LINKED ISSUE #162: One-line `if` statement transpiles incorrectly
`if coffee or relaxed then code() else sleep()` compiles to `if (coffee || relaxed) code() else sleep()` which is apparently invalid JS; it needs to be `if (coffee || relaxed) code(); else sleep()` (or have a newline before the `else`).

Milo123459:
Fixes #162


STRd6:
Thanks, great first PR!

PR #173, opened 2023-01-09
Try expression

### LINKED ISSUE #128: `try` expression
```coffee
y = try x()
```

```javascript
y = (function() { try { return x() } catch {}})()
```

aleksrutins:
Fixes #128

E.g.

```civet
foo := try potentiallyErroringThing() catch(e) {...} finally {...}
```

### DIFF
--- a/source/parser.hera
+++ b/source/parser.hera
@@ -39,6 +39,7 @@ ExpressionizedStatement
   IterationExpression
   SwitchExpression
   ThrowExpression
+  TryExpression
 
 # https://262.ecma-international.org/#prod-Expression
 Expression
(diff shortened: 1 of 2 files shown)


source/parser.hera
```diff
@@ -5677,6 +5686,16 @@ Init
       }
     }
 
+    function processTryExpressions(statements) {
+      if (module.config.implicitReturns) {
+        // Add returns to TryExpressions
+        gatherRecursiveAll(statements, n => n.type === "TryExpression")
+        .forEach(({ block }) => {
+          insertReturn(block)
```
STRd6, line 5694: The catch block also needs returns inserted (if present).

STRd6:
Looks great overall! You just need to insert returns in catch blocks if present.

Thanks for the PR!

source/parser.hera
aleksrutins: This should be done at this point.

PR #188, opened 2023-01-10
Moved :: behind coffeePrototype config flag

### LINKED ISSUE #187: Deprecate `::` shorthand for `.prototype.` behind `coffeePrototype` flag
I'd propose that we hide the `.prototype.` shorthand `::` behind a `coffeePrototype` flag. `prototype` is a pretty rare thing to use these days, and `::` is valuable real estate, which is shaping up to a useful secondary typing operator.

We should keep an option for `::` for CoffeeScript compatibility.

_Originally posted by @edemaine in https://github.com/DanielXMoore/Civet/discussions/126#discussioncomment-4648615_

STRd6:
Fixes #187

### DIFF
--- a/source/parser.hera
+++ b/source/parser.hera
@@ -662,8 +662,8 @@ PropertyAccess
       children,
     }
 
-  # NOTE: Added CoffeeScript :: prototype shorthand
-  DoubleColon:p IdentifierName?:id ->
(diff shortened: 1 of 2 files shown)


edemaine:
Can you add mention of this to the README?

PR #201, opened 2023-01-11
Fix identifier used only in coffee for loop

### LINKED ISSUE #193: Bug in coffeeFor for loop generation code
```coffee
"civet coffeeCompat"
for i in [0...n]
  5
```

transpiles to

```js
var i
for (let i = i = 0, asc = 0 <= n; asc ? i < n : i > n; i = asc ? ++i : --i) {
  5
}
```

which crashes with `i used before assignment`.  If `i` is used within the body, it works: the second instance of `i` gets renamed to `i1`.  So maybe an issue with the ref mechanism?

edemaine:
Fixes #193

`CoffeeForDeclaration` was actually pretty broken: it modified the AST without duplicating the object! But it also removed the `Identifier` type.

Also fix a bug where CLI in `--ast` mode wasn't showing the full AST (just the first couple of levels like `console.log` does).

### DIFF
--- a/source/cli.coffee
+++ b/source/cli.coffee
@@ -139,7 +139,9 @@ repl = (options) ->
         when options.compile then '🐈> '
         else '🐱> '
     writer:
-      if options.compile and not options.ast
+      if options.ast
+        (obj) -> JSON.stringify obj, null, 2
+      else if options.compile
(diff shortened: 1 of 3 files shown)


STRd6:
Thanks!

Those coffee compat loops are definitely rough in places 😅

PR #207, opened 2023-01-11
Fixes #205

### LINKED ISSUE #205: Another implicit call bug
```coffee
function test(a)
  a = simplify a
  (x) => x+a
```

[transpiles to](https://civet.dev/playground?code=ZnVuY3Rpb24gdGVzdChhKQogIGEgPSBzaW1wbGlmeSBhCiAgKHgpID0%2BIHgrYQo%3D)

```js
function test(a) {
  return (a = simplify(a)(x)(() => x + a));
}
```

I expected:

```js
function test(a) {
  a = simplify(a)
  return (x) => x + a;
}
```

STRd6:
There may still be some caching related improvements but adding this negative lookahead prevents the parens of an arrow function on the following line from being interpreted as arguments in a call expression.

When I added no-cache to many rules the parser became too slow. Probably need to do some Hera and caching improvements in the future to really optimize.

### DIFF
--- a/source/parser.hera
+++ b/source/parser.hera
@@ -79,8 +79,8 @@ IndentedApplicationAllowed
 
 ArgumentsWithTrailingMemberExpressions
   # Since this is recursive Arguments must consume input to avoid infinite recursion
-  # NOTE: Do not allow trailing template literals to match
-  Arguments ( Samedent MemberExpressionRest )*
+  # NOTE: Assert "." to not match "?" or "!" as a member expression on the following line
+  Arguments ( Samedent &"." MemberExpressionRest )*
 
 # https://262.ecma-international.org/#prod-ArgumentList
 ArgumentList
--- a/source/parser.hera
+++ b/source/parser.hera
@@ -80,7 +80,7 @@ IndentedApplicationAllowed
 ArgumentsWithTrailingCallExpressions
   # Since this is recursive Arguments must consume input to avoid infinite recursion
   # NOTE: Do not allow trailing template literals to match
-  Arguments ( Samedent !Backtick CallExpressionRest )*
+  Arguments ( Samedent !Backtick !ArrowFunction CallExpressionRest )*
 
 # https://262.ecma-international.org/#prod-ArgumentList
 ArgumentList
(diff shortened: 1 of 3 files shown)


source/parser.hera
```diff
@@ -80,7 +80,7 @@ IndentedApplicationAllowed
 ArgumentsWithTrailingCallExpressions
   # Since this is recursive Arguments must consume input to avoid infinite recursion
   # NOTE: Do not allow trailing template literals to match
-  Arguments ( Samedent !Backtick CallExpressionRest )*
+  Arguments ( Samedent !Backtick !ArrowFunction CallExpressionRest )*
```
edemaine, line 83: Hmm, I don't quite see why `CallExpressionRest` is what we want here. Certaintly `MemberExpressionRest` makes sense with

```coffee
foo
.bar()
```

But why do we want a call on a future same-dented line?
```coffee
foo
(bar)
```
I feel like this should compile to `foo; bar` but it currently compiles to `foo(bar)`.

So I'd propose:

```suggestion
  Arguments ( Samedent MemberExpressionRest )*
```

Oh, does this still allow the member expression to be called?

STRd6: Maybe it should be `!"("` for the lookahead since that will cover parenthesized expressions as well and arguments on the following line should be prohibited/discouraged.

STRd6: I think you are right about `MemberExpressionRest`. The additional `CallExpressions` get handled because `ArgumentsWithTrailingCallExpressions` is always either followed by a `CallExpressionRest*` or inside `CallExpressionRest`. I've got a change matching your suggestion incoming.

edemaine: Could `!Backtick !"("` be replaced by a `&"."` assertion? Equivalent?

edemaine: Oh, I see you switched to trailing `MemberExpression`; that makes me happy.

STRd6: I also added the `&"."` assertion so that shorthand `?` and `!` won't match as `MemberExpressions` when on the following line.

PR #208, opened 2023-01-11
Docs: fix URI to base64 translation

Argeento:


edemaine (approved):
Nice catch and fix! I tested and it works well.

Unfortunately, I found unpkg.com to be down yet again. I've switched to jsdelivr.net; OK?

edemaine:
@STRd6 Does Build Docs need credentials to not hit rate limits?

STRd6:
@edemaine yes, take a look at `.github/workflows/static.yml env:`

We may need to set the `build-docs` environment to `gh-pages` to get access to the token (or a different environment with a token that has fewer permissions)

PR #211, opened 2023-01-12
`return` from expressionized blocks

STRd6:
Started cleaning up implicit return insertion.

Made `SingleLineStatements` match `[ws, statement, delimiter][]` interface.

Cleaned up `DefaultClause` and `ElseClause` block handling.

Added returnSymbol ref declaration.


edemaine (commented):
Sorry for the delay on reviewing this. These look like some really nice improvements! Much less special-casing toward `return`, and important cases like `switch`-in-`for`-via-`push` working.

PR #224, opened 2023-01-12
Fixes #221

### LINKED ISSUE #221: Multiple divisions are not RegExp
`Math.floor x / y / 2` parses as a regular expression instead of two divisions.

I think usually a space after `/` means "not a regular expression".

STRd6:


edemaine (changes requested):
I find this a pretty obscure way to fix the issue.  I'd rather see `RegularExpressionBody` follow https://262.ecma-international.org/#prod-RegularExpressionFirstChar which it seems to be trying to do; I think space just needs to be in the `![*\/\r\n]` negative lookahead?

STRd6:
The ambiguity comes from the interaction with the implicit function application since:

```coffee
x / 7 / 2
```

Get's interpreted as an implicit function call with a RegExp as the first argument (explicitly calling itself with 2 as an arg, a possible separate issue :P ).

This is valid JS:

```javascript
x = / /
```

The `RegularExpressionFirstChar` from the spec disallows line breaks (we probably should too but it's a also a different issue than this one).

edemaine (approved):
Ah, I understand now. `/` is a special character in that it can be a binary operator (verb) *and* the beginning of a literal (noun).

It looks like `//` isn't a valid regular expression, and non-heregexes can't have newlines, so I think I now believe this is the correct exclusion.

Sorry for the confusion and thanks for the explanation!

source/parser.hera
```diff
@@ -547,6 +547,8 @@ NonNullAssertion
 # ie: the 'of' in 'for x of ...'
 AdditionalReservedWords
```
edemaine, line 548: I do wonder whether there's a better name than `AdditionalReservedWords`, like `ForbiddenImplicitCalls` or something. There's no reference to the other `ReservedWord` rule so it's not really additional.

STRd6: I'll update the name.

STRd6:
No worries, the JS spec is vast and there is a lot to keep track of.

PR #225, opened 2023-01-13
Fixes #218. Optional type void => shorthand

### LINKED ISSUE #218: Argumentless => in types
I was trying to convert this TS code to Civet
```
interface Array<T> {
    listen(fu: ()=>void): void
}
```

but doing it like:
```
interface Array<T>
    listen(fu: =>): void
```

results in a `Failed to parse` message.

If we lose `: void` 
```
interface Array<T>
    listen(fu: =>)
```

it transpiles successfully, but produces the following:

```
interface Array<T> {
    state(): Array<T>
}
    listen({fu: () => {}})
```

Erik on Discord had the idea that maybe `=> void` (omitting parens when zero arguments) could be added to the type system.
https://discord.com/channels/933472021310996512/933472021310996515/1063137264554737684

STRd6:


edemaine (commented):
This seems correct, though I find the notation a little uncomfortable. It makes me worry about whatever coming after `=>` being treated as part of the type. But the tests show reasonable examples where it's not: following `)` or `,`s.  So I guess it's cool, and the parser will parse the natural way.

I *thought* the suggestion was to allow types like `=> number`, but I see that that already works.

edemaine (approved):
Also, the more I think about this, the more it matches our existing function shorthand (which can have an empty block), so seems like a good idea.

I'm starting to wonder what `->` type could mean...

PR #226, opened 2023-01-13
Fixes #160

### LINKED ISSUE #160: Crashes with import types
This type can't be used inside a .civet file:

```ts
type CompilerOptions = import("monaco-editor")
```

Assume it's  just not been added yet:

```
unknown:1:30 Failed to parse
Expected:
Dot "."
Whitespace /[\s]+/
JSMultiLineComment "/*"
```

STRd6:


edemaine (commented):
Looks good! A couple comments.

I'm a little confused why dot access is part of `TypePrimary` instead of a higher-level thing, like property access is.  But maybe that's how types work. In any case it's a separate issue.

PR #229, opened 2023-01-13
Move autolet to nearest statement

IamIpanda:
Hope that's the last PR for `auto-let` :-)

### What changed in this PR
Move `let` expression to just before the `statement` who contains the `AssignmentExpression`.    
The term `statement` here means the direct child of a `BlockStatement`.    
When the `statement` contains only one child which is an `AssignmentExpression` with only one variable name, *and not a ObjectBinding*, the `let` will be embedded.    
Added tests is a duplicate for `auto-var.civet`, but `let` version.

#### Why must iterate over the blocks, not just scan and process like `auto-var`
```
{ a = 2 }
a = 3
```

#### Why not to put the let declaration inner assignment, but before statement?
```
let a = (b = 2)
```

#### Why not hoist `let` to top of the block
```
// This code should not compile
if(a) { return 1 }
a = 2
```
When `let` a variable, it claims the declared names don't exist before the let statement, hoist will make unsafe code compiles.

#### Is insert before the statement just safe?
No. But it may be the best we can do.
```
// This code should not compile
console.log(a, a = 2)
```

#### Why embedded let can only declare one variable 
```
a = b = 2
```

#### Why change `ExportVarDec`
`auto-let` need to collect `var` declared variables by declaration to move them out from processing. For `ExportVarDec`, it don't contains the `var` token inner `Declaration`, and collector can't see them, That will make the test `export implicit let dec` fail.

#### Why Object Binding cannot use embedded let? <- **NEW!**
There will be a `()` inner ObjectBinding before `insertLetDec` step, so it will be:
```
{compile} = require 'civet'
↓↓↓
let ({compile} = require("civet"))
```
It would be quite complicated to squeeze the let inner parens, so I think disable it is reasonable.

#### More tests
+ Tested in my personal project.
+ Tested in full `coffeescript` , seems all let are in position where it should be.

#### What I still see
+ Copy indent for `let` declaration may lead to mapping wrong.
+ Bare blocks will hoist variable scope, may(though it's not right) be not work as expected (See test case `auto let inner if parens`)


source/parser.hera
```diff
@@ -2946,7 +2946,9 @@ ExportDeclaration
 
 # CoffeeScript style `export x = 3` -> `export var x = 3`
 ExportVarDec
-  InsertVar VariableDeclarationList
+  InsertVar VariableDeclarationList ->
+    $2.children.unshift($1)
```
STRd6, line 2950: This mutation isn't safe with the parser's caching. Instead use:
```javascript
return [$1, ...$2]
```

STRd6:
Looks great, thanks!

PR #231, opened 2023-01-13
Fixes #197

### LINKED ISSUE #197: Trailing call expressions should work following an identifier
```coffee
testCase """
  leading . after plain identifier
  ---
  data
  .map (x) => x + 1
  .forEach (x) => console.log
  ---
  data
  .map((x) => x + 1)
  .forEach((x) => console.log)
"""

testCase """
  nested leading . after plain identifier
  ---
  data
    .map (x) => x + 1
    .forEach (x) => console.log
  ---
  data
    .map((x) => x + 1)
    .forEach((x) => console.log)
"""
```

From @orenelbaum and @Argeento in Discord.

STRd6:
Suppress newline delimited trailing member properties inside of implicit function arguments. This makes use of a state stack to be robust to nested contexts and backtracking.

Moved hacky indentation check out of MemberExpressionRest.

Added NonSuppressedTrailingMemberExpressions to CallExpression.

This works and I think I almost understand it.


STRd6:
It may be possible to simplify a bit further if we make

This:
```coffee
f x
  .y
```

Transpile to:
```javascript
f(x)
  .y
```

Instead of:
```
f(x
  .y)
```

Which would be more compatible with CoffeeScript.

We may even be able to drop all the state tracking for trailing member properties.

source/parser.hera
```diff
@@ -80,15 +85,17 @@ ForbiddenImplicitCalls
...
-    return
-
 ArgumentsWithTrailingMemberExpressions
-  # Since this is recursive Arguments must consume input to avoid infinite recursion
+  Arguments TrailingMemberExpressions
+
+TrailingMemberExpressions
   # NOTE: Assert "." to not match "?" or "!" as a member expression on the following line
-  Arguments ( Samedent &"." MemberExpressionRest )*
+  MemberExpressionRest* ( ( Samedent / IndentedFurther ) &"." MemberExpressionRest )* ->
```
edemaine, line 93: Personally I think `?.` and maybe even `!.` could make sense as a trailing member expression.

source/parser.hera
```diff
@@ -529,14 +536,23 @@ CallExpression
...
         type: "CallExpression",
-        children: [$1].concat($2),
+        children: [member, ...trailing, ...rest],
       }
     }
-    return $1
+
+    if (trailing.length) {
+      return {
+        children: [member, ...trailing]
```
edemaine, line 550: I think this should be a `type: "MemberExpression"`.

You could plausible combine the two cases into `if (rest.length || trailing.length)` and have `type: rest.length ? "CallExpression" : "MemberExpression"`.

edemaine (approved):
Looks great!  A couple comments.  I also tested whether this worked inside parentheses; not sure it's worth another test.

```
🐈> hello world, (
...   hello world
...   .foo bar
... )
... .foo bar
```

source/parser.hera
STRd6: I think the preferred place for `!` would be as a suffix after the identifier but `?.` makes sense to me.

PR #232, opened 2023-01-13
Fix CLI behavior especially on Unix

### LINKED ISSUE #217: Building is broken from "scripts"
If I run
`./node_modules/@danielx/civet/dist/civet test.civet`
everything is **fine**.

But if I add a script in `package.json`:
```
"scripts": {
     "test": "civet < ./test.civet > ./test.ts"
}
```
and try to run it with
`npm run test`

I get:

```
/dev/fd/test.civet failed to load:
[Error: ENOENT: no such file or directory, open '/dev/fd/test.civet'] {
  errno: -2,
  code: 'ENOENT',
  syscall: 'open',
  path: '/dev/fd/test.civet'
}
```

edemaine:
Fixes #217. Tested on MacOS.

* Was previously overwriting `filename` and then trying to read it
* Use `readline` exactly in TTY mode
* Outside TTY mode, new "chunk" reader to read directly from `process.stdin`
* Re-enable `-` command-line option for reading stdin (was previously disabled by helpful error message)

### DIFF
--- a/source/cli.coffee
+++ b/source/cli.coffee
@@ -87,7 +87,7 @@ parseArgs = (args = process.argv[2..]) ->
       when '--'
         endOfArgs ++i  # remaining arguments are filename and/or arguments
       else
-        if arg.startsWith '-'
+        if arg.startsWith('-') and arg != '-'
           throw new Error "Invalid command-line argument #{arg}"
         if options.run
           endOfArgs i  # remaining arguments are arguments to the script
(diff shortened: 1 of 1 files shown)


STRd6 (approved):
Looks good!

edemaine:
Just tested on Windows too and all cases seem good. 😃 

We could definitely use those CLI tests... On GH Actions we can run them on all 3 operating systems.

Issue #233, opened 2023-01-14
Dedent to continue function arguments


edemaine:
```coffee
setTimeout =>
    console.log 'test'
, 100
```

I expect this to transpile to

```js
setTimeout(() =>{
    return console.log 'test'
}, 100)
```

but it transpiles to 

```js
setTimeout(() => {
    return console.log("test", 100)
})
```

PR #234, opened 2023-01-14
Fixes: #233. Dedented commas should not delimit

STRd6:


edemaine (approved):
Looks great!  I wonder about adding some of the following tests (which I checked and all work, and you've explicitly checked for):

```coffee
setTimeout =>
  console.log 'test'
  , 100
```

```coffee
setTimeout =>
    console.log 'test'
 , 100
```

Or maybe some tests like this already exist; I didn't check.

PR #235, opened 2023-01-14
New fast JSX parser

edemaine:
* In default JSX parsing mode, JSX children must be properly indented.
* `coffeeJSX` removes this constraint, but then requires that all JSX tags are explicitly closed (as in CoffeeScript)
* Now should be linear time in indented mode (1 zillion times faster!)
  * On my particular example of 25ish tags, compile time went from >6 seconds to ~0.2 seconds.


source/parser.hera
```diff
@@ -3864,44 +3864,93 @@ JSXTag
...
-  # NOTE: Auto-closing element with indented children
-  JSXOpeningElement:open JSXMixedChildren:children InsertNewline InsertIndent ->
-    return {
-      type: "JSXElement",
-      children: children.jsxChildren.length ? [...$0, ["</", open[1], ">"]]
-                                            : [open.slice(0, -1), " />"]
+  # NOTE: In default JSX mode, JSX children must be properly indented,
+  # and closing tags are optional (but still allowed).
+  !CoffeeJSXEnabled PushJSXOpeningElement:open JSXMixedChildren?:children JSXOptionalClosingElement:close ->
+    module.JSXTagStack.pop()
```
STRd6, line 3870: We might want to omit the assertion with `$0 = $0.slice(1)` above this line.

I think a later version of Hera will omit assertions from the results array but that's in the future.

STRd6:
Looks good!

STRd6 (approved):

source/parser.hera
edemaine: Oh, I assumed assertions were omitted. Thanks! I'll fix that in a bunch of places.

PR #239, opened 2023-01-16
wip: pattern matching rules

STRd6:
Starting on an actual pattern matching implementation based on the discussion: https://github.com/DanielXMoore/Civet/discussions/150

- Create parallel `MatchingPattern` rules similar to `BindingPattern`
- Use them inside `switch` statements or as part of a TBD "match operator"
- Match based on structure, regexp, inheritance, relational (`> 3`, `!= null`)
- Attach bindings to other properties
- Generate `if` conditionals with `const` bindings in blocks

```typescript
  switch obj
    [code is 200, result]
      // obj is array-like of length 2
      console.log(code, result) // code matched 200 and was bound, result was also bound
    {code: 200, result}
      // obj has properties code and result
      // code is 200 and is available in this block
      // result is bound in this block
    <? Result and {value}
      // obj is an instance of Result
      // value property of obj is bound in this block
    <? Error
      // obj is an instance of error
```

Lots more possibilities but those are the broad strokes.


STRd6:
Still not complete, but I want to get it into main because it has some related cleanup and so we don't diverge too much.

test/switch.civet
```diff
@@ -222,3 +238,112 @@ describe "switch", ->
...
+      basic
+      ---
+      switch x
+        1
+          console.log "it's one"
+        2
+          console.log "it's two"
+      ---
+      if(x === 1) {
+          console.log("it's one")}
```
edemaine, line 253: Note for future improvement: Not a fan of the brace placement here. Maybe `InsertNewline`?

test/switch.civet
```diff
@@ -222,3 +238,112 @@ describe "switch", ->
...
+      nested array pattern
+      ---
+      switch x
+        [a, [b, c]]
+          console.log a, b, c
+      ---
+      if(Array.isArray(x) && x.length === 2 && Array.isArray(x[1]) && x[1].length === 2) {
+          const [a, [b, c]] = x;
+          console.log(a, b, c)}
+    """
```
edemaine, line 349: Missing test cases for:
* `ObjectMatchingPattern`
* `RegularExpressionLiteral`

But maybe these only exists in the grammar and are not yet implemented.

edemaine (approved):
I just did a quick review, but this seems good to me! Exciting to see all this.

test/switch.civet
STRd6: Yep, not yet implemented.

test/switch.civet
STRd6: I was leaning towards fewer `InsertNewline`s to keep the output closer to the input but I could go either way.

PR #241, opened 2023-01-16
Wrap object expressions with parentheses at statement level

STRd6:
We now wrap `ObjectExpression` at the statement level with parentheses. Sometimes this results in an implicit `return` returning an unnecessarily wrapped object but we can optimize that further later.


STRd6:
> I think you can update the comment `// NOTE: currently not processing any non-binding pattern ObjectExpression or ArrayExpressions` in `processAssignments`.

That comment is more related to https://262.ecma-international.org/#sec-assignment-operators-static-semantics-early-errors about the LeftHandSide of an Assignment. It should throw errors if not a binding pattern.

> This makes me wonder whether we should extend `testCase` to confirm that the output is valid JS. I think `new vm.Script` can compile but not eval a given piece of code.

We probably should, even just `Function(src)` will parse it. The challenge is that it will only be reliable in `--js` mode or if we aren't using any TS features.

PR #245, opened 2023-01-16
Import assertions

### LINKED ISSUE #243: Import Assertions
https://v8.dev/features/import-assertions

edemaine:
Fixes #243

* Add support for static `import` with assertions, including object literal shorthand.
* Dynamic import assertions already worked; add tests.

### DIFF
--- a/source/parser.hera
+++ b/source/parser.hera
@@ -2902,7 +2902,7 @@ FromClause
 # https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-5.html#import-assertions
 # NOTE: Forbid newline before `assert` which could mean a function call
 ImportAssertion
-  _? "assert" NonIdContinue __ ObjectLiteral
+  _? "assert" NonIdContinue _? ObjectLiteral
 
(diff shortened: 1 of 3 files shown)


source/parser.hera
```diff
@@ -2898,6 +2898,12 @@ NamedImports
 FromClause
   From __ ModuleSpecifier
 
+# https://github.com/tc39/proposal-import-assertions
+# https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-5.html#import-assertions
+# NOTE: Forbid newline before `assert` which could mean a function call
+ImportAssertion
+  _? "assert" NonIdContinue __ ObjectLiteral
```
STRd6, line 2905: Using `_*` instead of `__` should allow nested object literals to work here.

edemaine: I notice you use `_*` in a few places but `_?` seems sufficient, as `_` already has a `+` in it.

STRd6: `_?` is sufficient and probably leads to a shallower parse tree so maybe we should tend towards that.

PR #246, opened 2023-01-17
Re-order docs to ease people into the Civet mindset

STRd6:
Emphasizing the "better ES" features first. The goal is to ease
people into the language concepts by first presenting things that
most people feel are obvious, if minor improvements. Then as they
read further present more cool/powerful ideas. They might not
notice the shorthand until it is too late... and then down comes
the net.

### DIFF
--- a/civet.dev/cheatsheet.md
+++ b/civet.dev/cheatsheet.md
@@ -18,24 +18,113 @@ The TypeScript code on
 
 [[toc]]
 
+## TC39 Left a Lot on the Table
+
+### Default to `const` for Iteration Items
+
+<Playground>
+for (item of [1, 2, 3, 4, 5]) {
+  console.log(item * item);
+}
+</Playground>
+
+### Block Strings / Templates
+
+[TC39 Proposal: String Dedent](https://github.com/tc39/proposal-string-dedent)
+
+<Playground>
+text = """
+  This text is a string that doesn't include the leading
+  whitespace.
+"""
+</Playground>
+
+<Playground>
+text = ```
+  Also works for
+  templates!
+```
+</Playground>
+
+### Modulo Operator
+
+<Playground>
+let a = -3
+let b = 5
+let rem = a % b
+let mod = a %% b
+console.log rem, mod
+</Playground>
+
+### Spread in Any Position
+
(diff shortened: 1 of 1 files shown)


civet.dev/cheatsheet.md
```diff
@@ -18,25 +18,114 @@ The TypeScript code on
 
 [[toc]]
 
-## Basics
+## TC39 Left a Lot on the Table
```
edemaine, line 21: This heading seems a little ... controversial? I wonder if it would be better to be positive with something like "Making JavaScript More Convenient"?

civet.dev/cheatsheet.md
```diff
@@ -18,25 +18,114 @@ The TypeScript code on
...
 
-### Humanize Operators
+### Default to `const` for Iteration Items
```
edemaine, line 23: Should we start with bigger features like pipes instead? Maybe it's better to ease people in, like you say, but thought I'd mention the idea.

I assume it's intentional that you have braces and parens in the for loop here, to focus on just the `const` aspect.

edemaine (approved):
Overall, a big improvement! Just a couple thoughts.

civet.dev/cheatsheet.md
STRd6: Yes, I'm intentionally keeping the first examples very close to "normal" ES.

It might make sense to lead with bigger features but I think that I want the first one to be something that 95% of devs pretty much agree "yeah, that's strictly better". Even pipes was (is still) fairly contentious in the proposal.

civet.dev/cheatsheet.md
STRd6: Controversy can sometimes be good for attention :)

civet.dev/cheatsheet.md
tomByrer: I prefer more basic things are listed on top.

If the Table of Contents is at the top or side, they can 'find' the bigger features readily.  & if you (@edemaine ) think Civet pipes are a 'hidden gem' make blog posts, twitter feeds & videos displaying that feature!
(I'm a big pipes proponent since Gulp)

PR #249, opened 2023-01-17
Fixes #247

### LINKED ISSUE #247: Object pattern destructuring with multiple properties on the same line doesn't work
This doesn't work:
```js
({
  b, c
}: d) =>
  e
```

STRd6:


source/parser.hera
```diff
@@ -834,7 +834,9 @@ ObjectBindingPattern
 ObjectBindingPatternContent
   # NOTE: Added indentation based binding properties
   NestedBindingProperties
-  BindingProperty*:props BindingRestProperty?:rest BindingProperty*:after ->
+  BindingPropertyList*:props BindingRestProperty?:rest BindingPropertyList*:after ->
+    props = props.flat()
+    after = after.flat()
```
edemaine, line 839: It seems like `BindingPropertyList` already has a `+` so each `BindingPropertyList*` could be replaced by `BindingPropertyList?`, right? And then you could omit the `.flat()`.

source/parser.hera
```diff
@@ -906,6 +917,8 @@ NestedBindingProperties
   # NOTE: Add NestedBindingProperties after rest from CoffeeScript
   PushIndent NestedBindingProperty*:props NestedBindingRestProperty?:rest NestedBindingProperty*:after PopIndent ->
     if (!(props.length || rest)) return $skip
+    props = props.flat()
+    after = after.flat()
```
edemaine, line 921: I think each `NestedBindingProperty*` should be `NestedPropertyList?`, given that there's no longer a delimiter in `BindingProperty`? (and omit `.flat()`)

source/parser.hera
```diff
@@ -919,20 +932,21 @@ NestedBindingProperties
     }
 
 NestedBindingProperty
-  Nested:indent BindingProperty:prop ->
-    return {
+  Nested:ws ( BindingProperty ObjectPropertyDelimiter )+:inlineProps ->
```
edemaine, line 935: Seems like `( BindingProperty ObjectPropertyDelimiter )+` should be replaced by equivalent `BindingPropertyList`. Oh, you're doing something different in the code block. But maybe just the whitespace insertion could be done in this code block, and avoid duplicating code?

edemaine (changes requested):

source/parser.hera
STRd6: This needs the `*` to absorb multiple lines of `Nested`.

edemaine: Maybe I'm missing something, but it seems like (at least in this case) there's no `Nested` being called, it's just a rule with `+` which can't actually consume `EOS`.

STRd6:
Things were getting cray cray so I decided to just refactor it into a less crazy way.

- Moved BindingRestProperty into BindingProperty
  - No longer enforcing exactly one rest property at the parser level (letting TSServer handle the error notices)
  - Moved delimiter checking into container rules
- Moved all reordering logic into `module.reorderBindingRestProperty`

source/parser.hera
```diff
@@ -5688,20 +5684,49 @@ Init
...
+      props.forEach(({type}, i) => {
+        if (type === "BindingRestProperty") {
+          if (restIndex < 0) restIndex = i
+          restCount++
+        }
+      })
 
-        rest = {...rest, children: [...rest.children.slice(0, -1), lastDelim]}
-        after = [...after.slice(0, -1), {...lastAfterProp, children: [...lastAfterChildren.slice(0, -1), restDelim]}]
+      if (restCount === 1) {
```
edemaine, line 5699: If you wanted to, you could throw a parse error when `restCount > 1`. Maybe beneficial? But not a big deal.

edemaine (approved):
Oh wow, this is so much cleaner! Looks good to me; one idea for better feedback.

PR #250, opened 2023-01-17
Civet loader for Bun (integration needs fixing)

zolomatok:
@STRd6 

I spent a lot time with it, but couldn't get the integration working. I feel like I'm out of my depth. I have very little xp with npm packages and bundlers and I think the crux of the issue is in that area.

The loader itself works 100% (I tested it separately), but when I try to import it from the Civet repo, `bun` gets stuck on some code in that repo. But when I use the loader as-is, as just a file in my own project (albeit a TS version), it works no problem.

I wager you guys can see what the problem is more apparently. Feel free to give pointers and reject the PR, or try the integration yourself.

Alternatively, you can just copy the loader out of the PR, convert it to TS and try it as is. Just create a TS file like `test.ts`, write some code like:

_test.ts_
```
import "./bun-civet.ts"
import * as file from './file.civet'
console.log(file.animal)
```

_file.civet_
```
export let animal
let a = ->
    animal = '🐈'
a()
```

then just do `bun test.ts` (plus install Civet in that project ofc)

You can even import other civet files in your civet files, bun applies the loader to every imported `.civet`.

### DIFF
--- a/build/esbuild.coffee
+++ b/build/esbuild.coffee
@@ -83,3 +83,22 @@ esbuild.build({
     heraPlugin
   ]
 }).catch -> process.exit 1
+
+esbuild.build({
+  entryPoints: ['source/bun-civet.coffee']
+  bundle: true
+  sourcemap
+  minify
+  watch
+  platform: 'node'
+  format: 'esm'
+  outfile: 'dist/bun-civet.mjs'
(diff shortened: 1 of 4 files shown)


STRd6:
Fixed in #252 and released as Civet v0.5.47

PR #261, opened 2023-01-21
Change `is in` operator to use `includes` method

edemaine:
As discussed in Discord a while back:

Unlike `coffeeCompat`'s `in` operator, `is in` is not restricted by CoffeeScript backwards compatibility.  By mapping `a is in b` to `b.includes(a)`, we enable the ability for the user to override this operator, while still maintaining compatibility with arrays and `NodeList` which already have `includes` methods. Also, `in` is a prefix of `includes` so it's intuitive.

Breaking change: For strings, `is in` now checks for substrings instead of individual characters.  This is in line with Python's `in` operator (and how it's currently documented in the cheatsheet), but different from CoffeeScript's `in` operator.

### DIFF
--- a/source/parser.hera
+++ b/source/parser.hera
@@ -1959,14 +1959,23 @@ BinaryOpSymbol
       $loc,
       token: "instanceof",
       special: true,
+      negated: true,
     }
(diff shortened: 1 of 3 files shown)


STRd6:
Thanks, looks great!

PR #262, opened 2023-01-21
Change `is` operator to `Object.is` call

### LINKED ISSUE #230: Change `is` operator to `Object.is` call
[Object.is](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/is) is better behaved than `===`. In particular, `x is NaN` reports the same result as `Number.isNaN(x)` (`isNaN(x)` detects other things like `undefined`) and `x is 0` is a bit more sane.

`Object.is` also has the same name as our `is` operator so it feels pretty natural.

[Benchmark](https://jsbench.me/rllcurvdgh/1) shows negligible performance difference (maybe `is` ref is actually faster).

We'd need a `coffeeIs` compatibility flag for CoffeeScript.

edemaine:
Fix #230

`coffeeIs` switches back to old `===` behavior of `is`, and `!==` behavior of `is not`.


source/parser.hera
```diff
@@ -4945,6 +4967,14 @@ Reset
         // [indent, statement]
         module.prelude.push(["", ["const ", hasPropRef, typeSuffix, " = {}.hasOwnProperty", module.asAny, "\n"]])
       },
+      is(isRef) {
+        const typeSuffix = {
+          ts: true,
+          children: [": (a: any, b: any) => boolean"]
```
STRd6, line 4973: We may need an advanced TS assertion to get the same type narrowing that `===` provides.

edemaine: Ouch. I'm not sure that's doable. TypeScript's built-in type for `Object.is` [doesn't even realize type equality](https://www.typescriptlang.org/play?#code/GYVwdgxgLglg9mABMAFAQwFxhAWwEYCmATgDSJ4bgDWYcA7mAJSIDeAsAFCKIzCIoB5PACsC0AHQwAzujJ5Gzdl26IICKVFWIAvOU4rEAekMGAegH59iAL5Xe-NDu275rK9zVgNWl+6MmVCytbDmsgA), and I don't see how `is` assertions could fix this (I might be missing), but moreso I don't see how they could do [type narrowing](https://www.typescriptlang.org/play?#code/JYOwLgpgTgZghgYwgAgBrIN4FgBQz-IhwC2EAXMgESqW4C+uoksiKAmprgYSeVW7RwMcuGAFcQCMMAD2IZDAAUcMugA+yNgEpOeZMBjJFAeQBGAKwhSAdMADOy60VIAaKjS07se-Ajl2wZARkAF5kOC4CAHoo7gA9AH5I4XwDIzgnXlCQsOpKL0jff0DgsIifZBj4pL1hOiA).

So as much as I like this feature... it doesn't seem very TypeScript compatible. 😢

edemaine: [This definition](https://www.typescriptlang.org/play?ts=5.0.0-dev.20230121&ssl=2&ssc=3&pln=3&pc=49#code/MYewdgzgLgBAlhAXAKBjAFAHlJWBBAGhgCEYBTADyjLABMIY8A+dAQ0UaICMPiBKGAF4mMVvAb8YAMlQZs4aCSJ5yVGvRIt2nGDxIDhu8YwMwA8lwBWZYFAB0CUQ1ZgAngG5kyHIqiuADmQc0ABOcGAA5kIwAORgILRkMZ7IcABmGAjofoFEcQlJfAIA3rJoOWSeaGgA9DUwAHoA-MgAvqkZ6Fn5iTFEFUUwpdUwFVXVdY0t7cgANmSwFBxgAK4AtlxkIXMLMK7L65shMAA+MKHhER0YFEKCgnslZTA+sABe0RTPk9XNz68wADu0Vc33qv2m1y6EHQriIFEGw2qAI+Dy+Ix+aD+IwBwIeoIx4Kx0yAA) works in TS5. Should we switch to TS5 internally and require it for our output? Or wait for TS5 to be released?

STRd6: I think we can merge this, then open a new issue to update when TS5 is released.

edemaine: I've added a non-`const` version which [still works pretty well](https://www.typescriptlang.org/play?ts=5.0.0-dev.20230121#code/GYVwdgxgLglg9mABDAzgHgIIBpECFECmAHlAWACYqIYB8iAFAIYBc1OARq7gJSuPJVcAbgBQoSLAQDMhEmUp4cuOk1bZEnPLw0Dqo8dHhJUDFokZgAnhz5Xt7OHAA2BC4gDeIxIgBOBKCA+SADy7ABWBNAAdKhMWOzcIgC+IiIQCChQiFCWAA4ErJk+MGAA5ogAvIgA5GBw5ATVoiIwwAyxOfk4tfWN3NweXt7ZeQSiw4gA9JOIAHoA-MktbfSxPQ3VOJ0E-YMT2+PD03OLKSINEE6MfojpYJmIRKxgIAC27AQ+55FXN3cPlmebw+PkQAB9EEUSqVlu0UPQiDhLLtPBN-lkAF6VR5DbzHYYLXG3DJZADu2MsRPx3kJKVacPo1keKKJ6MQWKqRCpMwJizRJMQ5KqlIm1JOySAA). Then we can add `const` when TS5 is ready.

STRd6: I think it needs another overload for when there is no overlap in types. Also it doesn't intersect union types in the same way `===` does: [TS Playground](https://www.typescriptlang.org/play?ts=5.0.0-dev.20230121#code/MYewdgzgLgBAlhAXAKBjAFAHlJWAVAGhgFUYBTADyjLABMIY8A+dAQ0UaICMPiBKGAF4mMVvAb8YAMlQZs4aJxLkqNeoxbsS3DngHCYXcSQEy06Nh1ZgAnjtG39IgGasANhDJ9ZgmAHkuACsyYCgAOgRRBmsbAG5kZBxFKBsABzIOaAAnODAAcyEYAHIwEFoyIvjkOGcMBHQU9KISsoq+AQBvWTRGsni0NAB6QZgAPQB+ZABfatr0epbyoqJe9pgugZhe-oHhscmZ5DcyWAoOMABXAFsuMiyjk5gbc+vbrJgAHxhs3LzZjAoQkEvhsnW6MCSsAAXoUKOC9gMJuDITAAO6FGzwkaIg7-eYQdB2GAUNYbAYomG+OGbBFoJGbFHokFYzZIw7lYBuVhZMgQhSwACOMA4ACJWCLPjARVwRcgOVyeXzcDB3qKZZKRcBZTUMELgb4smDNgKsfSBvcaYM2Qkdfj0AKiIajQMTZazWgLbsrZNpkA)

edemaine: Good catch. I can get [half of it to work](https://www.typescriptlang.org/play?ts=5.0.0-dev.20230121#code/MYewdgzgLgBAlhAXAKBjAFAHgIIBoYBCAfBgIaIx4wBGFBAlDALwmnwSUwBkhjXqGHPmJkKVWr2Ylq7TjwbcB6dORikwAT3wT1GxixgAzUgBsIAU3oCmMAPLUAVueBQAdAjUddAbmTJQkLBQGgAO5hTQAE5wYADmzDAA5GAgACbmib7IcIYYCOjBYfjJaRn0jADeAmiF5r5oaAD0jTAAegD8yAC+2bno+SXpifi15TBVDTC19Q3NbZ09yOnAJqSR5jAm5rAAHhRgAK4AttTmkUvOq+ub2zAa+8enkTAAPjBRMbG9GDvMTDZ6cbVGABaAwABeCR2wLmDQ6wNBsAA7gkNDCWnCFt9+hB0FoYDsxhMGoiIVD0ZN4ZNSSiARTMd0-MsrhtSQBHGAUABEpC5rxgXOoXIuKzWrPAYOe3KF-K5wGFOQwHP+NkilWBbPRVIa50mzXhi0VOPQbPwavVk01esa2rQutmNs63SAA) but seems like more magic is needed...

edemaine: Here's my proposal:

```ts
const is: {
  <B, A extends B> (a: A, b: B): b is A,
  <A, B> (a: A, b: B): a is A & B
} = Object.is as any;
```

If the second argument is more general, it narrows it.  Otherwise, narrow the first argument in all cases.  Until https://github.com/Microsoft/TypeScript/issues/26916 this seems to be the best we can do.

[Works well in TS5](](https://www.typescriptlang.org/play?ts=5.0.0-dev.20230121#code/MYewdgzgLgBAlhAXDA3gKBjAPAIQDQwCCMApgB5QlgAmEMOAfDABQCGyhBARsjgJTIu8OpwzZO9Jmw7deAmK2FEYAMnpoAvjAC8MAPJcAViWBQAdAgV1WYAJ4BuNGlCRYUWwAcSyaACc4YADmOjAA5GAg1CShjmhwAGYsCMzuXgThkdF8fKhimKkkjpiYAPQlMAB6APyacYnMyRlRoQQF2bnFMAVFxWWVNRpOUcAANqy+JDAjJLBkyGAArgC2XCS+aMNjE1MzMLbzy6u+MAA+MH4BgXUsZDraurY56J1keTBl1W+2bx8D1w0QZhkAiPJ5vV6dX5fH4lT6DBJJQG2AhkdrPYoQ3qwmqdb6Q7G1DYmLaTFzQGAARxgyAARKwaacYDSuDSiaNxqTwOTjrSWYyacBWQjmFT7rpfGDOhSYZ9Out8XCnMLkhSCBLJcVpQqccV5VjPpogA), still [works pretty well in TS4](https://www.typescriptlang.org/play?ts=4.9.4#code/MYewdgzgLgBAlhAXDA3gKBjAPAIQDQwCCMApgB5QlgAmEMOAfDABQCGyhBARsjgJTIu8OpwzZO9Jmw7deAmK2FEYAMnpoAvjAC8MAPJcAViWBQAdAgV1WYAJ4BuNGlCRYUWwAcSyaACc4YADmOjAA5GAg1CShjmhwAGYsCMzuXgThkdF8fKhimKkkjpiYAPQlMAB6APyacYnMyRlRoQQF2bnFMAVFxWWVNRpOUcAANqy+JDAjJLBkyGAArgC2XCS+aMNjE1MzMLbzy6u+MAA+MH4BgXUsZDraurY56J1keTBl1W+2bx8D1w0QZhkAiPJ5vV6dX5fH4lT6DBJJQG2AhkdrPYoQ3qwmqdb6Q7G1DYmLaTFzQGAARxgyAARKwaacYDSuDSiaNxqTwOTjrSWYyacBWQjmFT7rpfGDOhSYZ9Out8XCnMLkhSCBLJcVpQqccV5VjPpogA). We don't seem to need `const` in the output for TS5 to do better.

Actually, I think I can make TS4 better by adding `as const` to the `is` arguments when they are literals.

STRd6:
Since the improvement of `Object.is` over `===` is minor and the loss of any type narrowing is somewhat significant, I propose making this an experimental `"civet objectIs"` config flag and defaulting to `===` for now. In the future we can make `Object.is` the default for `is` once the type narrowing is as good as it is with `===`.

PR #264, opened 2023-01-21
Braceless type = and implicit inline braceless types

### LINKED ISSUE #158: TypeScript `type`s without braces
```ts
type Point =
  x: number
  y: number
```

This works with `interface` and no `=`, but it's also natural to expect it to work with `type`, viewing the rest as an object literal.

Daniel suggests it would just require building an `InlineImplicitInterface` rule mirroring object literals.

STRd6:
Fixes #158 

Added support for:

```typescript
type Point =
  x: number
  y: number

type Point = x: number, y: number

type Point =
      x: a: number, b: number
      y: c: [string], d: =>
```


edemaine (approved):
Nice!

PR #274, opened 2023-01-22
Interface fixes

### LINKED ISSUE #254: interface can't have more than 1 extends; class can't have both extends & implements!
Civet fails to transpile `interface A < B, C` to TS.

Civet transpiles `class Q < W implements A, B` as `class Q extends W(implements(A, B))`.

BtW, it'd be nice for Civet to have a shorthand for keyword `implements` as maybe `<<`:
`class Q < W << A, B`

https://Civet-Web.Vercel.app/#aW50ZXJmYWNlIEEgPCBCCiAgICBmb286IHN0cmluZwoKLy9pbnRlcmZhY2UgQSA8IEIsIEMgLy8gZmFpbHMKLy8gICAgZm9vOiBzdHJpbmcKCmNsYXNzIFEgPCBXCiAgICBmb286IHN0cmluZwoKY2xhc3MgUSBpbXBsZW1lbnRzIEEsIEIKICAgIGZvbzogc3RyaW5nCgpjbGFzcyBRIDwgVyBpbXBsZW1lbnRzIEEsIEIgLy8gVyhpbXBsZW1lbnRzKEEsIEIpKQogICAgZm9vOiBzdHJpbmc=

edemaine:
Fixes #254:
* `interface` support for multiple `extends`
* Fix `class` with `extends` and `implements`

I tried adding a `<<` shorthand for `implements`, but `class A < B << C` is currently treated as `class A extends (B << C)` and it seems a bit difficult to bypass (and maybe once we have operator overloading, we wouldn't want to?).  Let me know if you have thoughts on this.

### DIFF
--- a/source/parser.hera
+++ b/source/parser.hera
@@ -4494,7 +4494,7 @@ TypeDeclarationBinding
   ( BindingPattern / BindingIdentifier ) TypeSuffix?
 
 InterfaceExtendsClause
-  ExtendsToken InterfaceExtendsTarget
+  ExtendsToken InterfaceExtendsTarget ( Comma InterfaceExtendsTarget )*
 
 InterfaceExtendsTarget
   ImplementsTarget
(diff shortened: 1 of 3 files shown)


STRd6:
Looks good!

Maybe some other symbol pair would make a good implements shorthand? `<:`, perhaps?

PR #275, opened 2023-01-22
`<:` shorthand for implements

edemaine:
Here's my implementation of a shorthand for `implements`, as originally suggested in #254 and using the `<:` notation suggested in #274.

We might want to see if there are any better notations suggested before committing to `<:`.  Or we could merge this and change it later.


STRd6:
Looks good! I think this is a good syntax. We can change it later if there is a good reason.

STRd6 (approved):

Issue #238, opened 2023-01-16
Object literal shorthand


edemaine:
## [DONE] `{foo.bar().baz}`

> The following has been implemented! (#276)

`{foo}` for `{foo: foo}` is nice. As discussed in Discord, let's extend to:
* `{foo.bar}` for `{bar: foo.bar}`
* `{foo()}` for `{foo: foo()}` (note: avoid conflict with methods with empty bodies)
* `{foo.bar().baz}` for `{baz: foo.bar().baz}`

One catch is that this (specifically `foo()`) forbids methods with empty bodies. I think this is a useful trade-off, but we might want to figure out a nice empty body notation (perhaps `do`, as in `foo() do`).

## [DONE] JSX without braces

All of these would automatically work as JSX attributes. Furthermore, they are all unambiguous without the braces in JSX, so should be supported without them (just like `...foo` is currently shorthand for `{...foo}`). This is especially valuable for things like `props.foo` (meaning `foo={props.foo}`) or `foo()` (meaning `foo={foo()}`), both common patterns in SolidJS.  But a bare `foo` would remain as-is, instead of `foo={foo}`.

This is now implemented! (#508)

## [DONE] Glob notation

Related, we also discussed one of the following notations for common stems:
* `foo.{a,b,c}` for `{foo.a, foo.b, foo.c}` i.e. `{a: foo.a, b: foo.b, c: foo.c}`
* `foo{a,b,c}`
* Or allow both (similar to `?.(` and `?(`)

We can also imagine more complex chains like `foo.{a,b}.{c,d}`, which I imagine expanding to `{foo.a.c, foo.a.d, foo.b.c, foo.c.d}`. I'm less certain this is useful.  But something like `foo.{a,b.c}` for `{foo.a, foo.b.c}` almost certainly would be useful.

This is now implemented! (#333, #331, #327)

## [PARTLY DONE] Assignment globs

We should also allow `foo.{a,b,c} = bar` (and maybe `foo{a,b,c} = bar`) to mean `foo.a = bar.a, foo.b = bar.b, foo.c = bar.c` or perhaps `Object.assign(foo, bar.{a,b,c})`. Actually, `{a: foo.a, b: foo.b, c: foo.c} = bar` works, just like the non-assingment case.

We can support spreads in this case: `foo.{a,b,...rest} = bar` means `{a: foo.a, b: foo.b, ...foo.rest} = bar`.

Above is now implemented! (#333, #343)

[...]

PR #276, opened 2023-01-23
Braced object literal shorthand

edemaine:
Implements the following braced object literal shorthand (the first part of #238):

* `{props.name}` → `{name: props.name}` (and `name={props.name}` in JSX context)
* `{computed()}` → `{computed: computed()}`
* `{x.y?.z?()}` → `{z: x.y?.z?()}`

This PR newly forbids methods with no body. However, `method(){}` specifies an empty body.

Some potentially confusing behavior is that `{ console.log('hello', name) }` is now potentially an object literal (or a block), namely `{ log: console.log('hello', name) }`. This broke one test case that used a weird mix of `for` loop notation: no parens around the condition, but braces around the block:

```coffee
for [i, j] of x {
  console.log(i)
}
```

This is still broken: doesn't parse, because `for` loop has no body. Proposal is that `for` loops should use parens *and* braces, or neither.

### DIFF
--- a/source/parser.hera
+++ b/source/parser.hera
@@ -86,7 +86,12 @@ ForbiddenImplicitCalls
   AtAt # experimentalDecorators
 
 ArgumentsWithTrailingMemberExpressions
-  Arguments TrailingMemberExpressions
+  Arguments:args TrailingMemberExpressions:trailing ->
+    const call = {
+      type: "Call",
+      children: args,
+    }
+    return [ call, ...trailing ]
 
 TrailingMemberExpressions
   # NOTE: Assert "." to not match "?" or "!" as a member expression on the following line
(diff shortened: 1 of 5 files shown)


source/parser.hera
```diff
@@ -1166,6 +1184,15 @@ Arrow
     return { $loc, token: $1}
 
 ExplicitBlock
+  __ OpenBrace __ CloseBrace ->
```
STRd6, line 1187: I've been starting to replace the loose leading `__` with `( Samedent / _ )?`. The `__` in between open and close in fine.

Not something that needs to be done in this PR just something I wanted to share as we move towards more strict/robust indenting.

STRd6 (approved):

PR #277, opened 2023-01-23
Insert semicolons between lines that JS would combine

### LINKED ISSUE #272: Arrow functions and parentheticals on their own line
The following input transpiles to itself:

```coffee
x
(z) => z
```

But JavaScript has trouble with it, trying to parse it as a function call `x(z)` and then failing with `SyntaxError: Malformed arrow function parameter list`.

Perhaps even grosser is that the following gets treated like a function call, but I don't think it should.

```coffee
x
(z)
```

Perhaps we can automatically add semicolons in these cases?

edemaine:
Fixes #272.  As discussed there, this quick fix adds some extra semicolons, as it is based on input instead of generated code.  But I think I covered most if not all cases, including a potential object literal on the next line that might get parenthesized.

### DIFF
--- a/test/examples.civet
+++ b/test/examples.civet
@@ -543,7 +543,7 @@ describe "examples (from real life)", ->
       debug: {
         module: serverModule,
       },
-    }
+    };
 
     const clientOptions: LanguageClientOptions = {
       documentSelector: 1,
--- a/source/parser.hera
+++ b/source/parser.hera
@@ -3575,6 +3575,10 @@ ExpressionDelimiter
 
 StatementDelimiter
   SemicolonDelimiter
+  # NOTE: Avoid automatic continuation onto lines that start with
+  # certain characters by adding an explicit semicolon.  See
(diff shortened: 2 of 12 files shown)


test/examples.civet
```diff
@@ -543,7 +543,7 @@ describe "examples (from real life)", ->
       debug: {
         module: serverModule,
       },
-    }
+    };
```
STRd6, line 546: We can get rid of this one by refining the rule:

```
( ... / ( !LexicalDeclaration ObjectLiteral ) ) InsertSemicolon
```

edemaine: That makes a lot of sense, but it breaks a lot of tests, presumably because of caching. I guess that's another danger of this hack, that `StatementDelimiter` now uses `Samedent` which might make it and all parents uncachable (with the current caching system). I'm not sure why the current hack doesn't break tests, but maybe best to leave it as is for now? Or feel free to help debug it. 😅

STRd6 (approved):

PR #281, opened 2023-01-23
Fix semicolon method body

edemaine:
Fix `SingleLineStatements` to stay on the same line.

Also cleanup some duplicated code in the grammar.

### DIFF
--- a/source/parser.hera
+++ b/source/parser.hera
@@ -1291,7 +1291,7 @@ BracedBlock
 SingleLineStatements
   # NOTE: Statement can start with __ via AssignmentExpression.
   # Force staying on the same line via !/\n/ assertion.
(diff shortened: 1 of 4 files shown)


source/parser.hera
```diff
@@ -1297,7 +1289,9 @@ BracedBlock
 
 # NOTE: SingleLineStatements includes the empty case
 SingleLineStatements
-  ( _* Statement SemicolonDelimiter )*:stmts ( _* Statement SemicolonDelimiter? )?:last ->
+  # NOTE: Statement can start with __ via AssignmentExpression.
+  # Force staying on the same line via !/\n/ assertion.
+  ( ( _? !/\n/ ) Statement SemicolonDelimiter )*:stmts ( ( _? !/\n/ ) Statement SemicolonDelimiter? )?:last ->
```
STRd6, line 1294: I think `!EOS` is a better way to assert same line since it picks up `\r\n` as well.

STRd6 (approved):

PR #282, opened 2023-01-24
move beyond tc39 to home page, slight reorder

STRd6:
In line with some of @tomByrer's suggestions. Keeping cheatsheet simpler, moving "why Civet?" type features to the home page.

### DIFF
--- a/civet.dev/cheatsheet.md
+++ b/civet.dev/cheatsheet.md
@@ -18,112 +18,6 @@ The TypeScript code on
 
 [[toc]]
 
-## Beyond TC39
-
-### Default to `const` for Iteration Items
-
-<Playground>
-for (item of [1, 2, 3, 4, 5]) {
-  console.log(item * item);
-}
-</Playground>
-
-### Block Strings / Templates
-
-[TC39 Proposal: String Dedent](https://github.com/tc39/proposal-string-dedent)
-
-<Playground>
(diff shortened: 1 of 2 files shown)


gustavopch:
IMHO the home page and the cheatsheet should be merged. I can’t think of any reason to keep them separate.

STRd6:
Duplicating on the homepage is fine, I'm thinking of it like a special highlights.

Current cheatsheet should cover everything though maybe splitting into a 2-4 page "cheatsheet" and an exhaustive "handbook" would be a nice way to give people an appropriate reference scaled to their needs.

The goal is to highlight attention catching features while assuaging concerns that Civet is a radical departure from JS/TS (it may or may not be a radical departure but the home page shouldn't scare interested yet timid JS/TS devs). For this current iteration showcasing how we've already added pipes, dedent, expanded object shorthand (should add), pattern matching (coming soon) will illustrate the value of Civet. Keeping the examples close to existing TS syntax and less like CoffeeScript should prevent people from an initial gut reaction where they judge before understanding the core value.

`import`/`export` could probably be cut from the Homepage. `throw` expression is definitely a weaker example... maybe an "Everything is an expression" section would be better. It's not quite true that everything is an expression but we intend it to be and the homepage can be slightly aspirational.

**under_construction.gif**

PR #283, opened 2023-01-24
Integer property access

edemaine:
* `x.0.0` -> `x[0][0]`
* `x.-1` -> `x.at(-1)`

There was some discussion on Discord about whether `.5` should become uniformly `[5]` or `.at(5)` for all integers `5`.  I went with a pragmatic approach: `[]` property access is usually what you want, it's probably faster, it matches `."string"` notation, and it works with more objects; while `.-1` is clearly asking for the last element. It also makes sense in the grammar, because `-` is already not part of the integer literal but rather an operator.

(Counterargument would be that `.at` is more overridable, so maybe mapping everything to `.at` is interesting... but I'm not sure who would really want to override this operator.)

I also wondered about a generic `x@y` shorthand for `x.at(y)`, but Oren pointed out that `-1` is *the* main use of `.at`, so instead the `.-1` notation is probably sufficient.

### DIFF
--- a/civet.dev/cheatsheet.md
+++ b/civet.dev/cheatsheet.md
@@ -85,6 +85,8 @@ Property access shorthand:
 <Playground>
 json.'long property'
 json.`${movie} name`
+matrix.0.0
+array.-1
 </Playground>
 
 ### Arrays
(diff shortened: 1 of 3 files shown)


STRd6 (approved):
Looks great, thanks!

PR #284, opened 2023-01-24
`{x[y]}` shorthand for `{[y]: x[y]}`

edemaine:


STRd6:
Looks good! You may want to add a TODO comment to use a ref for complex expressions inside `[]` for the future.

STRd6 (approved):

edemaine:
`// TODO: Should use a ref instead of duplicating the expression.` ✅

PR #288, opened 2023-01-24
Escape `*/` in `###...###`, document `###` and `coffeeBoolean`

edemaine:
This started as a documentation commit, but I discovered that `###...###` didn't support `*/` inside. That matches CoffeeScript behavior but not the existing behavior of JSX `<!--...-->` comments, and it was an easy copy/paste to make it work in both.  I think it's nice to have a way to comment out a block of code that already has `/*...*/` comments.

### DIFF
--- a/civet.dev/cheatsheet.md
+++ b/civet.dev/cheatsheet.md
@@ -594,6 +594,15 @@ numbers[1...-1] = []
 a + b = c
 </Playground>
 
+### Block Comments
+
+<Playground>
+###
+block comment
+/* nested comment */
+###
+</Playground>
+
 ### `Object.is`
 
 The `"civet objectIs"` directive changes the behavior of the `is` operator to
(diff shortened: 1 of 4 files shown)


STRd6:
We might also want to allow nested `/*` comments but that will need to be a separate mechanism.

Looks good!

edemaine:
> We might also want to allow nested `/*` comments but that will need to be a separate mechanism.

What do you mean? `/* /* ... */` is a valid JS comment, i.e., nested `/*`s are already OK. I believe just the `*/`s need to be escaped.

STRd6:
> What do you mean? `/* /* ... */` is a valid JS comment, i.e., nested `/*`s are already OK. I believe just the `*/`s need to be escaped.

I meant allow nested JS comments like:

```typescript
/* A comment
/* I'm nested inside the comment! */

Still in the comment */

inTheCodeNow := true
```

edemaine:
Oh, got it!  Yes, that'd be neat too.

PR #295, opened 2023-01-26
Identifier operators

edemaine:
* `operator a, b, c` to bless already existing functions into infix operators
* `import {operator x} from ...` to import and bless an imported function
* `operator op(a, b) ...` to define and bless a new function
* `operator {min, max} := Math` or any lexical declaration to simultaneously bless
* `op(a, b)` is always a function call

I haven't added to cheatsheet yet, but can do so later.

### DIFF
--- a/source/parser.hera
+++ b/source/parser.hera
@@ -1307,8 +1307,9 @@ OperatorDeclaration
       operator: true,
     }
   # `operator id` alone blesses `id` as an operator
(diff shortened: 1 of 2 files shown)


STRd6:
he actually did it the absolute madman

STRd6:
Seriously though, it looks great!

PR #296, opened 2023-01-26
wip object pattern matching

STRd6:
Pattern matching is getting closer.

### DIFF
--- a/source/parser.hera
+++ b/source/parser.hera
@@ -6293,6 +6293,7 @@ Init
           const spliceRef = module.getRef("splice")
 
           blockPrefix = {
+            type: "PostRestBindingElements",
             children: ["[", module.insertTrimmingSpace(after, ""), "] = ", spliceRef, ".call(", restIdentifier, ", -", after.length.toString(), ")"],
             names: after.flatMap(p => p.names),
           }
@@ -6739,6 +6740,7 @@ Init
                 case "ArrayMatchingPattern":
                 case "ObjectMatchingPattern":
                 case "RegularExpressionLiteral":
+                case "Literal":
                   getPatternConditions(e, subRef, conditions)
                   break
                 case "BindingMatchElement":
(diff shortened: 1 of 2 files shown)


edemaine (approved):
Great progress!

As discussed in Discord, I think we should think about whether it's "right" that arrays are exact matching (require exact length instead of just "at least pattern's length") while objects are inexact matching (require at least these keys, but could have more).  There are two parallels to draw:

* TypeScript types: object types are inexact lower bounds (at least these keys are available), while arrays (as tuple patterns) [are exact matches](https://www.typescriptlang.org/play?#code/CYUwxgNghgTiAEYD2A7AzgF3gSwFzxQFcBbAIxBgBp4ArfIsigKGXSwDMkl4BeeAbWzUa1bAF14UNIlSYmAennxl8AHoB+FrKylY+fg3JUCJIxL6ckTIA). This is what you're following now.
* JS destructuring: both object types and arrays are inexact lower bounds. You can write `[a, b] = [1, 2, 3]`.

The nice thing about exact patterns is that you can always build inexact patterns out of them using a rest parameter.  Given that `[a, b]` forces array of length 2, you can use `[a, b, ...]` to ask for array of length at least 2.  But there's currently no way to specify an exact object type.  Possibly related to Flow's `{| ... |}` exact object type.

STRd6:
I'm leaning towards TS semantics for the defaults. An opt-in exact match does sound useful in some situations and should be simple enough to add on later.

edemaine:
Sounds good. I'm really looking forward to when I can write this:

```ts
interface point
  x: number
  y: string
switch v
  point
    ...
---
interface point {
  x: number
  y: string
}
if (typeof v === "object" && v != null && typeof(v.x) === "number" && typeof(v.y) === "string") {
  ...
}
```

Before we understand type references though (looking up `point` is tricky), I wonder if there's a way to specify "something of type `number`"?  Maybe this (based on one of the #126 proposals):

```ts
switch v
  {x:: number, y:: string}
```

STRd6:
I was thinking of something like:

```typescript
switch v
  {x: <? "number", y: <? "string" }
```

making use of the expanded `<?`/`instanceof`

It would also allow thing like:

```typescript
switch x
  <? Error
    ...
  <? Result
    ...
```

But that's slightly trickier to parse right now.

PR #297, opened 2023-01-27
pushing flagging props to the limit

STRd6:


test/object.civet
```diff
@@ -98,6 +98,42 @@ describe "object", ->
     })
   """
 
+  testCase """
+    flagging shorthand with numeric, string, template, computed keys
+    ---
+    {
+      -1
```
edemaine, line 105: Maybe add `+1` and `!1` as examples here?

edemaine (approved):
I was worried this would be too xtreme but the examples all look quite sensible. Like a bunch of object literal changes lately, it feels like we're cleaning up the edge cases to no longer be edge cases, which feels nice.

PR #301, opened 2023-01-28
operator= assignment

edemaine:
This PR transforms `x op= y` into `x = op(x, y)`, with the idea that `op` is a user-defined binary infix operator.

However, I have not restricted this to identifiers that have been blessed via `operator`. It just works for all identifiers. I think this is a good idea, but it does have the following impact:

> `x y= z` used to parse as `x(y=z)` but now parses as `x=y(x, z)`

But note that spaces are required before and after `y=`, and the space before can't include a newline, which prevented falsely matching all existing tests.  For example, `x y=z` still means `x(y=z)`.

If we think this is too risky, we could protect it behind blessed operators with a one-line change.

### DIFF
--- a/civet.dev/cheatsheet.md
+++ b/civet.dev/cheatsheet.md
@@ -292,6 +292,20 @@ export operator has(x, y)
   y contains x
 </Playground>
 
+### Operator Assignment
+
+Even without blessing a function as an `operator`, you can use it in
+an assignment form:
+
+<Playground>
+{min, max} := Math
+smallest = Infinity
+largest = -Infinity
+for item in items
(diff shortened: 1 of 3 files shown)


STRd6 (approved):

STRd6:
Looks good!

STRd6:
It might be nice to see a test for something like:

```typescript
x not contains= y
```

edemaine:
I don't think that currently works... Should it mean `x = !contains(x, y)`?

STRd6:
I'm not sure... but having a test documenting the current behavior would be nice.

PR #307, opened 2023-01-31
Fix #306: don't rewrite .ts imports in deno

### LINKED ISSUE #306: Deno: Don't change import path file extensions
Civet seems to change import paths from `.ts` to `.js`, even when you're specifying `-o .ts` output.

I'm using Civet on a Deno project and this breaks Deno logic. For example,

```ts
// Civet
import { DB } from "https://deno.land/x/sqlite/mod.ts"
```
becomes
```js
// JS
import { DB } from "https://deno.land/x/sqlite/mod.js"
```

STRd6:


edemaine (approved):
LGTM. I guess we'll change the default of `rewrite-ts-imports` eventually (once TS is sane).

PR #318, opened 2023-02-02
Thicc pipe

STRd6:
Working on thick pipes. Starting with test cases to figure out some of the details.

Added:
- [x] Single `|>=` assignment in first position
- [x] `||>` in any position

Maybe TODO:
- [ ] Refs
- [x] Code cleanup
- [ ] Unwrap & blocks
- [ ] Kind of weird but makes sense with implicit function application (the pipes go inside the function parameters)
```typescript
array ||> .push 3 ||> .shift() ||> .sort()
---
($ => $.push(($1 => $1.shift())(3),($2 => $2.sort())(3),3))(array),array
```
- [ ] Even worse with newlines and definitely shouldn't happen this way
```typescript
  array
  ||> .push 3
  ||> .shift()
  ||> .sort()
  ---
  ($ => $.push(($1 => $1.shift())(3),($2 => $2.sort())(3),3))(array),array
```


edemaine:
`array ||> .push 3 ||> .shift() ||> .sort()`

I expected `||>` to have lower precedence than (implicit) function calls... at least, I find the existing parse pretty confusing. An existing example of this is postfix `if`.

This is an issue with `|>` as well, so not specific to this PR. But maybe we can fix it by using `NonPipelineExtendedExpression` in all implicit function calls?

edemaine (approved):
This looks great! I think we can merge this now. We should then probably open an issue for the remaining issues: refs, unwrapping, and preventing pipelines in implicit function arguments.  Refs and implicit arguments seem particularly important at this point. (I thought I'd opened an issue for unwrapping, but looks like it was just https://github.com/DanielXMoore/Civet/pull/83#issuecomment-1364739339 )

PR #321, opened 2023-02-02
Improve LSP queuing behavior

edemaine:
Continuing discussion from #304:

* No more `setInterval`, so no work when idle
* Schedule updates to other documents after an entire batch completes

I tested and this seems to work well.  The delay of 50 does seem to be necessary though.

### DIFF
--- a/lsp/source/server.mts
+++ b/lsp/source/server.mts
@@ -79,7 +79,8 @@ const ensureServiceForSourcePath = async (sourcePath: string) => {
 }
 
 // TODO Propagate this to an extension setting
-const diagnosticsPropagationDelay = 100;
+const diagnosticsDelay = 16;  // ms delay for primary updated file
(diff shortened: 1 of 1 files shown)


lsp/source/server.mts
```diff
@@ -373,27 +373,46 @@ documents.onDidOpen(async ({ document }) => {
...
+    scheduleExecuteQueue()
+  } else {
+    scheduleUpdateDiagnostics(changed)
+  }
+}
+async function scheduleExecuteQueue() {
+  // Schedule executeQueue() if there isn't one already running or scheduled
+  if (executeTimeout) return
+  if (!changeQueue.size) return
+  await (executeTimeout = setTimeout(50))
```
STRd6, line 404: Let's change this to `25` or `16` since it is the minimum time before processing rather than the maximum.

STRd6 (approved):

PR #323, opened 2023-02-02
Function fixes

### LINKED ISSUE #308: Coffeescript bug: postfix if
### Coffeescript bug (parse error)

```
"civet coffeeCompat"

[1,2,3].forEach((value) -> console.log(value) if true)      
```
gives
```
"civet coffeeCompat"

[1,2,3].forEach((value) -> console.log(value) if true)      
                                                     ^ ParseError
```

### Playground links

Coffeescript: https://coffeescript.org/#try:%5B1%2C2%2C3%5D.forEach((value)%20-%3E%20console.log(value)%20if%20true)%20%20%20%20%20%20
Civet: https://civet.dev/playground/?code=ImNpdmV0IGNvZmZlZUNvbXBhdCIKClsxLDIsM10uZm9yRWFjaCgodmFsdWUpIC0%2BIGNvbnNvbGUubG9nKHZhbHVlKSBpZiB0cnVlKSAgICAgIA%3D%3D

edemaine:
The following changes are in separate commits for easier review:

* Allow suffixed expressions and statements in more places
  * Same-line body of arrow functions (fixes #308)
  * JSX braced expressions and children
  * But not e.g. in `try` expression which wants postfix to be outside, and `do` expression which needs to treat `while` postfix as something completely different.
  * This is how this PR started, but then I ran into a zillion other bugs along the way...
* Forbid `while`/`until` as beginning of implicit calls
  * Matches existing behavior of `for`
  * Otherwise conflicts with postfix `while`/`until` notation
  * For example, `x while y;` got treated as a call `x(while y {})` before
* Allow semicolon-separated lines in one-line `=>`
  * For example, `=> x; y`, matching existing behavior for `->`
  * Previously lots of weird special cases for FatArrowBody
  * Now just one case for no brace wrapping and then the usual BracedBlock
* Fix implicit return in `else` clause
  * Wasn't working when semicolon was inserted between `then` and `else`.
  * Improve AST so this is less brittle.
  * For example, `-> if x then y else z` wasn't returning `z`
* Fix more cases of no implicit return via semicolon
  * `-> x while y;` was still returning
* Wrap top-level anonymous `function` expressions in parens
  * Otherwise JS complains: `SyntaxError: Function statements require a function name`


source/main.coffee
```diff
@@ -78,6 +78,7 @@ uncacheable = new Set [
   "NestedJSXChildExpression"
   "NestedObject"
   "NestedPropertyDefinitions"
+  "NonSingleBracedBlock"
   "NonSuppressedTrailingMemberExpressions"
   "ObjectLiteral"
   "PopIndent"
```
STRd6, line 84: We should probably also add `PostfixedExpression` to no-cache rules since `AssignmentExpression` and `ExtendedExpression` are no-cache.

STRd6:
Nice! Great work cleaning up those edge cases, lots of great fixes here!

STRd6 (approved):

source/main.coffee
edemaine: Good point. I added `PostfixedExpression` and `PostfixedStatement`.

PR #325, opened 2023-02-02
typeof shorthand

### LINKED ISSUE #319: typeof shorthand
We discussed this on Discord but it never got written down (or implemented):

* `x <? "string"` → `typeof x === "string"`
* `x instanceof "string"` → `typeof x === "string"`
* In general, change the behavior of `instanceof` when the right-hand side is a string literal, which is currently a useless check (throws `Uncaught TypeError: Right-hand side of 'instanceof' is not an object`), into a useful check. Doesn't work for non-literals.

This will pair nicely with pattern matching when it supports `<?` / `instanceof`.

edemaine:
Fixes #319

### DIFF
--- a/civet.dev/cheatsheet.md
+++ b/civet.dev/cheatsheet.md
@@ -585,6 +585,15 @@ a !<? b
 a <? b !<? c
 </Playground>
 
+### `typeof` shorthand
+
+<Playground>
+a <? "string"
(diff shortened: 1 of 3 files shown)


STRd6 (approved):
Nice, glad it ended up being so simple!

PR #327, opened 2023-02-04
Object glob patterns obj.{a,b}

edemaine:
Part 3 of #238

* `obj.{a,b}` → `{a:obj.a, b:obj.b}`
* `obj.{a: b}` → `{b: obj.a}`
* No ref (yet), but still probably useful to merge now
* `obj.{a, ...rest}` forbidden, though this could maybe be added in the future via something like `{a, ...rest} = obj, {a, rest}`
* `obj.{@a}` and `obj.{a=5}` forbidden (I think this makes sense)

Bug: automatic semicolon insertion doesn't work before glob patterns which become object literals. I don't see an easy fix, so maybe finally time to add a `processBlock` to do proper semicolon insertion after all processing takes place (but I'd suggest a different PR for this).

### DIFF
--- a/source/parser.hera
+++ b/source/parser.hera
@@ -567,27 +567,26 @@ LeftHandSideExpression
 CallExpression
   # NOTE: Tracking trailing member expressions based on indentation level for implicit arguments
   "super" ArgumentsWithTrailingMemberExpressions CallExpressionRest*:rest ->
-    return {
+    return module.processGlob({
       type: "CallExpression",
       children: [$1, ...$2, ...rest.flat()],
-    }
+    })
   # Dynamic import(), with optional parentheses when not used at top level.
   # (At top level, ImportDeclaration will match first.)
   # NOTE: Using ExtendedExpression to allow for If/Switch expressions
   "import" ArgumentsWithTrailingMemberExpressions CallExpressionRest*:rest ->
-    return {
+    return module.processGlob({
       type: "CallExpression",
       children: [$1, ...$2, ...rest.flat()],
-    }
+    })
 
   MemberExpression:member NonSuppressedTrailingMemberExpressions:trailing CallExpressionRest*:rest ->
     if (rest.length || trailing.length) {
       rest = rest.flat()
-      return {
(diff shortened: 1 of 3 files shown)


source/parser.hera
```diff
@@ -567,27 +567,26 @@ LeftHandSideExpression
...
       type: "CallExpression",
       children: [$1, ...$2, ...rest.flat()],
-    }
+    })
   # Dynamic import(), with optional parentheses when not used at top level.
   # (At top level, ImportDeclaration will match first.)
   # NOTE: Using ExtendedExpression to allow for If/Switch expressions
   "import" ArgumentsWithTrailingMemberExpressions CallExpressionRest*:rest ->
-    return {
+    return module.processGlob({
```
STRd6, line 578: Not necessary for this PR but something I've been moving towards is moving the processing functions to the end of the parse ( inside processProgram ) and having the intermediates be better AST nodes which could be useful for linting/prettier type things.

STRd6:
Looks good!

I assume `foo.{a, b.c}` is something for the future as well. I agree that `obj.{a=5}` doesn't really make sense but `obj.{a?=5}` might (also for later).

STRd6 (approved):

STRd6:
One more note `{b: obj.a}` is opposite of LiveScript. We may want to switch it since ObjectLiterals use the name on the left and object destructuring uses the name on the right. Or just explore more parallels and document why it's different.

edemaine:
I agree, I think it's backwards. Specifically, I like the idea of `obj.{a().b, c.d}` but then how would you rename? We can't have `obj.{a().b: x, c.d: y}` so it has to be `obj.{x: a().b, y: c.d}` and this also matches the existing `{x: a().b, y: c.d}` or the more unusual `{a().b, c.d}`.

Will fix this tomorrow if you don't beat me to it. :-)

source/parser.hera
edemaine: Yeah, that's probably a good idea. In this case, the paren wrapping around object literals is currently done by a `->` code block, so I stuck with that. But we could change them both together.

PR #330, opened 2023-02-04
Make ampersand optional for some common blocks

STRd6:


source/parser.hera
```diff
@@ -1323,7 +1323,7 @@ FunctionExpression
     }
 
   # Ruby/Crystal style block shorthand
-  AmpersandUnaryPrefix?:prefix Ampersand AmpersandBlockRHS?:rhs ->
+  AmpersandUnaryPrefix?:prefix ( Ampersand / ( !NumericLiteral &( QuestionMark? Dot ) ) ) AmpersandBlockRHS?:rhs ->
```
edemaine, line 1326: What does the `!NumericLiteral` do?

test/call-expression.civet
```diff
@@ -132,11 +132,15 @@ describe "call-expression", ->
...
-    it "should not parse with !. on newline after plain identifier", ->
-      throws """
-        data
-        !.map (x) => x + 1
-      """
+    testCase """
+      !. on newline after plain identifier is a separate fn block
+      ---
+      data
+      !.map (x) => x + 1
```
edemaine, line 139: Took me a while to understand what this does... Unary operators with implicit `&` is a little crazy. But I guess it makes sense. (Though in this case doesn't make sense to negate an array :))

test/function-block-shorthand.civet
```diff
@@ -17,6 +17,22 @@ describe "&. function block shorthand", ->
...
+    x.map($ => $.name)
+  """
+
+  testCase """
+    omit &
+    ---
+    x.map ?.name
+    ---
+    x.map($ => $?.name)
+  """
```
edemaine, line 34: These are definitely cool!

edemaine (approved):

source/parser.hera
STRd6: It prevents `.1` from being interpreted as `&.1`.

test/call-expression.civet
STRd6: Yeah, it's a little strange but falls out naturally from the existing rules.

source/parser.hera
edemaine: Does it? Wouldn't it have to come after the Dot, inside the positive lookahead assertation?

edemaine: Oh, I see, I was thinking of `/[0-9]/`, but `NumericLiteral` includes a leading dot. Got it!

PR #340, opened 2023-02-07
xor/^^ and xnor/!^ operators

edemaine:
As suggested by @samualtnorman:
* `^^` and `xor` are logical XOR operators (but `xor` is not a reserved word), to `^` as `||` is to `|`.
* `!^` and `!^^` and `xnor` are logical XNOR operators (logical equality), by analogy to `!=`/`!==`. (Should I add `not xor` too?)
* Assignment versions of all of the above, e.g., `xor=`

Normally I'd write these as `Boolean(a) !== Boolean(b)` and `Boolean(a) === Boolean(b)`, respectively. But `xor` has a much more clear meaning, and these should shortcut properly and use `a` or `b` as values instead of `true` and `false` whenever possible, similar to `&&` and `||`.

In a separate commit, added semicolons to prelude ref declarations; otherwise a `(` on the next line will prevent automatic semicolon insertion. For example, [`(x %% y)` in the first line compiles incorrectly](https://civet.dev/playground?code=KHggJSUgeSk%3D).  (We may need to do this for `autoVar` too.)

### DIFF
--- a/source/parser.hera
+++ b/source/parser.hera
@@ -5558,7 +5558,7 @@ Reset
           children: [": <T>(this: T[], searchElement: T) => boolean"]
         }
         // [indent, statement]
-        module.prelude.push(["", ["const ", indexOfRef, typeSuffix, " = [].indexOf", module.asAny, "\n"]])
+        module.prelude.push(["", ["const ", indexOfRef, typeSuffix, " = [].indexOf", module.asAny, ";\n"]])
       },
       hasProp(hasPropRef) {
         const typeSuffix = {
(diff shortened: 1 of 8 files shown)


source/parser.hera
```diff
@@ -5581,15 +5610,31 @@ Reset
...
-        module.prelude.push(["", ["const ", moduloRef, typeSuffix, " = (a, b) => (a % b + b) % b", "\n"]])
+        module.prelude.push(["", ["const ", moduloRef, typeSuffix, " = (a, b) => (a % b + b) % b;", "\n"]])
+      },
+      xor(xorRef) {
+        const typeSuffix = {
+          ts: true,
+          children: [": (a: unknown, b: unknown) => boolean"]
+        }
+        // [indent, statement]
+        module.prelude.push(["", ["const ", xorRef, typeSuffix, " = (a, b) => a && b ? false : a || b;", "\n"]])
```
STRd6, line 5629: For the future: it should be possible to get a much more accurate return type by using conditional types.

STRd6 (approved):

PR #341, opened 2023-02-08
wip: attempting caching based on parser state

STRd6:
Made some progress with caching based on parser state but still having trouble.

- This isn't any faster than the existing no-cache list. Maybe computing the key as a string is too expensive?
- The errors are a sign that there is a mistake with tracking the state. I should use them to learn.

Will probably close this soon but have it as a reference if exploring again in the future.

### DIFF
--- a/source/main.coffee
+++ b/source/main.coffee
@@ -111,6 +111,27 @@ uncacheable = new Set [
   "UpdateExpression"
 ]
 
+uncacheable = new Set [
+  # Meta
+  "Debugger"
+
+  # Indentation
+  "InsertIndent"
+  "PushIndent"
+  "PopIndent"
+  "TrackIndented"
+  # Nested Application
+  "PopIndentedApplication"
+  "SuppressIndentedApplication"
+  # Trailing Member Expressions
+  "SuppressTrailingMemberProperty"
+  "PopTrailingMemberProperty"
+  # JSX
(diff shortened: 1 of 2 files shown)


edemaine:
Curious. I don't see an obvious bug. I had independently realized that `PopTrailingMemberProperty` needed to be added.

I tried this out on my 600-line Civet art app, and it does slow things down: 1.45 seconds instead of 0.9 seconds to compile. On the other hand, turning on verbose mode, it seems like the number of cache misses goes down from 201,829 to 102,015. And it does seem to remove all the obvious duplication of effort. For example, in the current compiler I see this line repeated 19 times in a row:
```
Parsed "" [pos 172-172] as CoffeeCommentEnabled
```
But I see it only once with this branch.

Maybe we should have a set of state-independent rules (stuff like `CoffeeCommentEnabled` which are used many many times) that don't need these fancy keys. We could also restrict the JSX state to JSX rules, for example.

Alternatively, we could rebuild the state string only when it changes, which is presumably somewhat rare (indentations are probably most common), and have a completely a different set of caches for each state, so instead of having to combine keys we dereference (`get`) twice. But we'd only need to change the current set of caches when the state changes, so most of the time would be just one dereference...

STRd6:
Completed in https://github.com/DanielXMoore/Civet/pull/645

PR #343, opened 2023-02-08
Spreads in object globs

edemaine:
(More of part 4 of #238.)

### Assignment form:
`obj.{a, b, ...rest} = data` → `{a:obj.a, b:obj.b, ...obj.rest} = data`
* This is clearly the right thing.

### Object literal form:
`obj.{a, b, ...c}` → `{a:obj.a, b:obj.b, ...obj.c}`
* This is useful e.g. to merge subobjects: `obj.{...defaults, ...globals, ...user}` → `{...obj.defaults, ...obj.globals, ...obj.users}`
* Also clearly mirrors the assignment form.
* Originally I thought that `obj.{a, b, ...c}` should mean something like `{a, b, ...c} = obj, {a, b, c}`, i.e., extract the "rest" into `c` and make it into an attribute. But I think this is a little too magical, restricted to a single `...spread`, and isn't actually the intuitive thing for a non-assignment. The fact that the implemented syntax does this for actual assignments suggests it's the right thing in both cases.

### DIFF
--- a/source/parser.hera
+++ b/source/parser.hera
@@ -2001,6 +2001,7 @@ PropertyDefinitionList
     return $0.map(([prop, delim]) => {
       return {
         ...prop,
+        delim,
         children: [...prop.children, delim],
       }
     })
@@ -2047,7 +2048,8 @@ PropertyDefinition
       type: "SpreadProperty",
       children: [ws, dots, exp],
       names: exp.names,
-      value: [dots, exp],
+      dots,
+      value: exp,
     }
   # NOTE: Added `{x.y?.z()}` shorthand for `{z: x.y?.z()}`
   # NOTE: this needs to be at the bottom to prevent shadowing NamedProperty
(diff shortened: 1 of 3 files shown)


source/parser.hera
```diff
@@ -2047,7 +2048,8 @@ PropertyDefinition
       type: "SpreadProperty",
       children: [ws, dots, exp],
       names: exp.names,
-      value: [dots, exp],
+      dots,
+      value: exp,
```
STRd6, line 2052: I'd name this `expression` instead of `value`.

STRd6 (approved):

source/parser.hera
edemaine: This is to match regular `Property` AST nodes.  In particular, this code relies on it when checking `if (part.value && !["CallExpression", "MemberExpression", "Identifier"].includes(part.value.type)) {` — this way, it's the same check for both `Property` and `SpreadProperty`.

But I agree that `SpreadProperty` is special, so perhaps this doesn't make sense.  Do you still think `expression` is best?

STRd6: In that case either is fine.

PR #344, opened 2023-02-08
Switch prelude declarations from const to var

### LINKED ISSUE #339: CLI can't use ref functions multiple times
The CLI (run simply as `civet`) runs commands multiple times in the global context. The result is that multiple `const` declarations cause an error like so:

image

Should we change automatic ref functions from `const` to `var`? Or get the parser to return prelude content separately so we can deduplicate in the CLI?

edemaine:
Fixes #339

image

### DIFF
--- a/source/parser.hera
+++ b/source/parser.hera
@@ -2811,7 +2811,7 @@ CoffeeForStatementParameters
       // TODO: Exp ref if exp is not a literal or identifier
 
       if (declaration.own) {
-        const hasPropRef = module.getRef("hasProp")
+        var hasPropRef = module.getRef("hasProp")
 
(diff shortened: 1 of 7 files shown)


source/parser.hera
```diff
@@ -2811,7 +2811,7 @@ CoffeeForStatementParameters
       // TODO: Exp ref if exp is not a literal or identifier
 
       if (declaration.own) {
-        const hasPropRef = module.getRef("hasProp")
+        var hasPropRef = module.getRef("hasProp")
```
STRd6, line 2814: This can still be `const.`

STRd6 (approved):

PR #345, opened 2023-02-08
Support hex and other numbers in ranges

### LINKED ISSUE #342: Coffeescript bug: can't handle hex in range expression
### Coffeescript bug (parse error)

```
"civet coffeeCompat"

x = [0...0xa]
```
gives
```
Unexpected non-whitespace character after JSON at position 1
```

### Playground links

[Coffeescript link](https://coffeescript.org/#try:x%20%3D%20%5B0...0xa%5D)
[Civet link](https://civet.dev/playground/?code=ImNpdmV0IGNvZmZlZUNvbXBhdCIKCnggPSBbMC4uLjB4YV0%3D)

edemaine:
Fixes #342

Originally, I used `eval`, because at this point we should have guaranteed that the `Literal` is of a reasonable form.  But this seems a little risky.  Instead I switched to `parseInt` which seems to cover the cases we care about (given CoffeeScript's weird semantics with floats).

I tried to get BigInt literals to work too, but they're annoying to work with because even `+= 1` doesn't work (can't mix BigInts with regular numbers). Anyway, they're unlikely to be useful, as they likely have length more than 20.


STRd6:
Needs one more check to support:

```
[0..0b10]
[0..0o10]
```

https://coffeescript.org/#try:%5B0..0b10%5D%0A%5B0..0o10%5D

edemaine:
Oof, I didn't realize `parseInt` doesn't automatically recognize binary or octal literals.

I ended up writing a general `module.literalValue` helper. Could be worth another review.

STRd6 (approved):

PR #348, opened 2023-02-09
Inner assignments within assignment chains

edemaine:
One idea from #317:

```js
(obj[key] ?= 0) += 1
---
(obj[key] ?= 0), obj[key] += 1
```

Now though I wonder whether assignment expressions is actually the right time to do this, because another natural example `(x ?= 0)++` does not work.  (Well, Civet is fine with it and leaves it as is, but JS is not fine with it, as `x ?= 0` is not a valid LHS, which is what this PR is trying to fix.)  I'd appreciate suggestions on *where* to do this transformation.  Currently the AST doesn't have the ability to go to the parent node, which makes it hard to look at all assignments and then replace itself.  I guess the `AssignmentExpression` node could rewrite its own contents (and even become a different type of node), but I don't know how (or even where) to add the assignment operation as a "pre" operation...

This might still be worth merging for now, as it does work, and is presumably useful. But also happy to rework it.

### DIFF
--- a/source/parser.hera
+++ b/source/parser.hera
@@ -6880,7 +6880,7 @@ Init
     function processAssignments(statements) {
       gatherRecursiveAll(statements, n => n.type === "AssignmentExpression" && n.names === null)
       .forEach(exp => {
-        let {lhs: $1, exp: $2} = exp, tail = [], i = 0, len = $1.length
+        let {lhs: $1, exp: $2} = exp, pre = [], tail = [], i = 0, len = $1.length
 
         // identifier=
         if ($1.some((left) => left[left.length-1].special)) {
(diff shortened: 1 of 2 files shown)


STRd6:
I think this is worth adding and we can refine it further later.

As for transforming the AST nodes I think we should add search helpers / query selectors that let us grab a root node and all relevant children. That would come in handy for adding ref declarations to the enclosing blocks among other things.

Very roughly:
```
querySelectRootWithChildren(ast, ".block", "Ref", (rootNodeWithABlockProperty, refsArray) => {
  // do stuff
})
```

With inspiration from xpath, jq, etc. and maybe a couple different methods to affect how recursing happens.

STRd6 (approved):

edemaine:
querySelect sounds like a cool idea. I do worry about adding a lot of distinct traversals through the AST, as just traversing the tree is overhead. So when multiple traversals can be done in parallel (e.g. they operate on distinct node sets), maybe we could do them together? A querySelector API could maybe be designed to take multiple queries and resolve them all at the same time.

In the case of this specific feature, I'm not even sure where it makes the most sense to move the assignment... I guess it's just operations that need a left-hand side, so that's assignments (done) and ++/-- and maybe not much else. OK, this is halfway there then. And I can just move it left slightly as in `(x ?= 0)++` -> `(x ?= 0), x++`.

PR #349, opened 2023-02-09
Test helper `throws` supports description and ---

edemaine:
While still offering backward-compatible mode of just code (still used in some batch tests)


STRd6 (approved):
Looks good, probably don't even need backward compat.

Issue #351, opened 2023-02-09
Integrating Svelte?


levitabris:
Svelte is gaining popularity among people who appreciate simplicity and easy composability. Since its functionality is not based on a js import, would you consider making Civet compatible with .svelte components?

I would recommend a syntax like the following:

`<script lang="civet">
   'hello world!' |> console.log
</script>
`

edemaine:
Sounds like a great idea!  Perhaps we can mimic [svelte-add/coffeescript](https://github.com/svelte-add/coffeescript)?  Let us know if you'd like to work on this; it doesn't look too difficult.

STRd6:
We'll need to add support for `$:` labels too. Should be straightforward.

edemaine:
Looking at [Svelte's TypeScript support](https://svelte.dev/blog/svelte-and-typescript), the transpilation may be possible now with our existing [Vite plugin](https://github.com/lorefnon/vite-plugin-civet) which should also work as a Rollup plugin.

levitabris:
> Sounds like a great idea! Perhaps we can mimic [svelte-add/coffeescript](https://github.com/svelte-add/coffeescript)? Or is svelte-add still active / the right approach?
> 
> Let us know if you'd like to work on this; it doesn't look too difficult.

Not sure if it is the right approach. But I can experiment with mimicing the svelte-add coffeescript and see what happens.

edemaine:
Great! Feel free to suggest a better approach as well if you find one.

FYI, we've added support for `$:` to Civet already.

levitabris:
There is a bit learning curve here. My WIP version:
 
https://github.com/levitabris/svelte-add

levitabris:
Update: 
It seems that `svelte-add` coffee script support is just a wrapper around `svelte-proprocess`(feel free to correct me if wrong), which implemets the original coffee support for svelte. I'm looking into how it was done in the svelte-preprocess. 

BTW, the `svelte-preprocess` seems to be straight forward to config itself🦦

STRd6:
@levitabris Have you taken a look at our new Rollup and Vite plugins? https://github.com/DanielXMoore/Civet/tree/main/integration/unplugin/examples please try them out and let us know if you have any questions or encounter any errors, thanks!

adam2am:
Hey, any updates or use-cases with svelte? All I found is an attempt https://github.com/moka-ayumu/svelte-civet 2 years ago

adam2am:
[...]

PR #354, opened 2023-02-09
Support for labeling statements

edemaine:
`:label` or `$:` for Svelte

This is enough for Svelte compatibility (#351).

But we should also add label support to `break` and `continue`. I just think we should discuss the other potential uses mentioned in  https://github.com/DanielXMoore/Civet/discussions/312#discussioncomment-4832062 first.

### DIFF
--- a/source/parser.hera
+++ b/source/parser.hera
@@ -2554,6 +2554,10 @@ Statement
 
   EmptyStatement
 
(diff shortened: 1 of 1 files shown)


STRd6 (approved):

STRd6:
We should probably add some tests for this soon.

edemaine:
Oops, that means I forgot to `git add` them. Will do once I'm on my computer!

PR #356, opened 2023-02-11
Cleanup flag stacks, re-allow stuff inside parens/brackets/braces

edemaine:
* General rename from `Suppress` to `Forbid`
* Add complementary `Allow` rules
* Add `Restore` rules for popping `Forbid`/`Allow`.
* New `AllowAll`/`RestoreAll` for re-enabling everything.
* No more popping inside rules except `Pop`/`Restore` rules (for eventual new caching).


STRd6 (approved):
Great cleanup!

PR #358, opened 2023-02-11
Allow newline before type inside parens

STRd6:


source/parser.hera
```diff
@@ -5255,7 +5255,7 @@ TypePrimary
   _? IdentifierName (Dot IdentifierName)* TypeArguments?
   # NOTE: Check FunctionType before parenthesized in order to distinguish between (a: T) => U and
   # A parenthesized inline interface (a: T) ---> ({a: T})
-  __ OpenParen Type __ CloseParen
+  __ OpenParen ( Type / ( EOS Type ) ) __ CloseParen
```
edemaine, line 5258: This could be rewritten as follows:

```suggestion
  __ OpenParen ( EOS? Type ) __ CloseParen
```

Or maybe even simpler:
```suggestion
  __ OpenParen __ Type __ CloseParen
```

I'm a little confused why `Type` doesn't allow leading `__` already.  Ah, I see, it's because `TypePrimary` has lots of `_?` prefixes but sometimes-not-always a `__` prefix.  Seems a bit messy, but this should work...

edemaine (approved):
Nice catch!

source/parser.hera
STRd6: I want it to check `Type` first to handle nested implicit interfaces. `__ Type` would swallow the `EOS` and prevent `Nested` from matching.

edemaine: Oh, I see. Interesting! Resolved.

edemaine: (but maybe worth adding a comment)

PR #360, opened 2023-02-12
Fix multiple spread object regression

STRd6:


edemaine (approved):
Is there a test for multiple spreads in object pattern throwing an error, which I assume this error is intended for? I couldn't find one in `object.civet` but may have missed it.

source/main.coffee
```diff
@@ -161,7 +161,7 @@ export compile = (src, options) ->
 
   if options.errors?.length
     #TODO: Better error display
-    throw new Error "Parse errors: #{options.errors.join("\n")} "
+    throw new Error "Parse errors: #{options.errors.map((e) -> e.message).join("\n")} "
```
edemaine, line 164: Do these come through on the LSP now too? It'd be especially cool if we could have error recovery in addition to reporting. Like we return an error in the parse tree but the parser keeps going. I'm thinking of `foo.` with no property after, to enable autocompletion.

test/helper.civet
```diff
@@ -66,8 +66,8 @@ throws := (text: string, opt?: "only" | "skip") ->
 
     """
 
-throws.only = (text: string) -> testCase text, "only"
-throws.skip = (text: string) -> testCase text, "skip"
+throws.only = (text: string) -> throws text, "only"
+throws.skip = (text: string) -> throws text, "skip"
```
edemaine, line 70: Oops, thanks for fixing this copypasta! At least it failed in a noisy way.

source/main.coffee
STRd6: They're not surfaced in the LSP yet but that's the planned direction.

PR #364, opened 2023-02-12
return.value and return =

### LINKED ISSUE #361: return =
It's common to want to do some cleanup before actually returning a value.  Declaring a variable to store that value seems superfluous, and less declarative.  Let's use `return` like a variable we can assign to!

```coffee
function f(list)
  for item of list
    if match item
      return = item
      break
  if (return.value) return += 1
  list.destroy()
  return
---
function f(list) {
  var returnValue
  for (item of list) {
    if (match(item)) {
      returnValue = item
      break
    }
  }
  if (returnValue) returnValue += 1
  list.destroy()
  return returnValue
}
```

* [x] `return = foo` changes the default return value for a bare `return`
* [x] `return += foo` modifies that value
* [x] `return.value` lets you query that value. So `return = foo` is shorthand for `return.value = foo`.
* [ ] `return?` could test for whether a value has been specified i.e. shorthand for `return.value?`

Open question: If there's no explicit `return` at the end of a function using `return =`, should we return the last expression as usual, or should we return `return.value`?  For example, if there wasn't a bare `return` in the example above, we could plausibly compile to:

```coffee
  return ref = list.destroy(), returnValue ?? ref
```

I'm not sure this is particularly intuitive, though.  Unlike JavaScript and Python, our functions don't have an implicit `return` at the end.  And I think it makes more sense for `return = foo` to set the default behavior of `return`, not implicit return.

On the other hand, I think we could override in the case of no implicit return, like so:

```coffee
  list.destroy();
---
  list.destroy();
  return returnValue
```

edemaine:
I used what I learned from writing #359 to fix #361 and this time keep things particularly clean.

Extra features:

* Simpler global ref resolution from #359 (necessary if `return.value` is used within a loop, and `ret` is used only outside)
* Add parent pointers and `findAncestor` helper
* Remove parent pointers during prune
* CLI error handling when AST fails to `JSON.stringify` (helped figure out that I need the previous)
* `getIndent` helper extracted from `insertReturn`
* `gatherRecursive` gains a `skipPredicate` (not strictly necessary but convenient)
* `gatherRecursiveWithinFunction` to search just within this function (probably would simplify `autoVar` code)
* `isFunction` helper

### DIFF
--- a/source/parser.hera
+++ b/source/parser.hera
@@ -6877,7 +6877,7 @@ Init
         return node.flatMap((n) => gatherRecursive(n, predicate))
       }
 
-      if (skipPredicate?.(node)) return
+      if (skipPredicate?.(node)) return []
 
       if (predicate(node)) {
         return [node]
--- a/source/parser.hera
+++ b/source/parser.hera
@@ -7626,22 +7626,7 @@ Init
     }
 
     function populateRefs(statements) {
-      const refNodes = gatherNodes(statements, ({type}) => type === "Ref")
-      const blockNodes = new Set(gatherNodes(statements, ({type}) => type === "BlockStatement"))
-      const forNodes = gatherNodes(statements, ({type}) => type === "ForStatement")
-
-      // Populate refs from inside out
-      forNodes.forEach(({declaration, block}) => {
-        // Need to include declarations with block because they share scope
-        if (block.type === "BlockStatement") {
-          populateRefs([declaration, ...block.children])
-        } else { // single non-block statement
-          populateRefs([declaration, ...block])
-        }
-        blockNodes.delete(block)
-      })
-
-      blockNodes.forEach(({expressions}) => populateRefs(expressions))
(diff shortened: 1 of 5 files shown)


source/parser.hera
```diff
@@ -6859,13 +6870,15 @@ Init
...
     // i.e. if the predicate matches a node it is not recursed into further
-    function gatherRecursive(node, predicate) {
+    function gatherRecursive(node, predicate, skipPredicate) {
       if (node == null) return []
 
       if (Array.isArray(node)) {
         return node.flatMap((n) => gatherRecursive(n, predicate))
       }
 
+      if (skipPredicate?.(node)) return
```
STRd6, line 6880: We might want to `return []` here so that:

```
gatherRecursive(s, ->, -> true).forEach(...)
```

will work. Also it won't insert `undefined` into the flatmap.

STRd6 (approved):
Just need to update the return value when matching `skipPredicate` to `[]`.

PR #366, opened 2023-02-12
Declare and update return.value

edemaine:
Sequel to #364. I started writing documentation and then added a few intended features:
* `let return: T`
* `return .= init`
* `return++`

Also cleans up the code in a single `ReturnValue` rule.

### DIFF
--- a/source/parser.hera
+++ b/source/parser.hera
@@ -594,7 +594,7 @@ CallExpression
     })
   # NOTE: Special `return.value` (and `return =` shorthand)
   # for changing the automatic return value of function
-  ( "return.value" NonIdContinue ) / ( "return" &WAssignmentOp ) ->
+  ( "return.value" NonIdContinue ) / ( Return &( WAssignmentOp / UpdateExpressionSymbol ) ) ->
     return { type: "ReturnValue", children: $0 }
 
   MemberExpression:member AllowedTrailingMemberExpressions:trailing CallExpressionRest*:rest ->
@@ -3441,7 +3441,7 @@ KeywordStatement
 
   # https://262.ecma-international.org/#prod-ReturnStatement
   # NOTE: Modified to leave room for `return.value` and `return =`
-  Return !("." / WAssignmentOp) MaybeNestedExpression?:expression -> {
+  Return !("." / WAssignmentOp / UpdateExpressionSymbol) MaybeNestedExpression?:expression -> {
     type: "ReturnStatement",
     expression,
     children: $0,
--- a/source/parser.hera
+++ b/source/parser.hera
@@ -592,10 +592,7 @@ CallExpression
       type: "CallExpression",
       children: [$1, ...$2, ...rest.flat()],
     })
-  # NOTE: Special `return.value` (and `return =` shorthand)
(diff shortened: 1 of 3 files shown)


source/parser.hera
```diff
@@ -608,6 +605,19 @@ CallExpression
...
+# NOTE: Special `return.value` (and `return =` shorthand)
+# for changing the automatic return value of function
+ReturnValue
+  ( "return.value" NonIdContinue ) / ( Return &AfterReturnShorthand ) ->
+    return { type: "ReturnValue", children: [$1[0]] }
+
+AfterReturnShorthand
+  WAssignmentOp
+  UpdateExpressionSymbol
+  __ Colon
```
STRd6, line 617: This could pick up a `Label` on the following line, we may want to disallow `__` and enforce same line or use a more specific rule.

STRd6:
We may also be able to use the return type annotation of a function, if present, to set the default type annotation for the return value.

edemaine:
> We may also be able to use the return type annotation of a function, if present, to set the default type annotation for the return value.

Great idea!  I implemented this, which involved improving the AST for types a bit.  Not sure I got the best names for things (`t` is particularly annoying, because `type` is already taken).  Feel free to suggest (or make) improvements.

source/parser.hera
```diff
@@ -608,6 +605,20 @@ CallExpression
...
+# for changing the automatic return value of function
+ReturnValue
+  ( "return.value" NonIdContinue ) / ( Return &AfterReturnShorthand ) ->
+    return { type: "ReturnValue", children: [$1[0]] }
+
+AfterReturnShorthand
+  WAssignmentOp
+  UpdateExpressionSymbol
+  # NOTE: Colon indicates potential type annotation, e.g. `let return: T`
+  _? Colon
```
STRd6, line 618: This could be `ReturnTypeSuffix`. There isn't really any downside to being more explicit.

source/parser.hera
```diff
@@ -7024,11 +7061,18 @@ Init
 
       // Add declaration of return.value after {
       if (!declared) {
+        let returnType = func.returnType ?? func.signature?.returnType
+        if (returnType) {
+          const {t} = returnType
+          if (t.type === "AssertsType" || t.type === "TypePredicate") {
+            returnType = undefined
```
STRd6, line 7068: The return type should be `boolean` if `t.type === "TypePredicate"`.

source/parser.hera
edemaine: Given that the context here is a `LexicalDeclaration`, I think `TypeSuffix` is appropriate. Agreed?

edemaine:
4469fab prevents newlines before colons in a lot of cases. (I searched for all `__ Colon` and changed most of them — the only remaining case is ternary `?:`.) This actually fixed a bunch of bugs with label parsing (see new tests in `test/label.civet`), and didn't seem to break any tests, so it's hopefully OK in practice.  If needed, in the future, we could allow newlines before the colon if we then check for a space after the colon to ensure it's not a label.

source/parser.hera
STRd6: 👍

PR #368, opened 2023-02-13
Trailing member properties in blocks

### LINKED ISSUE #367: Chained call on next line in arrow function
I call .filter() on a variable in an arrow function with an added level of indentation under it:
```coffee
useEffect :void =>
  criteria.value = debouncedWorkingCriteria
    .filter (_, i) => validCriteria.value.has(i)
```
Civet (playground) interprets this as calling .filter() after the useEffect call:
```ts
useEffect((): void => {
  criteria.value = debouncedWorkingCriteria;
}).filter((_, i) => validCriteria.value.has(i));
```
I expected it to be within the anonymous function like this:
```ts
useEffect((): void => {
  criteria.value = debouncedWorkingCriteria
    .filter((_, i) => validCriteria.value.has(i));
});
```

edemaine:
Fixes #367

* Explicit braces re-enable everything
* Indented block re-enables everything
* Add tests for trailing member properties on/off

### DIFF
--- a/source/parser.hera
+++ b/source/parser.hera
@@ -1487,11 +1487,15 @@ ExplicitBlock
     })
 
 ImplicitNestedBlock
-  InsertOpenBrace NestedBlockStatements:block InsertNewline InsertIndent InsertCloseBrace ->
-    return Object.assign({}, block, {
-      children: [$1, ...block.children, $3, $4, $5],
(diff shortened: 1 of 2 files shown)


STRd6 (approved):
Nice, great fix!

Issue #346, opened 2023-02-08
Pipe operator to-do


edemaine:
- [ ] Refs to avoid duplicate computation, e.g. `x |> f ||> g` evaluates `f(x)` twice
- [x] For efficiency, unwrap `&` blocks (or arbitrary single-argument functions) as described in https://github.com/DanielXMoore/Civet/pull/83#issuecomment-1364739339
  ```typescript
  x |> &+1 |> return
  ---
  return (x)+1
  ```
- [x] Weird behavior that kind of makes sense with implicit function application (the pipes go inside the function parameters):
  ```typescript
  array ||> .push 3 ||> .shift() ||> .sort()
  ---
  ($ => $.push(($1 => $1.shift())(3),($2 => $2.sort())(3),3))(array),array
  ```
  I would argue that `|>` and `||>` should have lower precedence than implicit function calls, similar to postfix `if`. Maybe we can fix it by using `NonPipelineExtendedExpression` in all implicit function calls?
- [x] Even worse with newlines and definitely shouldn't happen this way:
  ```typescript
  array
  ||> .push 3
  ||> .shift()
  ||> .sort()
  ---
  ($ => $.push(($1 => $1.shift())(3),($2 => $2.sort())(3),3))(array),array
  ```
  
(porting leftover issues from #318)

PR #370, opened 2023-02-14
#346 Unwrap amp blocks

STRd6:


edemaine (approved):
Wow, this is pretty slick! Not much code, and beautiful results. Pipe output becomes intelligible! 

I would someday like to unwrap all single-argument functions, but I see that's a lot harder, as the argument could be used more than once (so really needs a ref) and doesn't already share a ref object. So definitely a future PR.

PR #371, opened 2023-02-14
x[..] slice

### LINKED ISSUE #365: x[..]
`x[..]` is a convenient way in CoffeeScript to duplicate an array (like `[...x]`)

Even more useful is the assignment form: `x[..] = []`

Currently, ranges seem to need a value at the end.

Workaround: `x[...x.length]`

edemaine:
Fixes #365

I went with the CoffeeScript approach of just allowing `..`, not `...`, to avoid conflict with spreads, though we could add both...

### DIFF
--- a/source/parser.hera
+++ b/source/parser.hera
@@ -776,7 +776,7 @@ SliceParameters
       children: [start, [...ws, {...sep, token: ", "}], [end, ...inc]]
     }
 
-  Loc:l __:ws DotDot ->
+  Loc:l __:ws ( DotDot / DotDotDot ) &( __ CloseBracket ) ->
     const start = {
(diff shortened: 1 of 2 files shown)


STRd6:
I think we should also add `...` to be more compatible with CoffeeScript. We can also add a `&( __ CloseBracket )` assertion to avoid a potential spread conflict.

edemaine:
Thanks for taking care of this. I don't think `&( __ CloseBracket )` is necessary because `SliceParameters` is only used within `MemberBracketContent`. I'd actually worry more about content beforehand, like `foo...`, but that's not possible in `MemberBracketContent`.

PR #372, opened 2023-02-15
implicit return and results aggregation in pattern matching

STRd6:


test/switch.civet
```diff
@@ -629,3 +629,68 @@ describe "switch", ->
...
+        switch x
+          1
+            2
+          else
+            3
+      ---
+      (function() {
+        if(x === 1) {
+            return 2}
+      else  {
```
edemaine, line 646: Is the indentation off here? Maybe it's getting inserted after the `else` instead of before?

edemaine (approved):
That was nice and easy! This makes pattern matching far more useful.

test/switch.civet
STRd6: Yeah, the indentation is way off on the else branches currently.

PR #374, opened 2023-02-16
Namespace and declare support

### LINKED ISSUE #155: TypeScript `namespace` and `export interface`
Here's a small example [from TypeScript](https://www.typescriptlang.org/docs/handbook/declaration-merging.html#merging-namespaces):

```ts
namespace Animals {
  export interface Legged {
    numberOfLegs: number;
  }
  export class Dog {}
}
```

* [x] ~~`namespace` doesn't have an indentation form; it needs braces~~
* [x] `export interface` doesn't parse in general
* [x] `export class Dog` doesn't parse inside `namespace Animals {...}`

### LINKED ISSUE #156: TypeScript `declare`
* [x] `declare let $: JQuery` doesn't parse
* [x] `declare global {}`, `declare module "hot-new-module";` and `declare module "path" {}` parse as function calls
* [x] `declare class`

edemaine:
Fix #155 and fix #156 and various other TypeScript compatibilities discovered when copy/pasting examples.

* `namespace` block is like a top-level module, e.g., can support top-level `export` and can have arbitrary executable code. (Previously could just contain type declarations.)
* `declare` blocks can only contain type declarations, and everything inside is automatically as such (as if it had a `declare` prefix, and can't contain another `declare` prefix).
* `declare class` is a new thing which I call `ClassSignature` by analogy to `FunctionSignature`. It can have methods but they act like function signatures so don't get braced blocks.
* `declare var` was missing (previously only `let`/`const`).
* Function arguments can have a type of just `?` without any colon part, e.g. `function f(arg?)`.


STRd6 (approved):

STRd6:
Looks great, thanks for this!

Issue #335, opened 2023-02-05
Pattern matching bugs


edemaine:
* [x] Empty array pattern `[]`
* [ ] Implicit return:
  ```coffee
  function categorize(x)
    switch x
      1
        'one'
  ```
* [ ] Expression form:
  ```coffee
  y :=
    switch x
      1
        'one'
  ```
* [ ] Multiple use of same property makes duplicate `const`:
  ```coffee
  switch x
    [{type: "text"}, {type: "image"}]
      x
  ```
* [ ] Use of keyword `class` makes invalid `const`:
  ```coffee
  switch x
    {class: 'hi'}
      x
  ```

STRd6:
For `class` and multiple of the same keys we should have some way of aliasing but I'm not sure of the best syntax for it yet.

edemaine:
Perhaps `{class: klass === "hi"}` or `{class: klass is "hi"}`? Eh, maybe that looks like an expression that should be evaluated and passed in as the class vakue to match.

I think we also need to improve the default behavior when there are duplicate or keyword names. Some options:

1. Keyword names could get an automatic `_` or `1` suffix or something.
2. Just the first instance of a name could get a variable.
3. First instance gets a `1` suffix, second gets a `2` suffix, etc.
4. Name gets assigned an array of values (but this changes substantially between between one and multiple instances of tge name). [as you've suggested in Discord]
5. Name with suffix of `0` gets an array of matches, like how in Hera `$0` is `[$1, $2, ...]`.

I also feel like the most common case of duplicate names is when we have bound them to constants already. Something that would help in this sort of scenario is to not actually bind variables if they don't get used in the scope of the match. This would also pair nicely with Option 5 above; we'd like to build this array only if it gets used.

edemaine:
#375 implemented Option 4 above, which I agree seems like a good choice.

A simple workaround for `class` and other keywords, at least for now, would be to just not bind those (or use a Ref that the user can't access). At least then the code would compile.

eevleevs:
What about adding the same functionality as in #444? To allow something like:

```js
switch a
  m := /(\d)/
    return m[1]
```

JUSTIVE:
any plans for guard clauses other than simple binary operators?

JUSTIVE:
Great to hear that! I just needed the latter one (invoking a function), and the `isCool(&)` notation looks fine. 
[...]

PR #375, opened 2023-02-16
Aggregate duplicate pattern matching bindings

STRd6:
This aggregates duplicate bindings with the same identifier in pattern matching. Since there can only ever be one identifier with the same name in a scope merging more than one bindings into an array makes a lot of sense to me. Adding an implicit suffix seems like it would be hard to discover when adding or removing identifiers. TS should provide decent support and highlighting if you weren't expecting an array.

There are probably some other cases where this doesn't work but I'll fix those at a later time.

Part of #335

### DIFF
--- a/source/parser.hera
+++ b/source/parser.hera
@@ -7579,18 +7579,18 @@ Init
       const props = gatherRecursiveAll(bindings, n => n.type === "BindingMatchProperty")
       const declarations = []
 
-      const propsGroupedByName = props.reduce((acc, p) => {
+      const propsGroupedByName = new Map
+
+      for(const p of props) {
         const { name } = p
 
         const key = name.name
-        if (acc.has(key)) {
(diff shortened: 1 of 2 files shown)


source/parser.hera
```diff
@@ -7575,6 +7575,48 @@ Init
...
+
+        const key = name.name
+        if (acc.has(key)) {
+          acc.get(key).push(p)
+        } else {
+          acc.set(key, [p])
+        }
+
+        return acc
+      }, new Map())
```
edemaine, line 7593: My personal preference for writing this would be to have a `const acc = new Map` outside and use `for(...of...)` or `forEach` to iterate through `props`. But `reduce` is fine too; just slightly less efficient (?) and I find it a bit harder to read.

edemaine (approved):
Looks great!!

Issue #202, opened 2023-01-11
try expressions don't support return


edemaine:
```coffee
compiled :=
  try
    compiled = require('@danielx/civet').compile str, options
  catch err
    return "error"
```

transpiles to

```js
const compiled = (() => {
  try {
    return (compiled =
      require("@danielx/civet").compile(
        str,
        options
      ));
  } catch (err) {
    return "error";
  }
})();
```

I would have expected the `return` in the `catch` block to escape the entire function, not just the IFFE.

This is a general issue with IFFEs. I'm not sure what the best solution is. For now I guess we could throw an error if there's a `return` inside the `try` block...

STRd6:
CoffeeScript's solution is to raise a parse error but I think we can do better.

```javascript
const returnSymbol = Symbol("return")

const compiled = (() => {
  try {
    return {[returnSymbol]: (compiled =
      require("@danielx/civet").compile(
        str,
        options
      ))
    };
  } catch (err) {
    return "error";
  }
})();

if (typeof compiled === "object" && returnSymbol in compiled) return compiled[returnSymbol]
```

Something along those lines.

edemaine:
Agreed! Another option would be to use exceptions (again 😅 ):

```js
try {
  const compiled = (() => {
    try {
      return (compiled =
        require(["@danielx/civet"](https://www.npmjs.com/package/@danielx/civet)).compile(
          str,
          options
        ));
    } catch (err) {
      throw new ReturnError("error");
    }
  })();
} catch (e) {
  if (e instanceof ReturnError) return e;
  throw e;
}
```

But I think I prefer your solution so we don't mess with the source of exceptions...

STRd6:
Also related `break label` inside of nested iteration iifes (contingent on our eventual label support).

juh9870:
The best solution mige be not using IFFEs for this at all.
Code in the original post can be compiled into something like
```ts
let compiled$temp;

try {
  compiled$temp = require("@danielx/civet").compile(str, options);
} catch (err) {
  return "error";
}
const compiled = compiled$temp;
```

PR #376, opened 2023-02-16
do expressions wrapping in iffe

### LINKED ISSUE #58: `do` expressions without `while`/`until`
Oren wanted something like this on Discord: (originally suggested without `do` keyword)

```coffee
x = do
  y := 5
  y + 1
```

It does seem like a natural extension to `do ... while` currently supported, i.e., make the `while` optional. It also corresponds to [this proposal](https://github.com/tc39/proposal-do-expressions) (though we don't need that proposal for `if`/`while` expressions).

Related, I expected this to work, but it doesn't parse: (it works without the `:=`)

```coffee
x = if true
  y := 5
  y + 1
```

Current workarounds, which work but are a bit inefficient as they create arrays:

```coffee
x = (do
  y := 5
  y + 1
while false)[0]
```

```coffee
x = (for nothing of [1]
  y := 5
  y + 1
)[0]
```

edemaine:
Fixes #58, via the simpler approach. It's equivalent to wrapping in braces, unless it's used as an expression, in which case it gets an IFFE.

We can later explore the plan to remove IFFEs when possible, but this is nice for allowing declarations (even local `var`s) in the block. I think we also want to do this implicitly for `if` like CoffeeScript does, but that's (yet) another issue.

#202 tracks the issue that `return`s are forbidden within expressioned statements like this.

### DIFF
--- a/source/parser.hera
+++ b/source/parser.hera
@@ -2797,6 +2797,8 @@ IterationStatement
   # NOTE: Added `loop` from CoffeeScript
   LoopStatement
   !CoffeeDoEnabled DoWhileStatement -> $2
+  # DoStatement must come after DoWhile statement
+  !CoffeeDoEnabled DoStatement -> $2
   WhileStatement
   ForStatement
 
@@ -2804,6 +2806,7 @@ IterationExpression
   IterationStatement ->
     return {
       type: "IterationExpression",
+      subtype: $1.type,
       children: [$1],
       block: $1.block,
     }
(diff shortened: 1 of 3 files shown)


source/parser.hera
```diff
@@ -6284,37 +6296,53 @@ Init
...
-        base: "results",
-        id: "results",
-      }
+      if (statement.type === "DoStatement") {
+        if (outerRef) {
+          insertPush(statement.block, outerRef)
+        } else {
+          insertReturn(statement.block)
+        }
+      } else {
```
STRd6, line 6326: We can avoid `else` and nesting further with an early return at the end of the `if`.

STRd6:
We may later want to support `break`/`continue` inside `do` blocks. Not necessary for this PR but something to consider in the future.

STRd6 (approved):

PR #382, opened 2023-02-17
Docs: Comparison to JavaScript/TypeScript

edemaine:
I'm tempted to move the comparison to CoffeeScript in here too... What do you think?


STRd6:
It looks like the  `git add` ghost struck again 👻

edemaine:
Oops, sorry about that. Someday I'll learn...

edemaine:
Another possibility for the Comparison to CoffeeScript document would be for it to be a separate docs page, but instead of it being linked from the navbar, the only link is from this page.

STRd6 (approved):
Nice, looks good!

PR #385, opened 2023-02-18
Fixes #369

### LINKED ISSUE #369: returning value from try
The following Civet code...
```
function someFunction()
  return 5
let x = try someFunction()
```

... generates ...
```
function someThrowingFunction() {
  return 5;
}

let x = (() => {
  try {
    return someThrowingFunction();
  } catch (e) {}
})();
```

... which runs into the error `Not all code paths return a value.(7030)` because the catch block doesn't include `return undefined` or `return null`.

STRd6:
This adds implicit `return`/`results.push` inside implicit catch blocks.

A couple things to note:

- CoffeeScript doesn't push for implicit catch so we may want to call that out in the docs.
- Truly implicit returns may imply that `->` should become `function(){return}` but I'll need to think it through a bit before committing to such a thing.

### DIFF
--- a/source/parser.hera
+++ b/source/parser.hera
@@ -3350,28 +3350,53 @@ ImpliedColon
 TryStatement
   Try:t NoPostfixBracedOrEmptyBlock:b CatchClause?:c FinallyClause?:f ->
     if (!c && !f) {
+      const e = []
+      const emptyCatchBlock = {
+        type: "BlockStatement",
+        expressions: e,
+        children: ["{", e, "}"],
+        bare: false,
+        empty: true,
+      }
+
+      c = {
+        type: "CatchClause",
+        children: [" catch(e) ", emptyCatchBlock],
+        block: emptyCatchBlock,
+      }
+
       return {
         type: "TryStatement",
-        children: [t, b, " catch(e) {}"]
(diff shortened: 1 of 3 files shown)


edemaine (approved):
Looks great!

I don't see a harm in adding an implicit return to an empty function block, though it seems TypeScript doesn't care about that case. I've noticed `return` added already at the end of code lile this:

```coffee
function f
  if x
    5
```

...though I've never quite understood why. Maybe it was also to appease TypeScript?

STRd6:
The reason for the return after the `if` without an `else` is to prevent the "Not all code paths return a value" error from TS: https://www.typescriptlang.org/play?#code/G4QwTgBCBcUHYE8BQSBmBXOBjALgSwHs4JUAKASggG8kII8yRKa66wBTHdMYgRlogBfJIKA

Longer term as we sort out the blocks better I think we can converge the complicated switch statements into something slightly simpler by having nodes with blocks have them exposed as a key and an additional boolean `exhaustiveBlocks` to indicate if a trailing `push`/`return` is needed.

I think I'd prefer `->` to transpile to `function(){}` without the return for the sake of minimizing the output since it all seems to be the same to TS.

PR #390, opened 2023-02-18
Ligatures font toggle

edemaine:
* Ligatures off by default (I've gotten a bunch of feedback on this).
* Toggle in bottom right of every Playground box. (I tried to explore doing this via locales and the topnav, but doesn't seem to be possible.)

image

image

TODO: Keep toggle state in localstorage. (I feel like a serious Vue noob here...)

TODO: I think the fonts should be unified; they're substantially different heights. @STRd6 Do you like my choice of Hack? If so, there's a [FiraCode Ligatured Hack](https://github.com/gaplo917/Ligatured-Hack) that I can replace FireCode with.


STRd6:
I personally prefer Fira Mono if we want to unify the ligature/non-ligature fonts.

edemaine:
Ah, I didn't know about Fira Mono.  That works.

I prefer Hack (taller letters, less space between lines, somewhat bolder weight) but am fine going with Fira. I eventually get used to whatever I look at, but switching between them is really striking. (Try it with the two images above.)

STRd6 (approved):

edemaine:
Sorry, could I get a clarification of whether you strongly prefer Fira or leave it up to me? I assume the former but want to make sure.

PR #395, opened 2023-02-19
Parent pointer cleanup

STRd6:
- Don't add parent pointers to arrays
- Update parent pointers after adjusting subtrees in pipe and pattern matching


edemaine (approved):
I did wonder whether it was appropriate to add properties to arrays. It works, but they certainly print funny. And this is a good way to encourage use of more AST object nodes.

PR #398, opened 2023-02-19
Indented application

edemaine:
At this point, this PR is more of a proof-of-concept, but it does seem possible to unambiguously enable

```coffee
f
  a
  b
  g
    c
    d
---
f(
  a,
  b,
  g(
    c,
    d,))
```

Unfortunately, the number of rules added to the no-cache set is probably too big; tests get substantially slower, and there's one test that needs >2 seconds, so Mocha times out unless you specify `yarn test --timeout 5000`.

My guess is that there's something too repetitive in the way that arguments are parsed (e.g. in `NonPipelineArgumentList`) that requires too much revisiting of the same code.  The challenge here seems to be keeping the indentation around long enough; if we end up parsing the innards again after popping the indentation, we mistakenly treat something as indented.  An example that was hard to get working:

```coffee
f [
  o a, -> x
  p b, -> y
]
```

If `x` gets parsed a second time after the array indentation has been lost, it treats `p` as an argument.

Maybe this is worth revisiting after #341 gets fixed?

### DIFF
--- a/source/main.coffee
+++ b/source/main.coffee
@@ -70,6 +70,7 @@ uncacheable = new Set [
   "MemberExpression"
   "MemberExpressionRest"
   "Nested"
+  "NestedArgumentList"
   "NestedBindingElement"
   "NestedBindingElements"
   "NestedBlockExpression"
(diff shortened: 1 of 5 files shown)


STRd6:
I think the current arguments parsing is pretty hacky so there may be a way to simplify it greatly, especially as we've become more strict about the whitespace of things that go into arguments. I think there is currently some needlessly complicated recursion that is very non-optimal and could be fixed with enough wherewithal.

STRd6:
This might be in a position to revisit after the improved caching.

STRd6:
Finished in #677

PR #399, opened 2023-02-20
Automatically await/async expressionized statements with await

### LINKED ISSUE #355: Expressionized statements and await
Should  expressionized statements that contain an await also await?

```civet
let a = try { await b } catch { c }
```

```
let a = await (async () => {
  try {
    return await b;
  } catch () { return c }
})();
```

- [ ] Fix `catch { c }`
- [x] Implicit `async`
- [x] Add implicit `await` when `await` is used in an expressionized statement.

edemaine:
Fixes #355 by replacing all IIFE code (that I could find anyway) with a new `module.wrapIIFE` helper that checks for `await` in the wrapped expression (without crossing a function boundary) and wraps accordingly.

One thing: I'm not sure what was meant by "Fix `catch { c }`" in #355. Is this about whether `catch` should return a value? Should we split that off into a separate issue?

### DIFF
--- a/source/parser.hera
+++ b/source/parser.hera
@@ -1409,10 +1409,14 @@ FunctionExpression
     }
 
     const { ref } = rhs
+    const children = [ref, " => ", prefix, rhs]
+    if (module.hasAwait(rhs)) {
+      children.unshift("async ")
+    }
 
     return {
       type: "ArrowFunction",
-      children: [ref, " => ", prefix, rhs],
+      children,
       ampersandBlock: true,
     }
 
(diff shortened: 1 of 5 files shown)


STRd6:
Here's a minimal example of the `catch` issue: https://civet.dev/playground/?code=dHJ5IGEgY2F0Y2ggeyBjIH0%3D

```
try a catch { c }
--- actual
try {
  a;
} catch ({ c }) {}
--- expected (or similar)
try {
  a;
} catch(e) {
  c;
}
```

It should be split off as a separate GH issue.

STRd6:
Looks good! Also check out adding to ampersand blocks inside `FunctionExpression` around line 1374 in `parser.hera`

PR #402, opened 2023-02-20
async do, async for

### LINKED ISSUE #292: async do
It's currently annoying to build a promise from async code: you need to wrap it in a function. CoffeeScript's `do` is a nice way to do this:

```coffee
await Promise.all(
  for item in array
    do ->
      data = await op1(item)
      await op2(item, data)
)
```

I propose that we support this in non-`coffeeCompat` `do` as well, like so:

```coffee
await Promise.all(
  for item in array
    async do
      data = await op1(item)
      await op2(item, data)
)
```

(And `async` maybe becomes optional when we have an `autoAsync` flag.)

edemaine:
`async` variation of iteration expressions, including simple `do`, which build a promise for the results by making the IIFE wrapper `async`.

For example, `async do` lets you build a Promise easily using `await` syntax:

```coffee
promise := async do
  await f1()
  await f2()
---
const promise = (async ()=>{ {
  await f1()
  return await f2()
}})()
```

Fixes #292 which shows a nice example of using this in a `for` loop to do `await Promise.all` in parallel.

Similarly, `async for` returns a promise that resolves to the returned array, once all `await`s are complete (in sequence):

```coffee
promise := async for x of y
  await f(x)
---
const promise = (async ()=>{const results=[];for (const x of y) {
  results.push(await f(x))
}; return results})()
```

Having this as an option is essentially what @gustavopch was asking for in #355.  I think it makes sense to `await` by default, because there isn't obviously a function involved here.  But it *also* makes sense to allow the user to ask that a task be done `async` and get that promise.  An `async` prefix seems to reasonably naturally mirror `async function`, and given that I had previously come up with the same idea in #292 (and then promptly forgotten) it seems pretty natural. 🙂

### DIFF
--- a/source/parser.hera
+++ b/source/parser.hera
@@ -13,7 +13,7 @@ Program
 TopLevelStatement
   EOS? ModuleItem StatementDelimiter
 
-# Expressions with If and Switch
+# Expressions with If and Switch, but no comma operator
 ExtendedExpression
   NonAssignmentExtendedExpression
   AssignmentExpression
@@ -3514,9 +3514,12 @@ RestoreAll
 ExpressionStatement
   # NOTE: semi-colons are being handled elsewhere
   # NOTE: Shouldn't need negative lookahead if shadowed in the proper order
-  # NOTE: Using ExtendedExpression instead of Expression to allow for
-  # e.g. await forms of IterationExpressions that can't be IterationStatement
-  ExtendedExpression
+  # NOTE: Allow for e.g. await forms of IterationExpression that weren't
+  # already parsed as IterationStatement. Must be before Expression to avoid
+  # treated `async do ...` like a function call `async(do ...)`.
+  IterationExpression
+  # NOTE: Expression enables , operator
+  Expression
 
 KeywordStatement
   # https://262.ecma-international.org/#prod-BreakStatement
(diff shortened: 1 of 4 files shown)


source/parser.hera
```diff
@@ -3513,7 +3514,9 @@ RestoreAll
 ExpressionStatement
   # NOTE: semi-colons are being handled elsewhere
   # NOTE: Shouldn't need negative lookahead if shadowed in the proper order
-  Expression
+  # NOTE: Using ExtendedExpression instead of Expression to allow for
+  # e.g. await forms of IterationExpressions that can't be IterationStatement
+  ExtendedExpression
```
STRd6, line 3519: This change will forbid `a,b` expressions pretty much everywhere. I'm not necessarily against it just want to note that it's a bigger change than it appears.

STRd6: Adding to `Expression` is probably the minimally invasive change.

STRd6: Actually maybe inside `AssignmentExpression` adjacent to`PipelineExpression`.

STRd6 (approved):

source/parser.hera
STRd6: Looks good!

PR #403, opened 2023-02-20
Pipe ref

STRd6:
Adds pipe references.

- [ ] Ref declaration hoisting
- [x] |>= refs

### DIFF
--- a/source/parser.hera
+++ b/source/parser.hera
@@ -7186,6 +7186,31 @@ Init
       }
     }
 
+    function removeParentPointers(node) {
+      if (node == null) return
+      if (typeof node !== "object") return
+
+      // NOTE: Arrays are transparent and skipped when traversing via parent
+      if (Array.isArray(node)) {
+        for (const child of node) {
+          removeParentPointers(child)
+        }
+        return
+      }
+
+      node.parent = null
+      if (node.children) {
+        for (const child of node.children) {
+          removeParentPointers(child)
+        }
+      }
+    }
+
+    function clone(node) {
+      removeParentPointers(node)
+      return structuredClone(node)
+    }
+
     // Find nearest strict ancestor that satisfies predicate,
     // aborting (and returning undefined) if stopPredicate returns true
     function findAncestor(node, predicate, stopPredicate) {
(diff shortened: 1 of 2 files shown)


STRd6:
Hoisting ref decs is still TODO but that could be a separate follow on.

source/parser.hera
```diff
@@ -7951,33 +7952,102 @@ Init
...
+                children: [ lhs, children ],
+                names: null,
+                lhs,
+                assigned: arg,
+                exp: children,
+              })
+
+              // Clone so that the same node isn't on the left and right because splice manipulation
+              // moves things around and can cause a loop in the graph
+              arg = structuredClone(arg)
```
edemaine, line 8013: I'm worried what happens with parent pointers here. Does structuredClone end ip traversing the entire AST? I wonder if there's a way to avoid this... Can you just clone what you end up changing?

source/parser.hera
```diff
@@ -7951,33 +7952,102 @@ Init
...
+            let initRef
             if (i === 0) {
-              // TODO: assignment node
-              children.push(arg, " = ")
+              switch(arg.type) {
+                case "MemberExpression":
+                  // If there is only a single access then we don't need a ref
+                  if (arg.children.length <= 2) break
+                case "CallExpression":
+                  const access = arg.children.pop()
```
edemaine, line 7976: Don't you need to check whether this is an access, not a call? Or because this is a left-hand side it can't be a call?

I think it can also be a "?" or a "!" which you probably want to deal with... There's related code in the object literal shorthand. (I'm glad my work for that to canonicalize the children of a MemberExpression/CallExpression is paying off. But we might want to extract a helper function to find rhe last access...)

edemaine (commented):
Looking great so far! I think we probably need ref hoisting before we can merge, or we break some existing pipelines used as expressions.

edemaine:
> I think we probably need ref hoisting before we can merge, or we break some existing pipelines used as expressions.

On second thought, as this only affects fat pipes, which really need refs to be useful, I agree with merging this sooner and leaving hoisting for the next PR.

source/parser.hera
STRd6: I could null out the parent pointers, then `structuredClone`. They'll get added back at the end of the method.

source/parser.hera
STRd6: Yeah, I should check the `access` type. If it is a property or slice then it is fine, but other things not so much.

source/parser.hera
edemaine: Or write our own simple clone method for ASTs... Not sure which is faster.

STRd6: Our own clone could be handy. It should keep refs the same for example.

PR #408, opened 2023-02-23
Leave plain JSX strings alone, including newlines

### LINKED ISSUE #405: Multiline strings in JSX attributes
Apparently this JSX:

```jsx
<div class="hello
world"/>
```

gets translated to a string with a newline in it. (At least [with Solid](https://playground.solidjs.com/anonymous/dd826685-a801-45c2-b389-3984761340b7).)

However, Civet translates it to

```jsx
<div class="hello\nworld">
```

but [`\` escape characters don't work in JSX strings](https://facebook.github.io/jsx/#sec-jsx-string-characters), so this ends up having a literal `\` in it. (At least [with Solid](https://playground.solidjs.com/anonymous/b7e791eb-1c3b-4cd5-b2f6-d612881ff7b6).)

Probably just need to remove newline handling in JSX strings without braces, but need to test the various cases.

edemaine:
Fixes #405 by not processing plain JSX strings at all.

> Another option would be to convert the `\n` to ` ` in JSX.

Interestingly, this is what [CoffeeScript does](https://coffeescript.org/#try:%3Cdiv%20class%3D%22hello%0A%0A%0Aworld%5Cn%5Ct%22%2F%3E). (It even seems to convert multiple newlines into a single space.)  I guess it's fine with `class`, but I like better the approach of just leaving the string alone in this case.

One inconsistency I noticed is that `\t` and other escape characters are normally ignored (the backslash remains a backslash) in JSX strings.  But if you use an interpolated string (CoffeeScript's `"#{foo}"` or Civet's always-available `"""#{foo}"""`), which turn into a braced template string, then the backslashes are processed.  This is [consistent with CoffeeScript](https://coffeescript.org/#try:%3Cdiv%20class%3D%22hello%0A%0A%0A%23%7Bfoo%7D%0A%0Aworld%5Cn%5Ct%22%2F%3E) so I think it's OK.  Just a bit weird.

### DIFF
--- a/source/parser.hera
+++ b/source/parser.hera
@@ -5070,12 +5070,15 @@ JSXAttributeValue
     // Check for string literal resulting from CoffeeScript interpolated
     // double-quoted string that didn't end up actually interpolating.
     if ($2.children?.length === 1 && $2.children[0].type === "StringLiteral") {
-      return $2.children[0]
+      // Instead parse using the following string literal rule,
+      // which avoids e.g. converting newlines into \n.
+      return $skip
     }
     return $0
   # NOTE: InlineJSXAttributeValue which contains TemplateLiteral must be before
   # StringLiteral, so that CoffeeScript interpolated strings get checked first.
-  StringLiteral
+  # NOTE: JSX strings allow newlines, and they get passed through as-is.
+  /"[^"]*"|'[^']*'/
 
 # JSX shorthand to avoid explicit braces when unnecessary (no whitespace)
 InlineJSXAttributeValue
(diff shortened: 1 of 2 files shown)


STRd6:
I think that's a bug in CoffeeScript and the backslashes should be escaped when put into template strings but we can adjust them as a separate PR.

STRd6 (approved):

edemaine:
Maybe? In JSX, `="\n"` is unprocessed but `={"\n"}` and ``={`\n`}`` are processed.  So if you think of an interpolated string as an implicitly braced template string (which is what it compiles to), it makes sense to process the `\n`s.

If we really wanted consistency, I'd be tempted to process backslashes even in unbraced strings (as we implicitly brace a lot of expressions). But then it's less HTML/XML compatible, so I'm not sure it's a good idea.

FWIW I've never run into this in practice (never tried to put backslash sequences in a JSX string), so hopefully not a big deal either way. (But I did see someone in the SolidJS Discord using newlines in a `class` field recently, because it's a nice way to list a lot of classes.)

STRd6:
Those are some good points. We can wait until real use cases come up before deciding.

PR #409, opened 2023-02-23
Pattern Matching

STRd6:
Allow pattern matching for reserved word keys by binding to inaccessible refs.

Link: #335

### DIFF
--- a/source/parser.hera
+++ b/source/parser.hera
@@ -7845,6 +7845,25 @@ Init
       }
 
       propsGroupedByName.forEach((shared, key) => {
+        if (!key) return
+        // NOTE: Allows pattern matching reserved word keys by binding to inaccessible refs
+        // HACK: using the parser's ReservedWord rule here
+        if (ReservedWord({
+          pos: 0,
+          input: key,
+        })) {
+          shared.forEach((p) => {
+            const ref = {
+              type: "Ref",
+              base: `_${key}`,
+              id: key,
+            }
+            p.children.push(": ", ref)
+          });
+          // Don't push declarations for reserved words
+          return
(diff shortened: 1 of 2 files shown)


source/parser.hera
```diff
@@ -7845,6 +7845,25 @@ Init
...
+        // HACK: using the parser's ReservedWord rule here
+        if (ReservedWord({
+          pos: 0,
+          input: key,
+        })) {
+          shared.forEach((p) => {
+            const ref = {
+              type: "Ref",
+              base: `_${key}`,
+              id: key,
```
edemaine, line 7859: This seems like as good a time as any to ask what the difference between `base` and `id` in Refs. I've never understood what `id` gets used for.

source/parser.hera
```diff
@@ -7845,6 +7845,25 @@ Init
       }
 
       propsGroupedByName.forEach((shared, key) => {
+        if (!key) return
+        // NOTE: Allows pattern matching reserved word keys by binding to inaccessible refs
+        // HACK: using the parser's ReservedWord rule here
+        if (ReservedWord({
+          pos: 0,
+          input: key,
+        })) {
```
edemaine, line 7854: This is fun!

edemaine (approved):
Nice solution.

source/parser.hera
STRd6: `id` is only used in `adjustAtBindings` so I could probably omit it in all other cases. It's purpose there is to hold onto the property key to expand `{@class}` into `{<id>: <name>}` where name is computed by finding the first `base` that is available, incrementing a suffix until found.

PR #410, opened 2023-02-23
enum support

edemaine:
Fixes #227.

The fun part here is the JavaScript code generation.  I tried to (roughly) follow the rules in [the documentation](https://www.typescriptlang.org/docs/handbook/enums.html#computed-and-constant-members), but I didn't enforce as many rules because I believe these simple rules work whenever TS considers the enum valid, and we can rely on TS checking for validity.

Note that TS [doesn't generate back references for string values](https://www.typescriptlang.org/play?#code/CYUwxgNghgTiAEEQBd4A8Bc8CuA7A1rgPYDuuANPAJ5Z6GkXwBmRRWUuVAsAFC8i5sAW3gARAJZwwycUVzwA3r3jwAqgAd4AXngAibOt3llYhtvgADYAwvGeKgDIgmqHQEY7KgEriA5gAtXPRg-QKNeAF9eIA) (and it doesn't support mixing strings with "computed" types, so we don't need to detect e.g. a variable of type `string`).

TS also puts all of this inside a function wrapper, but I don't see the point of that. There can't be any declarations, this can't be an expression, and this code always goes right after the declaration.

### DIFF
--- a/source/parser.hera
+++ b/source/parser.hera
@@ -1715,6 +1715,7 @@ Literal
   LiteralContent ->
     return {
       type: "Literal",
+      subtype: $1.type,
       children: $0,
       raw: $1.token,
     }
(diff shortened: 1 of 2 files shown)


STRd6 (approved):
Looks great!

We're getting closer and closer to 99% TS compatibility...

PR #411, opened 2023-02-23
Call with unparenthesized iteration expression argument

edemaine:
I've long wanted to be able to do this (CoffeeScript doesn't allow it):

```coffee
func for x of y
  x * 2
```

As @STRd6 suggested, this checks for `IterationExpression`s without any body (e.g. a single semicolon) and skips in that case to enable a postfix iteration. Seems relatively easy!

I am a little worried about weird interaction with caching if we end up parsing the iteration body multiple times and getting the indentation wrong the second time (which happened in #398)... but so far it seems to work in all cases. (And this would hopefully be fixed by #341.) A workaround if it ever happens would be to add an explicit semicolon after a postfix iteration.

### DIFF
--- a/test/for.civet
+++ b/test/for.civet
@@ -181,24 +181,30 @@ describe "for", ->
     postfix
     ---
     console.log(i) for let i = 0; i < 10; i++
+    nextLine
     ---
     for (let i = 0; i < 10; i++) { console.log(i) }
+    nextLine
   """
 
   testCase """
     postfix for in
     ---
     console.log(i) for i in x
+    nextLine
     ---
     for (const i in x) { console.log(i) }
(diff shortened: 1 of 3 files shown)


test/for.civet
```diff
@@ -181,24 +181,30 @@ describe "for", ->
     postfix
     ---
     console.log(i) for let i = 0; i < 10; i++
+    nextLine
```
STRd6, line 184: We should keep some postfix tests without the next line as well.

STRd6 (approved):

STRd6:
I agree about the potential caching issues as an intermediate concern but long term the solution is to fix the caching directly.

PR #413, opened 2023-02-24
Decorators with member and calls

STRd6:


source/parser.hera
```diff
@@ -2262,7 +2262,8 @@ ComputedPropertyName
     }
 
 Decorator
-  AtAt IdentifierReference Arguments?
+  # Might want to disallow import, super, and return CallExpressions
```
edemaine, line 2265: You can use decorators inside a class method, for example, so I *think* they're all plausible...

edemaine (approved):

PR #414, opened 2023-02-25
New top-level statement system

edemaine:
* Enforce consistent indentation at the top level.
* Allow comments at the beginning of statements.

This was surprisingly difficult to get right, especially while producing output compatible with the past.

I explored redoing statements into actual AST nodes like `{type: "Statement", indent: "  ", statement: <statement without indentation>, children: [eos, indent, leadingComments, statement]}` instead of relying on `[indent, ...rest]`, but it was getting to be a rather huge effort.

### DIFF
--- a/source/cli.coffee
+++ b/source/cli.coffee
@@ -169,7 +169,8 @@ repl = (options) ->
           output = compile input, {...options, filename}
         catch error
           #console.error "Failed to transpile Civet:"
-          return callback error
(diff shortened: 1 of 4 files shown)


STRd6 (approved):
Nice, this makes things a lot better!

Don't worry, one day we'll finally clean up the last remaining whitespace and statement strangeness :)

PR #417, opened 2023-02-28
Fixes #416: Fix handling of backslash escapes inside strings that convert to template strings

### LINKED ISSUE #416: Triple-quoted literals don't escape backslashes
Inserting a backslash into string literals quoted with `'''` only results in one in the output; therefore, a backslash inserted before an escape (which is itself escaped because it is inside backticks) results in the expression being interpolated.
https://civet.dev/playground/?code=Y29uc29sZS5sb2cgJycnCiAgPGRpdj4KICAgIFwkezMgKyA0fQogIDwvZGl2PgonJyc%3D

STRd6:
In the string processing we weren't pulling the backslash escaped characters with the backslash so they ended up double escaping.

### DIFF
--- a/source/parser.hera
+++ b/source/parser.hera
@@ -7153,8 +7153,13 @@ Init
         .replace(/(\r?\n|\n)[ \t]*$/, "")
       }
 
-      // escape backtick, $
-      str = str.replace(/(`|\$\{)/g, "\\$1")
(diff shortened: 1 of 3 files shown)


edemaine (approved):
Nice catch and fix!

PR #421, opened 2023-03-01
structuredClone isn't available in node <17

STRd6:


edemaine:
It seems it's also not available on Safari (desktop and mobile). So potentially worth writing clone ourselves...

STRd6:
Closing in favor of #426

PR #425, opened 2023-03-02
Fixes #420 Decorate exported class

### LINKED ISSUE #420: Cannot directly export classes with decorators
Playground:

- [`@foo class Bar`](https://civet.dev/playground/?code=aW1wb3J0IHsgaHRtbCwgY3NzLCBMaXRFbGVtZW50IH0gZnJvbSAnbGl0JwppbXBvcnQgeyBjdXN0b21FbGVtZW50LCBwcm9wZXJ0eSB9IGZyb20gJ2xpdC9kZWNvcmF0b3JzLmpzJwoKQEBjdXN0b21FbGVtZW50KCdzaW1wbGUtZ3JlZXRpbmcnKQpjbGFzcyBTaW1wbGVHcmVldGluZyBleHRlbmRzIExpdEVsZW1lbnQKICBAc3R5bGVzID0gY3NzYHAgeyBjb2xvcjogYmx1ZSB9YAoKICBAQHByb3BlcnR5KCkKICBuYW1lID0gJ1NvbWVib2R5JwoKICByZW5kZXIoKQogICAgaHRtbGA8cD5IZWxsbywgJHt0aGlzLm5hbWV9ITwvcD5gCg%3D%3D)
- [`@foo export class Bar`](https://civet.dev/playground/?code=aW1wb3J0IHsgaHRtbCwgY3NzLCBMaXRFbGVtZW50IH0gZnJvbSAnbGl0JwppbXBvcnQgeyBjdXN0b21FbGVtZW50LCBwcm9wZXJ0eSB9IGZyb20gJ2xpdC9kZWNvcmF0b3JzLmpzJwoKQEBjdXN0b21FbGVtZW50KCdzaW1wbGUtZ3JlZXRpbmcnKQpleHBvcnQgY2xhc3MgU2ltcGxlR3JlZXRpbmcgZXh0ZW5kcyBMaXRFbGVtZW50CiAgQHN0eWxlcyA9IGNzc2BwIHsgY29sb3I6IGJsdWUgfWAKCiAgQEBwcm9wZXJ0eSgpCiAgbmFtZSA9ICdTb21lYm9keScKCiAgcmVuZGVyKCkKICAgIGh0bWxgPHA%2BSGVsbG8sICR7dGhpcy5uYW1lfSE8L3A%2BYAo%3D) (ParseError)
- [`export @foo class Bar`](https://civet.dev/playground/?code=aW1wb3J0IHsgaHRtbCwgY3NzLCBMaXRFbGVtZW50IH0gZnJvbSAnbGl0JwppbXBvcnQgeyBjdXN0b21FbGVtZW50LCBwcm9wZXJ0eSB9IGZyb20gJ2xpdC9kZWNvcmF0b3JzLmpzJwoKZXhwb3J0IEBAY3VzdG9tRWxlbWVudCgnc2ltcGxlLWdyZWV0aW5nJykKY2xhc3MgU2ltcGxlR3JlZXRpbmcgZXh0ZW5kcyBMaXRFbGVtZW50CiAgQHN0eWxlcyA9IGNzc2BwIHsgY29sb3I6IGJsdWUgfWAKCiAgQEBwcm9wZXJ0eSgpCiAgbmFtZSA9ICdTb21lYm9keScKCiAgcmVuZGVyKCkKICAgIGh0bWxgPHA%2BSGVsbG8sICR7dGhpcy5uYW1lfSE8L3A%2BYAo%3D) (Incompatible with TypeScript)

STRd6:
This allows decorators in TS "before export" and TC39 "after export" positions and doesn't try to do anything smart about it yet.

I added another state flag for when we are inside `Decorators` to prevent `@@decorator class ...` from being treated as an implicit call passing a class expression to the decorator function. If someone really wants to do that they'll need to use parentheses and be explicit.

### DIFF
--- a/source/parser.hera
+++ b/source/parser.hera
@@ -5349,6 +5349,9 @@ TypeDeclaration
   ( Export _? )? ( Declare _? )  TypeLexicalDeclaration -> { ts: true, children: $0 }
   ( Export _? )? ( Declare _? )? TypeDeclarationRest    -> { ts: true, children: $0 }
 
+  # NOTE: TS Exported classes can have decorators before export but not after
+  Decorators ( Export _? ) !Decorators ClassDeclaration -> { ts: true, children: $0 }
+
 # NOTE: These are declarations even without a `declare` prefix
 TypeDeclarationRest
   # TODO: ( __ Type ) can be refined further to check for consistently nested binary ops, etc.
@@ -5459,7 +5462,7 @@ NestedDeclareElement
 
 # NOTE: Variation on TypeDeclaration where Declare already applied.
 DeclareElement
-  ( Export _? )? TypeLexicalDeclaration -> { ts: true, children: $0 }
+  Decorators? ( Export _? )? TypeLexicalDeclaration -> { ts: true, children: $0 }
   ( Export _? )? TypeDeclarationRest    -> { ts: true, children: $0 }
 
 EnumDeclaration
--- a/source/parser.hera
+++ b/source/parser.hera
@@ -129,6 +129,8 @@ ForbiddenImplicitCalls
   /(as|of|satisfies|then|when|implements|xor|xnor)(?!\p{ID_Continue}|[\u200C\u200D$])/
   # NOTE: Don't allow non-heregex regexes that begin with a space as first argument without parens
   "/ "
+  # Don't treat @@decorator class ... as an implicit call
+  ForbidClassImplicitCall Class
   AtAt # experimentalDecorators
   Identifier "=" Whitespace
   Identifier !"(" ->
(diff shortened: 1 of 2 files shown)


source/parser.hera
```diff
@@ -5349,6 +5349,9 @@ TypeDeclaration
   ( Export _? )? ( Declare _? )  TypeLexicalDeclaration -> { ts: true, children: $0 }
   ( Export _? )? ( Declare _? )? TypeDeclarationRest    -> { ts: true, children: $0 }
 
+  # NOTE: TS Exported classes can have decorators before export but not after
+  Decorators ( Export _? ) !Decorators ClassDeclaration -> { ts: true, children: $0 }
```
edemaine, line 5353: I don't think this should be `ts: true`. Exported classes should be in JS code too, just like other decorated classes.

Related, why is this here and not in `ExportDeclaration`?

Ah, I see, this type of decorating is inconsistent with the [TC39 decorator proposal](https://github.com/tc39/proposal-decorators)...

edemaine:
Could you explain (maybe with a test case) why we need another flag here? Oh, was this a consequence of extending to CallExpression in a previous PR? Kind of surprised any tests passed...

edemaine (approved):

STRd6:
Yeah, the idea was to keep allowing `fn class ...` as an implicit application generally but disallow it in inline decorators.

PR #426, opened 2023-03-02
Use a simple deep copy because structuredClone isn't available on node <17

STRd6:


edemaine (approved):
Great, thanks!

If there's a released version of the LSP with a Civet version with structuredClone, we should probably release a new version of the LSP with updated deps.

STRd6:
It was only released as a Civet version, not bundled into the LSP with that version yet.

PR #428, opened 2023-03-03
Codeblocks!!!!!

STRd6:
Proof of concept using Hera's new codeblocks support. Now we can `require` any other source file. This will make it much easier to split out the complicated processing steps from the AST parsing and we can even get typechecking or use .civet for our transforms soon.

### DIFF
--- a/source/lib.js
+++ b/source/lib.js
@@ -54,8 +54,75 @@ function removeParentPointers(node) {
   }
 }
 
+// Gather child nodes that match a predicate
+// while recursing into nested expressions
+// without recursing into nested blocks/for loops
+function gatherNodes(node, predicate) {
+  if (node == null) return []
+
+  if (Array.isArray(node)) {
+    return node.flatMap((n) => gatherNodes(n, predicate))
+  }
+
+  if (predicate(node)) {
+    return [node]
+  }
+
+  switch (node.type) {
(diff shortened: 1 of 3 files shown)


edemaine (commented):
This looks great! It's going to make things like configuration a lot easier too. (E.g. CLI can more easily specify compiler flags.)

Two questions:

1. Can we use ESM import/export or is there a reason we need to use require? (I thought most of the codebase was rewritten to ESM...)

2. I wonder if "util" is optimal naming. Maybe we can start with an "ast" file which is for AST manipulation. I'm not sure what else we would have, but at least "ast" is more descriptive. Of course we can always change this later too.

STRd6:
We could (and should) switch to ESM eventually. The current thing holding it back is that Hera is assuming commonjs but that can be fixed in another Hera PR.

Eventually we can break up the pieces and organize them into any number of files/folders. It might make sense for all the pipeline stuff to even be in one file.

edemaine (approved):
Sounds good! I like the idea of e.g. a pipeline specific file. I wonder if we even want a `source` subdirectory for the parser. Just wanted to make sure `util` doesn't become a ln "all functions go here", hence suggesting more specific names. But mainly important that we share a vision for future organization.

PR #430, opened 2023-03-04
Replace TrailingComment* with _? except after statement delimiters.

STRd6:
TrailingComment includes // comments and those can't appear in the middle of a single line statement (it would comment out the rest of the line). So now we only check for inline comments in most places except statement delimiters which should have the rest of line trailing comments attached to them so they don't get attached to enclosing blocks.

### DIFF
--- a/source/parser.hera
+++ b/source/parser.hera
@@ -4346,6 +4346,8 @@ Comment
   MultiLineComment
   SingleLineComment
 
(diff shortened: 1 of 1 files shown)


edemaine (approved):
Wow, this is a great cleanup. I was pretty confused about when to use `_` versus `TrailingComment`. The new definition makes this very clear. I agree these all seem more correct.

PR #431, opened 2023-03-04
Ampersand pipes

### LINKED ISSUE #423: Function composition / flow operator
I use ramda a lot and I was curious if civet could support a few things natively. Something I do a lot with ramda is create a function using pipe. For example:

```typescript
const doTheThing = R.pipe(
  R.pluck('myprop'),
  R.map(R.toUpper)
)

console.log(doTheThing([{myprop: 'hello'}])) // outputs [HELLO]
```

My understanding is that the pipe operator in civet currently requires a start value, so one can't easily compose functions using pipe. You can see my attempts at this in https://github.com/DanielXMoore/Civet/issues/75#issuecomment-1448930717 

It would be really neat if there was some sort of function composition / flow operator (see [pipe vs flow in fp-ts](https://rlee.dev/practical-guide-to-fp-ts-part-1) and [F# Pipelining vs. Composition](https://spin.atomicobject.com/2021/08/26/fsharp-operators-pipeline-composition/) for good explanations of these two concepts).

F# uses `>>` for composition. So these two would be equivalent:

```typescript
input := [myprop: 'hello']

// using pipe, |>
input
  |> .map .myprop
  |> .map .toUpperCase() 
  |> console.log

// using flow, >>
doTheThing := .map .myprop >> .map .toUpperCase()
console.log doTheThing input

STRd6:
This could probably use some more test cases but seems to work pretty well so far.

Fixes #423

### DIFF
--- a/source/parser.hera
+++ b/source/parser.hera
@@ -423,14 +423,16 @@ PipelineExpression
         base: "$",
       }
 
+      const arrowBody = {
+        type: "PipelineExpression",
+        children: [ws, ref, body]
+      }
+
       return {
         type: "ArrowFunction",
-        children: [ref, " => ", {
-          type: "PipelineExpression",
-          children: [ws, ref, body]
-        }],
+        children: [ref, " => ", arrowBody],
         ref,
-        body,
+        body: [arrowBody],
         ampersandBlock: true,
       }
     }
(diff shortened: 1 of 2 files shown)


edemaine (commented):
Looks great!

Maybe worth adding a test (`throws` I assume) for `&` in the middle of a pipeline.

Another nice test might be `x ||> (& |> f)` which I think will unwrap to `f(x), x` given the `ampersandBlock: true` set here.

edemaine (approved):

edemaine (approved):
Nice improvements.

PR #442, opened 2023-03-09
For loop optimizations and generalizations

### LINKED ISSUE #412: For loop of range literal
We should optimize

```coffee
for x of [1..n]
  ...
```

in the same way we optimize

```coffee
"civet coffee-compat"
for x in [1..n]
  ...
```

as they iterate the exact same values. This is just a matter of refactoring some code, but I wanted to make this issue before I forget. 😅 

---

Then we should also support `for [1..n]` for the same loop but without access to the index. This is one approach to #312 and is part of CoffeeScript compatibility.

### LINKED ISSUE #440: Coffee for loops don't work with one-line bodies
CoffeeScript `for` loops often need to add a line to the loop body, via the `blockPrefix` mechanism. When the body is a single line, this isn't done correctly:

```coffee
"civet coffee-compat"
for i in x then i
---
var i
for (let i1 = 0, len = x.length; i1 < len; i1++)  i = x[i1]
 i
```

[Playground link](https://civet.dev/playground?code=ImNpdmV0IGNvZmZlZS1jb21wYXQiCmZvciBpIGluIHggdGhlbiBpCg%3D%3D)

(Note that I don't normally write code this way. Just encountered it when exploring other features that would add code to `for` loops in general.)

edemaine:
Fixes #412 and fixes #440

* Factor out RangeLiteral loop generation code into `lib.js` (new `forRange` helper)
* Cleanup ref handling with new `needsRef` and `maybeRef` helpers
* Fix missing use of ref for `asc` and `step` (previously would shadow/use the wrong variables)
* Fix `blockPrefix` mechanism with bare block (#440)
* Optimize `for x of [a..b]` in the same way that we did for CoffeeScript `for x in [a..b]` (#412)
* Support `for [a..b]` (a CoffeeScript feature) both inside and outside `coffee-compat` mode (#412)

### DIFF
--- a/source/lib.js
+++ b/source/lib.js
@@ -194,6 +194,42 @@ function maybeRef(exp, base = "ref") {
   }
 }
 
+// Convert (non-Template) Literal to actual JavaScript value
+function literalValue(literal) {
+  let {raw} = literal
+  switch (raw) {
+    case "null": return null
+    case "true": return true
+    case "false": return false
+  }
+  if (
+    (raw.startsWith('"') && raw.endsWith('"')) ||
+    (raw.startsWith("'") && raw.endsWith("'"))
+  ) {
+    return raw.slice(1, -1)
+  }
+  const numeric = literal.children.find(
+    (child) => child.type === "NumericLiteral"
+  )
+  if (numeric) {
+    raw = raw.replace(/_/g, "")
+    const {token} = numeric
+    if (token.endsWith("n")) {
+      return BigInt(raw.slice(0, -1))
+    } else if (token.match(/[\.eE]/)) {
+      return parseFloat(raw)
+    } else if (token.startsWith("0")) {
+      switch (token.charAt(1).toLowerCase()) {
+        case "x": return parseInt(raw.replace(/0[xX]/, ""), 16)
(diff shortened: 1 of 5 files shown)


source/lib.js
```diff
@@ -140,15 +140,199 @@ function gatherRecursiveWithinFunction(node, predicate) {
   return gatherRecursive(node, predicate, isFunction)
 }
 
+// Trims the first single space from the spacing array or node's children if present
+// maintains $loc for source maps
```
STRd6, line 144: This also inserts string `c` after removing the first space or not but we only ever use it with `""`. We may want to rename it later. I think it was originally used to insert an open paren while removing a space if present but we do that a different way now.

STRd6:
Lots of great changes here, thanks!

STRd6 (approved):

source/lib.js
edemaine: I did actually find one instance of `insertTrimmingSpace("case ", ws)`. But I agree it's a pretty confusing name given that it's mostly trimming, not inserting. Should we just call it `trimSpace` or `trimLeadingSpace`, and make `""` the default value for the second argument?

(I'm going to merge this PR given that this was just a function move, but can address this comment in a future PR.)

PR #443, opened 2023-03-09
Semicolon-separated statements in blocks

edemaine:
`a; b` was supported at the top level but not in deeper blocks, e.g., within an `if`!

This adds support to both regular blocks and if/else expressions.  It also refactors the top-level support to be cleaner, before I copied it two more times.

### DIFF
--- a/source/parser.hera
+++ b/source/parser.hera
@@ -1842,7 +1842,6 @@ NestedBlockStatements
     }
 
 NestedBlockStatement
-  # NOTE: _? allows for leading inline comments
   Nested:nested BlockStatementPart+:statements ->
     return [
       [nested, ...statements[0]],
@@ -1850,6 +1849,7 @@ NestedBlockStatement
     ]
 
 BlockStatementPart
+  # NOTE: _? allows for leading inline comments
   !EOS _?:ws StatementListItem:statement StatementDelimiter:delimiter ->
     if (ws) {
       statement = {...statement, children: [ ws, ...statement.children ]}
--- a/source/parser.hera
+++ b/source/parser.hera
@@ -53,11 +53,11 @@ NestedTopLevelStatements
 
 # Multiple top-level statements on the same line
 TopLevelSingleLineStatements
-  ( !EOS TopLevelStatement )+:statements ->
-    return statements.map(([, statement]) => statement)
+  TopLevelStatement+
 
 TopLevelStatement
-  _?:ws ModuleItem:statement StatementDelimiter:delimiter ->
+  # NOTE: _? allows for leading inline comments
+  !EOS _?:ws ModuleItem:statement StatementDelimiter:delimiter ->
     if (ws) {
       statement = {
         ...statement,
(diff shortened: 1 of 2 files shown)


source/parser.hera
```diff
@@ -53,11 +53,11 @@ NestedTopLevelStatements
...
 # Multiple top-level statements on the same line
 TopLevelSingleLineStatements
-  ( !EOS TopLevelStatement )+:statements ->
-    return statements.map(([, statement]) => statement)
+  TopLevelStatement+
 
 TopLevelStatement
-  _?:ws ModuleItem:statement StatementDelimiter:delimiter ->
+  # NOTE: _? allows for leading inline comments
+  !EOS _?:ws ModuleItem:statement StatementDelimiter:delimiter ->
```
STRd6, line 60: We may be able to remove `!EOS` and use `SemicolonDelimiter` since it looks like `!EOS` is only preventing the `StatementDelimiter` choices that make use of `EOS`.

source/parser.hera
```diff
@@ -1839,11 +1842,19 @@ NestedBlockStatements
     }
 
 NestedBlockStatement
-  Nested:nested _?:ws StatementListItem:statement StatementDelimiter:delimiter ->
+  # NOTE: _? allows for leading inline comments
```
STRd6, line 1845: This note is disconnected from its target.

source/parser.hera
edemaine: We do want to allow `EOS` (and no semicolon) at the end. I could use special code for the last statement, though this seemed a bit simpler...

source/parser.hera
```diff
@@ -1839,11 +1842,19 @@ NestedBlockStatements
...
+  Nested:nested BlockStatementPart+:statements ->
+    return [
+      [nested, ...statements[0]],
+      ...statements.slice(1).map(s => ["", ...s]),
+    ]
+
+BlockStatementPart
+  !EOS _?:ws StatementListItem:statement StatementDelimiter:delimiter ->
```
STRd6, line 1853: Similar to above we could remove `!EOS` and use `SemicolonDelimiter`.

source/parser.hera
STRd6: In that case maybe just add a note and remove/change the note on `TopLevelSingleLineStatements` since it's not quite clear to me.

test/if.civet
```diff
@@ -86,14 +86,45 @@ describe "if", ->
...
+    if (x) {
+      /*a*/ a; /*b*/ b
+    }
+    else {
+      /*c*/ c; /*d*/ d
+    }
+  """
+
   // TODO
   testCase.skip """
```
STRd6, line 122: Should this still be `.skip`?

STRd6 (approved):

source/parser.hera
edemaine: Revised.

If I recall correctly (this really comes from a previous PR that fixed top-level indentation), there might have also been an issue where `Expression` and hence `ModuleItem` can also start with `__`, and this prevents jumping to the next line here. Indeed, `Expression` → ... → `AssignmentExpression` → `__ AssignmentExpressionTail`. I've added this to the note too.

test/if.civet
edemaine: It still doesn't work, but probably easyish to fix. Currently `if x then` compiles to `if (x)`, so it's missing either `;` or `{}`.

PR #447, opened 2023-03-11
@ type parameter with other parameters

### LINKED ISSUE #446: `this` type parameter only works if there are no other parameters
```typescript
fn := (this: T, x: X) ->
```

Actual:
```typescript
const fn = { this: T, x: X }(function () {})
```

Expected:
```typescript
function fn({ this: T, x: X }) {}
```

STRd6:
Fixes #446

### DIFF
--- a/test/types/js.civet
+++ b/test/types/js.civet
@@ -188,9 +188,9 @@ describe "Types", ->
       testCase.js """
         plain function with @ shorthand
         ---
-        function (@: T) {}
+        function (@: T, x: X) {}
         ---
-        (function () {})
+        (function ( x) {})
       """
 
     testCase.js """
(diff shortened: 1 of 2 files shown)


test/types/js.civet
```diff
@@ -188,9 +188,9 @@ describe "Types", ->
       testCase.js """
         plain function with @ shorthand
         ---
-        function (@: T) {}
+        function (@: T, x: X) {}
         ---
-        (function () {})
+        (function ( x) {})
       """
```
edemaine, line 194: Seems worth keeping the original (no-argument) case as well.
Hmm, though I guess the `this` (not `@`) case covers that.
Or we could have all four cases...

edemaine (approved):

PR #448, opened 2023-03-11
Configurable and chainable civet esbuild plugin

### LINKED ISSUE #429: Improve ESBuild Plugin + ESBuild Plugin Types.
This is related to a discussion that I had with Daniel in the `help` channel in the Civet discord.

Currently, the ESBuild plugin doesn't support compiling into JSX, nor does it support piping through other ESBuild plugins, such as SolidJS's babel transpilation through `esbuild-plugin-solid` (https://github.com/amoutonbrady/esbuild-plugin-solid).

We found a temporary solution to this by specifically replacing the current ESBuild plugin with a new form that directly injects SolidJS's babel transpilation into the outputted contents.

```ts
const { readFile } = require('fs/promises')
const path = require('path');

// NOTE: this references the built version of the module, not the source
const { compile } = require("@danielx/civet")

const { transformAsync } = require("@babel/core");
const solid = require("babel-preset-solid");
const ts = require("@babel/preset-typescript");

module.exports = {
  name: 'civet',
  setup(build) {
    return build.onLoad({
      filter: /\.civet$/
    }, async function (args) {
      try {
        const source = await readFile(args.path, 'utf8')
        const filename = path.relative(process.cwd(), args.path)
        const compiled = compile(source, {
          filename,
          inlineMap: true,
          js: true
        })

        const result = await transformAsync(compiled, {
          presets: [[solid, {}], [ts, {}]],
          filename,
          sourceMaps: "inline",
        });
        if (result?.code === void 0 || result.code === null) {
          throw new Error("No result was provided from Babel");
        }
        return { contents: result.code, loader: "js" };
      }
      catch (e) {
        return {
          errors: [{
            text: e.message
          }]
        }
      }
    })
  }
}
```

This is only a temporary solution to pipe Civet into SolidJS to transpile directly with ESBuild. The permanent solution proposed is to use an API like so:
```ts
build({
// additional esbuild config...
plugins: [civet({ next: solid() })]
})
```

This would allow the Civet ESBuild plugin to take SolidJS's ESBuild plugin and pipe it through to go from Civet code to SolidJS's `createElement` in one single build step. If you wouldn't like to use SolidJS in your Civet project, then you could create or find another ESBuild plugin that would do your desired behaviour - this allows for reasonable flexibility.

Additionally, the types from `types.d.ts` for the ESBuild Civet plugin seem to not be going through correctly and an error like this is shown:
`Cannot find module '@danielx/civet/esbuild-plugin' or its corresponding type declarations. ts(2307)`

A fix for this should be found as well.

Thanks heaps for all of the hard work put into Civet!

STRd6:
Fixes #429

This could probably use some testing by someone who is using esbuild/solid. Also worth testing with vanilla-extract and others.

TODO
---
- [ ] Compose sourcemaps
- [x] cannot assign name to function
- [x] mkdtmp
- [x] cleanup?

### DIFF
--- a/source/esbuild-plugin.civet
+++ b/source/esbuild-plugin.civet
@@ -72,36 +72,34 @@ civet := (options: Options) ->
   return {
     name: "civet"
     setup(build)
-      build.onLoad {
-      filter
-    }, (args) ->
-      try
-        source := await readFile args.path, 'utf8'
-        filename := path.relative(process.cwd(), args.path)
-        compiled := compile source, {
-          filename
-          inlineMap
-          js
-        }
+      build.onLoad { filter }, (args) ->
+        try
+          source := await readFile args.path, 'utf8'
+          filename := path.relative(process.cwd(), args.path)
+          compiled := compile source, {
+            filename
+            inlineMap
+            js
+          }
 
-        if !dirCreated
-          await mkdir tmpPath
-          dirCreated = true
+          if !dirCreated
+            await mkdir tmpPath
+            dirCreated = true
 
-        // I'd prefer not to use temp files but I can't find a way to pas a stream to fs.readFile which is what
-        // most esbuild plugins use
-        await writeFile path.join(tmpPath, filename), compiled
-        await nextTransform {
-          path: path.join(tmpPath, filename)
-        }
+          // I'd prefer not to use temp files but I can't find a way to pass a stream to fs.readFile which is what
+          // most esbuild plugins use
+          await writeFile path.join(tmpPath, filename), compiled
+          await nextTransform {
+            path: path.join(tmpPath, filename)
+          }
 
-        return
-          contents: compiled
-      catch e
-        return
-          errors: [{
-            text: e.message
-          }]
+          return
+            contents: compiled
+        catch e
+          return
+            errors: [{
+              text: e.message
+            }]
   }
 
 // Default zero-config plugin
--- a/source/esbuild-plugin.civet
+++ b/source/esbuild-plugin.civet
@@ -35,20 +35,23 @@ esbuild.build({
 ```
 */
 
+import type { Plugin, OnLoadArgs, OnLoadResult } from 'esbuild'
+
+interface Options {
+  filter?: RegExp
+  inlineMap?: boolean
+  js?: boolean
+  next?: Plugin
+}
+
 { readFile, writeFile, mkdir } := require 'fs/promises'
 path := require 'path'
 
 // NOTE: this references the built version of the module, not the source
 { compile } := require "../dist/main.js"
 
-interface Options
-  filter?: RegExp
-  inlineMap?: boolean
-  js?: boolean
-  next?: unknown
-
 // NOTE: this function is named civet so esbuild gets "civet" as the name of the plugin
-civet := (options: Options) ->
+civet := (options: Options = {}): Plugin ->
   {
     filter=/\.civet$/
     inlineMap=true
@@ -56,7 +59,7 @@ civet := (options: Options) ->
     next
   } := options
 
-  let nextTransform
+  let nextTransform: (args: OnLoadArgs) => OnLoadResult | Promise<OnLoadResult | null | undefined> | null | undefined
   let tmpPath: null | string
   let dirCreated = false
 
@@ -65,7 +68,19 @@ civet := (options: Options) ->
     tmpPath = path.join tmpdir(), `civet-${process.pid}`
 
     next.setup {
-      onLoad: (_, handler) ->
+      onEnd()
+        ;
+      onStart()
+        ;
+      resolve()
+        ;
+      onResolve()
+        ;
+      initialOptions()
+        ;
+      esbuild()
+        esbuild
+      onLoad(_, handler)
         nextTransform = handler
     }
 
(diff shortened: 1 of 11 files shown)


source/esbuild-plugin.civet
```diff
@@ -13,16 +15,91 @@ esbuild.build({
...
+      onLoad: (_, handler) ->
+        nextTransform = handler
+    }
+
+  return {
+    name: "civet"
+    setup(build)
+      build.onLoad {
+      filter
+    }, (args) ->
```
edemaine, line 75: Did we lose a level of indentation here? Hard to read.

source/esbuild-plugin.civet
```diff
@@ -13,16 +15,91 @@ esbuild.build({
...
+    js=true
+    next
+  } := options
+
+  let nextTransform
+  let tmpPath: null | string
+
+  if next
+    { tmpdir } := require 'os'
+    tmpPath = path.join tmpdir(), 'civet'
```
edemaine, line 64: Don't we need to `mkdir` this directory?

I also worry that this will lead to permission issues when multiple people on the same machine try to use this same directory.  Perhaps naming it `` `civet-${process.pid}` `` would be better?

source/parser.hera
```diff
@@ -5918,8 +5918,9 @@ TypeParameterDelimiter
   &( __ ">" )
   &EOS InsertComma -> $2
 
+# TypeScript's this: T syntax in function parameters
```
edemaine, line 5921: Not sure why 919c680fc5bf6c507a6df29a8d80fe2aca1cbb44 is appearing here again; isn't it already merged into `main`?

source/esbuild-plugin.civet
```diff
@@ -13,16 +15,91 @@ esbuild.build({
...
 // NOTE: this references the built version of the module, not the source
 { compile } := require "../dist/main.js"
 
-module.exports = {
-  name: 'civet'
+interface Options
+  filter?: RegExp
+  inlineMap?: boolean
+  js?: boolean
+  next?: unknown
```
edemaine, line 48: I'm not very familiar with the esbuild plugin system, but I wonder if it makes sense to support multiple `next` transforms, as an array.  I needed this for the analogous option in the Vite plugin (see https://github.com/lorefnon/vite-plugin-civet/pull/5) to e.g. process JSX and then add compile-time constants.  But if most plugins already support a `next` option, it's probably not necessary...

source/parser.hera
STRd6: I branched off a branch by mistake but its should be merged now.

source/esbuild-plugin.civet
STRd6: I could be possible though I don't know of any uses off the top of my head but I'm sure some might come up. Most esbuild plugins are unchainable by default. Maybe even extracting this tmp file technique to arbitrarily chain esbuild plugins could be useful more generally than civet.

source/esbuild-plugin.civet
```diff
@@ -13,34 +15,115 @@ esbuild.build({
...
+      onEnd()
+        ;
+      onStart()
+        ;
+      resolve()
+        ;
+      onResolve()
+        ;
+      initialOptions()
+        ;
```
edemaine, line 80: FWIW, I think the `;` can go on the same line as the method name.

source/esbuild-plugin.civet
```diff
@@ -13,34 +15,115 @@ esbuild.build({
...
+          filename := path.relative(process.cwd(), args.path)
+          compiled := compile source, {
+            filename
+            inlineMap
+            js
+          }
+
+          if next
+            if !dirCreated
+              await mkdir tmpPath
```
edemaine, line 102: I was looking up whether `mkdir` crashes if the directory already exists, and saw [mkdtemp](https://nodejs.org/api/fs.html#fspromisesmkdtempprefix-options) which seems slightly better as it combines the unique directory-name generation with directory making.

I still like the idea of re-using the directory when the plugin gets used multiple times, so nice to delay it until here. Perhaps `dirCreated` gets replaced by a `dir` string variable that gets set when it's first needed.

source/esbuild-plugin.civet
```diff
@@ -13,34 +15,115 @@ esbuild.build({
...
+              dirCreated = true
+
+            outputFileName := filename + js ? '.jsx' : '.tsx'
+
+            // I'd prefer not to use temp files but I can't find a way to pass a stream to fs.readFile which is what
+            // most esbuild plugins use
+            await writeFile path.join(tmpPath, outputFileName), compiled
+            return await nextTransform {
+              ...args
+              path: path.join(tmpPath, filename)
```
edemaine, line 112:
```suggestion
              path: path.join(tmpPath, outputFilename)
```

Actually, this join is used twice, so perhaps it should be stored in a variable, say `outputPath`.

source/esbuild-plugin.civet
```diff
@@ -13,34 +15,115 @@ esbuild.build({
...
+            }
+
+          return
+            contents: compiled
+        catch e
+          return
+            errors: [{
+              text: e.message
+            }]
+  }
```
edemaine, line 122: Is there a "cleanup" mechanism in esbuild plugins, so we could delete the temporary directory?  Not a big deal but it'd be nice.

source/esbuild-plugin.civet
edemaine: We can always leave this for later when a need arises, as it did for the Vite plugin.

greghuc:
Not wishing to explode this task, just asking innocent question :-)

Are there imminent plans for the esbuild civet plugin to return more complete error+location information on an exception? Currently the error returned to esbuild just contains ```{ text: e.message }```, and not the full esbuild spec of:
```
interface Message {
  text: string;
  location: Location | null;
  detail: any; // The original error from a JavaScript plugin, if applicable
}

interface Location {
  file: string;
  namespace: string;
  line: number; // 1-based
  column: number; // 0-based, in bytes
  length: number; // in bytes
  lineText: string;
}
```

FYI, the reason I'm asking is that - as I [commented in a previous bug report](https://github.com/DanielXMoore/Civet/issues/342#issuecomment-1423292313) - I'm trying to migrate some coffeescript codebases to civet. Their underlying build system is browserify, but I've now concluded that whole ecosystem is effectively dead, so it's not worth writing ```civetify``` (I got a version working based on ```coffeeify```). Everything is brittle, with browserify plugins breaking on newer versions of JS, depending on which AST parser they are using internally.

So I'm trying to replace browserify with esbuild+[esbuild-coffeescript](https://github.com/johnie/esbuild-coffeescript). This is looking very promising, except I've just found the reported error locations are busted as per [this bug](https://github.com/johnie/esbuild-coffeescript/issues/3) reported by @STRd6 last April.

Navigating source maps looks pretty horrendous, so I'm wondering if I can hold off until the esbuild civet plugin has solid error reporting.

And thanks for all your hard work :-)

Issue #444, opened 2023-03-10
Declaration inside conditional


eevleevs:
I would use often if 

```js
if a := 1
  something
```

translated to something like

```js
{
  const a = 1
  if (a) {
    something
  }
}
```

all combinations with `let`, `while`, `unless`, etc. also welcome.

edemaine:
Agreed, this would be super useful; we've discussed it on Discord before.  In particular, I'd use this a lot with regular expressions (`if match := regex.exec ...`).

A more likely compilation for your example would be:

```js
let a
if (a = 1) {
  something
}
```

This is useful because it preserves the order of operations.  For example, if you had the expression `f() + (a := g())`, we'd expect `f()` to be called before `g()`, so the assignment should happen within the expression instead of in the declaration.

To preserve the `const` nature (no reassignment), @STRd6 has proposed this sort of compilation:

```js
let a1
if (a1 = 1) {
  const a = a1
  something
}
```

In addition to implementing correct `const` semantics (except within the `if` clause itself, where I guess we'd need to check for a single assignment), this would make `a` accessible only within the `if` block, not in an `else` block or outside the `if...else` clause.  It's similar to the behavior of `for (const ...)`, which scopes the variable to exist only within the `for` loop.

STRd6:
This could also work with destructuring inside the conditional:

```coffee
if [a, b, c] := x
  a + b + c
```

```typescript
let ref
if (ref = x) {
  const [a, b, c] = ref
  a + b + c
}
```

edemaine:
That would be super useful!  The number of times I've assigned `match` from a regular expression execution, checked for it being defined, only to immediately destructure it and ignore `match`...

STRd6:
Fixed in #451

PR #451, opened 2023-03-14
Declarations inside conditionals

STRd6:
#444

- [x] Hoist Refs
- [x] Expand to while loops

### DIFF
--- a/source/lib.js
+++ b/source/lib.js
@@ -7,21 +7,18 @@
  */
 
 /**
- * Duplicate a block and attach statements from the clause to be prefixed to the block.
+ * Duplicate a block and attach statements prefixing the block.
  * Adds braces if the block is bare.
  *
- * @returns the duplicated block with prefix statements attached.
+ * @returns the duplicated block with prefix statements attached or the unchanged block.
  */
-function blockWithPrefix(clause, block) {
-  if (!clause) return block
-  const { blockPrefix } = clause
-
-  if (blockPrefix && blockPrefix.length) {
+function blockWithPrefix(prefixStatements, block) {
+  if (prefixStatements && prefixStatements.length) {
     const indent = getIndent(block.expressions[0])
     // Match prefix statements to block indent level
-    const prefixStatements = indent
-      ? blockPrefix.map((statement) => [indent, ...statement.slice(1)])
-      : blockPrefix
+    if (indent) {
+      prefixStatements = prefixStatements.map((statement) => [indent, ...statement.slice(1)])
+    }
 
     const expressions = [...prefixStatements, ...block.expressions]
 
@@ -95,6 +92,16 @@ function removeParentPointers(node) {
     for (const child of node.children) {
       removeParentPointers(child)
     }
+  }
+}
+
+// Find nearest strict ancestor that satisfies predicate,
+// aborting (and returning undefined) if stopPredicate returns true
+function findAncestor(node, predicate, stopPredicate) {
+  node = node.parent
+  while (node && !stopPredicate?.(node)) {
+    if (predicate(node)) return node
+    node = node.parent
   }
 }
 
@@ -184,22 +191,30 @@ function hasYield(exp) {
 }
 
 function hoistRefDecs(statements) {
-  gatherRecursiveAll(statements, (s) => s.hoistableDec)
+  gatherRecursiveAll(statements, (s) => s.hoistDec)
     .forEach(node => {
-      const { hoistableDec } = node
+      const { hoistDec } = node
 
       // TODO: expand set to include other parents that can have hoistable decs attached
       const outer = closest(node, ["IfStatement", "IterationStatement"])
+      if (!outer) {
+        node.children.push({
+          type: "Error",
+          message: "Can't hoist declarations inside expressions yet."
+        })
+        return
+      }
+
       const block = outer.parent
 
       // NOTE: This is more accurately 'statements'
       const { expressions } = block
       const index = expressions.indexOf(outer)
       const indent = getIndent(outer)
-      if (indent) hoistableDec[0] = indent
-      expressions.splice(index, 0, hoistableDec)
+      if (indent) hoistDec[0][0] = indent
+      expressions.splice(index, 0, hoistDec)
 
-      node.hoistableDec = null
+      node.hoistDec = null
     })
 }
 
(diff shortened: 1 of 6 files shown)


edemaine:
Oh sweet! I was aiming to get here as well, so let me mention what my plan was so we can avoid overlap:

I have an unpushed branch that's refactoring statements into their own AST node (instead of `[indent, content, delimitet]`). I was thinking this would be helpful for walking up the tree from a declaration in an expression to find the containing statement and put the declaration there. So I was focusing on that first, and hadn't worked on actual declaration moving yet.

Would Statement AST nodes be useful for this, or did you have another way? I can hopefully get the Statement thing working in another couple of days.

STRd6:
Statement AST wouldn't be too useful for this specifically, I just ended up using the statement tuple style.

Also refactored some of the block prefix code so some for loops have better indentation now. I removed the empty assignment hack since now we are tracking the actual assignments inside for loop headers.

Last thing is to add ref-hoisting.

The statement AST updates should fit in fine with this, I'll just need to adjust a couple places where we're inserting the indent.

edemaine:
I guess my question was: how are you planning to do ref hoisting? I was thinking it would make sense to follow parent pointers until you hit a Statement. But maybe you can do the same thing with Blocks?

STRd6:
I was going to find the `parent` of the `IfStatement` and then splice a the hoisted ref statement before the `IfStatement` in the parent's children.

We'll probably need to do a whole additional pass on expressionizing statements later and I think the Stamement AST nodes could help a lot with that.

test/if.civet
edemaine, line 1: Could you add a test of an if expression with a declaration?

test/for.civet
```diff
@@ -121,7 +120,7 @@ describe "for", ->
       console.log(i)
     ---
     for (let i1 = 1; i1 <= 10; ++i1) {
-    var i = i1;
+      var i = i1;
```
edemaine, line 123: Thanks for fixing these!

source/lib.js
```diff
@@ -6,6 +6,50 @@
...
 
+/**
+ * Duplicate a block and attach statements from the clause to be prefixed to the block.
+ * Adds braces if the block is bare.
+ *
+ * @returns the duplicated block with prefix statements attached.
+ */
+function blockWithPrefix(clause, block) {
+  if (!clause) return block
+  const { blockPrefix } = clause
```
edemaine, line 17: This API feels a little weird, but I guess it leaves room for passing in content other than `blockPrefix`? It might be cleaner to just pass in `clause?.blockPrefix` though.

From the name "blockWithPrefix", I'd also expect the arguments to be in the other order.

source/lib.js
```diff
@@ -6,6 +6,50 @@
...
+  return block
+}
+
+function closest(node, types) {
+  do {
+    if (types.includes(node.type)) {
+      return node
+    }
+  } while (node = node.parent)
+}
```
edemaine, line 51: This is a special case of [findAncestor](https://github.com/DanielXMoore/Civet/blob/9b28e3a351d130ce031c7b28b50bcb6d2a499315/source/parser.hera#L7376). I'm fine with a separate implementation that's more efficient, but perhaps we could name it similarly (`findAncestorOfType` might be more descriptive), and maybe move `findAncestor`'s code to be next to this in lib?

source/lib.js
```diff
@@ -129,6 +183,26 @@ function hasYield(exp) {
   return gatherRecursiveWithinFunction(exp, ({ type }) => type === "Yield").length > 0
 }
 
+function hoistRefDecs(statements) {
+  gatherRecursiveAll(statements, (s) => s.hoistableDec)
+    .forEach(node => {
+      const { hoistableDec } = node
+
+      // TODO: expand set to include other parents that can have hoistable decs attached
```
edemaine, line 191: Yeah, I worry that currently a declaration in an expression that is deeply inside an `if` body (with a function declaration or various other things in between) will end up attaching to the `if` here. But reading below, I guess that won't happen with the current grammar.  In the future I guess this is where `Statement` AST nodes would be helpful, and would enable declarations in more general places in expressions.

source/parser.hera
```diff
@@ -2921,22 +2933,31 @@ IfStatement
...
+    children.push(block)
     if (block.bare && e) children./**/push(";")
     if (e) children./**/push(e)
+
     return {
       type: "IfStatement",
       then: block,
       else: e,
       children,
+      condition: clause.condition,
```
edemaine, line 2947: Would it make sense to instead start this object literal with `...clause`? Then could also omit `type` here.

source/parser.hera
```diff
@@ -3628,7 +3640,22 @@ CatchParameter
...
+    }
   ParenthesizedExpression !( _? ( BinaryOp / AssignmentOp / Dot / QuestionMark ) ) !( _ OperatorAssignmentOp ) -> $1
+  # NOTE: DeclarationCondition must come before the ExpressionCondition because we need to look ahead to potentially match `:=` or `.=`
+  InsertOpenParen:open DeclarationCondition:expression InsertCloseParen:close ->
+    return {
+      type: "ParenthesizedExpression",
+      children: [open, expression, close],
+      expression,
+    }
+
```
edemaine, line 3658: Hmm, I expected this to be allowed in an arbitrary Expression, so you could do things like `if x + (y := ...)` but I guess that can be an extension for later.

source/parser.hera
```diff
@@ -3639,6 +3666,24 @@ Condition
...
+      type: "Ref",
+      base: "ref",
+    }
+
+    const { binding, initializer } = dec
+
+    const initCondition = {
+      type: "AssignmentExpression",
+      children: [ref, " ", initializer],
+      hoistableDec: [["", ["let ", ref], ";\n"]],
```
edemaine, line 3681: "Hoistable" suggests that it *could* be hosted but doesn't need to be. Maybe `hoistDec` would make more sense?

edemaine (approved):
Looks great! I just have a few minor comments.

source/lib.js
STRd6: I named this one after the https://developer.mozilla.org/en-US/docs/Web/API/Element/closest API and eventually want to add full query selector support. `findAncestorOfType` is a better description of what it is doing currently, however.

source/lib.js
STRd6: 👍 I imagine this will get refined a bit as we fill in more refs and do more with blocks and hoisting.

test/if.civet
STRd6: I don't think it works in expressions but I'll add a test with TODO indicating that.

PR #463, opened 2023-03-19
Fixes #461

### LINKED ISSUE #461: ref hoisting bug
I was writing docs for the new declarations within conditionals and loops, and bumped into this:

```coffee
node .= linkedList.head
while {data, next} := node
  console.log data
  node = next
```

## Current compilation

```js
let node = linkedList.headlet ref;

while (ref = node) {
  const {data, next} = ref;
  console.log(data)
  node = next
}
```

There's a missing newline (or semicolon) before `let ref;`. And probably an extra newline after `let ref;`.

STRd6:


source/parser.hera
```diff
@@ -3678,7 +3678,7 @@ DeclarationCondition
     const initCondition = {
       type: "AssignmentExpression",
       children: [ref, " ", initializer],
-      hoistDec: [["", ["let ", ref], ";\n"]],
+      hoistDec: [["", ["let ", ref], ";"]],
```
edemaine, line 3681: This is counterintuitive but I think right, because the EOS separator is in the beginning not the end of a line. Could you maybe add a test where this happens inside an indented block to confirm it works well in that case? (and doesn't e.g. output `let ref;    indented block`)

edemaine (approved):

source/parser.hera
STRd6: It does output that way when it is the first statement in a block. I think the issue is that the first block statement isn't getting the EOS attached so that could be done on a separate indentation cleanup more generally.

edemaine: Interesting. There's actually code that removes the leading EOS of the first statement in a block.  There's a comment that says it's for "easier insertion into the block". But perhaps it'd be more useful not to?

https://github.com/DanielXMoore/Civet/blob/02deeb31b567d12d7be66b265891f08267e0c856/source/parser.hera#L1849-L1856

PR #465, opened 2023-03-19
Refactor block indentation

STRd6:
- [x] Don't split EOS from indent in first block statement
- [x] Make block prefix statements consistent
- [x] Map indent into each prefix statement
- [x] Update getIndent to get all non-comment ws tokens

A lot of the block additions are more consistent now and mostly improved. One liners remain one liners (so we have to insert more semi-colons). Indentation is mapped from a statement in the block into the statements in `blockPrefix`.

### DIFF
--- a/source/lib.js
+++ b/source/lib.js
@@ -170,12 +170,21 @@ function gatherRecursiveAll(node, predicate) {
 }
 
 /**
- * Gets the indentation node from a statement.
+ * Gets the indentation node from a statement. Includes newline,
+ * excludes comments, strips location info.
  */
 function getIndent(statement) {
+  debugger
   let indent = statement?.[0]
-  // Hacky way to get the indent out of [EOS, indent] pair
-  if (Array.isArray(indent)) indent = indent[indent.length - 1]
+  if (Array.isArray(indent)) {
+    indent = indent.flat(Infinity)
+
+    return indent.filter((n) => n && !(n.type === "Comment")).map((n) => {
+      if (typeof n === "string") return n
+      if (n.token != null) return n.token
+      return ""
+    })
+  }
   return indent
 }
 
--- a/source/lib.js
+++ b/source/lib.js
@@ -174,7 +174,6 @@ function gatherRecursiveAll(node, predicate) {
  * excludes comments, strips location info.
  */
 function getIndent(statement) {
-  debugger
   let indent = statement?.[0]
   if (Array.isArray(indent)) {
     indent = indent.flat(Infinity)
--- a/source/lib.js
+++ b/source/lib.js
@@ -376,7 +376,7 @@ function forRange(open, forDeclaration, range, stepExp, close) {
     } else { // const or var: put inside loop
       // TODO: missing indentation
       blockPrefix = [
-        ["", forDeclaration, " = ", counterRef, ";\n"]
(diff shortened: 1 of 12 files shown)


source/lib.js
```diff
@@ -170,12 +170,21 @@ function gatherRecursiveAll(node, predicate) {
 }
 
 /**
- * Gets the indentation node from a statement.
+ * Gets the indentation node from a statement. Includes newline,
+ * excludes comments, strips location info.
  */
 function getIndent(statement) {
+  debugger
```
edemaine, line 177: This probably needs to be removed.

source/parser.hera
```diff
@@ -6872,7 +6863,11 @@ Init
           // else block
           if (exp.else) insertReturn(exp.else[2])
           // Add explicit return after if block if no else block
-          else exp.children.push(["\n", indent, wrapWithReturn()])
+          else exp.children.push([{
```
edemaine, line 6866: Shouldn't there be a first element of "" here for the indent?

source/parser.hera
```diff
@@ -8436,7 +8432,12 @@ Init
       if (varIds.length) {
         // get indent from first statement
         const indent = getIndent(statements[0])
-        statements.unshift([indent, "var ", varIds.join(", "), "\n"])
+        let delimiter = ";"
+        if (statements[0][1]?.parent?.root) {
+          delimiter = "\n"
```
edemaine, line 8437: This seems dangerous to me. Automatic semicolon insertion may not apply. Why would the root be special? Oh, maybe you want `";\n"`.

test/compat/coffee-for-loops.civet
```diff
@@ -247,12 +245,9 @@ describe "coffeeForLoops", ->
     "civet coffeeCompat"
     coffees = (s for s in scripts when s.type in coffeetypes)
     ---
-    var coffees, s
-    var indexOf: <T>(this: T[], searchElement: T) => boolean = [].indexOf as any;
+    var coffees, s;var indexOf: <T>(this: T[], searchElement: T) => boolean = [].indexOf as any;
```
edemaine, line 248: The lack of newline between the two var statements here is sad, but not terrible.

test/function.civet
```diff
@@ -270,9 +270,7 @@ describe "function", ->
       ---
       (@a, @b) ->
       ---
-      (function(a, b) {this.a = a;
-      return this.b = b;
-      })
+      (function(a, b) {this.a = a;return this.b = b;})
```
edemaine, line 273: This was a bug before but I don't think `b` (or anything) should be returned in this function.

edemaine (approved):
Nice refactor!

Sadly I suspect this will conflict a lot with my Statement branch, but it's worth it (and that branch still isn't working). 

The compilations are sometimes uglier now, but mainly it's the one-line examples, and keeping those one line seems nice.

source/parser.hera
STRd6: I just didn't want to update 57 more tests but I'll do it :)

STRd6: I think ASI would always apply since it is a var declaration.

edemaine: Ah, I think you're right: it's just initalized declarations that can have a problem with ASI. An uninitialized declaration should be safe.

PR #470, opened 2023-03-21
Fixes #469

### LINKED ISSUE #469: const declaration in conditions bug
```coffee
if const match = regex.exec string
  console.log match
```

compiles to

```ts
let ref;if (ref ) {
  = ref;
  console.log(match)
}
```

I expected this to work like

```coffee
if match := regex.exec string
  console.log match
```

which compiles to

```ts
let ref;if (ref = regex.exec(string)) {
  const match = ref;
  console.log(match)
}
```

STRd6:


STRd6:
I think I'll need to do a follow up with splats and `@` bindings since they may not work currently.

PR #471, opened 2023-03-23
splices and this assignments in conditional bindings

STRd6:


source/parser.hera
```diff
@@ -4069,7 +4072,18 @@ ImplicitExportSpecifier
 Declaration
   HoistableDeclaration
   ClassDeclaration
-  LexicalDeclaration
+  LexicalDeclaration:d ->
+    if (d.thisAssignments?.length)
+      return {
+        ...d,
+        children: [...d.children, ...d.splices, ";", ...d.thisAssignments]
```
edemaine, line 4079: Should we just have a general mechanism for postfixing statements here? Or is there a need to have splices and thisAssignments separated?

edemaine, line 4079: Will this throw if splices is undefined? Maybe worth building up an array and protecting splices and thisAssignments pushing with ifs.

source/parser.hera
```diff
@@ -4121,11 +4139,15 @@ LexicalBinding
...
 
+    const [splices, thisAssignments] = gatherBindingCode(binding.children)
+
     return {
       children: [binding, initializer],
       names: binding.names,
       binding,
       initializer,
+      splices: splices.map(s => [",", s]),
+      thisAssignments: thisAssignments.map(s => ["", s, ";"]),
```
edemaine, line 4150: Ah, maybe these different separators are a good reason to keep them separate.

edemaine (approved):

source/parser.hera
STRd6: The splices become part of the declaration so I don't think it can be generalized.

PR #475, opened 2023-03-25
Indentation after await operator

### LINKED ISSUE #473: Indentation after unary operator
One of the motivations for `await.all` (until we have #398) was to be able to write this:

```coffee
await.all
  for item of array
    async do
      ...
```

Currently, though, we don't allow a newline after a unary operator. Perhaps we could allow `IndentedFurther` as an alternative to `_`? (Hoping this is an easy fix but can't try right now.)

edemaine:
Fixes #473

I tried just adding support directly to `AwaitOp`:

```diff
diff --git a/source/parser.hera b/source/parser.hera
index 77ab5ec..ae55026 100644
--- a/source/parser.hera
+++ b/source/parser.hera
@@ -2850,7 +2850,7 @@ UnaryOp

 # https://github.com/tc39/proposal-await.ops
 AwaitOp
-  Await:a ( Dot IdentifierName )?:op ( _ / &OpenParen ):ws ->
+  Await:a ( Dot IdentifierName )?:op ( &OpenParen / _ / IndentedFurther ):ws ->
     if (op) {
       return {
         ...a,
```

But `UnaryExpression` only allows `UpdateExpression` children, but we need `ExpressionizedStatement`.  So I've added part of `NonAssignmentExtendedExpression` (which already had nice `Nested` logic) as a possible child for `UnaryExpression`.

### DIFF
--- a/source/main.coffee
+++ b/source/main.coffee
@@ -94,6 +94,7 @@ uncacheable = new Set [
   "NestedJSXChildExpression"
   "NestedModuleItem"
   "NestedModuleItems"
+  "NestedNonAssignmentExtendedExpression"
   "NestedObject"
   "NestedPropertyDefinitions"
   "NonSingleBracedBlock"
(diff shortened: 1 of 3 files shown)


STRd6 (approved):
Looks good!

PR #477, opened 2023-03-25
CLI can run ESM scripts via import

edemaine:
* `civet filename.civet` now works with code that `import`s `.civet` or `.js` files (in new ESM mode)
* `@danielx/civet/esm` now transpiles with `js: true` so we don't *always* need ts-node

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,7 +35,7 @@ civet
 civet -c
 # Compile Civet source file to TypeScript
 civet < source.civet > output.ts
-# Execute a simple .civet script (no imports)
+# Execute a .civet script
 civet source.civet ...args...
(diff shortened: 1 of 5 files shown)


edemaine:
I first tried to add ESM support via `vm`'s [SourceTextModule](https://nodejs.org/api/vm.html#class-vmsourcetextmodule), which also required the `civet` CLI to run a Node subprocess in order to turn on `--experimental-vm-modules`. But it seems to be extremely difficult to actually `link` modules this way; at least, I couldn't find a way to provide the default `loader` behavior.

STRd6 (approved):

PR #479, opened 2023-03-26
Inline implicit object literals can't end with comma

### LINKED ISSUE #478: Inline implicit object literals in function arguments bug
This code should be a function call with two arguments, one object literal and another variable:

```coffee
f x: y, z
```

CoffeeScript [agrees](https://coffeescript.org/#try:f%20x%3A%20y%2C%20z).
But Civet parses the `x: y,` (note the trailing comma) as an object literal, which ends up with this compilation:

```js
f({x: y,}(z))
```

Do we need yet another flag for this, to avoid eating the trailing comma?  Seems like there ought to be a better way...

Maybe `InlineObjectLiteral` should just never consume the final comma; it seems like it should be treated as incorrect.

edemaine:
Fixes #478 by forbidding commas at the end of an inline object literals or implicit `export` list.
The intuition is that commas indicate a continuation onto the next line in these cases, so we can't have arbitrary content on the next line, only more items.

These are now forbidden:

```coffee
a: 1,
b: 2,
```

```coffee
a: 1, b: 2,
```

This matches CoffeeScript behavior, which [forbids](https://coffeescript.org/#try:a%3A%201%2C) [both](https://coffeescript.org/#try:a%3A%201%2C%0Ab%3A%202%2C).

**Open question:** I'm not sure that the above really need to be forbidden, when at the top level like this.  Mainly I want to forbid the second one in an inline context like function arguments.

Non-inline object literal with extra commas are still allowed, [as in CoffeeScript](https://coffeescript.org/#try:f%0A%20%20a%3A%201%2C%0A%20%20b%3A%202%2C):

```coffee
f
  a: 1,
  b: 2,
```

---

**Open question:** Currently, the comma lets you break indentation: the following parses to a single object literal.

```coffee
if condition
  x: 1,
y: 2
```

This is [different from CoffeeScript](https://coffeescript.org/#try:if%20condition%0A%20%20x%3A%201%2C%0Ay%3A%202%0A). Should I replace `__` with `Samedent / _?`?

### DIFF
--- a/source/main.coffee
+++ b/source/main.coffee
@@ -38,6 +38,7 @@ uncacheable = new Set [
   "ConditionalExpression"
   "Declaration"
   "Debugger"
+  "Dedented"
   "ElementListWithIndentedApplicationForbidden"
   "ElseClause"
   "Expression"
--- a/source/parser.hera
+++ b/source/parser.hera
@@ -2240,24 +2240,20 @@ NestedPropertyDefinition
     }))
 
 InlineObjectLiteral
-  InsertInlineOpenBrace:open SnugNamedProperty ImplicitInlineObjectPropertyDelimiter ( ( Samedent / _* ) ImplicitNamedProperty ImplicitInlineObjectPropertyDelimiter )* InsertCloseBrace:close ->
+  InsertInlineOpenBrace:open SnugNamedProperty:first ( ImplicitInlineObjectPropertyDelimiter ImplicitNamedProperty )*:rest InsertCloseBrace:close ->
     return {
       type: "ObjectExpression",
-      children: [open, $2, $3, ...$4, close],
+      children: [open, first, ...rest, close],
     }
 
 # This is different from ObjectPropertyDelimiter because the braces are implicit so we can't look ahead to find the closing one
 # Instead we see if the next line matches a NamedProperty and if so we insert a comma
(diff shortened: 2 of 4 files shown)


STRd6:
v0.5.90 currently matches the CoffeeScript behavior for dedent after a trailing comma: https://civet.dev/playground?code=aWYgY29uZGl0aW9uCiAgeDogMSwKeTogMgo%3D so we should probably keep that behavior.

Instead of having the `__` after the comma consuming trailing space you can keep `( Samedent / _? )` inside the `*:rest` iteration in `InlineObjectLiteral`.

edemaine:
OK, I attempted to fix both open problems.  Now, like CoffeeScript, trailing commas are allowed everywhere except the top level.  More precisely, they're allowed if there's a following dedent. For example, this works:

```coffee
if x
  a: 1,
```

But this still doesn't:

```coffee
a: 1,
```

This seems like a reasonable trade-off (CoffeeScript didn't seem to mind) because it's not very helpful to have an object literal at the very top level, whereas it's very useful inside a function (via implicit return).

Naturally, I also prevented commas from being followed by a dedent as a condition. `ImplicitInlineObjectPropertyDelimiter` is still doing the newline handling because they're still not symmetric.

STRd6 (approved):

PR #482, opened 2023-03-29
implicit JSX element

STRd6:
Implicit JSX elements.

Defaults to `div` but can be set with `"civet defaultElement=span"` etc.

```
<.main>
```

```
<div class="main" />
```

```
"civet defaultElement=span"
<.app>
```

```typescript
<span class="app" />
```

Currently limited to requiring `.class` or `#id` because it could be ambiguous with some JSX attributes.


edemaine (approved):
Nice! Thanks for implementing this!

PR #484, opened 2023-03-30
fix #483

### LINKED ISSUE #483: `class` doesn't parse as key when in implicit object literal
```
class: ->
  "active"
```

STRd6:


edemaine (approved):

source/parser.hera
```diff
@@ -3829,7 +3829,7 @@ KeywordStatement
 
   # https://262.ecma-international.org/#prod-ReturnStatement
   # NOTE: Modified to leave room for `return.value` and `return =`
-  Return !("." / AfterReturnShorthand) MaybeNestedExpression?:expression -> {
+  Return !":" !("." / AfterReturnShorthand) MaybeNestedExpression?:expression -> {
```
edemaine, line 3832:
```suggestion
  Return !(":" / "." / AfterReturnShorthand) MaybeNestedExpression?:expression -> {
```

Avoids another undefined in $0 (but not a big deal)

source/parser.hera
```diff
@@ -4650,7 +4650,7 @@ Catch
     return { $loc, token: $1 }
 
 Class
-  "class" NonIdContinue ->
+  "class" !":" NonIdContinue ->
```
edemaine, line 4653: 1. Should we be consistent about whether colon is excluded from the low-level rule like this, or where it's used (like how you moved Void)?
2. Any chance it makes sense to put colon exclusion in NonIdContinue? Or would that break other things?

edemaine (approved):

source/parser.hera
STRd6: Ideally they would all be higher up (like in the unary ops).

edemaine (approved):

PR #495, opened 2023-04-01
Fixes #488

### LINKED ISSUE #488: Class member initializers ending with type parameters
This [doesn't parse](https://civet.dev/playground?code=Y2xhc3MgRmV0Y2hlcgogIGNhY2hlID0gbmV3IE1hcDxJRCwgUGVyc29uPgo%3D):

```coffee
class Fetcher
  cache = new Map<ID, Person>
```

These [do](https://civet.dev/playground?code=Y2xhc3MgRmV0Y2hlcgogIGNhY2hlID0gbmV3IE1hcDxJRCwgUGVyc29uPigpCmNsYXNzIEZldGNoZXIKICBjYWNoZSA9IG5ldyBNYXA%3D):

```coffee
class Fetcher
  cache = new Map<ID, Person>()
class Fetcher
  cache = new Map
```

STRd6:


source/parser.hera
```diff
@@ -982,10 +981,10 @@ AfterReturnShorthand
 # The parameters in a function definition
 Parameters
   NonEmptyParameters
-  "" ->
+  TypeParameters?:tp Loc:p ->
```
edemaine, line 984: This changes the current input: `<div>=> something`
which currently compiles to `<div>{() => something}</div>` but now gets treated as a parameterized arrow function

But perhaps that JSX is not something that should have been supported in the first place, because it shadows a TypeScript arrow function. In fact, `<For>(item) => foo` already wasn't supported as JSX; the function had to start on the next line.

For symmetry, though, I think this JSX should probably be supported, and currently isn't with this PR:
```coffee
<div>
  => something
```
Perhaps another `__` that should be a `_?`?

Also, `<T>=> x` and `<T>(x) => x` currently output arrow functions starting with `<T>` but that's not valid TSX. The output should start with `<T,>`.  I thought we fixed this, but I can't find any code that does it...  We ought to do it for both `TypeParameters` and `TypeArguments`, when there's one parameter that's just a single word.

source/parser.hera
```diff
@@ -422,8 +422,8 @@ ArrowFunction
 
 FatArrow
   # Ensures at least one space before arrow
```
edemaine, line 424: This comment seems inaccurate (both before and after), so perhaps should remove it (or remove the `?` from `_?`).
I *think* it makes sense to allow `<T>=>` though I'm not totally sure.

STRd6: It's "ensure" and not "require" it's mainly describing that inside the handler a `" "` is added before the `=>` if none present.

edemaine: Ohh, I see. Never mind then!

Could maybe clarify the comment by mentioning the "output". But it's OK as is.

edemaine (approved):

PR #496, opened 2023-04-01
Fix #493

### LINKED ISSUE #493: Missing parentheses in pipe call when starting with paren
```coffee
id
|> (@ as any) as (id: ID) => string
```

[compiles to](https://civet.dev/playground?code=aWQKfD4gKEAgYXMgYW55KSBhcyAoaWQ6IElEKSA9PiBzdHJpbmcK)

```ts
(this as any) as (id: ID) => string(id)
```

but should be

```ts
((this as any) as (id: ID) => string)(id)
```

I take it there's a test for whether the content starts with `(`, or maybe whether one child is a `ParenthesizedExpression` (as something like `@ as any` does get wrapped in parens), but it should really check whether the content is exactly one `ParenthesizedExpression`.

Though the code above looks weird, the actual application where I saw this failure is:

```coffee
id
|> (@ as any)[methodName] as (id: ID) => string
```

STRd6:


source/lib.js
```diff
@@ -640,18 +641,9 @@ function processUnaryExpression(pre, exp, post) {
...
-  } else if (Array.isArray(exp)) {
-    const children = [...pre, ...exp]
-    if (post) children./**/push(post)
-    return { children }
-  } else {
-    const children = [...pre, exp]
-    if (post) children./**/push(post)
-    return { children }
+  return {
+    type: "UnaryExpression",
```
edemaine, line 645: Maybe worth adding `exp,` property here? Or we could wait till we need it.

edemaine (approved):

PR #497, opened 2023-04-01
source-map-support dependency

### LINKED ISSUE #489: ESM loader requires @cspotcode/source-map-support
Currently `dist/esm.mjs` includes the line

```js
import sourceMapSupport from "@cspotcode/source-map-support"
```

Was this supposed to be inlined via esbuild?

If not, I guess anyone using `--loader @danielx/civet/esm` needs to have `@cspotcode/source-map-support` installed locally.

Unfortunately, this means that the ESM version of the CLI still isn't working without locally installing the dependency. (It was only working in my `npm link` testing, because that gives access to Civet's dependencies, not just `dist`.)

For what it's worth, even when I do install the dependency locally, I get the error message `No source map found in code` in console, and none of the line numbers are correctly remapped. Probably a separate issue though.

STRd6:
Fixes #489

### DIFF
--- a/package.json
+++ b/package.json
@@ -37,6 +37,9 @@
   },
   "author": "Daniel X. Moore",
   "license": "MIT",
+  "dependencies": {
+    "@cspotcode/source-map-support": "^0.8.1"
+  },
   "devDependencies": {
     "@danielx/hera": "^0.8.0",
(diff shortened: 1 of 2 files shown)


edemaine (approved):
I don't have a good way to test this until we release, but I can do so afterward.

PR #506, opened 2023-04-15
x@y and @@x bind shorthand, plus JSX fixes

edemaine:
As discussed in Discord:
* `x@y` and `x@.y` → `x.y.bind(x)`
* `@@x` and `@@.x` → `this.x.bind(x)`
* Longer chains, optional property access, and non-null assertions also work
* No shadowing of `@@` decorators (this came for free, I guess because they are checked first)
* (No refs yet)

I also noticed that a bunch of updates to how properties work hadn't made it to the JSX fork. So I also fixed unbraced attribute values with:
* Additional access or calls for `super` properties and dynamic `import`
* Object globs
* New bind shorthand
* Bug fix: `<input checked/>` didn't work because of our "assert there's a following space" (to avoid invalid class name shorthand) didn't allow for `/>`

### DIFF
--- a/civet.dev/cheatsheet.md
+++ b/civet.dev/cheatsheet.md
@@ -991,6 +991,14 @@ id := @id
 obj := { @id }
 </Playground>
 
+### Bind
+
+<Playground>
+bound := object@.method
+bound := object@method
+bound := @@method
+</Playground>
+
 ### Static Fields
 
 <Playground>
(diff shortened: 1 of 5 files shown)


STRd6 (approved):
Looks good, glad it ended up being so easy!

For a future JSX follow-up we can add:

```
<Component @onClick />
<Component @@onClick />
```

Shorthands as well

PR #516, opened 2023-05-08
handle non-binding multiple pattern alternatives

STRd6:


source/parser.hera
```diff
@@ -3450,6 +3440,17 @@ CaseClause
       children: $0
     }
 
+PatternExpressionList
+  ( BindingPattern / ConditionFragment ):first ( _? Comma _? ( BindingPattern / ConditionFragment ) )*:rest ->
```
edemaine, line 3444: It might make sense to make `BindingPattern / ConditionFragment` into a separate rule (`PatternExpression`?). Plausibly `ConditionFragment` could be renamed into this, unless we want to use it elsewhere.

source/parser.hera
```diff
@@ -7989,22 +7990,34 @@ Init
             return
           }
 
-          let { pattern, block } = c
+          let { patterns, block } = c
+          let pattern = patterns[0]
```
edemaine, line 7994: I think this definition of `pattern` isn't used. (I also can't figure out what it would be for... maybe indentation measuring?)

source/parser.hera
```diff
@@ -7989,22 +7990,34 @@ Init
...
               return [" && ", ...c]
-            }), ")"],
-            expression: conditions,
+            })
+
+            if (i === 0) return conditionArray
+            return [" || ", ...conditionArray]
+          })
+
+          const condition = {
```
edemaine, line 8017: We could use `module.makeLeftHandSideExpression` here, which avoids parens in some simple cases. Though maybe we need a variety of such parenthesization functions...

edemaine (approved):
Nice, this worked out quite easily! Super helpful and more compatible with `when` clauses.

source/parser.hera
edemaine: Ohh, I see. Maybe worth adding a `testCase.skip` or adding to/opening new issue to document this. (In a perfect world, an error message would be useful, but that doesn't seem that important assuming we'll fix it soon enough.) ((Actually, how do you plan to transpile that...?))

source/parser.hera
edemaine: Oops, sorry, makes sense. We could probably still make a "parenthesize" helper, but we're building nodes all over the place so not sure that's worth singling out.

source/parser.hera
STRd6: The current plan is:

```
if (Array.isArray(p) && p.length === 2) {
  const [x, y] = p;
  ...
else if (typeof p === "object" && p != null && "x" in p && "y" in p
) {
  const { x, y } = p;
  ...
}}
```

I'll add a test case.

Essentially adding a new condition and duplicating the block body with a different binding prefix.

edemaine: Oh nice, that seems like it would work well. A bit if a code parsing penalty from the code duplication, but at least it won't penalize network time much thanks to gzip.

PR #517, opened 2023-05-12
Pattern matching gather duplicate bindings

STRd6:
This fixes duplicate binding merging for ArrayBindingPatterns and adds support for merging with BindingRestElement as well.

### DIFF
--- a/source/lib.js
+++ b/source/lib.js
@@ -10,16 +10,21 @@
  * Adjust the alias of a binding property, adding an alias if one doesn't exist or
  * replacing an existing alias. This mutates the property in place.
  */
-function aliasBindingProperty(p, ref) {
+function aliasBinding(p, ref) {
   if (p.type === "Identifier") {
+    // Array element binding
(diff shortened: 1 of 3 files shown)


source/lib.js
```diff
@@ -6,6 +6,30 @@
...
+    // TODO: This ignores `name` and `names` properties of Identifier and
+    // hackily converts it to a container for a Ref.
+    p.children[0] = ref
+  } else if (p.type === "BindingRestElement") {
+    aliasBinding(p.binding, ref)
+  } else if (p.value?.type === "Identifier") {
+    // aliased property binding
+    p.value = ref
+    // TODO: nasty
+    p.children[5] = ref
```
edemaine, line 25: Indeed 😂 Could you add to the comment a pointer to which rule this is?

edemaine (approved):
Impressive!!

source/lib.js
STRd6: I found a better way to handle it :)

PR #518, opened 2023-05-13
Pattern spread

STRd6:


civet.dev/cheatsheet.md
```diff
@@ -849,6 +849,16 @@ switch e
...
 
+::: info
+Patterns can aggregate duplicate bindings.
+:::
+
+<Playground>
+switch x
+  {a: y, b: y}
+    y
+</Playground>
```
edemaine, line 860: This is reasonable, though a more realistic example would be something like:

```js
switch x
  [{type}, {type}]
    type
```

edemaine (approved):

PR #526, opened 2023-05-20
Fix a typo when loading `civetPlugin`

iliocatallo:
It [looks like](https://github.com/DanielXMoore/Civet/blob/b57e75e6c6dbd8acbbf542c7e67cc017958fbecd/lsp/build/build.civet#L19) `civetPlugin` is a function. I changed the documentation accordingly. 

ciao!


STRd6:
Thanks!

PR #529, opened 2023-05-23
ref decs for pattern matching. Fixes #523

### LINKED ISSUE #523: Pattern matching throwing a ReferenceError
[Playground Link](https://civet.dev/playground?code=c3dpdGNoICJ0dXJuIGxlZnQiLnNwbGl0KCIgIikKIFsidHVybiIsIGRpcmVjdGlvbl0KICBjb25zb2xlLmxvZyBgVHVybmluZyAke2RpcmVjdGlvbn0hYA%3D%3D)

```civet
switch "turn left".split(" ")
 ["turn", direction]
  console.log `Turning ${direction}!`
```
outputs
```ts
if (
  Array.isArray(m) &&
  m.length === 2 &&
  m[0] === "turn"
) {
  const [, direction] = m;
  console.log(`Turning ${direction}!`);
}
```
where m is never defined.

STRd6:


edemaine (approved):
Looks good! 

Maybe add a test for when the `switch` is used as an expression? Though I think this doesn't involve any actual hoisting because it gets wrapped in an iife.

source/parser.hera
```diff
@@ -6649,20 +6649,31 @@ Init
...
+        prefix = "(()=>"
+        suffix = ")()"
       }
-      if (Array.isArray(exp)) {
-        return [prefix, ...exp, suffix]
-      } else {
-        return [prefix, exp, suffix]
+
+      // TODO: This rest prevents an infinite recursion bug, ideally it wouldn't be necessary
+      const expressions = Array.isArray(exp) ? [[...exp]] : [exp]
```
edemaine, line 6663: Could we get away with `[...exp]` in the first case?

I admit I don't yet see where infinite recursion would arise, but trust that aliasing is a tough problem...

edemaine (approved):

source/parser.hera
STRd6: `[...exp]` caused a problem of expanding `[ws, exp, delimiter]` statement tuples that were passed, causing a problem with inserting the hoisted ref. Unifying the statement nodes should help with that.

Now that we're using `lib.js` we could even update to `lib.ts` or `lib.civet` someday.

PR #530, opened 2023-05-23
post rest trailing commas and elisions. Fixes #525

### LINKED ISSUE #525: Empty elements after rest
```js
[..., nextToLast, ] = array
// ↓↓↓ Civet 0.6.2
([...ref] = array), ([nextToLast] = ref.splice(-1));
```

This seems incorrect; I expected `nextToLast` to be the next-to-last element, not the last one, as in:

```js
[..., nextToLast, last] = array
// ↓↓↓ Civet 0.6.2
([...ref] = array), ([nextToLast, last] = ref.splice(-2));
```

I'm guessing this is a consequence of how we deal with trailing commas in array literals. We shouldn't do the same in array patterns.

However, I notice that [CoffeeScript has the same issue](https://coffeescript.org/#try:%5B...%2C%20nextToLast%2C%20%5D%20%3D%20array), so it's perhaps not obvious what the right decision is here. Thoughts?

STRd6:


edemaine (approved):
Nice!

PR #535, opened 2023-05-28
refs for object glob left hand side

STRd6:


source/lib.js
```diff
@@ -546,10 +546,20 @@ function processCallMemberExpression(node) {
...
+      // add ref to ensure object base evaluated only once
+      if (prefix.length > 1) {
+        const ref = {
+          type: "Ref",
+          base: "ref",
+        }
+        hoistDec = [["", ["const ", ref, " = ", prefix], ";"]]
+        prefix = [ref]
+      }
+      prefix = prefix.concat(glob.dot)
```
edemaine, line 561: I think you could just do `prefix.push(glob.dot)` here because in all cases `prefix` is a fresh array.

source/lib.js
```diff
@@ -989,7 +1000,21 @@ function hasYield(exp) {
...
+      switch (type) {
+        case "ObjectExpression":
+          node.hoistDec = null
+          while (node.parent?.type !== "BlockStatement") {
+            node = node.parent
+          }
+          if (node.parent) {
+            insertHoistDec(node.parent, node, hoistDec)
+          } else {
+            throw new Error("Couldn't find block to hoist declaration into.")
```
edemaine, line 1014: I wonder how robust this is. For example, maybe wrong when in an `if` expression that should should circuit? I think it would be better/safer to hoist a `var` declaration, which is always possible, and do assignment as a comma operator.

test/object.civet
```diff
@@ -835,13 +835,12 @@ describe "object", ->
...
 
-    // TODO: needs ref
     testCase """
       complex left-hand side
       ---
       x.y()?.z.{a,b}
       ---
-      ({a:x.y()?.z.a,b:x.y()?.z.b})
+      const ref = x.y()?.z;({a:ref.a,b:ref.b})
     """
```
edemaine, line 844: I'd like to see more contexts for hoisting tested...

edemaine (commented):
Thanks for taking this on! Great to expand the hoisting infrastructure. Some questions though...

source/lib.js
STRd6: `prefix` is a shared ref in the `hoistDec` array so we need a clean ref to not mutate it.

source/lib.js
STRd6: I agree with this criticism. I think that changing `hoistDec` to `["let ", ref]` and hoisting that to the closest enclosing block should work. Then in the places where it is used have `["(", ref, " = ", complexValue, ",", exp, ")"]`. This should make it work with arbitrary expressions as well as simplifying this whole deal. Hopefully I can get to it pretty soon.

edemaine (approved):
New revision looks great!  It'd still be nice to have more tests, including maybe one like `if condition and x := expensive()` that would benefit from this change.  But also good as is.

PR #537, opened 2023-05-31
Use refs for indexed object shorthand. Fixes #415

### LINKED ISSUE #415: Non-idempotency in braced literals
Braced literal expressions involving function calls duplicate them; this is problematic in the case that those functions do IO or observe or mutate state, as seen below:
https://civet.dev/playground/?code=aSAuPSAwCmYgOj0gKCkgPT4gaSsrCnQgOj0gWyJhIiwgImIiXQpjb25zb2xlLmxvZyB7dFtmKCldfQ%3D%3D

STRd6:


test/object.civet
```diff
@@ -774,6 +774,14 @@ describe "object", ->
       ({[y]: x[y]})
     """
 
+    testCase """
+      member access with expression
+      ---
+      {x[f(y)]}
+      ---
+      let ref;({[ref = f(y),ref]: x[ref]})
```
edemaine, line 782: I just tested this on Chrome and it seems that JavaScript doesn't allow the comma operator inside brackets like this?? It looks like we need to wrap in parentheses like so:

```js
let ref;({[(ref = f(y),ref)]: x[ref]})
```

STRd6: Thanks, great catch!

edemaine (approved):

PR #540, opened 2023-06-02
Allow for method and constructor overloading in classes. Fixes #538

### LINKED ISSUE #538: How do I do constructor overloading in civet?
First there is another parsing error on this test file: 

```ts
class Test 
  variable: string[] 
  constructor({x: number, z: Nesting, y: SubElement[]})
```
(yes this isn't a valid class yet without an implementation but the error is a parse error not a syntax error)
```
test.civet:3:52 Failed to parse
Expected:
	Whitespace /[\s]+/
	JSMultiLineComment "/*"
	CoffeeHereCommentStart /###(?!#)/
	JSSingleLineComment /\/\/(?!\/)[^\r\n]*/
	Equals "="
	NonNewlineWhitespace /[ \t]+/
	InlineComment "/*"
	Comma ","
	ObjectPropertyDelimiter "}"
	EOL /\r\n|\n|\r|$/
Found: "[]})"
ts
```

And in general I am curious how best to implement overloaded constructors in civet. Didn't see it in the cheatsheet. 
I want to allow passing in a single object with all params and passing it params as args, so something like: 

```ts
class Test < SomeOther
  # In TS i would have to write the clunkier constructor(obj: {x:number, y:number, z:number} = {x: 0, y: 1, z: []}), this is an easy syntax improvement win (although would need an error like "need to define defaults for all values of an object param")
  @(obj: {x: number = 0, y: number = 1, z: number[] = []})
  # do i do @(public x: number = 0, ...) ? 
  @(p: number = 0, y: number = 1, z: number[] = [])
      if (y == undefined) {
        let {x, y, z} = p;
      } else {
        let x = p;
      }
      super(x,y)
      this.z = z
```

(with the parse error I can't tell whether this works or not)

it would be quite amazing if civet could intelligently construct a "unified" implementation from overloads (functions and methods and constructors) that have different types and arities, like so: (the fact that typescript doesn't do this automatically kind of surprises me)

```ts
function test(obj: {x: number, y:number=1, z:number[]=[]}): Type
  test(x, y, z)
function test(x: number, y:number = 1, z: number[]=[]): Type
  // ... does something
``` 

would turn into something like: 
```ts
function test(obj: {x:number, y:number, z:number[]} = {y:1, z:[]}): Type;
function test(x: number, y:number=1, z:number[] = []): Type; 
function test(xOrObj: number | {x:number, y:number, z:number[]}, y:number=1, z:number[] = []): Type {
  if (y == undefined) { 
     // boilerplate destructure
     let {x,y,z} = xOrObj; 
     // copies implementation here
     return test(x, y, z) 
  }
  // ... does something [copied here]
}

```
could also make two dummy functions which copy both implementations and call them in place with the correct arity.
```ts
function test(xo: number | {x:number, y:number, z:number[]}, y: number=1, z: number[] = []) {
  if (y == undefined) {
    // copy implementation here as IIFE
    return (({x:number, y:number, z:number[]}) => test(x, y, z)})(xo)
  } 
  return ((x,y,z) => {
     // ... does something
  })(x, y, z)
}
```
although of course first option is preferred.

Idk if civet understands types but even doing this for different arities would be a big dev boon

STRd6:


edemaine (approved):
The fix looks good, making constructors consistent with methods and functions with respect to declarations vs. implementations.

Unfortunately, this change really shows a limitation to our decision, because it's quite common to have a constructor with no body. Indeed, I naturally wrote several examples in [constructor documentation](https://civet.dev/cheatsheet#constructor). This PR should at least change those, perhaps to add `;` at the end (my preferred "empty block" notation).

Going forward (perhaps for another PR), I wonder if we should look over the equally named methods/functions in a block, and if all of them have no body (and no `declare` or other explicit declaration syntax, like being in a `declare` block), add an empty body to the last one. I think we can do this for functions everywhere and methods in everything except braced object literals (where `foo()` means `foo: foo()`), and would be a better overall experience. What do you think?

edemaine:
Actually, I wonder about addressing this issue before we merge, because this change would probably actually break Civet code in the wild. In general that's not terrible, but if it's needless we might better avoid it.

STRd6:
Adding an implicit empty body to the last method that shares a name seems simple enough and a good idea. I'll add it in.

edemaine:
Nice! The code looks like it requires the method declarations to be consecutive, though. Perhaps we could instead check the next item in the `gatherRecursiveAll`, if it has the same parent? Ah, even that requires them to be consecutive among the method declarations. Ideally we'd cluster by method names, maybe via a Map. And ideally we'd do so for function declarations too.

STRd6:
In TS overloads have to be immediately followed by the implementation.

edemaine:
Oh, I didn't realize that. Nice! (Technically, it still compiles, but generates an error. But I guess what we generate will generate an error too.)

Just need to extend to functions then.

PR #542, opened 2023-06-03
Function implicit bodies

edemaine:
Following up on #540 discussion, this PR makes `function`s also have an implicit empty block when they're used in an expression context (e.g. not a `declare` block, and not a braced object literal), similar to methods behavior added in #540. Also added tests for both.


STRd6 (approved):
Nice, I was thinking about the same thing.

PR #545, opened 2023-06-03
Fix implicit calls with bind and decorators

### LINKED ISSUE #539: Bind in implicit function call
`@@foo` works, but `f @@foo` fails to parse

edemaine:
Fixing #539 turned out to be easier than expected because we already had a flag. The code was buggy though. 😅 

Fixes #539, and fixes `ClassImplicitCall` rules which used the wrong flag and were backwards. Whoops!


STRd6 (approved):
Nice, thanks for the fix! I was wondering why it was behaving wacky when I implemented the feature. 😹

edemaine:
My bad for not spotting it in review. 😹

Issue #536, opened 2023-05-30
Babel plugin / Using Civet with metro bundler in react native


bennidi:
It is itching under my finger nails to try out civet in one of my personal projects to build a react native mobile application.
Is there a plugin available for the metro bundler? What would be the recommended way of integration?
Does anybody have experience with Civet and react native stack?

edemaine:
@bennidi Can you try the new Babel plugin to see how it works for you?

https://github.com/DanielXMoore/Civet/blob/main/source/babel-plugin.mjs

bennidi:
Thank you so much for that work. I am currently on a two month vacation. I
can not promise anything but I definitely keep it in mind for the next
programing sessions.

etherealHero:
I'm trying to use civet in react native and I get the following error:

![babel-need-cjs](https://github.com/user-attachments/assets/24de0433-233d-4e4f-9756-5cf2f55906c1)

I tried to solve the problem of setting up babel config for react native to make it work, but I couldn't find anything. 

Then I tried modifying civet's source a bit so that it also created the commonjs babel plugin, and it worked fine.

Here's what you need to do in your react native project to support .civet (if civet will also create babel in commonjs):

```diff
// install civet in react native project
yarn add -D @danielx/civet

// include .civet files to module resolution in metro.config.js
-const config = {};
+const config = {
+  resolver: {
+    sourceExts: ['js', 'jsx', 'json', 'ts', 'tsx', 'civet'],
+  },
+};

// add plugin in babel.config.js
 module.exports = {
   presets: ['module:@react-native/babel-preset'],
+  plugins: [['@danielx/civet/babel-plugin']],
+  sourceMaps: 'inline',
 };
```

I'll try to post a PR with babel commonjs added. I can also make a repository with an example react native project using civet to add a link to the civet site if required (Just like others have done with Solid templates)

bennidi:
@etherealHero This is amazing. Thank your for picking up on this and finding a solution. I hope the PR gets merged soon.

Side question: Will you use Civet full on or just for specific parts? Do you have experience using civet in other react projects?

etherealHero:
[...]

PR #550, opened 2023-06-08
basic babel plugin

STRd6:
Started on a basic babel plugin. It looks like it works but I still need to add automated testing and some more complex test cases.


STRd6:
Fixes #536 when complete.

edemaine (commented):
Another thought: is it possible for the plugin to have options, which could get passed to Civet `compile`?

Also wonder whether inline sourcemaps should enabled by default. (Probably should test.)

STRd6:
This works fine with simple setups and inline maps. It transpiles to JS then hooks into the default babel parse. In theory it could better hook into the TypeScript babel plugins but I have no idea how they work despite reading through the code.

https://github.com/babel/babel/blob/5f74b510ff251af5bf7fbbf25c1c934ffcb6df52/packages/babel-plugin-syntax-typescript/src/index.ts
https://github.com/babel/babel/blob/5f74b510ff251af5bf7fbbf25c1c934ffcb6df52/packages/babel-plugin-transform-typescript/src/index.ts
https://github.com/babel/babel/blob/5f74b510ff251af5bf7fbbf25c1c934ffcb6df52/packages/babel-preset-typescript/src/index.ts
https://github.com/babel/babel/blob/5f74b510ff251af5bf7fbbf25c1c934ffcb6df52/packages/babel-plugin-syntax-jsx/src/index.ts

Still need to add JSX support.

edemaine:
I wonder if you could do this kind of code to enable TypeScript when we don't have `javascript` turned on?

https://github.com/babel/babel/blob/5f74b510ff251af5bf7fbbf25c1c934ffcb6df52/packages/babel-plugin-syntax-typescript/src/index.ts#L60-L61

Alternatively, does TypeScript (and JSX) just work if those plugins are enabled along with this one? Seems fine to require the user to enable the relevant plugins.  And given that you're passing `opts` through as is, I would have expected code like the above to pass the correct options to the parser. (What error are you getting?)

STRd6:
I'm not going to work on this any more right now but it may be useful for some projects using Babel. If anyone tries it and hits specific issues I can try to look into them when I have time. I figure that getting this out is better that it just sitting in limbo forever.

PR #551, opened 2023-06-09
Implement config file

esthedebeste:
Add `--config` option to specify a config file,
otherwise looks for a `config.civet`, `.civetconfig`, or `civet.json` file in the current directory and all parent directories. If found, it is loaded and used to override the default options.

config.civet and .civetconfig get compiled & run, but can't import other civet files.
 - `export default`s the config object

civet.json is just parsed as json

### DIFF
--- a/lsp/source/lib/typescript-service.mts
+++ b/lsp/source/lib/typescript-service.mts
@@ -435,10 +435,7 @@ function TSService(projectURL = "./") {
   CivetConfig.findConfig(projectPath).then(async configPath => {
     if (configPath) {
       console.info("Loading Civet config @", configPath)
-      const envLsp = process.env.CIVET_LSP
-      process.env.CIVET_LSP = "true"
       const config = await CivetConfig.loadConfig(configPath)
-      process.env.CIVET_LSP = envLsp
       console.info("Found civet config!")
       civetConfig = config
     } else console.info("No Civet config found")
--- a/lsp/source/lib/typescript-service.mts
+++ b/lsp/source/lib/typescript-service.mts
@@ -6,6 +6,7 @@ import type {
 	CompileOptions,
 } from "@danielx/civet"
 import BundledCivetModule from "@danielx/civet"
+import BundledCivetConfigModule from "@danielx/civet/config"
 
 import ts, {
   CompilerHost,
@@ -431,10 +432,13 @@ function TSService(projectURL = "./") {
   }
 
   let civetConfig: CompileOptions = {}
-  CivetConfig.findConfig(projectPath).then(async (configPath: string) => {
+  CivetConfig.findConfig(projectPath).then(async configPath => {
     if (configPath) {
       console.info("Loading Civet config @", configPath)
+      const envLsp = process.env.CIVET_LSP
+      process.env.CIVET_LSP = "true"
       const config = await CivetConfig.loadConfig(configPath)
+      process.env.CIVET_LSP = envLsp
       console.info("Found civet config!")
       civetConfig = config
     } else console.info("No Civet config found")
@@ -479,10 +483,10 @@ function TSService(projectURL = "./") {
 
   function transpileCivet(path: string, source: string) {
     return Civet.compile(source, {
-			...civetConfig,
-			filename: path,
-			sourceMap: true,
-		})
+      ...civetConfig,
+      filename: path,
+      sourceMap: true,
+    })
   }
 }
 
--- a/lsp/source/lib/typescript-service.mts
+++ b/lsp/source/lib/typescript-service.mts
@@ -1,8 +1,11 @@
 import assert from "assert"
 import path from "path"
 
+import type {
+	SourceMap as CivetSourceMap,
+	CompileOptions,
+} from "@danielx/civet"
 import BundledCivetModule from "@danielx/civet"
-import type { SourceMap as CivetSourceMap } from "@danielx/civet"
 
 import ts, {
   CompilerHost,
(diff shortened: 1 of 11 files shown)


source/cli.coffee
```diff
@@ -33,14 +34,13 @@ if process.argv.includes "--help"
...
       -o / --output XX Specify output directory and/or extension, or filename
       -c / --compile   Compile input files to TypeScript (or JavaScript)
+      --config XX      Specify a config file (default finds the nearest config.civet, .civetconfig or civet.json file)
       --js             Strip out all type annotations; default to .jsx extension
       --ast            Print the AST instead of the compiled code
       --inline-map     Generate a sourcemap
       --no-cache       Disable compiler caching (slow, for debugging)
 
     You can use - to read from stdin or (prefixed by -o) write to stdout.
-
```
STRd6, line 42: These newlines should stay in, they are necessary to insert a newline after printing out the help message.

source/config.coffee
```diff
@@ -0,0 +1,34 @@
...
+
+export loadConfig = (path) ->
+    config = await fs.readFile path, 'utf8'
+    if path.endsWith '.json'
+        JSON.parse config
+    else
+        js = compile config, js: true
+        
+        tmpPath = path + ".civet-tmp-#{Date.now()}.mjs"
+        await fs.writeFile tmpPath, js
```
edemaine, line 26: I wonder if you could share code with `cli.coffee`?

source/parser.hera
```diff
@@ -6300,6 +6300,8 @@ Reset
       verbose: false,
     }
 
+    module.config = parse.config = Object.assign({}, DEFAULT, parse.config)
```
edemaine, line 6303: Will this interact correctly with virtual attributes like `coffeeConfig`? I worry it will just overwrite them, but perhaps not.

edemaine (commented):
On the topic of config naming: I don't like that `config.civet` and ``civet.json` have such different basenames. I feel like that's nonstandard? Maybe `civetconfig.{civet,json}` would be better, with or without `.` prefix? It is a bit redundant though, so I'm willing to be overruled/counterargued here. (Maybe we should support *both* `civetconfig.civet` and `config.civet`...)

Also, there's a movement to allow configuration files in `.config` subdirectories of the project root, so we should ideally support that too. See e.g. https://github.com/cosmiconfig/cosmiconfig (I also heard about it on the SyntaxFM podcast)

Currently there's no `compilerOptions` wrapper in the config file, which seems reasonable. But I wonder if we want to allow specifying other overrides here, for example, including a tsconfig inside a civetconfig, or having LSP- or CLI-specific configuration. So perhaps a `civet` wrapper *would* be helpful?

source/config.coffee
esthedebeste: Was thinking this too, but we need to get the `export default` from the config file. `cli.coffee` only needs to run it, so it can just start a new node process.

(there is also the option of wrapping the entire CLI with the `.civet` ESM loader so that both are just an `await import`, but node.js loaders are experimental and can change at any time)

source/parser.hera
esthedebeste: fixed in d4ef779 👍

esthedebeste:
> On the topic of config naming: I don't like that `config.civet` and ``civet.json`have such different basenames. I feel like that's nonstandard? Maybe`civetconfig.{civet,json}` would be better, with or without `.` prefix? It is a bit redundant though, so I'm willing to be overruled/counterargued here. (Maybe we should support _both_ `civetconfig.civet` and `config.civet`...)

Yeah, some people have file extensions hidden as well. I'm up for adding `civetconfig.{civet,json}`, it also fits well with the `tsconfig.json` that will likely be next to it.

> Also, there's a movement to allow configuration files in `.config` subdirectories of the project root, so we should ideally support that too. See e.g. https://github.com/cosmiconfig/cosmiconfig (I also heard about it on the SyntaxFM podcast)

Sounds cool, adding it!



> including a tsconfig inside a civetconfig, or having LSP- or CLI-specific configuration

I feel like having both typescript and civet in the same codebase will be decently common.
For LSP/CLI-specific configuration, I'm thinking we could load the config file with `process.env.LSP="true"`? 
```coffee
lsp := process.env.LSP?
export default {
  js: !lsp
}
```

STRd6:
Thanks for the contribution!

Looks pretty good overall and I think we can merge this soon.

One thing I'd like is a `--no-config` option that prevents auto-loading the config file so that we don't have to specify an empty config to avoid scanning for many non-existent files.

That option should also be used in `build.sh` to keep the build scripts that use `civet` as fast as possible.

esthedebeste:
> One thing I'd like is a `--no-config` option that prevents auto-loading the config file so that we don't have to specify an empty config to avoid scanning for many non-existent files.
> 
> That option should also be used in `build.sh` to keep the build scripts that use `civet` as fast as possible.

Done 👍

lsp/source/lib/typescript-service.mts
```diff
@@ -416,17 +420,32 @@ function TSService(projectURL = "./") {
...
     console.info("USING BUNDLED CIVET")
-    Civet = BundledCivetModule
   }
 
+  let civetConfig: CompileOptions = {}
+  CivetConfig.findConfig(projectPath).then(async configPath => {
+    if (configPath) {
+      console.info("Loading Civet config @", configPath)
+      const envLsp = process.env.CIVET_LSP
+      process.env.CIVET_LSP = "true"
```
STRd6, line 439: Setting `process.env.CIVET_LSP` isn't safe in concurrent situations. It would be better to pass an option to `loadConfig`.

esthedebeste: How do you suggest passing information to the config file? `loadConfig` uses an `await import`. Would a global `const CIVET_LSP: boolean` or `const CIVET_CONFIG_INFO: { lsp?: boolean, ... }` suffice?

I'm also in favor of just removing the CIVET_LSP environment variable entirely, to always have CLI and LSP behave the same.

STRd6: You could pass a second optional parameter to `loadConfig` that way each call to load will have it's own copy even if they are concurrent. Removing `CIVET_LSP` is also fine for now.

Thanks for your work on this PR!

esthedebeste: Removed it

STRd6 (approved):

PR #553, opened 2023-06-09
Fixes #552. Don't hoist decs to bare blocks

### LINKED ISSUE #552: Declarations in Conditions break with `else if`
```civet
if x := 0
 console.log "hi"
else if y := 1
 console.log "hey"
 ```
 Compiles to
 ```ts
let ref;
if ((ref = 0)) {
  const x = ref;
  console.log("hi");
} else let ref1;
if ((ref1 = 1)) {
  const y = ref1;
  console.log("hey");
}
```
which doesn't function (`else let x;` is invalid) and also has wrong control flow

STRd6:


edemaine (approved):
Nice fix.

PR #556, opened 2023-06-22
Added optional prefixed union syntax for TypeScript. Fixes #544

### LINKED ISSUE #544: TypeScript incompatability: prefixed type union syntax
```ts
type X = 
  | number
  | string
```
is valid TS to represent the `number | string` type. Civet currently throws a ParseError on the initial `|`
[Playground Link](https://civet.dev/playground?code=dHlwZSBYID0gCiAgfCBudW1iZXIKICB8IHN0cmluZw%3D%3D)

STRd6:
I'm pretty sure this is the correct place in the grammar because of this example: https://www.typescriptlang.org/play?experimentalDecorators=true#code/C4TwDgpgBAKlC8UCuA7A1ig9gdxQKFEigFUEoBnYAJwEsUBzPAY0xUqgA8AuKAMligAfEmQBEITKKA

### DIFF
--- a/source/parser.hera
+++ b/source/parser.hera
@@ -5816,9 +5816,11 @@ Type
   TypeConditional
 
 TypeBinary
-  TypeUnary ( __ TypeBinaryOp __ TypeUnary )* ->
-    if ($2.length) return $0
-    return $1
+  ( __ TypeBinaryOp __ )?:optionalPrefix TypeUnary:t ( __ TypeBinaryOp __ TypeUnary )*:ops ->
+    if (!ops.length && !optionalPrefix) return t
+    if (!ops.length) return [optionalPrefix, t]
+    if (!optionalPrefix) return [t, ...ops]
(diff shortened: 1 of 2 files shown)


source/parser.hera
```diff
@@ -5816,9 +5816,11 @@ Type
   TypeConditional
 
 TypeBinary
-  TypeUnary ( __ TypeBinaryOp __ TypeUnary )* ->
-    if ($2.length) return $0
-    return $1
+  ( __ TypeBinaryOp __ )?:optionalPrefix TypeUnary:t ( __ TypeBinaryOp __ TypeUnary )*:ops ->
```
edemaine, line 5819: I find it a bit awkward that the leading `__` is inside the optional part... So only with the prefix does this allow leading whitespace.

But I see TypeUnary has the same issue so maybe this is cleanup for later.

edemaine (approved):

Issue #555, opened 2023-06-16
Unicode operator support


edemaine:
Some folks don't use ligatures, but still would like nice notation. Here are some operators that we could support in a simple transpilation sense:

| Unicode | ASCII equivalent |
|---|---|
| `≤` | `<=` |
| `≥` | `>=` |
| `≠` | `!=` |
| `≡` | `===` (I guess) |
| `≢` | `!==` |
| `²` | `** 2` |
| `²³` | `** 23` (ideally) |
| `²⁺³` | `** (2+3)` (ideally) |
| `≔` | `:=` |
| `⁇` | `??` |
| `⩲` | `+=` (I guess) |
| `⧺` | `++` (I guess) |
| `‖` | `\|\|` |
| `≪` | `<<` |
| `≫` | `>>` |
| `⋙` | `>>>` |
| `⊕` | `^` or `^^` (maybe, common notation for XOR) |
| `¬` | `!` (maybe) |
| `…` | `...` |
| `‥` | `..` |
| `≟` | `?=` |
| `⩻` | `<?` (instanceof shorthand) |
| `∈` | `is in` |
| `∉` | `is not in` |
| `▷` | `\|>` |

Let me know if I missed any! There are some pretty interesting opportunities with shorthand like `∈`. APL here we come.

Note: These Unicode characters are relative easy to type using a compose key. On Windows, I use [WinCompose](http://wincompose.info/), which emulates a feature built into X-Windows (I believe).

This is somewhat related to the (unwritten) proposal to support emoji in identifiers — ideally translating e.g. `😊` to `smiling` or similar, for JS interop, though it might be easier to translate it to something like `$1F60A`, to avoid a lookup table. It might be nice to "support" other mathematical operators like `≹` as identifiers, so that users can define them via `operator`. (Or maybe such Unicode operators are automatically infix, but need to be defined via `function` or `operator`.)

fabiospampinato:
If `num²` is equal to `num ** 2`, should `numᵖᵒʷ` be equal to `num ** pow`? Maybe it's not worth it.

GoToLoop:
Keyword `var` is still lacking a shorthand for it though.

orenelbaum:
make it configurable so I can use whatever characters and emojis I want, it would be the best feature in Civet

edemaine:
@orenelbaum Thoughts on what this would look like? I wonder if templates (#587) would be a good answer, if you could define your own symbols like so:
```js
template 🧪
  try
[...]

PR #558, opened 2023-06-26
Unicode operators ≤≥≠≢≡≣⩶⩵«»⋙‖⁇∈∉∋∌▷‥…≔→⇒

edemaine:
For fun, here are some simpler and less controversial Unicode operators from #555.

The only one I'm less sure about is `≡` mapping to `===`. But given that `≠` makes `!=` and `≢` makes `!==`, this feels natural. And anyway it's good to encourage `===` over `==`. Alternatively, we could use `≣` for `===` and `≡` for `==`...

Incidentally, TIL there are symbols for double and triple equals: `⩵` and `⩶`.

Going forward:
* I'm less sure about the symbols with equals and such because they look like comparisons to me, but Civet uses them as assignments.
* I think the power operators, specifically `²`, would be handy, but that's a little more complicated, so left for the future.
* If there are other Unicode operators you think I should add now, let me know!

### DIFF
--- a/civet.dev/cheatsheet.md
+++ b/civet.dev/cheatsheet.md
@@ -332,6 +332,7 @@ a?
 item is in array
 item is not in array
 substring is in string
+item1 ∈ container ∌ item2  // Unicode
 </Playground>
 
 ### Assignment Operators
@@ -354,6 +355,7 @@ fun?(arg)
 
 <Playground>
 a < b <= c
+a ≤ b ≤ c ≠ d  // Unicode
 a is b is not c
 a instanceof b not instanceof c
 </Playground>
--- a/civet.dev/cheatsheet.md
+++ b/civet.dev/cheatsheet.md
@@ -609,6 +609,12 @@ circle := (degrees: number): {x: number, y: number} =>
   y: Math.sin radians
 </Playground>
 
+You can also use Unicode arrows:
+
+<Playground>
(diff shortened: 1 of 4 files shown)


STRd6:
I think these are good. `≡` mapping to `===` makes sense to me. `∈` and `∉` are cool. Looking forward to exploring more in the future!

STRd6 (approved):

GoToLoop:
> , we could use `≣` for `===` and `≡` for `==`...

I prefer this combo, I guess. 💭

edemaine:
> we could use `≣` for `===` and `≡` for `==`...

This would follow the pattern: for every line beyond `=`, you add another `=`. @STRd6 what do you think? (With `coffeeEq`, I would imagine `≡` still maps to `===`, just like `==` does.)

STRd6:
> > we could use `≣` for `===` and `≡` for `==`...
> 
> This would follow the pattern: for every line beyond `=`, you add another `=`. @STRd6 what do you think? (With `coffeeEq`, I would imagine `≡` still maps to `===`, just like `==` does.)

This sounds good, either way is fine with me.

edemaine:
OK, I switched, and also added → and ⇒ for functions (obvious in hindsight!) and inclusion operators ∋ and ∌. Plus tests and docs. Let me know if there are any final comments before merging.

STRd6:
Looks good to me! 👍

PR #570, opened 2023-07-06
Fix overflow

VyrekXD:
(I dont know if it fixes it but probably does)

Shows overflow to fix *overflow* cof cof

image

### DIFF
--- a/civet.dev/.vitepress/components/Playground.vue
+++ b/civet.dev/.vitepress/components/Playground.vue
@@ -179,7 +179,7 @@ const playgroundUrl = computed(() => {
   background: transparent;
   resize: none;
   color: transparent;
-  overflow: auto;
+  overflow: hidden;
 }
 
 .compilation-info {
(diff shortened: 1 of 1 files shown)


edemaine:
I tested and this doesn't seem to do anything. You can test by modifying the CSS in the element inspector.

It looks like you might want to use the [`wrap` attribute of `<textarea>`](https://html.com/attributes/textarea-wrap/) (and *maybe*  `word-wrap` and/or `word-break` CSS properties)

VyrekXD:
> I tested and this doesn't seem to do anything. You can test by modifying the CSS in the element inspector.
> 
> It looks like you might want to use `word-wrap` and/or `word-break` CSS properties (and _maybe_ the [`wrap` attribute of `<textarea>`](https://html.com/attributes/textarea-wrap/) or `cols`?)
> 
> Or maybe you just need to set `width` also.

I forgot i could do that, let me fix it 👍

edemaine:
Please re-open when the PR has some changes. 🙂

PR #571, opened 2023-07-06
Fix empty objects in if statements

### LINKED ISSUE #559: An empty object in if statement
Hi,

An empty object inside a conditional statement breaks everything:
```coffee
if condition
  {}
```
compiles to
```js
if (condition)
  {
```
I needed this in a function to return an empty object on some condition, so
```coffee
f = -> 
  if condition
    {}
```
compiles to
```js
f = function() { 
  if (condition)
    {;return
}
```

[Playground](https://civet.dev/playground/?code=aWYgY29uZGl0aW9uCiAge30KCmYgPSAtPiAKICBpZiBjb25kaXRpb24KICAgIHt9)

edemaine:
Fix #559 as discussed in https://github.com/DanielXMoore/Civet/issues/559#issuecomment-1624411555

`{}` will be treated as an empty block if it starts on the same line as the `if`, and if it's on an unindented line.  But if it's on an indented line, it will be treated like an object literal.

I feel like this is sufficiently JS compatible, because the difference between doing nothing and creating an empty object literal (and similarly for doing something vs. creating an object literal with that something) only matters when interacting with implicit returns or `if` expressions, which are already differences from JS anyway.

I also ran into a weird bug with `if (x) {} else {}`, but that's for another issue/PR.

### DIFF
--- a/source/parser.hera
+++ b/source/parser.hera
@@ -1636,9 +1636,11 @@ ImplicitNestedBlock
 
 # NOTE: This is the body of if/else/for/case etc.
 Block
-  ExplicitBlock
   # NOTE: Added indentation based implied braces
   ImplicitNestedBlock
+  # NOTE: Explicit block after implicit block so that a properly indented `{}`
+  # gets treated like an object literal, not an empty block.
+  ExplicitBlock
 
   ThenClause
   # NOTE: !EOS prevents capturing a following unindented Statement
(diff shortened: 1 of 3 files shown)


STRd6 (approved):

STRd6:
Thanks, good catch! The missing `{` is pretty spooky.

PR #574, opened 2023-07-07
Fix missing closing braces

### LINKED ISSUE #572: Missing closing braces
` if (x) {} else {}` somehow transpiles to have no closing braces.

edemaine:
Fixes #572

This makes me wonder the following: is it safe to modify `$0` in place?  In this case it isn't pretty (`$0.splice(2, 0, expressions)`) but often we want to just replace a specific item in `$0` with a wrapped version.

### DIFF
--- a/source/parser.hera
+++ b/source/parser.hera
@@ -1612,7 +1612,7 @@ ExplicitBlock
     return {
       type: "BlockStatement",
       expressions,
-      children: [$1, expressions, $2],
+      children: [$1, $2, expressions, $3, $4],
       bare: false,
       empty: true,
     }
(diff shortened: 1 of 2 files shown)


STRd6 (approved):

STRd6:
> This makes me wonder the following: is it safe to modify `$0` in place? In this case it isn't pretty (`$0.splice(2, 0, expressions)`) but often we want to just replace a specific item in `$0` with a wrapped version.

It's generally not safe to mutate input parameters in a rule due to caching.

edemaine:
Yeah, I wasn't sure about `$0` specifically — whether that array was constructed during the "parent" or "child" rule.  Sounds like it's re-used though.

PR #575, opened 2023-07-07
Fix binary ops RHS in pattern matching switch

### LINKED ISSUE #573: <? Operator behaves weirdly while pattern matching
```ts
switch value
 <? "string"
  value
 []
  value.map(&.toString())
 true
  value

// ↓↓↓ Civet 0.6.11
if (value instanceof "string") {
  return value;
} else if (Array.isArray(value) && value.length === 0) {
  return value.map(($) => $.toString());
} else if (value === true) {
  return value;
}

```

edemaine:
Fixes #573

### DIFF
--- a/source/lib.js
+++ b/source/lib.js
@@ -832,13 +832,13 @@ function expandChainedComparisons([first, binops]) {
           endIndex = i + 1
         }
 
-        results = results.concat(first, ...binops.slice(start, endIndex))
+        results.push(first, ...binops.slice(start, endIndex).flat())
         first = [exp].concat(binops.slice(index + 1, endIndex))
         start = endIndex
       })
     } else {
       // Advance start if there was no chain
-      results = results.concat(first, ...binops.slice(start, i + 1))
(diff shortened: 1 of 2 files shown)


source/lib.js
```diff
@@ -832,13 +832,13 @@ function expandChainedComparisons([first, binops]) {
...
 
-        results = results.concat(first, ...binops.slice(start, endIndex))
+        results.push(first, ...binops.slice(start, endIndex).flat())
         first = [exp].concat(binops.slice(index + 1, endIndex))
         start = endIndex
       })
     } else {
       // Advance start if there was no chain
-      results = results.concat(first, ...binops.slice(start, i + 1))
+      results.push(first, ...binops.slice(start, i + 1).flat())
```
edemaine, line 841: These changes from `concat` to `push` turned out to not be necessary, but they would be if `first` were even an array (so that it gets treated like a single thing). I left the change in also because it should be faster to be in-place.

STRd6 (approved):
Looks good, thanks!

PR #576, opened 2023-07-08
Forbid implicit calls with braced argument in extends/if/else/for/when/case

### LINKED ISSUE #543: `class extends` doesn't work with braced body
```ts
export class CustomError extends Error {
 constructor(message: string) {
  super(message) 
  this.name = "CustomError"
 }
}
```
Compiles to a function call on the `Error` constructor;
```ts
export class CustomError extends Error({
  constructor(message: string) {
    super(message);
    this.name = "CustomError";
  },
}) {}
```

edemaine:
Fixes #543


STRd6 (approved):
This looks pretty good. We may need to refine it further in the future but this is a definite improvement for now.

One other thing to check:

```
class X extends Y
  z=1
```

Thanks for the fix!

PR #577, opened 2023-07-08
Fix nested object with function children (caching)

### LINKED ISSUE #533: Strange behavior when declaring single-line functions in an object
Hi,

First of all, I would like to thank you for what you are doing here. I wish the project all the best.

Here's a bug I think (it's hard to imagine that this is intended behavior). I found it in the playground, but just in case I checked it on the command line (civet, version 0.6.4). Tried to find among open issues. Maybe I didn't search well enough.

Сode:
```coffee
obj = 
  a: => 1
  b: => 2
  c: => 3
```

Transpiles to:
```js
obj =  {
  a: () => 1({
  b: () => 2,
  c: () => 3,
}),
}
```

Expected:
```js
obj =  {
  a: () => 1,
  b: () => 2,
  c: () => 3,
}
```

[Open it in the Playground](https://civet.dev/playground/?code=b2JqID0gCiAgYTogPT4gMQogIGI6ID0%2BIDIKICBjOiA9PiAzCgpvYmogPQogIGE6ID0%2BIAogICAgMQogIGI6ID0%2BIDIKICBjOiA9PiAz)

It seems that we can't currently use the first function as a one-liner

A workaround would be to use a line break:
```coffee
obj = 
  a: => 
    1
  b: => 2
  c: => 3
```
Transpiles to:
```js
obj = {
  a: () => {
    return 1;
  },
  b: () => 2,
  c: () => 3,
};
```

edemaine:
Fixes #533 by adding 3 more no-cache rules.  (All three are necessary for this test.)

I also changed one related `__` to `NotDedented` which seems like an improvement (assignment chains shouldn't dedent). But can also change back.


STRd6 (approved):
👏

PR #584, opened 2023-07-16
Allow trailing CallExpression after ExpressionizedStatement

### LINKED ISSUE #582: Trailing .method bug
## Input

```coffee
sorted :=
  for item of list
    item.toString()
.sort()
```

## Current Compilation

```js
const sorted = (() => {
  const results = [];
  for (const item of list) {
    results.push(item.toString());
  }
  return results;
})()(($) => $.sort());
```

## Expected Compilation

```js
const sorted = (() => {
  const results = [];
  for (const item of list) {
    results.push(item.toString());
  }
  return results;
})().sort()
```

## Workaround

```coffee
sorted := (
  for item of list
    item.toString()
).sort()
```

edemaine:
Fixes #582 by allowing on-their-own-line not-dedented `CallExpressionRest` after all `ExpressionizedStatement`, if `TrailingMemberPropertyAllowed` is true.

We had a facility for `TrailingMemberExpressions` but not for `TraillingCallExpressions`.  This sufficed in the past because the trailing member was previously used in a `CallExpression` context, so the call part got captured there.  Here we haven't started a `CallExpression` so I had to build some more rules.

This PR allows for this form from https://github.com/DanielXMoore/Civet/issues/582#issuecomment-1636989322:

```coffee
sorted :=
  for item of list
    item.toString()
  .sort()
```

In the third commit, I also modified the indentation rules in `NestedNonAssignmentExtendedExpression` to allow the originally requested indentation of #582:

```coffee
sorted :=
  for item of list
    item.toString()
.sort()
```

I hope I got these situations correct. It's possible that I'm allowing `CallExpressionRest` where it shouldn't be.

We may want to allow `TrailingCallExpressions` in more contexts too. For example, `for` loops can only get them in expression contexts, such as right-hand sides of assignments, but we might also want them in statement contexts like so:

```coffee
for item of list
  item.toString()
.outputToFile 'foo'
```

This would probably be most useful in combination with implicit returns from functions. The tricky part is that this forces them to transform from `IterationStatement` to `IterationExpression`, which I'm not quite sure how to do. I tried adding `!TrailingCallExpressions` to the end of certain `Statement` rules to force parsing as `ExpressionStatement` but haven't gotten it working yet.

In the first commit, I also enabled some "return-if" tests that were already essentially working.

### DIFF
--- a/source/lib.js
+++ b/source/lib.js
@@ -1439,6 +1439,7 @@ function makeLeftHandSideExpression(expression) {
     case "AmpersandRef":
     case "Identifier":
     case "Literal":
+    case "IterationExpression": // already a CallExpression
     case "CallExpression":
     case "MemberExpression":
     case "ParenthesizedExpression":
--- a/source/parser.hera
+++ b/source/parser.hera
@@ -115,15 +115,28 @@ NonPipelineExtendedExpression
 NonAssignmentExtendedExpression
   # Check for nested expressionized statements first
   NestedNonAssignmentExtendedExpression
-  __ ExpressionizedStatement ->
+  __ ExpressionizedStatementWithTrailingCallExpressions ->
+    if ($2.length) {
+      return [...$1, ...$2]
+    }
     return {...$2,
       children: [...$1, ...$2.children]
     }
 
 NestedNonAssignmentExtendedExpression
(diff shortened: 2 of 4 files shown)


STRd6 (approved):
Looks good!

A possible future optimization would be to unwrap the iife and chain on the results ref.

PR #597, opened 2023-07-30
Fix #596. async function expression

### LINKED ISSUE #596: Language server doesn't support async function expressions
```js
const f = async function() {}
```
`Cannot find name 'async'.`

STRd6:


edemaine (approved):
Thanks for fixing!

PR #600, opened 2023-07-31
Support indented RHS after binary op

### LINKED ISSUE #598: Unnatural indentation
I just ran into a parse error in this natural code:

```js
name := optionalName ??
  if file.name?
    name = file.name
  else
    throw new Error
```

Civet isn't catching the indentation after the `??`, and instead thinks the `else` should be unindented by 2 spaces.

edemaine:
Fixes #598

### DIFF
--- a/source/parser.hera
+++ b/source/parser.hera
@@ -307,10 +307,17 @@ BinaryOpRHS
   # Does not match
   # a
   # +b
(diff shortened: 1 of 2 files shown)


STRd6 (approved):
👍

PR #602, opened 2023-07-31
autoVar/autoLet should treat `=>` and methods same as functions

### LINKED ISSUE #588: autoVar ignores `=>` functions when hoisting
Simplest example:

```js
"civet autoVar"
=>
  x = 5
↓↓↓
var x;
() => {
  return (x = 5);
};
```

Original setting:

```js
"civet autoVar"
function foo
  =>
    x = 5
↓↓↓
function foo() {
  var x;
  return () => {
    return (x = 5);
  };
}
```

Civet treats `function` and `->` the same, but `=>` and methods different. By contrast, CoffeeScript [treats them the same](https://coffeescript.org/#try:-%3E%0A%20%20%3D%3E%0A%20%20%20%20x%20%3D%205).

edemaine:
Fixes #588

### DIFF
--- a/source/lib.js
+++ b/source/lib.js
@@ -2669,7 +2669,7 @@ function createVarDecs(statements, scopes, pushVar) {
     if (!hasDec(x)) return a.indexOf(x) === i
   }).forEach(pushVar)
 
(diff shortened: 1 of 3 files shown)


STRd6 (approved):
👏

PR #603, opened 2023-08-01
Possessive object access

edemaine:
Inspired by [this tweet](https://twitter.com/judofyr/status/1685580152565055488), [this follow-up thread](https://twitter.com/nicuveo/status/1685745407421284352), and [_hyperscript](https://hyperscript.org/expressions/possessive/), via Discord: `'s` followed by whitespace simulates `.`.

```js
mario's brother's name
↓↓↓
mario.brother.name
```

In Unicode, `’s` works too.

I think there are contexts (such as the example above) where this improves readability of code, so seems like a generally good thing. Of course, this might deserve some discussion about whether this feature is a good idea.

The one thing this breaks is template strings starting with `foo's `. Previously, `foo's s'` compiled to ``foo`s s` ``. My sense is that this is not a big deal though; single-quoted template strings (which can't have any interpolation) seems like a bit of a niche feature already (though natural and inherited from CoffeeScript), and starting one with `s` and a space feels especially unusual.

I experimented with adding Japanese possessive form, as in `marioのbrotherのname` (mentioned in the [the follow-up thread](https://twitter.com/nicuveo/status/1685745407421284352)) which would be kind of cool, but it's also kind of breaking: currently `marioのbrotherのname` parses as a single identifier, and I imagine it does in most programming languages, so perhaps Japanese authors use this when naming their variables.  (I kind of want to now...)  We could support a spaced form such as `mario の brother の name`, though that could also be considered an implicit function call to `の`, and I don't think it is very grammatical in Japanese (normally there are no spaces around `の`). So in the end I decided not to pursue this.

### DIFF
--- a/civet.dev/cheatsheet.md
+++ b/civet.dev/cheatsheet.md
@@ -166,13 +166,22 @@ To specify a blank body, use `;` or `{}`.
 
 :::
 
-### Property Access Shorthand
+### Property Access
+
+Many more values can appear after a `.` to access an object property:
 
 <Playground>
 json.'long property'
 json.`${movie} name`
 matrix.0.0
 array.-1
+</Playground>
+
+You can also write property access as an English possessive:
+
+<Playground>
+mario's brother's name
+json's "long property"'s `${movie} name`
 </Playground>
 
 ## Arrays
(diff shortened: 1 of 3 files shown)


STRd6 (approved):
I find this terrifying... so I guess it means we have to try it. 😨😆

PR #606, opened 2023-08-03
Implicit returns of (last) declaration

edemaine:
In the spirit of "everything is an expression", and e.g. `if x := 5`

BREAKING CHANGE: `x := 5` now implicitly returns `x`

More complex example:

```js
->
  let x = 5, y = x
↓↓↓
function() {
  let x = 5, y = x;return y
}
```

Oren motivates it as giving a name to the return value, which helps define a semantic meaning.  It can't really have much other purpose.

I'm pretty sure I remember @STRd6 saying he preferred this behavior. Could be worth a minor version bump though given that it would break some code that uses declarations to prevent return (but this seems like bad practice).

Also fix a bug in `names` computation for `var` declarations. If we don't merge this PR, we should at least fix that.

### DIFF
--- a/source/lib.js
+++ b/source/lib.js
@@ -1699,7 +1699,7 @@ function processAssignmentDeclaration(decl, id, suffix, ws, assign, e) {
   splices = splices.map(s => [", ", s])
   thisAssignments = thisAssignments.map(a => ["", a, ";"])
 
-  const initializer = [assign, e]
+  const initializer = [ws, assign, e]
   const binding = {
     type: "Binding",
     pattern: id,
@@ -1707,7 +1707,7 @@ function processAssignmentDeclaration(decl, id, suffix, ws, assign, e) {
     splices,
     suffix,
     thisAssignments,
-    children: [id, suffix, ...ws, initializer]
+    children: [id, suffix, initializer]
   }
 
   const children = [decl, binding]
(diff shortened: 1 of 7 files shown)


edemaine:
One potential issue: `[x, y] := obj` causes the return of `y`. Should it instead return `[x, y]`?

STRd6 (approved):
`[x, y] := obj` should return `[x, y]` but we can do that in a follow up PR.

The general pattern would be that the const declaration returns the last declaration such as:

`const a=1,b=2` returns `b`
`const {a, b} = c` returns `{a, b}`

If at some point we expand `:=` to allow `a, b := c, d` then that would return `b` as well.

edemaine:
I spent several (painful) hours getting this to work. Along the way, refactored the AST a lot in hopefully useful ways. `BindingProperty` in particular now gains a `delim` property for the delimiter, which makes some manipulation easier.

Took me a while to realize that `BindingElement` and `BindingElementList` had very specific tuple formats, but this is now documented. I think it's also a decent approach, because `RegularExpressionLiteral` needs a wrapper of some sort for the delimiter. (Previously I tried to just put the `delim` property directly on the `BindingElement`, similar to `BindingProperty`, but this made delimiter removal harder and finally broke with `RegularExpressionLiteral`.)

STRd6 (approved):
Thanks! These improvements look great.

I agree that some of the tuple formats could use better documentation. Something we could try is moving `lib.js` -> `lib.ts` and then we can use TypeScript to document them with editor support (should "just work"). I've also got an old exploration on Hera where I was trying to add TS output so that the rules would generate proper TS types but never quite finished it. I should try it again soon.

PR #614, opened 2023-08-06
Revamp braced object literals

### LINKED ISSUE #589: Unexpected function calls in braced object literal with indentation in middle
```js
{url, blob,
  width: img.width
  height: img.height
}
↓↓↓
({
  url,
  blob,
  width: img.width({
    height: img.height,
  }),
});
```

Original context was more like this:

```js
setPrototypes prototypes.length, {url, blob,
  width: img.width
  height: img.height
}
```

I think we explicitly added the ability to add indentation in the middle of an array literal, for cases like

```js
[1, 2
 3, 4]
```

Perhaps we need to do the same for braced object literals.

edemaine:
Fix #589

I should probably redo `ArrayLiteralContent` in a similar way. Currently, it forbids an indented block from starting after a comma (it can only start after an implicit comma from a newline).

### DIFF
--- a/source/parser.hera
+++ b/source/parser.hera
@@ -2195,38 +2195,55 @@ ObjectLiteral
   InlineObjectLiteral
 
 BracedObjectLiteral
-  OpenBrace:open AllowAll ( BracedObjectLiteralContent? __ CloseBrace )? RestoreAll ->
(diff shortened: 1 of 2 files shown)


STRd6 (approved):
Nice!

PR #619, opened 2023-08-07
Unify Samedent/Nested, and other indent cleanup

edemaine:


source/parser.hera
```diff
@@ -6743,17 +6715,31 @@ PopIndent
     }
     module.indentLevels./**/pop()
 
+# "Nested" actually means "new line at current indentation level"
```
STRd6, line 6718: I think it's actually closer to:

"new line at the indentation level set by TrackIndented, where TrackIndented requires that the indentation is more than previously indented"

STRd6 (approved):

PR #621, opened 2023-08-09
for item, index of list

edemaine:
This is part 1 of "easier array iteration".  It lets you do

```js
for item, index of array
↓↓↓
let i = 0;for (const item of array) {
  const index = i++;
```

Part 2 will be more efficient array-specific loops, as in CoffeeScript's `for..in`.


STRd6 (approved):
👏

PR #622, opened 2023-08-11
for key, value in object

edemaine:
This is the equivalent of CoffeeScript's `for key, value of object`.

It should properly use refs, and supports when `key` is actually a pattern (see tests).

### DIFF
--- a/test/for.civet
+++ b/test/for.civet
@@ -256,6 +256,55 @@ describe "for", ->
   """
 
   testCase """
+    in with two implicit declarations
+    ---
+    for key, value in array
+      console.log `${key}: ${value}`
+    ---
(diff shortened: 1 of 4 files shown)


test/for.civet
```diff
@@ -255,6 +255,55 @@ describe "for", ->
...
+    for (var key in object) {
+      let value = object[key];
+      console.log(`${key}: ${value}`)
+    }
+  """
+
+  testCase """
+    in with pattern matching key and value
+    ---
+    for var [a, b], let [x, y] in object
```
STRd6, line 285: 😲 !

STRd6 (approved):

PR #623, opened 2023-08-11
for each..of

edemaine:
Sequel to #621: more efficient array-specific loops, as in CoffeeScript's `for..in`.

Also some code cleanup via new `makeRef` helper function.


STRd6 (approved):
Nice, now we don't have to give up the good part's of CoffeeScript's for loops any more!

PR #627, opened 2023-08-13
Fix hoisting around IIFE

### LINKED ISSUE #624: Hoisting of generated variables
I was trying out the `for item, index in array` syntax and came across a hoisting bug in civet.
Using the loop as a statement by itself works fine
```coffee
for ai, i of a
     ai + b[i]
```
And so does using it as an expression in global scope
```coffee
(for ai, i of a
     ai + b[i]
)
```
but when using it as an expression inside a nested scope like function
```coffee
function x()
    (for ai, i of a
        ai + b[i]
    )
```
it throws this error `Couldn't find expression in block for hoistable declaration.`.

Another behaviour I noticed, assigning the expression to a new variable (`:=`,`.=`) makes it work:

```coffee
function x()
  b := (ai + b[i] for ai, i of a)
```
but assigning it to an already declared variable (`=`) throws the same error.

It occurs because the way `for item, index of array` is compiled, the `index` variable is declared outside the iife closure, rather than inside.

edemaine:
Fixes #624

* `hoistRefDecs` needed to happen after expressionizing
* Parent pointers needed updating after `wrapIIFE` so that new wrapping block gets discovered.
* In these situations, `hoistDec` needs to work with raw expressions that lack `[ws, exp, delim]` format.
* Remove some extra array wrapping

### DIFF
--- a/source/lib.js
+++ b/source/lib.js
@@ -30,6 +30,31 @@ function addParentPointers(node, parent) {
   }
 }
 
+/**
+ * Just update parent pointers for the children of a node,
+ * recursing into arrays but not objects.  More efficient version of
+ * `addParentPointers` which just injecting one new node.
+ */
+function updateParentPointers(node, parent, depth = 1) {
+  if (node == null) return
+  if (typeof node !== "object") return
+
+  // NOTE: Arrays are transparent and skipped when traversing via parent
+  if (Array.isArray(node)) {
+    for (const child of node) {
+      updateParentPointers(child, parent, depth)
+    }
+    return
+  }
+
+  if (parent != null) node.parent = parent
+  if (depth && node.children) {
+    for (const child of node.children) {
+      addParentPointers(child, node, depth-1)
+    }
+  }
+}
+
 function addPostfixStatement(statement, ws, post) {
   let children, expressions
   if (post.blockPrefix?.length) {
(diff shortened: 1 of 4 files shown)


source/lib.js
```diff
@@ -30,6 +30,31 @@ function addParentPointers(node, parent) {
...
+    for (const child of node) {
+      updateParentPointers(child, parent, depth)
+    }
+    return
+  }
+
+  if (parent != null) node.parent = parent
+  if (depth && node.children) {
+    for (const child of node.children) {
+      addParentPointers(child, node, depth-1)
```
STRd6, line 53: Should this be `updateParentPointers`?

source/lib.js
```diff
@@ -30,6 +30,31 @@ function addParentPointers(node, parent) {
   }
 }
 
+/**
+ * Just update parent pointers for the children of a node,
+ * recursing into arrays but not objects.  More efficient version of
+ * `addParentPointers` which just injecting one new node.
```
STRd6, line 36: "which" -> "when" ?

source/lib.js
```diff
@@ -2699,12 +2735,12 @@ function processProgram(root, config, m, ReservedWord) {
   processSwitchExpressions(statements)
   processTryExpressions(statements)
 
-  hoistRefDecs(statements)
-
   // Modify iteration expressions
   gatherRecursiveAll(statements, n => n.type === "IterationExpression")
     .forEach((e) => expressionizeIteration(e))
 
+  hoistRefDecs(statements)
```
STRd6, line 2742: Maybe add a note that it is important for this to be after converting iteration expressions.

STRd6 (approved):

PR #630, opened 2023-08-13
Postfix expressions inside indented implicit object literals

### LINKED ISSUE #628: Postfix conditional expressions shouldn't require parens inside object properties
Works:
```
BoardPlayerTemplate
  class: ("opponent" if i)
```

Doesn't work:
```
BoardPlayerTemplate
  class: "opponent" if i
```

edemaine:
Fix #628 using the "right" interpretation for both 1 and 2-property indented implicit object literals.  This turned out to be a one-line change. But it differs from CoffeeScript's interpretation for 1 property.

I didn't make the change for "inline" object literals (which uses `SnugNamedProperty`) because it would change the interpretation of this test from CS's `lexer.coffee`:

```coffee
addTokenData token, initialChunk: yes if i is 0
```

and generally seems dangerous.  Unfortunately, this means that the following doesn't do what you'd expect:

```coffee
function f()
  a: 1 if hasA
  b: 2 if hasB
```

(This is treated as two top-level statements, the second of which gets implicitly returned.)

### DIFF
--- a/source/parser.hera
+++ b/source/parser.hera
@@ -2414,7 +2414,7 @@ PropertyDefinition
 NamedProperty
   # NOTE: CoverInitializedName early error doesn't seem necessary with this parser
   # NOTE: Using PostfixedExpression to allow If/Switch expressions and postfixes
-  PropertyName:name _? Colon ExtendedExpression:exp ->
+  PropertyName:name _? Colon PostfixedExpression:exp ->
     return {
       type: "Property",
       children: $0,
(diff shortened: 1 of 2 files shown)


STRd6 (approved):
Looks good!

STRd6:
We may be able to fine tune the multiline object literal handling with a lookahead or some such.

edemaine:
I was wondering that... A newline after the postfix is a good sign, but the lexer example has that.  Another SnugNamedProperty afterward is good too, but how do you handle the last property? Maybe the first property needs to be handled differently from the others...

edemaine:
Actually, `SnugNamedProperty` is used only for the first property. I think this would work:
```
SnugNamedProperty
  PropertyName Colon PostfixedExpression:exp &( Nested NamedProperty )
  PropertyName Colon ExtendedExpression:exp
```

Unfortunately, with current caching behavior, it seems to cause exponential running time. 😢

PR #631, opened 2023-08-13
Postfix expressions inside inline object literals

edemaine:
Followup to #630 based on the discussion there. I found a way to handle postfix expressions in inline object literals, as long as the prop isn't the only prop. I also required a newline (`Nested`) after the postfix; commas seem like they could be ambiguous.

### DIFF
--- a/source/parser.hera
+++ b/source/parser.hera
@@ -2425,14 +2425,17 @@ NamedProperty
 
 # Named property but doesn't allow any space between name and colon
 # used to distinguish between braceless inline objects and ternary expression conditions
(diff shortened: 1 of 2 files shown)


STRd6 (approved):

STRd6:
One thing we may want to consider for the future: allow postfix expressions inside single line inline literals but have a `coffeeCompat` flag to keep CoffeeScript 2 compatibility since having single line and multi-line have the same precedence is good for consistency.

For people who want to keep the CS2 style behavior the more explicit:

```
{a: b} if y
```

seems more correct.

PR #632, opened 2023-08-15
Add civet unplugin

Mokshit06:
This PR adds an [unplugin](https://github.com/unjs/unplugin) for simultaneously create vite, rollup, esbuild, rspack and esbuild transformer for civet.


STRd6 (approved):
Looks good!

Some possible areas for future work:

- Update the Civet package build step so that people won't need to install a separate package and will be able to import any plugin directly.
- Add some basic tests for each plugin.

Overall this is very helpful as is, thanks!

edemaine (approved):
This is so exciting! Mainly some docs comments, plus I wonder:

Should `@danielx/civet` be a peerDependency (maybe with `*` or `>=` version) instead? Otherwise, wouldn't we need to publish this plugin after every Civet update? Hmm, or maybe package managers deal with this properly when versions don't mismatch. Indeed, I see `vite-plugin-civet` uses a dependency not a peerDependency, and I haven't had issues with using newer versions. But we still might want `*` or `>=` version dependency here so it will default to the latest or whatever dependency is in the project root (I hope?).

> Update the Civet package build step so that people won't need to install a separate package and will be able to import any plugin directly.

I *think* this is how it works already. Even peerDependencies seem to install by default though (despite not for a while).

Reading https://nodejs.org/en/blog/npm/peer-dependencies, it seems like peerDependency is the "right" thing here, but I'm not certain how it works with a regular dependency...

integration/unplugin/README.md
```diff
@@ -0,0 +1,92 @@
...
+    code: string,
+    id: string
+  ) => TransformResult | Promise<TransformResult>;
+}
+```
+
+- `dts`: `unplugin-civet` also supports generating `.d.ts` type definition files from the civet source, which is useful for building libraries.
+- `outputExtension`: Output filename extension to use. Default: `.civet.jsx`, or `.civet.tsx` if `js` is `false`.
+- `js`: Whether to transpile to JS or TS.
+- `transformOutput`: Replace the `civet.compile` tranformer with a custom transformer. It gets passed the civet source and filename, and should return valid TS/JS code.
```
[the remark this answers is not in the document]
Mokshit06: Good catch. I'll update this

integration/unplugin/README.md
```diff
@@ -0,0 +1,92 @@
...
+  transformOutput?: (
+    code: string,
+    id: string
+  ) => TransformResult | Promise<TransformResult>;
+}
+```
+
+- `dts`: `unplugin-civet` also supports generating `.d.ts` type definition files from the civet source, which is useful for building libraries.
+- `outputExtension`: Output filename extension to use. Default: `.civet.jsx`, or `.civet.tsx` if `js` is `false`.
+- `js`: Whether to transpile to JS or TS.
```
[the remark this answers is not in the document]
edemaine: A default of `js: true` would break `dts: true`. We could make `dts` imply `js: false`, but honestly `js: false` is safer; some TypeScript features are not correctly implemented by Civet with `js: true`, and it just costs a bit of time to use already existing TS integration. So I think `js: false` is best.

Mokshit06:
Thank you!

> Update the Civet package build step so that people won't need to install a separate package and will be able to import any plugin directly.

I assume you're talking about building this and outputting to `dist` like done for esbuild plugin in `build/build.sh`?

> Should @danielx/civet be a peerDependency

I thought that peer dependencies were not installed by default, didn't realise that changed in v7. If we update the Civet package and include it in `@danielx/civet` then we won't require it, but if not, then keeping it a peer dependency is probably better.

integration/unplugin/README.md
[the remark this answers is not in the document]
Mokshit06: We could give imply `js` based on the value of `dts` first,
```js
const transpileToJS = options.dts || options.js;
```
in which case both options should work. My main concern was that rollup and webpack can't process typescript without additional plugins, so would need to set `js: true` to make the setup work, but if `js: false` has better typescript features then let's go with that default.

edemaine:
Oh, sorry, now I understand Daniel's comment: make this part of the `@danielx/civet` NPM package. I feel like separately released NPM packages is more common here, but I imagine the extra JS here is pretty minimal so it might make sense (and be convenient) to just include it. This would mean a bunch of the `exports` mappings in `package.json` should go to the top level.

STRd6:
> I assume you're talking about building this and outputting to `dist` like done for esbuild plugin in `build/build.sh`?

Yes, that's correct. Also updating the root Civet `package.json` to expose `./unplugin`, `./vite`, `./esbuild`, etc. We could also replace the existing esbuild plugin that is already exported with this to cut down on duplication.

It will bulk up the published package a little bit compared to separate packages but should tree-shake out just fine thanks to the import mappings.

Let me know if you have any questions about the build step, it's currently manually building a few different files that end up in `dist`.

Thanks again!

integration/unplugin/README.md
```diff
@@ -0,0 +1,92 @@
...
+    code: string,
+    id: string
+  ) => TransformResult | Promise<TransformResult>;
+}
+```
+
+- `dts`: `unplugin-civet` also supports generating `.d.ts` type definition files from the civet source, which is useful for building libraries. Default: `false`.
+- `outputExtension`: Output filename extension to use. Default: `.civet.tsx`, or uses `.civet.jsx` if `js` is `true`.
+- `js`: Whether to transpile to JS or TS. Default: `false`.
+- `transformOutput`: Adds a custom transformer over jsx/tsx code produced by `civet.compile`.
```
edemaine, line 92: Minor point: It might be worth keeping the text that described the arguments to the function.

integration/unplugin/README.md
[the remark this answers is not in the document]
edemaine: Maybe we can include the `{js: true}` in the README where we give recommended workflows for Webpack and Rollup?

edemaine:
I think `tsup` devDependency is missing.  I get
```
'tsup' is not recognized as an internal or external command,
```

Mokshit06:
You have to run `yarn add` inside `integration/unplugin` as well

Mokshit06:
Yarn does support workspaces, but for that the root `package.json` should have `private: true`, which can't be the case for `@danielx/civet` as it needs to be published. I think we need to add it to the build script

edemaine:
That's too bad. Would it be better if there was no `package.json` within `integration/unplugin`? Maybe the script and dep should just move to the top level. I don't like the idea of running `yarn` every build...

Mokshit06:
Hm so we could either move the dependencies up to the top level (shouldn't require moving the script), or we can do `yarn install` inside `integration/unplugin` on `postinstall`. That way whenever `yarn add` is ran on the top level, the unplugin deps should also be installed

edemaine:
Either way seems workable to me, though I'm not sure for the motivation of a separate `package.json` if we're fully integrating. @STRd6 do you have a preference?

edemaine (changes requested):
Now there seems to be an infinite build loop, both for me and for CI:

```
yarn build                                                    (unplugin-civet $)
yarn run v1.22.18
$ bash ./build/build.sh
$ bash ./build/build.sh
$ bash ./build/build.sh
$ bash ./build/build.sh
$ bash ./build/build.sh
```

Maybe you're using a different version of Yarn? v1 is standard, not v2. Or maybe you didn't test. 😉 

Because there's no longer a `integration/unplugin/package.json`, I believe `yarn --cwd ./integration/unplugin build` just runs the same script as `yarn build`. I think you want to run `tsup` instead. Also, can you add type checking to the top-level `yarn test` script?

Mokshit06:
Yeah I think I didn't test this change 😆 I have added a `private` package.json for `integration/unplugin` as calling `tsup` from root package.json required rewriting all paths in `integration/unplugin/tsup.config.ts` to be relative to the root rather than the unplugin directory.

integration/unplugin/src/index.ts
```diff
@@ -1,4 +1,5 @@
 import { TransformResult, createUnplugin } from 'unplugin';
+// @ts-ignore
```
edemaine, line 2: This doesn't seem to be necessary. Right?

package.json
```diff
@@ -54,7 +54,7 @@
     "docs:build": "yarn build && vitepress build civet.dev",
     "docs:preview": "yarn build && vitepress preview civet.dev",
     "prepublishOnly": "yarn build && yarn test",
-    "test": "c8 mocha"
+    "test": "c8 mocha && tsc"
```
edemaine, line 57: @STRd6 This is going to interact with #638. But I guess we can always add to `exclude` until we're ready.

tsconfig.json
```diff
@@ -93,16 +93,18 @@
     // "allowUnusedLabels": true,                        /* Disable error reporting for unused labels. */
     // "allowUnreachableCode": true,                     /* Disable error reporting for unreachable code. */
     /* Completeness */
-    // "skipDefaultLibCheck": true,                      /* Skip type checking .d.ts files that are included with TypeScript. */
-    // "skipLibCheck": true                              /* Skip type checking all .d.ts files. */
+    "skipDefaultLibCheck": true,                      /* Skip type checking .d.ts files that are included with TypeScript. */
+    "skipLibCheck": true                              /* Skip type checking all .d.ts files. */
```
edemaine, line 97: Is this needed, given that you're excluding `dist` directories? It would be better to leave it false if possible. Ah, I see, it turns off checking of `node_modules`, which currently fails (from missing `@types` perhaps?). Still, seems like this turns off a lot of actual type checking... though we could leave it for later. Same for `skipDefaultLibCheck` which seems risky.

build/build.sh
```diff
@@ -19,5 +22,10 @@ rm dist/cli.js
 # types
 cp types/types.d.ts dist/types.d.ts
 
+# unplugin
+yarn --cwd ./integration/unplugin build
+cp ./integration/unplugin/dist/* ./dist
+
```
edemaine, line 28: For Windows, I realize `chmod a-x dist/*.*[jt]s` would be helpful.

Also, what is this new `dist/chunk-UZRN3RFA.mjs` file? Presumably it's not desired?

integration/unplugin/src/index.ts
Mokshit06: Hm yeah it looks like it works without this as well. That's weird because earlier it was throwing a type error that `@danielx/civet` is not available to `integration/unplugin`

tsconfig.json
Mokshit06: This is needed because `unplugin`'s dts imports `webpack`, `vite`, `rollup`, `rspack`, and `esbuild` out of which we don't have `webpack` and `rspack` required in Civet, and `@typescript/vfs` imports `lz-string`. So when typescript does type checking it errors out on these imports. We could also install types for these packages, or add them to dev deps instead

build/build.sh
Mokshit06: For the esm build, tsup splits the common chunks into a separate file by default. As all the bundler entrypoints `/esbuild`, `/rollup` etc. import the central `civetUnplugin` it splits that into this chunk and imports it in them. Mostly helps keep the bundle size smaller as its not duplicated

edemaine: Ah, that is nice. Can we choose a better name for that file? Given that there's only one, maybe a custom name rule would work?

Mokshit06: We can choose a better name for it https://esbuild.github.io/api/#chunk-names, but i'm not sure if we should. There currently only exists one of these, but might be more than that in the future. I would suggest keeping it as is, or just turning off splitting

tsconfig.json
STRd6: Adding dev dependencies for types sounds good.

package.json
edemaine: #638 is now merged. Can you merge `main` and test what happens now? We might need to `exclude` it for now.

build/build.sh
edemaine: I think it would be nice to rename it for now; we can always change it if we need multiple chunks in the future (seems unlikely for the unplugin?). Maybe `unplugin-shared`?

edemaine (commented):
I realize the examples from your repo aren't here. Probably they should be, for manual testing and for documentation?

Also, a reminder to finish the last couple tweaks here — would be great to get this out!

Mokshit06:
Should be good to land now!

STRd6:
TS errors may be fixed by merging latest main now that our internals are in Civet.

STRd6 (approved):
👍 Looks good!

@edemaine any further comments?

integration/unplugin/examples/rollup/package.json
```diff
@@ -0,0 +1,5 @@
+{
+  "scripts": {
+    "build": "npx rollup --config rollup.config.js"
+  }
```
edemaine, line 4: Shouldn't all the example `package.json`s have a dependency of `@danielx/civet`? Or are we waiting for a version that actually has the unplugin?

Actually, currently when I try to build an example I get e.g.
`Error: Cannot find module 'C:\Users\edemaine\Projects\Civet\node_modules\@danielx\civet\vite'`
(despite having Civet `npm link`ed)
I think it's finding the older Civet that we depend on for building. So this would be fixed once we release an unplugin version, so maybe worry about this later.

I did get the examples to build via `npm link @danielx/civet`. (Well, I tested only Vite and Astro.)

edemaine (approved):
Looks good to me! Just tried out building including some examples.

yarn.lock
```diff
@@ -182,51 +182,101 @@
   resolved "https://registry.yarnpkg.com/@esbuild/android-arm64/-/android-arm64-0.17.19.tgz#bafb75234a5d3d1b690e7c2956a599345e84a2fd"
   integrity sha512-KBMWvEZooR7+kzY0BtbTQn0OAYY7CsiydT63pVEaPtVYF0hXbUaOyZog37DKxK7NF3XacBJOpYT4adIJh+avxA==
 
+"@esbuild/android-arm64@0.18.20":
```
edemaine, line 185: `yarn.lock` might need updating; it changed when I ran `yarn`

integration/unplugin/examples/rollup/package.json
Mokshit06: They probably should be having `@danielx/civet` as a dependency. Without doing npm link in that specific folder currently it traverses up to the node_modules that do have `@danielx/civet`, which happens to link to older version.

Mokshit06:
Made the suggested changes!

PR #636, opened 2023-08-17
Added private this shorthand. Fixes #633

### LINKED ISSUE #633: Automatic `this.` with private class fields
@fabiospampinato suggests shorthand for `this.#x`, shorter than the existing `@#x`:

```js
#x → this.#x
// but beware:
obj.#x → obj.#x
#x in obj → #x in obj
```

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Private_class_fields

STRd6:
Added shorthand for `#a` -> `this.#a` except in the case of branded `in` check.

Could use docs.

### DIFF
--- a/civet.dev/cheatsheet.md
+++ b/civet.dev/cheatsheet.md
@@ -1173,6 +1173,20 @@ bound := object@method
 bound := @@method
 </Playground>
 
+### Private Fields
+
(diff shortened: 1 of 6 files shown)


edemaine (approved):
Oops, forgot to approve. Added docs. Feels very ergonomic! *Almost* makes me want to use private fields. 🙂

PR #638, opened 2023-08-18
source/lib.js -> source/lib.ts

STRd6:
This is the very beginning of switching `lib.js` -> `lib.ts`

I probably won't have the fortitude to finish it all in one go but it works as is and we can continue to improve it bit by bit.

We're no worse off than without the typechecking even if it is so noisy as to not be useful yet.

If we merge this then we can keep adding types and eventually it will be good.

### DIFF
--- a/source/lib.ts
+++ b/source/lib.ts
@@ -6,11 +6,12 @@
  * here eventually.
  */
 
-type ASTString = string & { children: undefined }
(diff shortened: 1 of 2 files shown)


edemaine (approved):
This seems like a great start!  I'm fine with doing this piecemeal, and we should probably merge before we get conflicts with other `lib` additions.

PR #644, opened 2023-08-20
for own..in

edemaine:
Yet another feature from CoffeeScript for loops:

```js
for own key in object
↓↓↓
for (key in object) {
  if (!Object.hasOwn(object, key)) continue;
```

I also switched from `Object.prototype.hasOwnProperty` to `Object.hasProp` (still without using `Object` as an explicit name),
 as recommended in https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/hasOwn

Also fixed up types some around this code (but there are still errors).

### DIFF
--- a/civet.dev/cheatsheet.md
+++ b/civet.dev/cheatsheet.md
@@ -987,6 +987,14 @@ for key, value in object
   console.log `${key} maps to ${value}`
 </Playground>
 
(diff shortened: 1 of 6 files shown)


STRd6 (approved):
Looks good, nice to see the type improvements too!

Issue #642, opened 2023-08-19
Extremely slow compilation


zolomatok:
Something's up with **Civet 0.6.22**  on **macOS** (or maybe on specifically **Apple Silicon**).

This:

```
let obj =
    irtiwo: ""
    fdsf: ""
    dfs: ""
    gdfgdf: ""
    fdfgdfg: ""
    fdssg: ""
    fgdfdgd: ""
    gdfgh: ""
    gjghj: ""
    zrz: ""
    vvd: ""
    oiul: ""
    ohgfh: ""
    plhd: ""
    bkd: ""
    rir: ""
```

compiles in **38 seconds** with
`civet < test.civet > out.ts`

Compilation time seems to grow in a non-linear fashion **with every added key/value pair**.
The keys are distinct and the type of value I don't think matters. I got the same result with random numbers as values. The playground on the website seem to not have this issue.

---

Adding just **1 more** key/value pair to the above code, the compilation reliably errors out after about **a minute**:

<--- Last few GCs --->

[1965:0x118008000]    59949 ms: Scavenge 892.6 (909.6) -> 876.7 (909.6) MB, 0.5 / 0.0 ms  (average mu = 0.912, current mu = 0.886) allocation failure 
[1965:0x118008000]    59973 ms: Scavenge 892.6 (909.6) -> 876.7 (909.6) MB, 0.6 / 0.0 ms  (average mu = 0.912, current mu = 0.886) allocation failure 
[1965:0x118008000]    59996 ms: Scavenge 892.6 (909.6) -> 876.7 (909.6) MB, 0.5 / 0.0 ms  (average mu = 0.912, current mu = 0.886) allocation failure 

<--- JS stacktrace --->

FATAL ERROR: invalid array length Allocation failed - JavaScript heap out of memory
 1: 0x101317585 node::Abort() (.cold.1) [/usr/local/bin/node]
 2: 0x1000b25c9 node::Abort() [/usr/local/bin/node]
 3: 0x1000b272f node::OnFatalError(char const*, char const*) [/usr/local/bin/node]
 4: 0x1001f6eb7 v8::Utils::ReportOOMFailure(v8::internal::Isolate*, char const*, bool) [/usr/local/bin/node]
 5: 0x1001f6e53 v8::internal::V8::FatalProcessOutOfMemory(v8::internal::Isolate*, char const*, bool) [/usr/local/bin/node]
[...]

PR #645, opened 2023-08-21
cache2: 2 cache 2 furious

STRd6:
Fixes https://github.com/DanielXMoore/Civet/issues/642


source/main.coffee
```diff
@@ -7,149 +7,53 @@ import * as util from "./util.coffee"
...
 uncacheable = new Set [
-  "ActualAssignment"
+  # Meta
+  "Debugger"
+  "Init"
+  "Program"
+  "Reset"
+
+  # Indentation
+  "InsertIndent"
```
edemaine, line 23: This shouldn't be needed, as it uses but doesn't modify state. I confirmed this locally, but I'm getting a timeout for the test `should sourcemap complex civet files` — perhaps we should increase our 2000ms timeout for slower machines?

source/main.coffee
```diff
@@ -7,149 +7,53 @@ import * as util from "./util.coffee"
...
+  "PushIndent"
```
edemaine, line 24: Shouldn't need this one either, because it calls `TrackIndented` that does the actual work. But I'm fine with keeping it.

source/parser.hera
```diff
@@ -6633,6 +6633,23 @@ Reset
...
+      type: "ParserMeta",
+      children: [],
+      getStateKey() {
+        const stateInt =
+          ((module.currentIndent.level % 256) << 8) |
+          (module.classImplicitCallForbidden << 7) |
+          (module.indentedApplicationForbidden << 6) |
+          (module.bracedApplicationForbidden << 5) |
+          (module.trailingMemberPropertyForbidden << 4) |
+          (module.newlineBinaryOpForbidden << 3)
```
edemaine, line 6647: Other than current indentation level, we shouldn't really need shifting here, just ORing, because we're working with booleans. I confirmed that this is about 30% faster here: (please double check)

https://jsbench.me/mdllkhsbo3/1

source/main.coffee
```diff
@@ -201,42 +105,44 @@ export compile = (src, options) ->
...
-          return {
-            cache: if result then { ...result }
-          }
+      return if uncacheable.has(ruleName)
+
+      key = [ruleName, state.pos, ...getStateKey()]
+
+      if stateCache.has(key)
+        # logs.push "".padStart(stack.length * 2, " ") + ruleName + ":" + state.pos + "💰"
+        result = stateCache.get(key)
```
edemaine, line 121: Looking at the implementation of `has` here, we're doubling our work. Maybe just call `get` and check for undefined?

Actually, I worry that the implementation of `has` isn't that good: it's not caching `undefined` results. Isn't that what we use to say that a rule didn't match? I'm worried we're currently not caching non-matches.

edemaine (commented):
Wow, great that this is passing now! Some comments for optimization, and I think an important caching bug (maybe making exponentials, though I could be wrong).

source/main.coffee
edemaine: Hmm, the `undefined` issue goes all the way to Hera. It generates:
```js
if (result.cache)
  return result.cache`
```
One solution would be to instead generate
```js
if ('cache' in result)
  return result.cache
```

source/parser.hera
STRd6: It only made around a ~1% difference for me. I think the cost of any bit operations is 100x less than the other stuff we're doing.

One possible optimization would be to try out different orders of the cache key parts to try and localize the partitions to where they are used the most in practice. (Splitting on pos, state, JSX, then rule maybe...)

source/main.coffee
STRd6: We do need both `PushIndent` and `TrackIndented`. If we cache `PushIndent` then sometimes we would use the cached one and not actually push the new indent after backtracking.

We need to cache the state modifying rules up to the point where they balance within a parent rule so we need to cache all the way up to `PushIndent` and `PopIndent` since they must be paired in the grammar.

source/main.coffee
STRd6: 👍

edemaine (approved):
Looking great! So exciting to have this worked out.

edemaine:
By the way, I tried implementing a fancier StateCache that doesn't store recursive Maps until there is more than one item in the bucket, at each recursive level. But in my experiments it seemed to be a bit slower. Maybe one-element Sets are really efficient in Node? I'm pasting my code here in case it inspires something.

<details><summary>Fancy <code>source/state-cache.mts</code></summary>

```ts
type Cache<T> = T | CacheMap<T> | CacheLeaf<T>;
type CacheMap<T> = Map<unknown, Cache<T>>;
type CacheLeaf<T> = unknown[] & { value: T };

// Returned by find() when not in cache
const MISSING: unique symbol = Symbol('MISSING');
type MISSING = typeof MISSING;

export default class StateCache<T> {

  private cache: Cache<T> = new Map();

  // Returned by find() when not in cache
  readonly MISSING: MISSING = MISSING;

  find(key: unknown[]): T | MISSING {
    const { length } = key; // assumed same for all keys
    let cache = this.cache;
    for (let i = 0; i < length; i++) {
      if (cache instanceof Map) {
        // Multiple entries here
        if (!cache.has(key[i])) return MISSING;
        cache = cache.get(key[i])!;
      } else if (Array.isArray(cache)) {
        // Single entry here. Check that remaining key matches.
        for (let j = i; j < length; j++) {
          if (cache[j] !== key[j]) return MISSING;
        }
        return cache.value;
      } else {
        throw new Error("Unexpected cache type");
      }
    }
    // Fully traversed; assume we've arrived at a result
    return cache as T;
  }

  set(key: unknown[], value: T): void {
    const { length } = key; // assumed same for all keys
    let cache = this.cache, parent: CacheMap<T>;
    for (let i = 0; i < length; i++) {
      if (cache instanceof Map) {
        // Multiple entries here
        if (!cache.has(key[i])) {
          // But no match here, so create a shallow cache
          cache.set(key[i],
            i == length - 1
            ? value  // just store object at end
            : Object.assign(key, { value })  // CacheLeaf
          );
          return;
        } else {
          parent = cache;
          cache = cache.get(key[i])!;
        }
      } else if (Array.isArray(cache)) {
        // Single entry here. Check how much of remaining key matches.
        // First check for full match => just update value
        for (let j = i; j < length; j++) {
          if (cache[j] !== key[j]) break;
          if (j === length - 1) { // full match
            cache.value = value;
            return;
          }
        }
        // Not full match
        for (let j = i; j < length; j++) {
          // Deepen with one more Map
          const map = new Map;
          map.set(cache[j], cache);
          parent!.set(key[j-1], map);
          if (cache[j] === key[j]) {
            // Match, so continue deepening
            parent = map;
          } else {
            // Mismatch, so we are done
            map.set(key[j],
              j == length - 1
              ? value  // just store object at end
              : Object.assign(key, { value })  // CacheLeaf
            );
            return;
          }
        }
        throw new Error("Unexpected cache termination 1");
      } else {
        throw new Error("Unexpected cache type");
      }
    }
    // Full match: just update value from final map
    parent!.set(key[length-1], value);
  }
}
```

Note: This code doesn't pass a couple of JSX tests, so it probably has a bug around the last level of recursion. Didn't seem worth fixing unless it actually performed better.

</details>

PR #648, opened 2023-08-24
Fix #599

### LINKED ISSUE #599: Variable declaration in inline block doesn't compile
this doesn't compile
```js
function f() { const x = 1 }
```
also those examples don't compile
```js
function f() {
  const x = 1 }

function f() { const x = 1
}
```
same with
```js
function f() { a; b }
function f() { ;b }
```
those examples do compile:
```js
function f() { a }
function f() { a
}
function f() {
  a }
```
but the last one compiles incorrectly to
```js
function f() {
  return;
  a;
}
```

also related, this doesn't work
```js
function f() const a = 1
try const a = 1
```

STRd6:
- Added assertion for upcoming closing brace to end statements.
- Made `BracedContent` into `SingleLineStatements`.
- Allowed for Declarations in `SingleLineStatements`
- Allowed for Declarations in `PostfixedSingleLineStatements`

Things to look into in the future:

- Declarations in `then` will need to check if they should add braces.
- `BracedContent` may actually need to be something closer to `( SingleLineStatements EOS )+`

### DIFF
--- a/source/parser.hera
+++ b/source/parser.hera
@@ -475,11 +475,14 @@ FatArrow
       return " =>"
     return [ $1, "=>" ]
 
+TrailingDeclaration
+  ( _* ( ConstAssignment / LetAssignment ))
+
 # NOTE Different from
 # https://262.ecma-international.org/#prod-ConciseBody
 FatArrowBody
   # If same-line single expression, avoid wrapping in braces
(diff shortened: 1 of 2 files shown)


edemaine (commented):
Sorry for the delay in reviews (I've been traveling this week). This looks like some nice grammar improvements!

PR #649, opened 2023-08-24
Initial auto-const

### LINKED ISSUE #527: add autoConst
Just a feature request as by default I like to use const if I don't need to mutate some variable. So having autoConst would be usefull.

STRd6:
This is about 80% of auto-const by modifying the existing auto-let.

When the assignment appears in an expression it may be very difficult to actually enforce it as const without possibly heavily rewriting the parse tree.

There are still some simple cases that we could improve for both auto-let and auto-const.

This will work great for common situations.

Fixes #527

### DIFF
--- a/source/lib.ts
+++ b/source/lib.ts
@@ -2875,7 +2875,9 @@ function processProgram(root: BlockStatement, config, m, ReservedWord) {
   statements.unshift(...m.prelude)
 
   if (config.autoLet) {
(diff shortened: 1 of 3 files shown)


edemaine (commented):
Nice that this can share the `let` code (until we have rewriting to enforce `const`, which I agree is second priority).

PR #651, opened 2023-08-26
Fix #639 hoistable thick pipe ref decs

### LINKED ISSUE #639: Implicit return is interacting poorly with thick arrow ref declaration
```
->
 x + 1 ||> a.add

// ↓↓↓ Civet 0.6.22

(function() {
 return let ref;a.add((ref = x + 1)),ref
})
```

STRd6:


edemaine (commented):
Nice fix. Ordering is hard!

STRd6:
I was just happy that we don't have too many ordering dependencies between the parser adjustment passes yet 😅

PR #657, opened 2023-08-28
🐈🐈🐈

### LINKED ISSUE #641: Config.civet reserved
I haven't found any mention of a `config.civet` either on the website or on the GitHub page, but I guess `config.civet` is a reserved filename?

I happened to have `config.civet` file in my working dir with my own stuff in it and every time I wanted to compile, an empty output file was created and the process seemingly never completed. Took me ages to figure out what's going on.

The issue occurs even if nothing imports the file. Renaming the file solves the issue. Naming it back resurfaces it.

I assume [https://civet.dev/config](https://civet.dev/config) is supposed to detail what goes inside.
Even before it does, let's mention this in the readme, so peeps don't fall into the same hole I did.

---

Additionally, can we maybe think about not calling it `config.civet`? To me that seems too much like a project file, rather than a sort of meta config file. I assume it's a _.civet_ file, cause it can have some dynamic fields or something? If not, can't we just have a `json` file instead?

----

After some more testing, I found it that it probably does get over it at some point (maybe), because the time it takes seems to depend on the length of the file. But my original config file of 99 lines did not finish in _13 minutes_ (after which I stopped the timer).

STRd6:
I scope creeped this to now allow for `.civet` in our Civet build.

Key ideas:

- No checked in generated files to clutter up diffs
- Allow us to use `.civet` as our source files when building and testing
- Not substantially slower or requiring extra bootstrapping or downloads
- It's how TS does it: a devDependency on a previously published stable version.

Also converted `lib.ts` -> `lib.civet` it uncovered several bugs but only needed minor changes. It is however a little slower (~5s) to parse and that is sad. Hopefully we can speed up the parser at it will be no worse than TS.

Fixes #641 Config update:
- Removed `config.civet` from name of config files to check
- If not found in `.config` dir keep looking rather than return `null`
- Added some basic tests
- Simpler dynamic `import`


edemaine (commented):
I still think having `.civet` as a format would be useful, partly because we don't (yet) support YAML, and Civet is a good approximation thereof. I agree it can only use the default config, but I think that's fine when all you're writing is

```coffee
export default
  coffeeCompat: true
```

but I'd still rather write that than

```json
{"coffeeCompat": true}
```

I'm hopeful that `civetconfig.civet` is a sufficiently specific name that it won't conflict with user's filenames... Probably `🐈.civet` too. But we obviously do need to remove `config.civet`.

build/coffee-esm.mjs
```diff
@@ -3,9 +3,7 @@
 import { readFile } from 'fs/promises';
 import { pathToFileURL, fileURLToPath } from 'url';
 
-import CoffeeScript, { compile } from "coffeescript";
-// Handle cjs .coffee files
-CoffeeScript.register()
```
edemaine, line 8: I think it's nice to keep this registration in case others want to use this as a CoffeeScript ESM handler. So I lean toward keeping it, but it's also OK to remove it and leave this for a CoffeeScript issue (actually https://github.com/jashkenas/coffeescript/pull/5447).

source/config.mts
```diff
@@ -0,0 +1,72 @@
+import path from "path";
```
edemaine, line 1: Any particular reason for the move from CoffeeScript? ~~I preferred that version, and it's more consistent with the rest of the code (e.g. `main.coffee`).~~ Ah, I see, but we want to move toward full typechecking, so this makes sense.

source/config.mts
```diff
@@ -0,0 +1,72 @@
...
+      }
+    }
+    if (entry.isFile()) {
+      const name = entry.name.replace(/^\./, ""); // allow both .civetconfig.civet and civetconfig.civet
+      if ([
+        "🐈.json",
+        "🐈.civet",
+        "civetconfig.json",
+        "civetconfig.civet",
+      ].includes(name)) {
```
edemaine, line 27: `name in ['config.civet', 'civet.json', 'civetconfig.civet', 'civetconfig.json']` has a more efficient compilation in CoffeeScript than this. But I guess we could do even better with a `Set` (hoisted to global scope).

edemaine:
Wait, how can `main` be `.civet`? Isn't it needed to compile with the esbuild plugin?

STRd6:
> Wait, how can `main` be `.civet`? Isn't it needed to compile with the esbuild plugin?

I added a devDependecy for a previously published civet version so we can us that in development as long as we reference it from `node_modules`.

source/config.civet
```diff
@@ -0,0 +1,72 @@
+import path from "path";
```
edemaine, line 1: Not a high priority, but now that we're using Civet instead of TS, we could make this code closer to the CoffeeScript version you're replacing here (but with type annotations). Could be saved for the future though.

edemaine (approved):
Oh right, I see it now. Amazing — we can now live the Civet future!!

Is it possible to switch `lib.ts` over too? Or is there an issue when mixing with Hera? That is where we could probably benefit most from Civet.

edemaine (approved):
So exciting!!! Civetception!

package.json
```diff
@@ -85,9 +86,8 @@
     ],
     "loader": [
       "ts-node/esm",
-      "./build/coffee-esm.mjs",
       "./build/hera-esm.mjs",
-      "./dist/esm.mjs"
+      "./node_modules/@danielx/civet/dist/esm.mjs"
```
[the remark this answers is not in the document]
STRd6: This is the version that will parse the source and test files. Once parsed the tests are run against `main.civet`, `lib.civet`, `parser.hera`, etc.

source/util.civet
```diff
@@ -225,7 +225,7 @@ export base64Encode = (src) ->
 
 # Accelerate VLQ decoding with a lookup table
 vlqTable = new Uint8Array(128)
-vlqChars = 'ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789+/'
+vlqChars = 'ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789+/';
 do ->
```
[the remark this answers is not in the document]
STRd6: Yeah that directive was from some earlier testing.

We should probably export these utils from Hera as the definitive source eventually.

package.json
```diff
@@ -43,15 +42,14 @@
     "@cspotcode/source-map-support": "^0.8.1"
   },
   "devDependencies": {
+    "@danielx/civet": "^0.6.26",
```
edemaine, line 45: I'd feel safer if this were a pinned dep (no `^`). Given how we tend to change things and just bump the patch number. 😅

package.json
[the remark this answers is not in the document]
edemaine: Ah, got it. (If only we could add a comment to a JSON file. 😉)

That seems fine, because we don't generally use fancy features in the test infrastructure, only inside the triple quotes.  Actually, in some ways, it will be an improvement; I've run into issues where I break the Civet compiler (and ran `yarn build`) and then `yarn test` crashes, with unhelpful error messages.  This should avoid that sort of issue.

package.json
STRd6: I believe the lockfile should take care of it but being explicit here would also be fine.

edemaine:
By the way, it's also really great what a stress test this was, for both JavaScript and CoffeeScript compatibility. Thanks for writing down all the bugs you encountered. Time for another bug squashing session! 😄

STRd6:
I'm still sad about the build getting slower with the conversion from `lib.ts` -> `lib.civet` but its not so bad as to hold this up.

We can try a few things in the future:

- Optimization of the generated parser/grammar at the Hera level
- Better optimized caching
- Optimizations in the civet grammar / handlers
- Convert to more idiomatic Civet in `lib.civet` (less code, fewer braces)

edemaine:
All good ideas. And I'd rather pay the price of slowness and be able to dogfood. Civet should speed our development too.

One thing we might try is profiling the compilation stage.  I have no idea, for example, whether the dominant cost is running Hera rules/caching, or the `parser.hera` code blocks (if we can distinguish these, maybe somewhat because most blocks now call `lib`), or the AST postprocessing we do. You may be right that it's the Hera parsing cost, but it also wouldn't surprise me if there is still some quadratic behavior in AST processing (like say calling `addParentPointers` at multiple recursive levels). Knowing which is dominant would help focus optimization.

PR #659, opened 2023-08-28
??? -> unknown

STRd6:


edemaine (approved):
Seems fine to me. I could also see just `?`, but maybe that's confusing with optional properties and such.

PR #672, opened 2023-08-31
Added void to improve types and opt out of implicit returns

STRd6:
115_void_civet_cyberpunk_f5b2fb95-1fa9-48de-869b-e459d37fc1d4


source/lib.civet
```diff
@@ -2267,7 +2267,7 @@ function attachPostfixStatementAsExpression(exp, post) {
   }
 }
 
-function getPatternConditions(pattern, ref, conditions) {
+function getPatternConditions(pattern, ref, conditions): void {
```
edemaine, line 2270: I would not have guessed this one from the function name...

edemaine (approved):
enter_the_void

PR #676, opened 2023-08-31
Add nextjs unplugin example

Mokshit06:


integration/unplugin/examples/nextjs/components/button.civet
```diff
@@ -0,0 +1,4 @@
+export function Button()
+  return
```
edemaine, line 2: I'd maybe omit `return`

edemaine (approved):
Nice! Should we link to this from `integration/unplugin/README.md`? Along with Astro?

edemaine:
Oops, CI build (type checking) is failing.

Mokshit06:
I think this type error is occurring uses `React.*` types and imports a few things from `next`: packages which are part of `examples/nextjs/package.json` which we don't install during build step.

edemaine:
Could we just ignore these in `tsconfig`?

PR #677, opened 2023-08-31
New try at indented application

STRd6:
Ended up not being to bad, thanks @edemaine for the previous starting point.


edemaine (approved):
Thanks for reviving this! And some nice improvements to the grammar rules.

There are a few more tests in #398 for arrays of functions; maybe worth copying here?

This is an exciting evolution!  I've wanted this in CoffeeScript so many times...

STRd6:
When I looked at the array tests they were already covered by some existing tests in the meantime.

PR #679, opened 2023-09-01
Fixes #664

### LINKED ISSUE #664: Multi-line comment between access and identifier fails to parse
```civet
x./**/y
```

Actual
```
x./**/y
  ^ ParseError
```

Expected
```
x./**/y
```

STRd6:
This builds on top of the indented application branch because some of the changes overlapped.


edemaine (commented):
I was worried about the AST effects for glob processing, and it turns out that `x./**/{a,b}` doesn't parse. Could leave for another PR, but would be nice to fix there too; maybe we should check for more `Dot` → `AccessStart` changes needed?

PR #680, opened 2023-09-01
fix #655

### LINKED ISSUE #655: `x = {{}}` creates exception in parser: Cannot read properties of undefined (reading 'children')
```civet
x = {{}}
```

This isn't valid but it shouldn't throw an exception either.

STRd6:
Also modified the test command to continue to allow passing of `yarn test --inspect-brk` through to mocha because it is critical to my workflow :)

### DIFF
--- a/test/object.civet
+++ b/test/object.civet
@@ -795,6 +795,22 @@ describe "object", ->
     """
 
     testCase """
+      empty
+      ---
+      {{}}
+      ---
+      ({})
+    """
+
+    testCase """
+      empty
+      ---
+      {{}, {}}
+      ---
+      ({})
+    """
+
+    testCase """
       call
       ---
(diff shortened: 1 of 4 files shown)


test/object.civet
```diff
@@ -794,6 +794,22 @@ describe "object", ->
...
+      ({})
+    """
+
+    testCase """
+      empty
+      ---
+      {{}, {}}
+      ---
+      ({})
+    """
```
edemaine, line 811: Not sure I understand... Is this being treated like `{...{}, ...{}}`? Not really sure why nested object literals should be allowed.

STRd6: I believe it's being treated like `a.{x, y}` but with both parts empty.

PR #685, opened 2023-09-01
Reference

STRd6:
Moved cheatsheet to reference.

I also tried to update vitepress but ran into some trouble. In the updated version they were sticklers about .js and .mjs so that's why I renamed those as well as I may try to update again in the future.


edemaine:
Thanks for changing this. I wonder if there's a way to make the old cheatsheet link redirect (for now), so that old links aren't broken? Until we have a new cheatsheet.

STRd6:
I'll try and add a new actual cheatsheet to this PR tomorrow.

civet.dev/cheatsheet.md
```diff
@@ -5,1638 +5,297 @@ aside: false
...
+people.map .name // people.map($ => $.name)
+numbers.filter & % 2 is 0
+// numbers.filter($ => $ % 2 === 0)
 
-<Playground>
-$: document.title = title
-"$": "dollar"
-{$: "dollar"}
-</Playground>
+// # Conditional declarations
```
edemaine, line 163:
```suggestion
// Conditional declarations
```

civet.dev/cheatsheet.md
```diff
@@ -5,1638 +5,297 @@ aside: false
...
+  update item
 
-### Unbraced Literals
+for key in object
+  log key
 
-When every property has a value, braces can be omitted.
+for own key in object
+  log `my ${key}`
+```
```
edemaine, line 137: I might add the two-binding forms of `for..of/in`. Not sure whether to modify the existing examples or add them separately.

edemaine (approved):
Sorry, I realize I never reviewed the cheatsheet. It looks great!! Just a couple comments.

I think we've reached too much stuff in the navbar. At some smaller screen widths, Getting started line-breaks (but before it's small enough to trigger a navbar menu):

image

Maybe we should drop Config? It doesn't actually say much, though it probably should...

We could also deal with this layout issue later. It's important to get this merged before we need to change any docs.

PR #689, opened 2023-09-02
Add type-checking to unplugin

Mokshit06:
Adds type-checking to the unplugin. This currently runs only when `dts` is enabled, though maybe this should be changed.
One of the issues with the current implementation though, which i'm not sure can be fixed either, is that the errors point to the location in the compiled ts/js source rather than the original civet source. Ideally tsc should be mapping them using the sourcemaps, but it doesn't do that here.


STRd6:
TypeScript doesn't support incoming source maps and probably never will. We can do the same thing we do in the LSP and remap the diagnostics by using our own source maps like this: https://github.com/DanielXMoore/Civet/blob/main/lsp/source/lib/util.mts#L425

Mokshit06:
@STRd6 How can I get source map lines from compiled civet source?

STRd6:
You can pass `sourceMap: true` to `Civet.compile`:

https://github.com/DanielXMoore/Civet/blob/main/lsp/source/lib/typescript-service.mts#L482
https://github.com/DanielXMoore/Civet/blob/main/source/main.civet#L106

Then it returns a `{code, sourceMap}` object instead of just the string of code.

edemaine:
There's also some utility code for sourcemap mapping. I think in util.civet?

Mokshit06:
Thanks, yeah this should be working now :)

edemaine:
Could you add some docs to the unplugin README about how this works?

Mokshit06:
Sure, will do that

integration/unplugin/src/index.ts
```diff
@@ -1,5 +1,17 @@
...
+  TransformResult,
+  createUnplugin,
+  SourceMapCompact as UnpluginSourceMap,
+} from 'unplugin';
+import civet, { SourceMap } from '@danielx/civet';
+import {
+  remapRange,
+  flattenDiagnosticMessageText,
+  rangeFromTextSpan,
+  // @ts-ignore
```
STRd6, line 11: Will we need to publish this first so we can use it in the build step?

edemaine: I guess this is why the change above switched the build step from the unplugin esbuild to the OG esbuild plugin.

integration/unplugin/src/index.ts
```diff
@@ -104,7 +144,12 @@ const civetUnplugin = createUnplugin((options: PluginOptions = {}) => {
           const sourceFile = program.getSourceFile(file)!;
           program.emit(
             sourceFile,
-            (filePath, content) => {
+            async (filePath, content) => {
+              const dir = path.dirname(filePath);
+              if (!pathExists(dir)) {
+                await fs.promises.mkdir(dir, { recursive: true });
+              }
```
edemaine: > Calling `fsPromises.mkdir()` when path is a directory that exists results in a rejection only when `recursive` is false.

So I don't think you need to check `pathExists`.

integration/unplugin/src/index.ts
Mokshit06: `@danielx/civet` is marked as external for the plugin, so we the build step should work without needing to publish, but yeah to remove `@ts-ignore` we will need to publish this first

integration/unplugin/src/index.ts
Mokshit06: Oh interesting, yep we can remove it then

edemaine:
Another minor thing: I think `typecheck` should not be camel-cased. It's more common this way, and I think it can be considered a single word.

integration/unplugin/src/index.ts
STRd6: 👍

integration/unplugin/README.md
```diff
@@ -75,14 +75,16 @@ interface PluginOptions {
...
+  typecheck?: boolean;
   transformOutput?: (
     code: string,
     id: string
   ) => TransformResult | Promise<TransformResult>;
 }
 ```
 
-- `dts`: `unplugin-civet` also supports generating `.d.ts` type definition files from the civet source, which is useful for building libraries. Default: `false`.
+- `dts`: This also supports generating `.d.ts` type definition files from the civet source, which is useful for building libraries. Default: `false`
```
edemaine, line 86:
```suggestion
- `dts`: Whether to generate `.d.ts` type definition files from the Civet source, which is useful for building libraries. Default: `false`
```

integration/unplugin/src/index.ts
```diff
@@ -34,21 +48,22 @@ const isCivet = (id: string) => /\.civet$/.test(id);
 const isCivetTranspiled = (id: string) => /\.civet\.(m?)(j|t)s(x?)$/.test(id);
 
 const civetUnplugin = createUnplugin((options: PluginOptions = {}) => {
-  if (options.dts && options.js) {
+  if ((options.dts || options.typecheck) && options.js) {
     throw new Error("Can't have both `dts` and `js` be set to `true`.");
```
edemaine, line 52: Update the error message to mention `typecheck`.

lsp/source/lib/util.mts
```diff
@@ -1,31 +1,31 @@
...
-  TextSpan,
-} from "typescript"
-const {
-  DiagnosticCategory,
-  ScriptElementKind,
-  ScriptElementKindModifier,
-} = ts
-import vs, {
+import ts from 'typescript';
+import type { NavigationBarItem, NavigationTree } from 'typescript';
```
edemaine, line 2: I think we have a general preference for omitting `;`s when we can.

package.json
```diff
@@ -35,6 +35,10 @@
       "require": "./dist/esbuild.js",
       "import": "./dist/esbuild.mjs"
     },
+    "./ts-diagnostic": {
+      "default": "./dist/ts-diagnostic.js",
```
edemaine, line 39: `ts-diagnostic` uses `import`, so this doesn't work in CJS repos (e.g. I tried in one of the unplugin examples). We need to build an `.mjs` version and link to here.

edemaine (commented):
This is looking good to me. I got stuck when trying to test it though, because CJS vs. ESM.

STRd6 (approved):

edemaine:
This seems like it should be done, but I'm still failing to run tests. Here's what I tried:

## integration/unplugin/examples/vite

```patch
diff --git a/integration/unplugin/examples/vite/src/main.civet b/integration/unplugin/examples/vite/src/main.civet
--- a/integration/unplugin/examples/vite/src/main.civet
+++ b/integration/unplugin/examples/vite/src/main.civet
@@ -1,3 +1,6 @@
 {a} from "./module.civet"

-console.log a
+const b = 5
+console.log b()
+
+console.log a + 5
diff --git a/integration/unplugin/examples/vite/vite.config.js b/integration/unplugin/examples/vite/vite.config.js
index 7d9ffb5..bc51b4a 100644
--- a/integration/unplugin/examples/vite/vite.config.js
+++ b/integration/unplugin/examples/vite/vite.config.js
@@ -2,5 +2,5 @@ import civetVitePlugin from '@danielx/civet/vite';
 import { defineConfig } from 'vite';

 export default defineConfig({
-  plugins: [civetVitePlugin({})],
+  plugins: [civetVitePlugin({typecheck: true})],
 });
```

Then I run

```sh
npm install
npm link @danielx/civet
npm run build
```

This builds, but doesn't generate any diagnostics. Should it?

## integration/unplugin/examples/astro

Just running:

```sh
npm install
npm link @danielx/civet
npm run build
```

crashes with error:

```

> astro-example@0.0.1 build
> astro build

2:46:58 PM [vite] Error when evaluating SSR module /@fs/C:/Users/edemaine/Projects/Civet/dist/unplugin-shared.mjs: failed to import "@danielx/civet/ts-diagnostic"
|- Error: Cannot find module '@danielx/civet/ts-diagnostic' imported from 'C:/Users/edemaine/Projects/Civet/dist/unplugin-shared.mjs'
    at nodeImport (file:///C:/Users/edemaine/Projects/Civet/integration/unplugin/examples/astro/node_modules/vite/dist/node/chunks/dep-df561101.js:56010:25)
    at ssrImport (file:///C:/Users/edemaine/Projects/Civet/integration/unplugin/examples/astro/node_modules/vite/dist/node/chunks/dep-df561101.js:55912:30)
    at eval (C:/Users/edemaine/Projects/Civet/dist/unplugin-shared.mjs:5:37)
    at async instantiateModule (file:///C:/Users/edemaine/Projects/Civet/integration/unplugin/examples/astro/node_modules/vite/dist/node/chunks/dep-df561101.js:55974:9)

2:46:58 PM [vite] Error when evaluating SSR module /@fs/C:/Users/edemaine/Projects/Civet/dist/vite.mjs: failed to import "/@fs/C:/Users/edemaine/Projects/Civet/dist/unplugin-shared.mjs"
|- Error: Cannot find module '@danielx/civet/ts-diagnostic' imported from 'C:/Users/edemaine/Projects/Civet/dist/unplugin-shared.mjs'
    at nodeImport (file:///C:/Users/edemaine/Projects/Civet/integration/unplugin/examples/astro/node_modules/vite/dist/node/chunks/dep-df561101.js:56010:25)
    at ssrImport (file:///C:/Users/edemaine/Projects/Civet/integration/unplugin/examples/astro/node_modules/vite/dist/node/chunks/dep-df561101.js:55912:30)
    at eval (C:/Users/edemaine/Projects/Civet/dist/unplugin-shared.mjs:5:37)
    at async instantiateModule (file:///C:/Users/edemaine/Projects/Civet/integration/unplugin/examples/astro/node_modules/vite/dist/node/chunks/dep-df561101.js:55974:9)

2:46:58 PM [vite] Error when evaluating SSR module C:\Users\edemaine\Projects\Civet\integration\unplugin\examples\astro\astro.config.mjs: failed to import "/@fs/C:/Users/edemaine/Projects/Civet/dist/vite.mjs"
|- Error: Cannot find module '@danielx/civet/ts-diagnostic' imported from 'C:/Users/edemaine/Projects/Civet/dist/unplugin-shared.mjs'
    at nodeImport (file:///C:/Users/edemaine/Projects/Civet/integration/unplugin/examples/astro/node_modules/vite/dist/node/chunks/dep-df561101.js:56010:25)
    at ssrImport (file:///C:/Users/edemaine/Projects/Civet/integration/unplugin/examples/astro/node_modules/vite/dist/node/chunks/dep-df561101.js:55912:30)
    at eval (C:/Users/edemaine/Projects/Civet/dist/unplugin-shared.mjs:5:37)
    at async instantiateModule (file:///C:/Users/edemaine/Projects/Civet/integration/unplugin/examples/astro/node_modules/vite/dist/node/chunks/dep-df561101.js:55974:9)

[astro] Unable to load your Astro config

 error   Cannot find module '@danielx/civet/ts-diagnostic' imported from 'C:/Users/edemaine/Projects/Civet/dist/unplugin-shared.mjs'
  File:
    C:\Users\edemaine\Projects\Civet\integration\unplugin\examples\astro\node_modules\vite\dist\node\chunks\dep-df561101.js:56010:25
  Code:
    56009 |         if (!resolved) {
    > 56010 |             const err = new Error(`Cannot find module '${id}' imported from '${importer}'`);
            |                         ^
      56011 |             err.code = 'ERR_MODULE_NOT_FOUND';
      56012 |             throw err;
      56013 |         }
  Stacktrace:
Error: Cannot find module '@danielx/civet/ts-diagnostic' imported from 'C:/Users/edemaine/Projects/Civet/dist/unplugin-shared.mjs'
    at nodeImport (file:///C:/Users/edemaine/Projects/Civet/integration/unplugin/examples/astro/node_modules/vite/dist/node/chunks/dep-df561101.js:56010:25)
    at ssrImport (file:///C:/Users/edemaine/Projects/Civet/integration/unplugin/examples/astro/node_modules/vite/dist/node/chunks/dep-df561101.js:55912:30)
    at eval (C:/Users/edemaine/Projects/Civet/dist/unplugin-shared.mjs:5:37)
    at async instantiateModule (file:///C:/Users/edemaine/Projects/Civet/integration/unplugin/examples/astro/node_modules/vite/dist/node/chunks/dep-df561101.js:55974:9)
```

I'm not sure why though; the export looks properly defined. (And I did build Civet, and I checked the link; in particular, `node_modules/@danielx/civet/dist/ts-diagnostic.{js,mjs}` exist.) Maybe Astro doesn't always respect `exports`?

Mokshit06:
Huh, so my first guess was that it might have something to do with `exports` not being supported with self-referencing in node (referencing `@danielx/civet/ts-diagnostic` inside unplugin) but that is indeed supported https://nodejs.org/api/packages.html#self-referencing-a-package-using-its-name. It looks like this error is occurring inside vite, so might be something specific to it. I'll have to dig deeper

edemaine:
Do you have a setting where type checking successfully reports errors? (Should I try another example?) I'd be happy to merge this and leave more situations for the future — I'd just like to see it report type errors in at least one setting. 😸

Mokshit06:
Vite should be working now 😃 When adding files to `fsMap` for typescript to walk over, I missed adding the case for `options.typecheck`

edemaine:
In the Vite example I'm now getting this error:

```
[unplugin-civet] Cannot read properties of undefined (reading 'length')
error during build:
TypeError: Cannot read properties of undefined (reading 'length')
    at createSourceFile2 (C:\Users\edemaine\Projects\Civet\node_modules\typescript\lib\typescript.js:29215:58)
    at parseSourceFileWorker (C:\Users\edemaine\Projects\Civet\node_modules\typescript\lib\typescript.js:29089:30)
    at Object.parseSourceFile (C:\Users\edemaine\Projects\Civet\node_modules\typescript\lib\typescript.js:28918:26)
    at Object.createSourceFile (C:\Users\edemaine\Projects\Civet\node_modules\typescript\lib\typescript.js:28103:23)
    at Object.getSourceFile (C:\Users\edemaine\Projects\Civet\node_modules\@typescript\vfs\dist\vfs.cjs.production.min.js:1:4253)
    at findSourceFileWorker (C:\Users\edemaine\Projects\Civet\node_modules\typescript\lib\typescript.js:120079:25)
    at findSourceFile (C:\Users\edemaine\Projects\Civet\node_modules\typescript\lib\typescript.js:119999:22)
    at C:\Users\edemaine\Projects\Civet\node_modules\typescript\lib\typescript.js:119948:24
    at getSourceFileFromReferenceWorker (C:\Users\edemaine\Projects\Civet\node_modules\typescript\lib\typescript.js:119917:28)
    at processSourceFile (C:\Users\edemaine\Projects\Civet\node_modules\typescript\lib\typescript.js:119946:7)
```

This seems like progress but also not correctly reporting an error. :-)

Mokshit06:
Weird, this seems to working for me
```

> build
> npx vite build

vite v4.4.9 building for production...
transforming (1) index.htmlsrc/main.civet.tsx:4:13 - error TS2349: This expression is not callable.
  Type 'Number' has no call signatures.

4 console.log(b())
              ~

✓ 4 modules transformed.
dist/index.html                0.38 kB │ gzip: 0.27 kB
dist/assets/index-733fc199.js  0.77 kB │ gzip: 0.43 kB
✓ built in 319ms
```

edemaine:
Good news: Vite indeed works for me on Linux.  So the error I get (still as above) is Windows-specific.

Annoyingly, it does not give a source line number, but with some instrumentation, I find that it's using `dist/vite.js` and it's failing during this call:
```js
const program = import_typescript.default.createProgram({
          rootNames: [...fsMap.keys()],
          options: compilerOptions,
          host: host.compilerHost
        });
```
I checked `fsMap.keys()` and it seems reasonable, though the paths are in Windows format. (`'C:\\Users\\edemaine\\Projects\\Civet\\integration\\unplugin\\examples\\vite\\src\\main.civet.tsx', 'C:\\Users\\edemaine\\Projects\\Civet\\integration\\unplugin\\examples\\vite\\src\\module.civet.tsx'`) Whereas on Linux, they are Unix paths (`'/.../edemaine/Projects/Civet/integration/unplugin/examples/vite/src/main.civet.tsx', '/.../edemaine/Projects/Civet/integration/unplugin/examples/vite/src/module.civet.tsx'`).

Ah yes, the problem is that someone is normalizing filenames to use `/` instead of `\`. When `processSourceFile` gets called (by Vite I believe), it gives the key of `C:/Users/edemaine/Projects/Civet/integration/unplugin/examples/vite/src/main.civet.tsx` whereas we stored the version with `\`s.

I think this is the issue we're running into: https://vitejs.dev/guide/api-plugin.html#path-normalization
It sounds like we might need to call `normalizePath` helpers specifically for Vite and Rollup, and maybe other build systems that do similar path normalization. It doesn't seem like unplugin offers a helper that abstracts this, though perhaps it should...

I think one fix could be this: when we call `fsMap.set(path.resolve(process.cwd(), id), code);` also call `fsMap.set(samePath.replace(/\\/g, '/'), code)`. I think this should only double the set size, which should be minimal (re-using the `code` object) and should be safe in all environments. Thoughts?

edemaine:
I added my proposed fix and this now works for me on both Windows and Linux. So I think we're done!

Mokshit06:
Great find! I remember this is something I had faced issues with when working on [Macaron](macaron.js.org) as well, didn't realise we were facing the same issue here. I wonder if this also fixes the Astro issue  🤔 though it seems unrelated to the path formatting

edemaine:
Amazingly, it did fix Astro too! Yay!!

By the way, do you think we should add `typecheck: true` to all the examples?

Mokshit06:
That's funny, Astro's error messages can be pretty weird sometimes I've seen. I was testing the plugin out again, and it looks like we need to move the `fsMap.set` part to be inside `load` itself as rollup expects the return value of `transform` to be valid JS.

> we should add typecheck: true to all the examples

Hmm I don't think so, I like build tools to not do type checking in CLI during dev builds as it slows it down by alot. Maybe we can do it in the examples based on `process.env.TS_CHECK`?

edemaine:
I'm not getting any errors with the Rollup example using `typecheck: true`. If I use `dts: true`, I do get errors:

```js
main.civet → dist...
(!) The emitted file "main.civet.d.ts" overwrites a previously emitted file of the same name.
(!) The emitted file "utils\module.civet.d.ts" overwrites a previously emitted file of the same name.
created dist in 361ms
```



> That's funny, Astro's error messages can be pretty weird sometimes I've seen. I was testing the plugin out again, and it looks like we need to move the `fsMap.set` part to be inside `load` itself as rollup expects the return value of `transform` to be valid JS.
> 
> > we should add typecheck: true to all the examples
> 
> I like build tools to not do type checking in CLI during dev builds as it slows it down by alot. Maybe we can do it in the examples based on `process.env.TS_CHECK`?

It would be natural to have a `test` script to do the type checking. So yeah, maybe the script could be `TYPECHECK=1 npx rollup --config rollup.config.js` and then the config file checks for the environment variable `TYPECHECK`.

Mokshit06:
That is because currently the rollup example doesn't have any TS-specific syntax—all of it is inferred. But as soon I add a type declaration to `rollup/main.civet`

```
{a} from "./utils/module.civet"

export b = a

export * from "./utils/module.civet"

export type X = string
```

I get this error

```
$ npx rollup --config rollup.config.js

main.civet → dist...
[!] RollupError: Unexpected token (Note that you need plugins to import files that are not JavaScript)
main.civet.tsx (7:7)
5: export * from "./utils/module.civet"
6: 
7: export type X = string
          ^
8: 
9: //# sourceMappingURL=data:application/json;charset=utf-8;base64,eyJ2ZXJzaW9uIjozLCJmaWxlIjoibWFpbi5jaXZldC50c3gudHN4I...
    at error (/Users/mokshitjain/dev/Civet/node_modules/rollup/dist/shared/rollup.js:353:30)
    at Module.error (/Users/mokshitjain/dev/Civet/node_modules/rollup/dist/shared/rollup.js:15206:16)
    at Module.tryParse (/Users/mokshitjain/dev/Civet/node_modules/rollup/dist/shared/rollup.js:15937:25)
    at Module.setSource (/Users/mokshitjain/dev/Civet/node_modules/rollup/dist/shared/rollup.js:15538:39)
    at ModuleLoader.addModuleSource (/Users/mokshitjain/dev/Civet/node_modules/rollup/dist/shared/rollup.js:25720:20)
```

edemaine:
Oh, right, I recall you've mentioned that issue before: Rollup needs `js: true`. Could you put together another PR when you get a chance?

Somewhat related: Given that we have TypeScript integration in this plugin, I wonder whether it would actually make sense to offer an option of transpiling TS → JS via TypeScript instead of Civet's `js: true` mode, which lacks some features (most recently, `using`). Not sure the best option name — perhaps `typescript: true`? (`ts: true` feels weird given the meaning of `js: true`...?) It could be implied by `typecheck: true` or `dts: true`...

Mokshit06:
That would disable all the other bundler plugins from further processing the code. It could lead to confusing behaviour, so inferring it from `typecheck: true` or `dts: true` might not be suitable. Though maybe we can add it as a separate config option if enough people feel the need for it?

edemaine:
I guess I meant "implied by `typecheck: true` or `dts: true` *when we also have `js: true`*". Alternatively, we could use something like `js: 'tsc'`.

PR #703, opened 2023-09-05
Fix #702: assignment dec in postfix if

### LINKED ISSUE #702: Postfix `if` doesn't add destructure declaration to block
```civet
x if { x } := y
```

Expected

```
let ref;
if ((ref = y)) {
  const { x } = y;
  x;
}
```

Actual
```
let ref;
if ((ref = y)) {
  x;
}
```

STRd6:


source/lib.civet
```diff
@@ -130,21 +130,23 @@ function updateParentPointers(node: ASTNode, parent?: ASTNodeBase, depth = 1): v
...
 }
 
-function addPostfixStatement(statement: StatementTuple, ws: ASTNode, post) {
+function addPostfixStatement(statement: StatementTuple, ws: ASTNode, post)
   let children, expressions
-  if (post.blockPrefix?.length) {
-    let indent = post.blockPrefix[0][0]
-    expressions = [...post.blockPrefix, [indent, statement]]
+
+  prefix := post.blockPrefix or post.condition?.expression.blockPrefix
```
edemaine, line 136: What if they both exist?  This might trigger it:
```js
(if x := f()
  y
else
  z
) if w := g()
```

STRd6: IfExpressions don't currently handler declaration condition at all, we need to statementize them first.

edemaine (approved):

PR #706, opened 2023-09-06
Use Typescript 5

krist7599555:
I think this pull is breaking change so please review it carefully.

- use `typescript@5`: some newly `.ts` file will error when lsp use `typescript@4`. For example type only import `.d.ts` without actual `.ts` file, this technique use by [svelte-kit generated `./$types`](https://kit.svelte.dev/docs/types#generated-types).
- use `esbuild@0.16`: after `esbuild@0.17` api will [break](https://github.com/evanw/esbuild/releases/tag/v0.17.0) on watch mode. so use version that not break.
- print typescript version that civet-lsp use, for clearly say that typescript is come from civet  extension it self.

Before `civet-lsp 0.37`
image

After `civet-lsp 0.38 (this pull)`
image

If you need reproduce for this problem just tell me.
By the way. Love this project 😁🎉

### DIFF
--- a/lsp/build/build.civet
+++ b/lsp/build/build.civet
@@ -1,5 +1,6 @@
 { build } from esbuild
-civetPlugin from ../../dist/esbuild-plugin.js
+civetPlugin_ from ../../dist/esbuild-plugin.js
+civetPlugin := civetPlugin_?.default ?? civetPlugin_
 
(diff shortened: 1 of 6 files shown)


STRd6:
Thanks, I made a few tweaks but this is good!

Publish LSP v0.3.8

PR #709, opened 2023-09-07
Fix #691: unary not with existential

### LINKED ISSUE #691: not vs ? precedence
It seems we do have a precedence issue with `not` (or maybe `?`).

## Input

```coffee
not foo?
```

In CoffeeScript, I often write this to test whether `foo == null`.

## Current compilation
```js
!foo == null
```

## Expected compilation
```js
!(foo == null)
```

STRd6:


edemaine (commented):
Wow, that's even nicer codegen than I expected; I was just expecting `x?` to always get wrapped in parens. Looks like CoffeeScript does this too.

Anyway, I guess this handles unary operators. I just checked and CoffeeScript actually treats `~x?` as `~(x?)`.

Now I'm worried about binary operators though.  `x? + y?` should be treated like `(x != null) + (y != null)` (useful for counting nonnull objects), but Civet apparently treats it as ... `x != null && x + (y != null)`??!

source/lib.civet
```diff
@@ -876,52 +876,55 @@ function isExistence(exp) {
   }
 }
 
+// all relational operators could be chained in theory, including in and instanceof
+relationalOps := ["==", "===", "!=", "!==", "<", "<=", ">", ">=", "in"]
+
+function isRelationalOp(op)
+  relationalOps.includes(op.token) or op.relational
```
edemaine, line 883: I'd prefer to use a `Set` (and `.has`) here. (Maybe also for `lowerPrecedenceOps`)

edemaine (approved):
Wow, glad that was easy to fix too! I was worried the parse tree had things in the wrong order.

PR #710, opened 2023-09-07
Fix #701

### LINKED ISSUE #701: Possible bug in returning a switch expression
It appears that you can mix explicit returns with return value assignment:
```
function foo(a)
   if bar[a] === 'three' then return 3
   return .= 9
   return.value *= a
```
compiles to
```
function foo(a) {
  if (bar[a] === "three") return 3;
  let ret = 9;
  ret *= a;
  return ret;
}
```
which works as expected. And it appears you can return a switch expression:
```
function foo(a)
   if bar[a] === 'three' then return 3
   if a < 0 then return switch bar[-a]
      'three' 3
   9
```
compiles to
```
function foo(a) {
  if (bar[a] === "three") return 3;
  if (a < 0)
    return (() => {
      let m;
      if (((m = bar[-a]), m === "three")) {
        return 3;
      }
    })();
  return 9;
}
```
which is also OK. But if you mix all these things, e.g. in
```
function foo(a)
   if bar[a] === 'three' then return 3
   if a < 0 then return switch bar[-a]
      'three' 3
   return .= 9
   return.value *= a
```
you then get from compilation the TypeScript code
```
function foo(a) {
  if (bar[a] === "three") return 3;
  if (a < 0)
    return (() => {
      let m;
      if (((m = bar[-a]), m === "three")) {
        return ret;
      }
    })();
  let ret = 9;
  ret *= a;
  return ret;
}
```
which fails to compile under tsc with `TS2454: Variable 'ret' is used before being assigned.` and fails to run in the Playground with `ReferenceError: can't access lexical declaration 'ret' before initialization`. Seems like a bug -- by analogy with the case without the return value assignment, it seems like the first occurrence of `return ret;` in this last compiled version should be `return 3;`. Unless I am trying to do something I shouldn't be able to?

STRd6:


source/lib.civet
```diff
@@ -3440,35 +3404,55 @@ function typeOfJSXFragment(node, config, getRef) {
...
+  updateParentPointers(block, fn)
+
+  exp :=
+    type: "CallExpression"
+    children: [ makeLeftHandSideExpression(fn), "()" ]
+
+  if prefix
+    return [ makeLeftHandSideExpression [prefix, exp] ]
+
+  return [ exp ]
```
edemaine, line 3455: Do we still need the array wrappers? I'm guessing we can change the return type to `ASTNode` (but haven't tried).

edemaine (approved):
Wow, a great set of improvements.

source/lib.civet
STRd6: It could probably be `ASTNode` or `StatementTuple[]` as long as it is the same type that gets passed to `insertReturn`. I'm leaning towards keeping more things as `StatementTuple[]` in preparation for statementizing more things soon.

STRd6: Oh, for this specific place we could either use the array wrapper up update all the use locations to wrap in a `children` array.

PR #712, opened 2023-09-08
Updating civet dep to reap some perf improvement

STRd6:


build/esbuild.civet
```diff
@@ -3,7 +3,7 @@
 esbuild = require "esbuild"
 heraPlugin = require "@danielx/hera/esbuild-plugin"
 # Need to use the packaged version because we may not have built our own yet
-civetPlugin = require "../node_modules/@danielx/civet/dist/esbuild-plugin.js"
+{ default: civetPlugin } = require "../node_modules/@danielx/civet/dist/esbuild-plugin.js"
```
edemaine, line 6: Why did this change?

edemaine (approved):

build/esbuild.civet
STRd6: I think it's from this change: https://github.com/DanielXMoore/Civet/commit/4665c87ca2d61cf9998d1c423bbc017d4d8d19b1#diff-6283862284e4ac2122a7c631d77050a1a6d2fa9a9952e2f5d2062eb3afd971d4

We may want to switch back for better `require` usability.

STRd6: Though it might also instead be from https://github.com/DanielXMoore/Civet/commit/4665c87ca2d61cf9998d1c423bbc017d4d8d19b1#diff-9d7f01123a2ec84b6da89e066949c22128f2762fefee232c2b1357743f2e1102R53

Where we are now running it through esbuild instead of building the cjs manually.

PR #713, opened 2023-09-08
Allow more ergonomic require of cjs esbuild plugin

STRd6:
This lets us use `civetPlugin = require "@danielx/civet/esbuild-plugin"` instead of needing to use `.default`.

We can pick it up in `esbuild.civet` after publishing.

### DIFF
--- a/source/esbuild-plugin.civet
+++ b/source/esbuild-plugin.civet
@@ -135,5 +135,6 @@ civet.setup = defaultPlugin.setup
 if typeof module !== 'undefined'
   // So we can `civetPlugin = require "@danielx/civet/esbuild-plugin"`
   module.exports = civet
+  module.exports.default = civet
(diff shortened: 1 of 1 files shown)


edemaine:
According to the tests, this seems to break `import esbuildPlugin from "../../source/esbuild-plugin.civet"`... ☹️

edemaine:
I read https://github.com/evanw/esbuild/issues/1719 which talks about the fun history here, and followed some of the back links. It seems like https://github.com/conventional-changelog/commitlint/pull/3614 might work for our context, though I don't quite understand why. 🙂

STRd6:
This weird hybrid export hack works but it does create a warning from esbuild.

The only reason for the `export default` is so ts-node can do an esm import for test coverage purposes.

🤷

PR #716, opened 2023-09-09
Fix #629: Unary op with late assignment

### LINKED ISSUE #629: Late assignment for unary prefix `!`
```
!x = 3
```

This matches CoffeeScript: https://coffeescript.org/#try:!x%20%3D%203

STRd6:
The previous code generated from `test/assignment.civet` wasn't valid JS so this seems somewhat better.

### DIFF
--- a/source/parser.hera
+++ b/source/parser.hera
@@ -341,7 +341,7 @@ UnaryExpression
   # NOTE: Merged AwaitExpression with UnaryOp
   # https://262.ecma-international.org/#prod-AwaitExpression
   # NOTE: Eliminated left recursion
-  UnaryOp*:pre ( UpdateExpression / NestedNonAssignmentExtendedExpression ):exp UnaryPostfix?:post ->
+  UnaryOp*:pre UnaryBody:exp UnaryPostfix?:post ->
     return processUnaryExpression(pre, exp, post)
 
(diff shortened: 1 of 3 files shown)


source/parser.hera
```diff
@@ -350,14 +350,29 @@ UnaryExpression
...
+
 UnaryPostfix
   QuestionMark
   ( __ ( As / Satisfies ) Type )+ -> { ts: true, children: $0 }
 
 # https://262.ecma-international.org/#prod-UpdateExpression
 UpdateExpression
   # NOTE: Not allowing whitespace betwen prefix and postfix increment operators and operand
-  UpdateExpressionSymbol UnaryExpression ->
+  UpdateExpressionSymbol UnaryWithoutParenthesizedAssignment ->
```
edemaine, line 375: Hmm, why does `++(x=5)` still work? Does it parse via another path?

edemaine (approved):

source/parser.hera
STRd6: The `ParenthesizedAssignment` is for wrapping `x = 5` in implicit parentheses. An explicitly parenthesized assignment will be a `ParenthesizedExpression` containing an `AssignmentExpression`.

PR #717, opened 2023-09-09
Unify pattern matching and declaration conditions

STRd6:
- [x] Refactor declaration conditions
- [x] Apply pattern matching conditions to declaration conditions

There's some possible follow on discussion/work to do.

Should

```civet
if {x} := y
  ...
```

also add a condition check for x being non-falsey?

What about:

```
if {x, y} := z
  ...
```

Should x and y be non-falsey?

### DIFF
--- a/source/lib.civet
+++ b/source/lib.civet
@@ -27,10 +27,23 @@ type ASTLeaf =
   $loc: Loc
   token: string
 
+type Children = ASTNode[]
+
 type StatementDelimiter = ASTNode
 type IndentNode = ASTNode
 
 type StatementTuple = [IndentNode, ASTNode, StatementDelimiter?]
+
+type IfStatement =
+  type: "IfStatement"
+  children: Children
+  condition: IfCondition
+  then: BlockStatement
+  else: [ ASTNode, ElseToken, BlockStatement ] | undefined
+  then: block
+  else: e
+
+type ElseToken = { $loc: Loc, token: "else" }
 
 type ASTRef =
   type: "Ref"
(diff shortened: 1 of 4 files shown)


test/if.civet
```diff
@@ -282,9 +282,7 @@ describe "if", ->
     ---
     x if { x } := y
     ---
-    let ref;if (ref = y) {
-    const { x } = ref;x
-    }
+    let ref;if ((ref = y) && 'x' in ref) { const { x } = ref;x }
```
edemaine, line 285: Whoa, this is pretty interesting. But I think makes sense; `:=` is finally becoming a pattern-matching operator.

test/if.civet
```diff
@@ -296,15 +294,13 @@ describe "if", ->
       z
     } if w := g()
     ---
-    let ref;if (ref = g()) {
-    const w = ref;if (ref1 = f()) {
+    let ref;if (ref = g()) { const w = ref;if (ref1 = f()) {
```
edemaine, line 297: This seems to be missing a `let ref1`.

test/if.civet
```diff
@@ -797,13 +793,25 @@ describe "if", ->
...
+      ---
+      if {x, ^y} := obj
+        console.log x
+      ---
+      let ref;if ((ref = obj) && 'x' in ref && 'y' in ref && ref.y === y) {
+        const {x, y} = ref;
+        console.log(x)
+      }
+    """
+
```
edemaine, line 814: 🤯 Very cool.

source/lib.civet
```diff
@@ -131,26 +144,18 @@ function updateParentPointers(node: ASTNode, parent?: ASTNodeBase, depth = 1): v
...
-
-  if prefix?.length
-    indent := prefix[0][0]
-    expressions = [...prefix, [indent, statement]]
-    children = [" {\n", ...expressions, "\n", indent?.slice?.(0, -2), "}"]
-  else
-    expressions = [["", statement]]
-    children = [" { ", ...expressions, " }"]
+  expressions := [
+    ...post.blockPrefix or []
```
edemaine, line 148: I wonder about the loss of `post.condition?.expression.blockPrefix` here... but maybe it's not the cause of the bug below.

edemaine (commented):
This is very cool! I think I found one bug in one of the examples though (my annoying double-`if` example from before).

IMO, `if {x, y} := z` should not check for truthiness of `x` and `y`; I think it should behave the same as
```js
if (
  switch z
    {x, y}
      true
    else
      false
)
```

test/if.civet
STRd6: It looks like it was missing that from the first time I added it, I'll see about adding it tomorrow.

source/lib.civet
STRd6: The condition expression block prefix is now added in a more uniform way inside `processDeclarationConditions`, this one here is to pick up the random prefixes from coffee style for loops until we move them later on in the parse tree processing.

PR #718, opened 2023-09-10
Warning-free hack for ergonomic require of cjs esbuild plugin

edemaine:
It's still a bit hacky, but at least it doesn't cause esbuild to generate a loud warning.

Tested:
* `require('./dist/esbuild-plugin.js')` in `node` CLI (produces function)
* `import plugin from './dist/esbuild-plugin.js'` in a file `test.mjs` (produces function)
* `yarn test`

Based on https://github.com/egoist/tsup/issues/572#issuecomment-1060599574. Also, it looks like if we used `tsup` for this (we already do for the unplugin), there's a cleaner solution: https://tsup.egoist.dev/#interop-with-commonjs

### DIFF
--- a/build/esbuild.civet
+++ b/build/esbuild.civet
@@ -61,6 +61,10 @@ rewriteCivetImports = {
       rewriteCivetImports
       civetPlugin()
     ]
+    footer: if name == 'esbuild-plugin'
+      # Rewrite default export as CJS exports object,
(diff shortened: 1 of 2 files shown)


STRd6 (approved):
Looks good, only thing to note is that this depends on having a single export. We could expand it later if we need to.

PR #720, opened 2023-09-10
Fix #719: existential property glob and get/set shorthand with existential glob

### LINKED ISSUE #719: `get` shorthand with optional access
I expected `get data?.{x,y}` to work for get shorthand syntax from #637, but it doesn't parse.

STRd6:


edemaine (approved):
Thanks! I didn't realize it was glob-specific.

PR #721, opened 2023-09-10
basic parser tracing

STRd6:
It would be nice to be able to output a trace of all the rules including cache hit counts.

This is a start.

- [x] Remove `fs` dep
- [x] Add cache counts
- [x] Figure out output format

### DIFF
--- a/source/main.civet
+++ b/source/main.civet
@@ -7,8 +7,6 @@ export { parse, generate, util }
 
 import StateCache from "./state-cache.civet"
 
-{ writeFileSync } from fs
-
 // Need to no-cache any rule that directly modifies parser state
 // indentation stack, jsx stack, etc.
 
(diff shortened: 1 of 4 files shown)


edemaine:
Definitely sounds like a useful tool for optimization. It reminds me of the current `civet verbose` mode, which I still use for debugging parsing sometimes. Maybe this could replace/improve on that? Or at least be helpful for debugging purposes too.

source/main.civet
```diff
@@ -81,16 +83,53 @@ export compile := (src: string, options?: CompilerOptions) ->
...
+
+  try
+    //@ts-ignore
+    parse.config = options.parseOptions or {}
+    ast = prune parse(src, {
+      filename
+      events
+    })
+  catch err
+    throw err
```
edemaine, line 105: These two lines aren't needed (and will perhaps overcomplicate stack traces): you can just do `try...finally`.

edemaine (approved):
Maybe we can add some documentation for this, perhaps in https://github.com/DanielXMoore/Civet/blob/main/CONTRIBUTING.md ?

PR #722, opened 2023-09-10
Fix #503: TS `using`

### LINKED ISSUE #503: using / Python like with statement
This statement would be better as it would increase readability and improve the language semantics

STRd6:


source/parser.hera
```diff
@@ -5904,6 +5909,43 @@ IndentedJSXChildExpression
...
 
+## Using Declaration
+
+UsingDeclaration
+  Using:decl _? UsingBinding:binding ( __ Comma __ UsingBinding )*:tail UsingJSModeError ->
+    const bindings = [binding].concat(tail.map(([,,,b]) => b))
+
+    return {
+      type: "Declaration",
+      children: $0,
```
edemaine, line 5920: Might be nicer to exclude `UsingJSModeError` here via `[decl, $2, tail]` or maybe name `$2` as `ws`.

edemaine (approved):

source/parser.hera
STRd6: We need to keep it in the parse tree all the way through to generate.

edemaine: Oh, whoops, of course.

PR #725, opened 2023-09-12
Files for testing bun plugin

STRd6:
Doesn't test automatically yet but can be tested manually with:

```bash
( cd integration/example; bun run test.civet )
```


edemaine (approved):
Tested and worked for me on Linux. (I was surprised to see that Bun still doesn't support Windows?)

I added a comment to `test.civet` to explain how to run it.

PR #729, opened 2023-09-13
Fix 'not in' after logical binop

STRd6:
I went to try

```civet
x and y is not in z
```

and found out it was busted.

Here's the fix.

### DIFF
--- a/source/lib.civet
+++ b/source/lib.civet
@@ -967,19 +967,19 @@ function expandChainedComparisons([first, binops]: [unknown, [unknown, ASTLeaf, 
 
   return results
 
-  function processChains(op) {
+  function processChains(op): void
     if first and isRelationalOp op
       first = expandExistence first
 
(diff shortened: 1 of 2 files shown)


source/lib.civet
```diff
@@ -994,23 +995,25 @@ function expandChainedComparisons([first, binops]) {
       })
     } else {
       // Advance start if there was no chain
-      results.push(first, ...binops.slice(start, i + 1).flat())
+      if first
+        results.push(first, ...binops.slice(start, i + 1).flat())
+      else
+        results.push(...binops.slice(start, i + 1).flat())
```
edemaine, line 1001: The then and else cases seem redundant...

```js
results.push first if first
results.push ...binops[start...i+1].flat()
```

edemaine (approved):
Nice catch and fix!

PR #734, opened 2023-09-16
sourcemap fix

STRd6:


integration/unplugin/src/index.ts
```diff
@@ -204,7 +204,7 @@ const civetUnplugin = createUnplugin((options: PluginOptions = {}) => {
       sourceMaps.set(path.resolve(process.cwd(), id), compiled.sourceMap);
 
       const jsonSourceMap = compiled.sourceMap.json(
-        path.basename(id.replace(/\.tsx$/, '')),
+        path.basename(id.replace(/\.[t]sx$/, '')),
```
edemaine, line 207: What is the purpose of this change? Is it supposed to be `[jt]`?

STRd6: This fixes the sourcemap `sources` key from being `main.civet.jsx` and makes it the correct `main.civet` when building using esbuild.

STRd6: > What is the purpose of this change? Is it supposed to be `[jt]sx`?

Oops, misread, yes it is supposed to be that. Now to figure out why it appeared to be working as is... :P

source/util.civet
```diff
@@ -68,6 +68,8 @@ export SourceMap = (sourceString) ->
       mappings: @renderMappings()
       names: []
       sourcesContent: [sourceString]
+      toString: ->
+        JSON.stringify(this)
```
edemaine, line 72: Should this be a method on the class or the `json` object? From the failing test, I assume the former, so this should be dedented one level.

edemaine (approved):

PR #735, opened 2023-09-16
Cli update

STRd6:
Convert to a more civet style.

Added propagation of `--inspect` to child process.

### DIFF
--- a/source/cli.civet
+++ b/source/cli.civet
@@ -286,16 +286,27 @@ cli = ->
             console.error "Could not write #{filename} for Civet ESM mode:"
             console.error e
             process.exit 1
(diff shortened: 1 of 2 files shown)


source/cli.civet
```diff
@@ -1,8 +1,7 @@
 "civet coffeeCompat"
```
edemaine, line 1: Is this flag still necessary?  (I see the comments got translated, for example.)

edemaine (approved):

source/cli.civet
STRd6: I haven't adjusted `of`/`in` yet.

PR #737, opened 2023-09-17
LSP Build refactor

STRd6:
Using dependencies instead of bundling for the lsp's actual dependencies Added @danielx/civet and TypeScript to dependencies Using published @danielx/civet rather than linked
Sped up build by removing ts-node from build step
Removed CoffeeScript support (moved to plugin)
Greatly reduced bundle size by removing CoffeeScript (it pulled in babel which was > 5MB) Struggled to improve breakpoints from VS Code (still doesn't work but connecting Dev Tools debugger does)

### DIFF
--- a/lsp/integration/project-test/.civet/coffee-plugin.mjs
+++ b/lsp/integration/project-test/.civet/coffee-plugin.mjs
@@ -1,4 +1,6 @@
 // Rudimentary CoffeeScript plugin
+import { compile as coffeeCompile } from "coffeescript"
+
 function compile(path, source) {
   const { js, sourceMap } = coffeeCompile(source, {
     bare: true,
--- a/lsp/integration/project-test/.civet/coffee-plugin.mjs
+++ b/lsp/integration/project-test/.civet/coffee-plugin.mjs
@@ -1,0 +1,58 @@
+// Rudimentary CoffeeScript plugin
+function compile(path, source) {
+  const { js, sourceMap } = coffeeCompile(source, {
+    bare: true,
+    filename: path,
+    header: false,
+    sourceMap: true
+  })
+
+  const convertedSourceMap = convertCoffeeScriptSourceMap(sourceMap)
+
+  console.log("COFFEE SOURCE MAP", sourceMap, convertedSourceMap)
+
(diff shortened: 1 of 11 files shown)


lsp/source/lib/util.mts
```diff
@@ -328,34 +326,3 @@ export function forwardMap(sourcemapLines: SourcemapLines, position: Position) {
   // console.warn(`couldn't forward map src position: ${[origLine, origOffset]}`)
   return position
 }
-
-import type { SourceMap as CoffeeSourceMap } from "coffeescript"
-export function convertCoffeeScriptSourceMap(sourceMap: CoffeeSourceMap): SourceMap["data"]["lines"] {
```
edemaine, line 333: Should this code go in the `coffee-plugin` somehow?

lsp/integration/project-test/.civet/coffee-plugin.mjs
```diff
@@ -0,0 +1,58 @@
+// Rudimentary CoffeeScript plugin
```
edemaine, line 1: Is there some way for the end-user to use this? Perhaps it could be released as a second VSCode plugin, and users wanting both could install both?

edemaine (approved):
Looks like some nice improvements. I'm a little worried about the loss of CoffeeScript functionality; though I personally haven't done any mixing of CoffeeScript and Civet, I think I might in the future, so it'd be nice if this is easy for users to do, e.g. via a second VSCode plugin that adds the CoffeeScript functionality (with an extra dep).

lsp/integration/project-test/.civet/coffee-plugin.mjs
STRd6: Copying this file to the `.civet` folder should enable CoffeeScript integration for the LSP the same way that `.hera` is supported in the root of the Civet repo.

edemaine: Can we document this? I had no idea about either hack... Though now I see the general concept is in the README! Perhaps we could add a specific link to `coffee-plugin` from the README?

PR #738, opened 2023-09-17
Fix go to definition

STRd6:
This continues from lsp build refactor.

Now we reverse map the definitions after forward mapping to look them up in transpiled files.

### DIFF
--- a/lsp/source/server.mts
+++ b/lsp/source/server.mts
@@ -276,7 +276,7 @@ connection.onDefinition(async ({ textDocument, position }) => {
   const service = await ensureServiceForSourcePath(sourcePath)
   if (!service) return
 
-  let definitions, meta: FileMeta | undefined
+  let definitions
 
   // Non-transpiled
   if (sourcePath.match(tsSuffix)) {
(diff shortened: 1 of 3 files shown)


lsp/source/server.mts
```diff
@@ -310,11 +309,23 @@ connection.onDefinition(async ({ textDocument, position }) => {
...
+    if (meta) {
+      const { sourcemapLines } = meta
+      return {
+        uri: service.getSourceFileName(fileName),
+        range: {
+          start: remapPosition(sourceFile.getLineAndCharacterOfPosition(textSpan.start), sourcemapLines),
+          end: remapPosition(sourceFile.getLineAndCharacterOfPosition(textSpan.start + textSpan.length), sourcemapLines),
+        }
+      }
+    }
```
edemaine, line 327: This is a fair amount of duplicated code with the blow. Perhaps worth something like:

```js
let start = sourceFile.getLineAndCharacterOfPosition(textSpan.start)
let end = sourceFile.getLineAndCharacterOfPosition(textSpan.start + textSpan.length)
if (meta) {
  const { sourcemapLines } = meta
  start = remapPosition(start, sourcemapLines)
  end = remapPosition(end, sourcemapLines)
}
return {
  uri: service.getSourceFileName(fileName),
  range: {start, end}
}

PR #740, opened 2023-09-17
convert build to civet style

STRd6:


source/ts-diagnostic.civet
```diff
@@ -1,12 +1,14 @@
-ts from 'typescript'
 type { Diagnostic, DiagnosticMessageChain, TextSpan } from 'typescript'
+enum DiagnosticCategory
+  Warning = 0
+  Error = 1
+  Suggestion = 2
+  Message = 3
```
edemaine, line 6: Any particular reason for this not to be imported?

edemaine (approved):

source/ts-diagnostic.civet
STRd6: TS add 8MB when bundled with esbuild because it is not (yet) tree-shakeable: https://github.com/microsoft/TypeScript/issues/32949

edemaine: Oh, I see; I thought it was just a type, but now I remember that `enum`s generate code.

Maybe worth including a comment:
```js
// { DiagnosticCategory } from typescript
```

PR #741, opened 2023-09-17
Unbundled only works in debug mode

STRd6:
Later we could add / check the debug flag to mark eternal in when debugging so 200k lines of typescript.js don't lag out the dev tools.

### DIFF
--- a/lsp/.vscode/launch.json
+++ b/lsp/.vscode/launch.json
@@ -1,7 +1,6 @@
 {
   "version": "0.2.0",
   "configurations": [
-
     {
       "args": [
         "--extensionDevelopmentPath=${workspaceFolder}"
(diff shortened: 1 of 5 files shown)


edemaine:
I'm confused - does this PR work for releasing the LSP, or does it only work in debug mode? It seems weird... but I'm no VSCode expert.

STRd6:
This is the configuration that works for creating a release build. The debug mode works with either configuration and adding external makes setting break points in dev tools much more manageable. I'd like to add an env check to do that automatically eventually.

edemaine (approved):
I see. Is it worth keeping the old version as `build-dev.civet` or something similar, until the env check works, so that you can use it during dev?

PR #746, opened 2023-09-23
Fix #733; Allow postfixed expressions in array literals

### LINKED ISSUE #733: Postfix if within array literal
## Input

```js
indices := [
  allText.lastIndexOf ' ', middle-1
  middle if allText[middle] is ' '
  allText.indexOf ' ', middle+1
]
```

## Expected

```js
const indices = [
  allText.lastIndexOf(' ', middle-1),
  allText[middle] === ' ' ? middle : void 0,
  allText.indexOf(' ', middle+1)
]
```

## Current Behavior

```js
const indices = [
  allText.lastIndexOf(' ', middle-1),
  middle((allText[middle] === ' ')?(
  allText.indexOf(' ', middle+1)
):void 0)
]
```

STRd6:


test/array.civet
```diff
@@ -239,6 +239,58 @@ describe "array", ->
...
+
+  testCase """
+    postfix if with splat
+    ---
+    [
+      ...a if x
+    ]
+    ---
+    [
+      ...(x)?a:void 0
```
edemaine, line 278: This is sadly a bit broken, because `[...undefined]` doesn't work (while `{...undefined}` does...). Doesn't need to be fixed here, but would be cool. (I'm also not quite sure how to fix it...)

edemaine (approved):
Thanks for fixing this!

test/array.civet
STRd6: One possible fix would be: if there is a `...` to fill an implicit `else` branch with a `[]`

edemaine: That would be super cool, also with non-suffix versions:

```js
[
  ...if x
    [x, y, z]
]
```

Honestly I'd kind of prefer this behavior for things like:

```js
[
  x if x?
  y if y?
]
↓↓↓
[
  ...(x != null ? [x] : []),
  ...(y != null ? [y] : [])
]
```

But I guess that might be counterintuitive / a bit too magical?  What do you think?  I've mentioned adding this kind of notation to object literals, but it's less ambiguous there:

```js
{
  x: y if y?
}
↓↓↓
{
  ...(y != null ? {x: y} : void0)
}
```

PR #762, opened 2023-10-02
Parenthesize if expressions, remove other excess parens

### LINKED ISSUE #757: Incorrect code generation for `void if`
This Civet code:

```civet
void if a
  b
else
  c
```

generates this JS code:

```js
void a ? b : c;
```

This is parsed as `(void a) ? b : c`, not `void (a ? b : c)`. Civet needs to wrap the conditional expression in parentheses in this case.

edemaine:
Fixes #757

This ended up also fixing some `while` pattern matching tests, though I can't say I exactly understand why. 😅 (probably the old unwrapping of `condition.children`)

### DIFF
--- a/source/lib.civet
+++ b/source/lib.civet
@@ -1,5 +1,5 @@
 /**
- * lib.js holds functions that are used inside parser.hera
+ * lib.civet holds functions that are used inside parser.hera
  *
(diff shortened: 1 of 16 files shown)


STRd6 (approved):
Nice!

PR #766, opened 2023-10-04
Make ts-diagnostic.civet independent of vscode dependencies

Mokshit06:
This PR removes the dependency of `ts-diagnostic.civet` on vscode-specific dependencies as those are only required for the LSP and moves that code back to the lsp dir.


STRd6 (approved):

edemaine:
@STRd6 Feel free to go ahead and merge + release; I won't be able to review this today. Thanks!

PR #771, opened 2023-10-06
Perf3

STRd6:
Some interesting performance improvements:

- Switching around cache order had a ~50% perf improvement
- Factoring common prefixes in some rules had a ~1-2% improvement

`./dist/civet < source/lib.civet`
2.2018s
`./node_modules/.bin/civet < source/lib.civet`
4.0975s

### DIFF
--- a/source/main.civet
+++ b/source/main.civet
@@ -203,7 +203,7 @@ makeCache := ({hits, trace}: CacheOptions = {}): CacheEvents ->
 
       return if uncacheable.has(ruleName)
 
-      [stateKey, tagKey] := [...getStateKey()]
+      [stateKey, tagKey] := getStateKey()
       key: CacheKey := [tagKey, stateKey, state.pos, ruleName ]
 
       // We cache `undefined` when a rule fails to match so we need to use `has` here.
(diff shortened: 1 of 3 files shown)


source/parser.hera
```diff
@@ -524,8 +524,8 @@ ConditionalExpression
 TernaryRest
   NestedTernaryRest
   # NOTE: Ternary `a ? b : c` is disabled if CoffeeScript binary existential `a ? b` is enabled
-  !CoffeeBinaryExistentialEnabled &" " _? QuestionMark ExtendedExpression __ Colon ExtendedExpression ->
-    return $0.slice(2)
+  !CoffeeBinaryExistentialEnabled _ QuestionMark ExtendedExpression __ Colon ExtendedExpression ->
```
edemaine, line 527: I'm slightly worried about the loss of `&" "`. Is an inline conment here sufficient? Maybe...

source/parser.hera
```diff
@@ -925,7 +925,8 @@ CallExpressionRest
 
 # NOTE: Added shorthand x?(3) -> x?.(3)
 OptionalShorthand
-  InlineComment* QuestionMark OptionalDot ->
+  # perf: assertion to exit early
+  /(?=[\/?!])/ InlineComment* QuestionMark OptionalDot ->
```
edemaine, line 929: `!` isn't needed here, unless we place this in a rule above (which might help?).

source/main.civet
```diff
@@ -203,7 +203,8 @@ makeCache := ({hits, trace}: CacheOptions = {}): CacheEvents ->
 
       return if uncacheable.has(ruleName)
 
-      key: CacheKey := [ruleName, state.pos, ...getStateKey()]
+      [stateKey, tagKey] := [...getStateKey()]
```
edemaine, line 206: I don't think we need the `[...]` wrapper here (also below).

I also wonder whether we could avoid a Set wrapper when the jsxTag is null, which is most of the time... Now that it's in front.

edemaine (approved):

Issue #775, opened 2023-10-11
Vite unplugin dev mode scripts


edemaine:
Reproduction:
* `cd integration/unplugin/examples/vite`
* `pnpm install`
* `pnpm run dev`
* Open localhost dev server in Chrome

Result in console:

> Failed to load module script: Expected a JavaScript module script but the server responded with a MIME type of "". Strict MIME type checking is enforced for module scripts per HTML spec.

Tested on Windows and Linux.

Issue #773, opened 2023-10-11
unplugin watch


edemaine:
The esbuild unplugin doesn't currently work with [`watch`](https://esbuild.github.io/api/#watch). I believe we need to return [`watchFiles`](https://esbuild.github.io/plugins/#on-resolve-results).

Progress so far:
- [x] Rollup
- [x] Webpack
- [x] Vite build mode
- [x] Vite dev mode
- [x] esbuild (waiting on https://github.com/unjs/unplugin/pull/345)

STRd6:
This probably requires a PR to https://github.com/unjs/unplugin/

https://github.com/unjs/unplugin/blob/39c0b22a30bdd7112962e377c39d2a24dcbc5847/src/esbuild/index.ts#L64 is setting the namespace to the plugin name and not `file` which is required by esbuild to watch paths.

https://esbuild.github.io/plugins/#on-resolve-results

STRd6:
Additionally there is currently no way in unplugin to pass through `watchFiles` or `watchDirs`.

edemaine:
unplugin [claims support for `this.addWatchFile`](https://github.com/unjs/unplugin/blob/main/README.md#supported-1) but not yet for esbuild. Maybe we can easily add it (via a PR to unplugin) by storing them in an array and returning it from [`getWatchFiles`](https://github.com/unjs/unplugin/blob/39c0b22a30bdd7112962e377c39d2a24dcbc5847/src/esbuild/utils.ts#L134).

STRd6:
Watch works in [`esbuild-plugin`](https://github.com/DanielXMoore/Civet/blob/main/source/esbuild-plugin.civet) but it has some other bugs (something with sourcemaps, and https://github.com/DanielXMoore/Civet/issues/750). We may be able to update it to be at parity and it would help us understand the build process better as well.

STRd6:
I think the core unplugin issue is that it sets `namespace` to be the name of the plugin instead of `file` which causes the files to not be visible to `watch`.

edemaine:
I'm not sure removing the `namespace` setting from unplugin would fix the problem, because our `resolveId` callback (`onResolve` in esbuild terminology) adds a `.tsx` or `.jsx` extension to the filename, so it will no longer be a valid filename. (I assume this means watch is broken in all systems? Also it maybe explains the choice of namespace.) I'm hoping implementing `addWatchFile` might be a better universal solution for all build systems.

edemaine:
[...]

PR #780, opened 2023-10-15
unplugin calls addWatchFile

edemaine:
Together with https://github.com/unjs/unplugin/pull/345, this will fix eventually #773.

I have tested this (`yarn link`ing to my unplugin fork), using the new esbuild example code.

For now, it might fix watch mode on other build systems. But I wasn't able to test because e.g. Vite dev mode is currently broken (#775).

Update: The types don't work until the unplugin PR gets merged. So maybe we should wait for that.

### DIFF
--- a/integration/unplugin/examples/esbuild/esbuild.js
+++ b/integration/unplugin/examples/esbuild/esbuild.js
@@ -1,0 +1,16 @@
+import * as esbuild from 'esbuild'
+import civetEsbuildPlugin from '@danielx/civet/esbuild'
+
+const options = {
+  entryPoints: ['src/main.civet'],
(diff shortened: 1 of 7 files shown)


edemaine:
This is now ready for review! I moved the `addWatchFile` call to the `load` callback, which is already supported by the types and API, and lets us merge this now.

I did some testing and this fixes watch mode on Rollup and Webpack! It should also fix esbuild (and #773) once https://github.com/unjs/unplugin/pull/345 gets merged. It still doesn't fix Vite (on Windows or Linux), for some reason I don't yet understand. But... progress?

STRd6 (approved):

PR #781, opened 2023-10-16
Fix #643

### LINKED ISSUE #643: Automatic semicolon insertion (ASI) bugs
Civet's semicolon insertion currently runs during initial parse, not after all transpilation, so it can miss some cases where semicolons are needed. Here's a collection of examples with incorrect ASI:

```coffee
->
  yield foo
->
  await foo
```

```coffee
foo := new Bar

async do
   console.log "hi"
```

```coffee
point = data.{x,y}
point.{x,y} = data
```

```coffee
merged := data.{...global, ...user}
data.{a, b, ...rest} = result
```

```coffee
x?
x.y[z]?
```

STRd6:
Fixes many semicolon insertion issues by moving semicolon insertion into a later stage of the parser.

### DIFF
--- a/source/lib.civet
+++ b/source/lib.civet
@@ -2580,6 +2580,8 @@ function insertSemicolon(statements: StatementTuple[]): void
         delim := s[2]
         if !delim
           s[2] = ";"
+        // If delim is an object assume it already has a semicolon
+        // TODO: trailing comments might violate this assumption
         else if typeof delim is "string" and !delim.match(/;/)
           s[2] = `;${delim}`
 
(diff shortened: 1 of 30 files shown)


source/lib.civet
```diff
@@ -2538,6 +2561,50 @@ function processAssignments(statements): void {
...
+function insertSemicolon(statements: StatementTuple[]): void
+  l := statements.length
+  statements.forEach (s, i) ->
+    if i < l - 1
+      if needsPrecedingSemicolon(statements[i + 1])
+        delim := s[2]
+        if !delim
+          s[2] = ";"
+        else if typeof delim is "string" and !delim.match(/;/)
+          s[2] = `;${delim}`
```
edemaine, line 2584: I assume `delim` can be an object, but only when it's already a semicolon so we don't need to do anything? Maybe add a comment, like `// If delim is an object, assume it's already a semicolon`.

source/lib.civet
```diff
@@ -2538,6 +2561,50 @@ function processAssignments(statements): void {
...
+    // since the js: true is on the AST node the semicolon delimiters will stack up with empty nodes in between
+    following = exp
+
+  return false unless following
+
+  if Array.isArray following // for chained conditionals
+    return needsPrecedingSemicolon following[0]
+
+  switch following.type
+    when "ParenthesizedExpression", "ArrowFunction", "TemplateLiteral"
```
edemaine, line 2601: I wonder about adding `IfExpression`, `DebuggerExpression`, `SwitchExpression`, `ThrowExpression`, `TryExpression` (a partial list from `makeLeftHandSideExpression`). Then again, if they've already been parenthesized at this stage, probably not needed.

Perhaps more importantly, I think `RegularExpressionLiteral` might need to be on this list; MDN lists this as an explicit case, and so our old code looked for leading `/`.

edemaine (approved):
This is awesome!! Thanks for doing this. It does not look like it was easy to get working, and we might run into missing statement tuples in the future (which we can hopefully eventually fix with types and then actual objects), but this is definitely the correct direction.

Amazing the number of bugs (missing semicolons) in the old test output. I wonder whether we should add more examples from #643 to the tests? I've already removed the ugly workaround semicolons from the docs.

STRd6:
It also added a ~7% performance boost by avoiding all those lookaheads every statement 📈

PR #783, opened 2023-10-20
Allow comments near Civet directives

p2edwards:
Preserve comments near Civet directives

xref [Discord discussion](https://discord.com/channels/933472021310996512/933472021310996515/1165013088697860146)

### DIFF
--- a/source/parser.hera
+++ b/source/parser.hera
@@ -6525,11 +6525,9 @@ Shebang
 
 CivetPrologue
   [\t ]* DoubleQuote CivetPrologueContent:content DoubleQuote SimpleStatementDelimiter _? EOL? EOS?:eos ->
-    content.children.push(eos)
-    return content
+    return {...content, children: [...content.children, eos]}
   [\t ]* SingleQuote CivetPrologueContent:content SingleQuote SimpleStatementDelimiter _? EOL? EOS?:eos ->
-    content.children.push(eos)
-    return content
+    return {...content, children: [...content.children, eos]}
 
 CivetPrologueContent
   "civet" NonIdContinue CivetOption*:options [\s]* ->
--- a/source/parser.hera
+++ b/source/parser.hera
@@ -6524,8 +6524,12 @@ Shebang
   /#![^\r\n]*/ EOL
 
 CivetPrologue
-  [\t ]* DoubleQuote CivetPrologueContent:content DoubleQuote SimpleStatementDelimiter EOS? -> content
-  [\t ]* SingleQuote CivetPrologueContent:content SingleQuote SimpleStatementDelimiter EOS? -> content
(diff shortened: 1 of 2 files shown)


source/parser.hera
```diff
@@ -6524,8 +6524,12 @@ Shebang
   /#![^\r\n]*/ EOL
 
 CivetPrologue
-  [\t ]* DoubleQuote CivetPrologueContent:content DoubleQuote SimpleStatementDelimiter EOS? -> content
-  [\t ]* SingleQuote CivetPrologueContent:content SingleQuote SimpleStatementDelimiter EOS? -> content
+  [\t ]* DoubleQuote CivetPrologueContent:content DoubleQuote SimpleStatementDelimiter _? EOL? EOS?:eos ->
+    content.children.push(eos)
```
STRd6, line 6528: Thanks for the PR! One small issue is that the way caching works in the parser we want to avoid mutating the objects passed to the handler rules.

For this case we could do something like:

```javascript
return {...content, children: [...content.children, eos]}
```

p2edwards: Thanks! I've incorporated your suggestion verbatim.

source/parser.hera
```diff
@@ -6524,8 +6524,10 @@ Shebang
   /#![^\r\n]*/ EOL
 
 CivetPrologue
-  [\t ]* DoubleQuote CivetPrologueContent:content DoubleQuote SimpleStatementDelimiter EOS? -> content
-  [\t ]* SingleQuote CivetPrologueContent:content SingleQuote SimpleStatementDelimiter EOS? -> content
+  [\t ]* DoubleQuote CivetPrologueContent:content DoubleQuote SimpleStatementDelimiter _? EOL? EOS?:eos ->
```
edemaine, line 6527: Why did you add `_? EOL?`? In particular, this will discard inline comments (in `_`), and seems like it can consume two newlines. But maybe I'm missing something?

p2edwards: Oh interesting! I overlooked that `_` included InlineComment.

With `_? EOL?` I was looking for a simple way to do these 2 things:

1. Trim initial whitespace from an end-of-line comment (or the start of a multiline one)
2. Omit the blank line if there's no comment after a directive

So that

```coffee
"civet -implicitReturns" // hello!
"civet autoLet"
// hi!
```

becomes

```ts
// hello!
// hi!
```

rather than 
```ts
 // hello!

// hi!
```

edemaine: Ah, I see. Thanks for the explanation! That makes sense now.

The `Init` rule where this gets used has (recently added) support for comments, so I wonder if the following would actually be a simpler solution to this problem:

```
... PrologueEOL [ \t]* ( EOL / &RestOfLine ) -> content
```

This should leave the comments on the remainder of the line for further processing, but only if there's something other than a newline, after trimming spaces/tabs.

STRd6 (approved):

STRd6:
Thanks!

Issue #774, opened 2023-10-11
Vite unplugin doesn't support project-absolute paths


edemaine:
> The paths also seem to be resolved differently if I'm building it, but only if it's a civet file. By default the script looks like this in `index.html`: `/src/main.ts`, note the starting `/`, which works normally with Vite and even works with `.ts` file even if I use the civet plugin. But if I use `/src/main.civet` it seems to try to resolve it from the machine's root, not the project root and doesn't find the file. Only in build. Changing it to `./src/main.civet` works but it is different from normal Vite though.

[Stackblitz Reproduction](https://stackblitz.com/edit/vitejs-vite-vy56su?file=index.html)

STRd6:
What's the proper way to define project-absolute paths so the TS LSP picks it up?

In my testing it (surprisingly) worked with the Civet LSP in `.ts` files but the TS LSP thinks they are actual absolute paths and I don't know any config option to set to make TS understand this pattern.

Any thoughts?

edemaine:
I think you can use `baseUrl` (unclear if it works with leading slash though). `paths` is also a common solution.

STRd6:
The Stackblitz web based editor seems to pick it up fine but the same project and same configuration don't work locally on VSCode 🙀 

![image](https://github.com/DanielXMoore/Civet/assets/18894/2a7e4d6d-4e93-4eba-a65a-c727de0e3d4e)

![image](https://github.com/DanielXMoore/Civet/assets/18894/c89e99ef-f350-4245-8c10-09ef9e830826)

PR #785, opened 2023-10-21
Transform Vite HTML imports for Civet

Mokshit06:
This PR transforms the index.html file when being used with vite to add `outExt?transform` to the end of the import so that it gets transformed by the unplugin

### DIFF
--- a/integration/unplugin/src/index.ts
+++ b/integration/unplugin/src/index.ts
@@ -44,7 +44,8 @@ export type PluginOptions = {
 );
 
 const isCivet = (id: string) => /\.civet$/.test(id);
-const isCivetTranspiled = (id: string) => /\.civet\.(m?)(j|t)s(x?)$/.test(id);
+const isCivetTranspiled = (id: string) =>
+  /\.civet\.(m?)(j|t)s(x?)(\?transform)?$/.test(id);
 const isCivetTranspiledTS = (id: string) => /\.civet\.(m?)ts(x?)$/.test(id);
 
 const civetUnplugin = createUnplugin((options: PluginOptions = {}) => {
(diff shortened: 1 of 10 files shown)


integration/unplugin/src/index.ts
```diff
@@ -252,6 +256,20 @@ const civetUnplugin = createUnplugin((options: PluginOptions = {}) => {
 
         return null;
       },
+      async transformIndexHtml(html) {
+        const cheerio = await import('cheerio');
+        const $ = cheerio.load(html);
```
edemaine, line 261: Can we avoid the `cheerio` dependency? Civet aims to have a minimum number of dependencies. I think in this case we could probably write a careful regex...

integration/unplugin/src/index.ts
```diff
@@ -252,6 +256,20 @@ const civetUnplugin = createUnplugin((options: PluginOptions = {}) => {
 
         return null;
       },
+      async transformIndexHtml(html) {
```
edemaine, line 259: Should we do this only in dev mode? Or is it fine to have in both?

Mokshit06: Should be fine in both

edemaine:
I replaced the cheerio dependency with what I think is a reasonable regex HTML parser (for the purpose of detecting `<script>` tags). However, I can't get this to work on Linux or Windows. I get this error:

```
[vite] Internal server error: ENOENT: no such file or directory, open '/src/main.civet'
```

Is this some form of #774? The path is relative in the HTML file, though.

It also seems that, by the time the id is passed to `resolveId`, it has lost the `?transform` suffix. So I think the code might be getting confused because of that... Did you actually get this working in some test environment?

Mokshit06:
Strange, I did have it working on MacOS. Let me check once again

Mokshit06:
This change should no longer be required with #786. Also I'm pretty sure I've come across a bug either in Unplugin or Vite regarding how the paths are passed in from `resolveId` to `load`. Will investigate further to check which package is causing it

edemaine (approved):
With #786 merged in, this now works for me. Great!

PR #786, opened 2023-10-27
Fix unplugin path resolution (#774)

### LINKED ISSUE #774: Vite unplugin doesn't support project-absolute paths

Mokshit06:
This PR should fix #774 and also make #785 not required. `path.resolve` would not consider the parent path if the child path starts with `/`, which made paths like `/src/main.civet` be considered as system-absolute path instead of project-absolute. `path.join` should take them into account always, and make this work.
The only issue this would face is not supporting system-absolute paths for civet, though I'm not sure if that is a feature that is actually used that much

### DIFF
--- a/integration/unplugin/src/index.ts
+++ b/integration/unplugin/src/index.ts
@@ -15,6 +15,7 @@ import * as fs from 'fs';
 import path from 'path';
 import ts from 'typescript';
 import * as tsvfs from '@typescript/vfs';
+import { UserConfig } from 'vite';
 
 const formatHost: ts.FormatDiagnosticsHost = {
   getCurrentDirectory: () => ts.sys.getCurrentDirectory(),
@@ -64,6 +65,7 @@ const civetUnplugin = createUnplugin((options: PluginOptions = {}) => {
   let fsMap: Map<string, string> = new Map();
   const sourceMaps = new Map<string, SourceMap>();
   let compilerOptions: any;
+  let viteConfig: UserConfig;
 
   return {
     name: 'unplugin-civet',
@@ -178,7 +180,7 @@ const civetUnplugin = createUnplugin((options: PluginOptions = {}) => {
 
       const relativeId = path.relative(
         process.cwd(),
-        path.join(path.dirname(importer ?? ''), id)
+        path.resolve(path.dirname(importer ?? ''), id)
       );
       const relativePath = relativeId + outExt;
 
@@ -239,7 +241,8 @@ const civetUnplugin = createUnplugin((options: PluginOptions = {}) => {
       return null;
     },
     vite: {
-      config(_config, { command }) {
+      config(config, { command }) {
+        viteConfig = config;
         // Ensure esbuild runs on .civet files
         if (command === 'build') {
           return {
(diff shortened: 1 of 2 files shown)


edemaine:
I tested this on Windows and Linux, and found `vite build` works now, when either `index.html` or `src/main.civet` uses absolute paths to refer to other Civet files.

However, `vite dev` still doesn't work: loading `http://localhost:5173/src/main.civet` in the `<script>` tag still gives me the raw Civet code, with no MIME type. So maybe #785 is still necessary? Did you find `yarn dev` worked for you somehow?

integration/unplugin/src/index.ts
```diff
@@ -178,7 +178,7 @@ const civetUnplugin = createUnplugin((options: PluginOptions = {}) => {
 
       const relativeId = path.relative(
         process.cwd(),
-        path.resolve(path.dirname(importer ?? ''), id)
+        path.join(path.dirname(importer ?? ''), id)
```
edemaine, line 181: Actually, I wonder whether this is correct... I would have guessed:
* If the path is relative, use `path.join` with `importer`, as you have it here
* If the path is absolute, use `path.join` with the project's `root` setting (which probably also needs to be `join`ed with `process.cwd()` because it too can be relative).

This is somewhat confirmed because I find `src/main.civet` can successfully import `/module.civet`, but I don't think it should be able to; it should have to import `/src/module.civet` I believe.

Mokshit06: Oops yeah you're correct, I overlooked that case.

Mokshit06:
I am very sure that `yarn dev` worked after rebuilding with this change, but it doesn't actually. I think this might be because vite's reusing the cache in `node_modules/.vite` 🤔 But yeah then #785 is indeed still required

integration/unplugin/src/index.ts
```diff
@@ -252,6 +255,19 @@ const civetUnplugin = createUnplugin((options: PluginOptions = {}) => {
 
         return null;
       },
+      resolveId(id, importer) {
+        if (/\0/.test(id)) return null;
+        if (!isCivet(id)) return null;
+
+        const absolutePath = path.isAbsolute(id)
+          ? path.join(path.join(process.cwd(), viteConfig.root ?? ''), id)
+          : path.join(path.dirname(importer ?? ''), id);
```
edemaine: I'm pretty sure the latter case should use `path.resolve`, not `path.join`.

I pushed a new commit that cleans up this code. Could you check it?

Mokshit06: The result should be the same with `path.resolve` and `path.join` as we're already handling the case where `id` is an absolute path which is the only case where resolve and join differ

Mokshit06: Rest of the changes look good 👍

PR #789, opened 2023-10-29
Fix dts generation for Vite & Rollup

Mokshit06:
Rollup doesn't support returning non-js source inside the loader, but the previous plugin returned TS from loader if dts or typecheck were enabled, and then added that to typescript vfs map in transform step.
This didn't really help much, so I've moved the transform step to be inside the loader as well. In addition, the loader now always returns JS, and if TS features are enabled, then it does an additional build with TS enabled and uses that for vfs.

I'm not sure how fast civet's compile step is though. Would it be faster if we just did a TS compile with civet, and then transform that to JS with esbuild?

### DIFF
--- a/integration/unplugin/README.md
+++ b/integration/unplugin/README.md
@@ -74,7 +74,6 @@ module.exports = {
 interface PluginOptions {
   dts?: boolean;
   outputExtension?: string;
-  js?: boolean;
   typecheck?: boolean;
   transformOutput?: (
     code: string,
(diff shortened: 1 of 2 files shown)


STRd6:
> I'm not sure how fast civet's compile step is though. Would it be faster if we just did a TS compile with civet, and then transform that to JS with esbuild?

It will likely be faster to do a single compile step with Civet and then use esbuild for the ts -> js transform.

edemaine:
I think this has been replaced by later PRs.

PR #793, opened 2023-11-03
Fix #792: TryExpression in conditional declaration

### LINKED ISSUE #792: if-const + try throws a compiler error
When using `if` with constant assignment from a `try` expression:
```coffee
if x := try hello()
  print "hello"
```

The compiler throws:
```
Cannot read properties of undefined (reading 'type')
```

This works properly with a non-constant assignment, however:
```coffee
if x = try hello()
  print "hello"
```

I found this while messing around in one of the playground boxes on the homepage.

STRd6:


edemaine (commented):
I was working on this too, and trying to add updateParentPointers in various appropriate places. (I think there are a lot of places where they are missing: pretty much every time in lib we make a node with `children` it makes sense to call it.)

This addParentPointers is expensive, but works for now. I assume it doesn't work with updateParentPointers? If I get my updateParentPointers version working, I'll send in a followup PR.

edemaine (approved):

PR #794, opened 2023-11-05
Update more parent pointers along the way

edemaine:
After the initial parse and parent pointer setting, we modify the AST a lot, and rarely do we update the parent pointers.  This PR is a step in the right direction, in particular refining the `addParentPointers` call just added in #792.

This PR adds a lot of missing `updateParentPointers` (a shallow operation that just looks at immediate object children), including via a new helper `makeNode` that we should probably call every time we make a new AST node that has `children` set to something with objects.

> Note: An alternative would be to merge `updateParentPointers` and `makeNode`, by having `updateParentPointers` always return the node. Let me know if you have a preference!

At the more subtle level, I found some missing `addParentPointers` that are needed when we pull part of an AST from a property like `blockPrefix` or `hoistDec` into `children`, so they weren't previously in the actual tree of `children`, so didn't get visited by the original `addParentPointers` pass. This is fundamentally why an `addParentPointers` is needed in a case like declaration conditions, but it can be applied a little more shallowly than #792.

### DIFF
--- a/source/lib.civet
+++ b/source/lib.civet
@@ -12,7 +12,7 @@ type ASTNodeBase =
   children: ASTNode[]
   blockPrefix?: unknown
   names?: string[]
-  parent: ASTNodeBase | undefined
+  parent?: ASTNodeBase | undefined
 
 type ASTError =
   type: "Error"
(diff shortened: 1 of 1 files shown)


source/lib.civet
```diff
@@ -2347,19 +2341,19 @@ function processDeclarationConditionStatement(s: IfStatement | IterationStatemen
         expression: ref
         parent: s
       s.children[1] = s.condition
+      updateParentPointers s
```
STRd6, line 2344: Does this need a `parent` passed in?

edemaine: I don't think so. `updateParentPointers(x)` updates `x.children` to point to `x`. The second argument is if you also want to set `x`'s parent, but it's optional. `s` is at the top level of what this function considers, so we don't know anything special about its parent.

STRd6: Ok, upon re-reading the method I see that you are correct, thanks!

STRd6 (approved):

PR #795, opened 2023-11-05
Forbid comma operator in one-line thin arrow functions

### LINKED ISSUE #751: Incorrect parsing with coffeeCompat and Array#reduce
[copied from discord]

> ```ts
> 'civet coffeeCompat'
> 
> let threads: Object[] # sparse array
> let threadCount: number
> 
> threadCount is threads.reduce (x) -> x + 1, 0
> 
> // ↓↓↓ Civet 0.6.39
> 
> let threads: Object[]; // sparse array
> let threadCount: number;
> 
> threadCount ===
>   threads.reduce(function (x) {
>     return x + 1, 0;
>   });
> ```
> CoffeeScript generates `threads.reduce(function (x) { return x + 1; }, 0)`

edemaine:
Fixes #751 by making `NoComma` variations of `Statements` used in one-line thin and fat arrow functions.

Another option (which is what I actually implemented first) would be to forbid top-level comma operators (requiring parentheses, say), but this seemed a bit unnecessarily strict.

Also allow comma operator inside parenthesized expressions, which seems desirable in any case.

Also refactor `Expression` names a bit:
* `CommaExpression` allows comma operators; formerly this was called `Expression`, but we use `...Expression` all over to forbid commas, so old terminology was confusing.
* `AssignmentExpression` forbids comma operators (as before)

Also fix handling of `{},{}` which is not a valid JavaScript statement (needs to be `({}),({})`). Similarly for anonymous function expressions.

### DIFF
--- a/source/lib.civet
+++ b/source/lib.civet
@@ -1958,6 +1958,23 @@ function makeLeftHandSideExpression(expression)
     implicit: true
   }
 
+// Wrap expression in parentheses to make into a statement when:
+// * object literal expression
+// * anonymous function expression
+// * comma operators applied to above
+function makeExpressionStatement(expression: ASTNode): ASTNode
+  if Array.isArray(expression) and expression[1]?[1]?.token is "," // CommaExpression
+    expression.map (elt, i) =>
+      if i == 0 // AssignmentExpression
+        makeExpressionStatement elt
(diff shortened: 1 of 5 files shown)


STRd6 (approved):
Looks great, thanks!

PR #796, opened 2023-11-06
Fix #704; better open paren whitespace handling in type declarations

### LINKED ISSUE #704: Parse error in function type aliases when line is split
```
type Fred = (a: string) => number
```
parses (and is left alone, as it is valid TypeScript), but neither of
```
type Fred =
    (a: string) => number
```
nor
```
type Fred
   (a: string) => number
```
parses (even though the first is valid TypeScript as it stands).

STRd6:


edemaine (approved):
I'm not quite sure why this works (who processes the EOS/Nested?) but I'm happy that it does. Less `__` is a win.

STRd6:
The `__` is checked in the alternate choice here: https://github.com/DanielXMoore/Civet/pull/796/files#diff-b95b0a26fe57cfe39663405477d1c9cbe418e6ffb2a45b5cdfe47f24d3c68d62L6033

PR #797, opened 2023-11-06
Check for existence of absolute path in unplugin

Mokshit06:
This PR fixes a bug introduced in #786 where all absolute paths were considered to be absolute to project which broke vitest as it instead passed system absolute paths. The plugin now first checks for the existence of the path relative to the project, and then to the system.


STRd6 (approved):
Looks good!

PR #799, opened 2023-11-07
Update xor typing

bbrk24:
(Marking as Draft for now until I update xnor as well)

This PR corrects the typing of the xor operator to reflect the fact that it returns one of the arguments unless they're both truthy, as discussed on Discord.

The Discord discussion originally suggested to make it output:

```ts
var xor = <A, B>(a: A, b: B) => (a ? !b && a : b) as unknown as A extends Falsy ? B : B extends Falsy ? (A | B) : (A | B | false);
```

However, interleaving the types and code like this makes the hera code significantly harder to read, and there's already precedent for using `as any` (e.g. `is` being `Object.is as any`).

### DIFF
--- a/source/parser.hera
+++ b/source/parser.hera
@@ -6967,12 +6967,21 @@ Reset
         module.prelude.push(["", [preludeVar, xorRef, typeSuffix, " = (a, b) => (a ? !b && a : b)", asAny, ";", "\n"]])
       },
       xnor(xnorRef) {
+        const Falsy = module.getRef("Falsy")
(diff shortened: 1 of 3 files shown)


edemaine:
Yeah, I think `as any` inside the function is fine; we just want nice external types.

I do wonder whether the type narrowing is actually useful (if the types guarantee truthy/falsey, why are you using `xor`?) but they don't hurt. An alternative would just be `A | B | false`.

STRd6:
Looks good, thanks!

PR #801, opened 2023-11-08
Added references to lsp

STRd6:


edemaine (commented):
The code looks good. If you'd like me to test too, what exactly is enabled by references? I'm guessing ctrl+click on a function name to go to the function? Or something else? (Probably also fine to merge without double-testing, unless Windows testing would be particularly helpful.)

STRd6:
It's the inverse of go to definition, so if you Ctrl+Click on a definition it shows you where it is used.

edemaine (approved):
Tested and works for me! Cool!

I think you can add another checkmark to the README.

PR #802, opened 2023-11-10
Fix #705

### LINKED ISSUE #705: Generated code appears to omit necessary parentheses
In the [Playground](https://civet.dev/playground/?code=ZnVuY3Rpb24gbWFrZUZuCiAgPT4geDogMSwgeTox),
```
function makeFn
  => x: 1, y:1
```
compiles to
```
function makeFn() {
  return () => {x: 1, y:1}
}
```
but the latter does not compile as TypeScript (or JavaScript, unless I am mistaken). I believe it needs to be
```
function makeFn() {
  return () => ({x: 1, y:1})
}

### PR DESCRIPTION
**Pull request by STRd6**

### CODE REVIEW

File: source/parser.hera [hunk 1]
```diff
@@ -518,7 +518,12 @@ TrailingDeclaration
 # https://262.ecma-international.org/#prod-ConciseBody
 FatArrowBody
   # If same-line single expression, avoid wrapping in braces
-  !EOS NonPipelinePostfixedExpression:exp !TrailingDeclaration !SemicolonDelimiter -> exp
+  !EOS NonPipelinePostfixedExpression:exp !TrailingDeclaration !SemicolonDelimiter ->
+    // Ensure object literal is wrapped in parens
+    if (exp.type === "ObjectExpression") {
```
edemaine, line 523: I think it would be better to use the new `makeExpressionStatement` helper, which also covers more cases.

STRd6: I don't think that's necessary since anonymous functions are fine here. I don't think comma expressions can get picked up here either, they'd be part of the outer expression.

edemaine: Ah, good point. I thought anonymous functions were an issue, but they're not.

edemaine (approved):

PR #807, opened 2023-11-15
Fix #72. Reset service when tsconfig changes

### LINKED ISSUE #72: LSP dependency on tsconfig.json
The LSP should re-parse everything when `tsconfig.json` changes. Currently I need to restart. By contrast, `.tsx` files update when I change `tsconfig.json`.

Ideally the same for `node_modules/@danielx/civet`, for upgrading Civet more seamlessly.

STRd6:
Still need to actually test but the approach seems correct.


edemaine (commented):
I tried this out and didn't see any reloading behavior, when editing or when saving `tsconfig.json`. I didn't even see an `onDidChangeContent` message. Perhaps the plugin isn't running on the `tsconfig.json` file, so doesn't see the change. (Currently only registered for `*.civet` and `*.ts`?)

STRd6:
Good point, probably need to register `.json` for the LSP for this method to work.

edemaine:
Sorry, I missed that this was ready to review for a while.

I just tested and I think the behavior is still not quite right. Here's what I see:
* As I edit `tsconfig.json`, Civet furiously plugin tries to update diagnostics for `.civet` files.
* But it's still loading `tsconfig.json` from disk (I think), so it's getting the old version.
* When I save `tsconfig.json`, it *doesn't* trigger Civet to update diagnostics, so things are still out-of-date.
* Only after making an (unsaved) change again to `tsconfig.json` does Civet actually update, finally with the new `tsconfig.json`.

I don't think we really need to update when `tsconfig.json` is edited, only when it's saved. (As it's edited, it probably has errors.) I wonder if, instead of including jsonc files in the plugin, we should use something like `createFileSystemWatcher`. It could watch for Civet config files too.

STRd6:
Thanks for the tip! `createFileSystemWatcher` seems like a much better approach. I've tested it quickly and it looks like it is working.

PR #810, opened 2023-11-20
Fix unplugin Typescript builds and update API

Mokshit06:
This PR:
- Changes the basic `js: boolean` option to a more configurable `ts: civet | esbuild | tsc | preserve` option instead. I feel like changing `js` to `ts` here makes more sense—we're configuring the transformation of _typescript_ source instead of js.

  The esbuild and tsc option both require a peer dependency which is lazily loaded if the option is used.
- Renames `dts` to `emitDeclaration`. With js renamed to ts, `dts` might be confusing naming. `emitDeclaration` follows naming similar to Typescript's compiler options which have `emitDeclarationOnly`

### DIFF
--- a/civet.dev/getting-started.md
+++ b/civet.dev/getting-started.md
@@ -160,7 +160,7 @@ esbuild.build({
       // Options and their defaults:
       // emitDeclaration: false,         // generate .d.ts files?
       // outputExtension: '.civet.tsx',  // replaces .civet in output
-      // ts: 'civet',                    // use Civet's TS -> JS transpiler?
+      // ts: 'civet',                    // TS -> JS transpilation mode
     })
   ]
 }).catch(() => process.exit(1))
--- a/civet.dev/getting-started.md
+++ b/civet.dev/getting-started.md
@@ -158,9 +158,9 @@ esbuild.build({
   plugins: [
     civetPlugin({
       // Options and their defaults:
-      // dts: false,                     // generate .d.ts files?
+      // emitDeclaration: false,         // generate .d.ts files?
       // outputExtension: '.civet.tsx',  // replaces .civet in output
-      // js: false,                      // use Civet's TS -> JS transpiler?
+      // ts: 'civet',                    // use Civet's TS -> JS transpiler?
     })
   ]
 }).catch(() => process.exit(1))
@@ -186,9 +186,11 @@ import civetPlugin from '@danielx/civet/esbuild';
 
 export default defineConfig({
   entryPoints: ['main.civet'],
-  esbuildPlugins: [civetPlugin({
-    // options
-  })],
+  esbuildPlugins: [
+    civetPlugin({
+      // options
+    }),
+  ],
 });
 ```
 
--- a/civet.dev/getting-started.md
+++ b/civet.dev/getting-started.md
@@ -51,6 +51,7 @@ To use TypeScript for type checking, create a `tsconfig.json` file. For example:
     "strict": true,
     "jsx": "preserve",
     "lib": ["es2021"],
+    "moduleResolution": "nodenext",
     "forceConsistentCasingInFileNames": true,
     "resolveJsonModule": true,
     "esModuleInterop": true
--- a/civet.dev/getting-started.md
+++ b/civet.dev/getting-started.md
@@ -139,7 +139,7 @@ To transpile within a CommonJS NodeJS app
 To transpile in the browser, you can load the browser build `dist/browser.js`
 via a `<script>` tag, and access the global variable `Civet`, as in
 `Civet.compile`.
-Alternatively, if you're using a build system, you can import `"danielx/civet"`
+Alternatively, if you're using a build system, you can import `"@danielx/civet"`
 normally, but you'll need to mark `"fs"` as an external dependency
 (see e.g. [esbuild instructions](https://esbuild.github.io/api/#external>)
 and [Vite instructions](https://vitejs.dev/guide/build#library-mode)).
--- a/integration/unplugin/README.md
+++ b/integration/unplugin/README.md
@@ -84,18 +84,16 @@ interface PluginOptions {
 }
 ```
 
-- `emitDeclaration`: Whether to generate `.d.ts` type definition files from the Civet source, which is useful for building libraries. Default: `false` (Requires installing `typescript`)
+- `emitDeclaration`: Whether to generate `.d.ts` type definition files from the Civet source, which is useful for building libraries. Default: `false`. (Requires installing `typescript`.)
 - `typecheck`: Whether to run type checking on the generated code. Default: `false`.
-- `outputExtension`: Output filename extension to use. Default: `".civet.tsx"`, or uses `".civet.jsx"` if `js` is `true`.
-- `ts`: Mode of transpilation of TS->JS source code. Default: `"civet"`. Options:
-
-  - `"civet"`: Use Civet's JS mode (Not all TS features supported)
-  - `"esbuild"`: Use esbuild's transpiler (Requires installing `esbuild`)
-  - `"tsc"`: Use the typescript compiler (Requires installing `typescript`)
+- `outputExtension`: Output filename extension to use. Default: `".civet.tsx"`, or uses `".civet.jsx"` if `ts` is `"preserve"`.
+- `ts`: Mode for transpiling TypeScript features into JavaScript. Default: `"civet"`. Options:
+  - `"civet"`: Use Civet's JS mode. (Not all TS features supported.)
+  - `"esbuild"`: Use esbuild's transpiler. (Fast and more complete. Requires installing `esbuild`.)
+  - `"tsc"`: Use the TypeScript compiler. (Slow but complete. Requires installing `typescript`.)
   - `"preserve"`: Don't transpile TS code.
-
-    Some bundlers like esbuild and Vite can handle TS directly. Also useful when using `transformOutput` to handle TS code, or using a plugin that modifies TS AST.
-
-    **Note** that this would require additional plugins for bundlers that can't handle TS. For example, for Webpack, you would need to install `ts-loader` and add it to your webpack config. This will **not** work for Rollup.
-
+    Some bundlers, like esbuild and Vite, can handle TS directly. Also useful when using `transformOutput` to handle TS code, or using a plugin that modifies TS AST.
+    Note that some bundlers require additional plugins to handle TS.
+    For example, for Webpack, you would need to install `ts-loader` and add it to your webpack config.
+    Unfortunately, Rollup's TypeScript plugin is incompatible with this plugin, so you need to set `ts` to another option.
 - `transformOutput(code, id)`: Adds a custom transformer over jsx/tsx code produced by `civet.compile`. It gets passed the jsx/tsx source (`code`) and filename (`id`), and should return valid jsx/tsx code.
--- a/integration/unplugin/README.md
+++ b/integration/unplugin/README.md
@@ -88,8 +88,14 @@ interface PluginOptions {
 - `typecheck`: Whether to run type checking on the generated code. Default: `false`.
 - `outputExtension`: Output filename extension to use. Default: `".civet.tsx"`, or uses `".civet.jsx"` if `js` is `true`.
 - `ts`: Mode of transpilation of TS->JS source code. Default: `"civet"`. Options:
-  - `civet`: Use Civet's JS mode (Not all TS features supported)
-  - `esbuild`: Use esbuild's transpiler (Requires installing `"esbuild"`)
-  - `tsc`: Use the typescript compiler (Requires installing `"typescript"`)
-  - `preserve`: Don't transpile TS code (Some bundlers like esbuild and vite can handle TS directly. Note that this would require additional plugins for bundlers that can't handle TS)
+
+  - `"civet"`: Use Civet's JS mode (Not all TS features supported)
+  - `"esbuild"`: Use esbuild's transpiler (Requires installing `esbuild`)
+  - `"tsc"`: Use the typescript compiler (Requires installing `typescript`)
+  - `"preserve"`: Don't transpile TS code.
+
+    Some bundlers like esbuild and Vite can handle TS directly. Also useful when using `transformOutput` to handle TS code, or using a plugin that modifies TS AST.
+
+    **Note** that this would require additional plugins for bundlers that can't handle TS. For example, for Webpack, you would need to install `ts-loader` and add it to your webpack config. This will **not** work for Rollup.
+
 - `transformOutput(code, id)`: Adds a custom transformer over jsx/tsx code produced by `civet.compile`. It gets passed the jsx/tsx source (`code`) and filename (`id`), and should return valid jsx/tsx code.
--- a/integration/unplugin/README.md
+++ b/integration/unplugin/README.md
@@ -84,12 +84,12 @@ interface PluginOptions {
 }
 ```
 
-- `emitDeclaration`: Whether to generate `.d.ts` type definition files from the Civet source, which is useful for building libraries. Default: `false`
+- `emitDeclaration`: Whether to generate `.d.ts` type definition files from the Civet source, which is useful for building libraries. Default: `false` (Requires installing `typescript`)
 - `typecheck`: Whether to run type checking on the generated code. Default: `false`.
-- `outputExtension`: Output filename extension to use. Default: `.civet.tsx`, or uses `.civet.jsx` if `js` is `true`.
-- `ts`: Mode of transpilation of TS->JS source code. Default: `civet`. Options:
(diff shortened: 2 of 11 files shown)


integration/unplugin/README.md
```diff
@@ -84,8 +84,12 @@ interface PluginOptions {
 }
 ```
 
-- `dts`: Whether to generate `.d.ts` type definition files from the Civet source, which is useful for building libraries. Default: `false`
+- `emitDeclaration`: Whether to generate `.d.ts` type definition files from the Civet source, which is useful for building libraries. Default: `false`
 - `typecheck`: Whether to run type checking on the generated code. Default: `false`.
 - `outputExtension`: Output filename extension to use. Default: `.civet.tsx`, or uses `.civet.jsx` if `js` is `true`.
-- `js`: Whether to transpile to JS or TS. Default: `false`.
+- `ts`: Mode of transpilation of TS->JS source code. Default: `civet`. Options:
```
edemaine, line 90: I'm also not sure what the best default is... maybe an `auto` mode that tries to `import('esbuild')`, and on failure, tries to `import('typescript')`, and on failure, uses Civet's transpiler? Alternatively, we could try to modify Civet's transpiler to emit a warning if it bumps into not-yet-implemented functionality, and try to use it first, and only on failure try the other systems. (But that would be another PR.)

Also the strings (`"civet"` etc.) should get quotes. Actually so are e.g. `".civet.tsx"`.

integration/unplugin/README.md
```diff
@@ -84,8 +84,12 @@ interface PluginOptions {
...
+  - `civet`: Use Civet's JS mode.
+  - `esbuild`: Use esbuild's transpiler. (Requires installing `esbuild`)
+  - `tsc`: Use the typescript compiler. (Requires installing `typescript`)
+  - `preserve`: Don't transpile TS code.
```
edemaine, line 94: Maybe mention some of limitations, like builders that don't support TS output, like Rollup. (But of course this mode will still work if the user doesn't use any TS features.)

integration/unplugin/README.md
```diff
@@ -84,8 +84,12 @@ interface PluginOptions {
 }
 ```
 
-- `dts`: Whether to generate `.d.ts` type definition files from the Civet source, which is useful for building libraries. Default: `false`
+- `emitDeclaration`: Whether to generate `.d.ts` type definition files from the Civet source, which is useful for building libraries. Default: `false`
 - `typecheck`: Whether to run type checking on the generated code. Default: `false`.
 - `outputExtension`: Output filename extension to use. Default: `.civet.tsx`, or uses `.civet.jsx` if `js` is `true`.
-- `js`: Whether to transpile to JS or TS. Default: `false`.
+- `ts`: Mode of transpilation of TS->JS source code. Default: `civet`. Options:
+  - `civet`: Use Civet's JS mode.
```
edemaine, line 91: We should probably mention limitations to this mode somewhere. For now we can just say "not all TS features supported", and later we can make a list (maybe under Compatibility) and link to it.

integration/unplugin/src/index.ts
```diff
@@ -13,49 +13,31 @@ import {
 } from '@danielx/civet/ts-diagnostic';
 import * as fs from 'fs';
 import path from 'path';
-import ts from 'typescript';
+import type { FormatDiagnosticsHost, Diagnostic, System } from 'typescript';
```
edemaine, line 16: Should we remove `typescript` as a dependency then? Or do we use it elsewhere?

integration/unplugin/src/index.ts
```diff
@@ -91,29 +73,37 @@ function resolveAbsolutePath(rootDir: string, id: string) {
...
   let fsMap: Map<string, string> = new Map();
   const sourceMaps = new Map<string, SourceMap>();
   let compilerOptions: any;
   let rootDir = process.cwd();
 
+  const tsPromise =
+    transformTS || options.ts === 'tsc'
+      ? import('typescript').then(m => m.default)
+      : null;
+  const getTS = () => tsPromise!;
```
edemaine, line 89: Not sure what `getTS` is for... why not just use `tsPromise!` directly? In fact, you do that at least once below already. 🙂

integration/unplugin/src/index.ts
```diff
@@ -238,26 +233,86 @@ const civetUnplugin = createUnplugin((options: PluginOptions = {}) => {
...
-      const jsonSourceMap = compiled.sourceMap.json(
-        path.basename(id.replace(/\.[jt]sx$/, '')),
-        path.basename(id)
-      );
+            compiled = { code: result.code, sourceMap: result.map };
+            break;
+          }
+          case 'tsc': {
+            const tsTranspile = (await getTS()).transpileModule;
+            const result = tsTranspile(compiledTS.code, { compilerOptions });
```
edemaine, line 285: If `transformTS` is set (e.g. typechecking is on), does this cause TypeScript to transpile *again*, or does it re-use the existing transpilation? Ideally we'd like the latter.

integration/unplugin/src/index.ts
```diff
@@ -238,26 +233,86 @@ const civetUnplugin = createUnplugin((options: PluginOptions = {}) => {
...
+          filename: id,
+          js: true,
+          sourceMap: true,
+        });
+      } else {
+        const compiledTS = civet.compile(rawCivetSource, {
+          filename: id,
+          js: false,
+          sourceMap: true,
+        });
```
edemaine, line 255: We shouldn't do this call when `options.ts === "civet"`: we don't want to run the Civet transpiler twice in this case. (Part of the point is for it to run faster! Also it's a little dicey whether the output of Civet is also valid input to Civet, as we don't necessarily preserve indentation.)

edemaine (commented):
Nice work! I think I like the new API. What do you think, @STRd6 ?

I wonder if we should support the old options (`js` and `tsc`) for backwards compatibility, or just go with fully breaking changes.

civet.dev/getting-started.md
```diff
@@ -158,9 +158,9 @@ esbuild.build({
   plugins: [
     civetPlugin({
       // Options and their defaults:
-      // dts: false,                     // generate .d.ts files?
+      // emitDeclaration: false,         // generate .d.ts files?
       // outputExtension: '.civet.tsx',  // replaces .civet in output
-      // js: false,                      // use Civet's TS -> JS transpiler?
+      // ts: 'civet',                    // use Civet's TS -> JS transpiler?
```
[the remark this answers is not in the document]
Mokshit06:
```js
civetPlugin({
  // Options and their defaults:
```
This should be mentioning the default values

integration/unplugin/README.md
Mokshit06: Maybe we can use civet's transpiler by default, and also show a warning when it's being used without being explicitly set?

integration/unplugin/src/index.ts
Mokshit06: We can remove it as a dependency, but we should add esbuild & typescript as dev dependencies for types locally.

integration/unplugin/src/index.ts
Mokshit06: Just so we don't have to use the non-null assertion everywhere. but now that i think of it, its probably better to remove it so we know that we are accessing something that might be null :)

integration/unplugin/src/index.ts
Mokshit06: With typechecking on, we only have typescript emit the declarations and not the transpiled source.
Generating declarations requires Typescript to have the entire program with imports resolved the file map, which we can only have in `buildEnd`. So this does cause Typescript to transpile _again_ as we can't reuse, but here it doesn't care about the diagnostics or the module resolution.

integration/unplugin/src/index.ts
```diff
@@ -238,26 +233,86 @@ const civetUnplugin = createUnplugin((options: PluginOptions = {}) => {
...
+            };
+            break;
+          }
+          case 'preserve': {
+            compiled = compiledTS;
+            break;
+          }
+          case 'civet':
+          default: {
+            compiled = civet.compile(compiledTS.code, {
```
esthedebeste, line 299: this JS compile should use the original civet code instead of the emitted typescript code. civet might output typescript that it cannot parse itself (or that is not equivalent). For example:
```ts
// ↓↓↓ input - valid civet code
:label for(const element of array) {
  if element === 5 then break label
  console.log(element)
}
// ↓↓↓ civet - valid emitted TS code
label: for (const element of array) {
  if (element === 5) break label;
  console.log(element);
}
// ↓↓↓ civet - invalid code.
({
  label: (() => {
    const results = [];
    for (const element of array) {
      if (element === 5) break label;
      console.log(element);
    }
    return results;
  })(),
});
```

integration/unplugin/src/index.ts
```diff
@@ -13,49 +13,31 @@ import {
...
 
-const formatHost: ts.FormatDiagnosticsHost = {
-  getCurrentDirectory: () => ts.sys.getCurrentDirectory(),
-  getNewLine: () => ts.sys.newLine,
-  getCanonicalFileName: ts.sys.useCaseSensitiveFileNames
-    ? f => f
-    : f => f.toLowerCase(),
-};
-
 export type PluginOptions = {
```
esthedebeste, line 21: I think it's good if the old `dts: boolean, js: boolean` is kept for a little bit, deprecated with a warning of course. `dts` would map plainly to `emitDeclaration`. `js: true` can map to `ts: "civet"`

esthedebeste (changes requested):

civet.dev/getting-started.md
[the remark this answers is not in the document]
edemaine: Ah, good point. Maybe the comment should change though:
```suggestion
      // ts: 'civet',                    // TS -> JS transpilation mode
```

integration/unplugin/src/index.ts
edemaine: Ah, I see. We could modify the compiler to be able to output both versions for this case; this just involves [running generate](https://github.com/DanielXMoore/Civet/blob/0efd56eb5fff8f35173a17623064279753409476/source/main.civet#L149) on the same parse tree with two different options.

(There's also still the double-transpilation issue mentioned below.)

integration/unplugin/src/index.ts
edemaine: Sounds good to me; it's really easy to still support them.

I like offering `js: true` maybe even long-term, because it corresponds to an option to the Civet compiler. (Someone was recently wondering why `inlineMap: true` didn't work. I'm not sure it's really useful for an unplugin, but it points toward mirroring Civet's compiler options.)

I think we could remove these options from the types, though; this is probably sufficient warning that there's a better way now.

STRd6:
Looks pretty good, I like the functionality of being able to choose the transform `"civet" | "esbuild" | "tsc" | "preserve"` for the emitted TypeScript and lazy loading the proper peer dependency.

Changing `dts` to `emitDeclaration` makes sense.

`js` as the name of the option feels more intuitive to me but maybe that's just because it's what I'm used to. I agree that `ts: "preserve"` is comparable to TypeScript's `jsx: "preserve"` option and makes sense from that perspective. I'm not too hung up on it.

I'm pretty sure we should keep `js: true` as an option at least for a little while, possibly with a deprecation warning. Maybe also for `dts` as well. I haven't used that option much but other people may.

Happy to merge this once @esthedebeste and @edemaine's feedback has been resolved.

integration/unplugin/src/index.ts
Mokshit06: Ah thanks for pointing out 🙂 That should've been `rawCivetSource` instead. Missed that point in @edemaine's comment

integration/unplugin/README.md
```diff
@@ -84,12 +84,12 @@ interface PluginOptions {
...
-  - `civet`: Use Civet's JS mode.
-  - `esbuild`: Use esbuild's transpiler. (Requires installing `esbuild`)
-  - `tsc`: Use the typescript compiler. (Requires installing `typescript`)
-  - `preserve`: Don't transpile TS code.
+- `outputExtension`: Output filename extension to use. Default: `".civet.tsx"`, or uses `".civet.jsx"` if `js` is `true`.
+- `ts`: Mode of transpilation of TS->JS source code. Default: `"civet"`. Options:
+  - `civet`: Use Civet's JS mode (Not all TS features supported)
+  - `esbuild`: Use esbuild's transpiler (Requires installing `"esbuild"`)
+  - `tsc`: Use the typescript compiler (Requires installing `"typescript"`)
+  - `preserve`: Don't transpile TS code (Some bundlers like esbuild and vite can handle TS directly. Note that this would require additional plugins for bundlers that can't handle TS)
```
edemaine, line 94:
```suggestion
  - `preserve`: Don't transpile TS code (Some bundlers like esbuild and Vite can handle TS directly. Note that this would require additional plugins for bundlers that can't handle TS, such as Rollup.)
```

Mentioning Rollup which isn't obvious because there is a TypeScript plugin, but (as you know) it doesn't work with additional plugins like ours.

Also, "vite" → "Vite".

edemaine:
Hmm, I notice there's also a failure to build in CI.

integration/unplugin/README.md
```diff
@@ -84,12 +84,12 @@ interface PluginOptions {
...
-- `ts`: Mode of transpilation of TS->JS source code. Default: `civet`. Options:
-  - `civet`: Use Civet's JS mode.
-  - `esbuild`: Use esbuild's transpiler. (Requires installing `esbuild`)
-  - `tsc`: Use the typescript compiler. (Requires installing `typescript`)
-  - `preserve`: Don't transpile TS code.
+- `outputExtension`: Output filename extension to use. Default: `".civet.tsx"`, or uses `".civet.jsx"` if `js` is `true`.
+- `ts`: Mode of transpilation of TS->JS source code. Default: `"civet"`. Options:
+  - `civet`: Use Civet's JS mode (Not all TS features supported)
+  - `esbuild`: Use esbuild's transpiler (Requires installing `"esbuild"`)
+  - `tsc`: Use the typescript compiler (Requires installing `"typescript"`)
```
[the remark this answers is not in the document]
Mokshit06: Oops, yeah I understood the intent but ended up quoting other things 🤦

integration/unplugin/README.md
Mokshit06: Hm maybe we should reword this statement if we are mentioning rollup, because in case of rollup it infact _won't_ work with even an additional plugin

Mokshit06:
The CI failure should be fixed now. I think it was caused by a missing `break` inside switch stmt.

edemaine (commented):
I tested all three modes. With `ts: "esbuild"` and a [Solid project](https://github.com/edemaine/character-layout/tree/unplugin), I found that esbuild was processing JSX with default options. Given that we are outputting JSX, I think it's safe to set `jsx: "preserve"` as I did in the latest commit.

With `ts: "esbuild"` and Vite (same project linked above), I get these error messages, suggesting something wrong with the sourcemaps:
```
Sourcemap for "src/index.civet.jsx" points to missing source files
Sourcemap for "src/App.civet.jsx" points to missing source files
```

About option naming:

I noticed that tsup [uses the `dts` option](https://tsup.egoist.dev/#copy-files-to-output-directory), which gives me some pause on the choice to switch to `emitDeclaration`.

And we still need to decide on `js` vs. `ts` naming. This may be linked... if we use `dts`, which is the name of a generated file, I can see arguing for the return to `js`. It'd also less disruptive. Then for backward compatibility, we'd just map `false` to `"preserve"` and `true` to `"civet"`.

Mokshit06:
Agree on the JSX part. We need to set `module` to `nodenext` for fixing the CI failure.

Re naming: I think we should try to match Typescript's config naming/behaviour rather than tsup. Same for `js` vs `ts`. Providing an option of `js: 'preserve'` doesn't make much sense. The `ts` option makes it similar in concept to tsconfig's `jsx` option where the options are `"preserve" | "react" | "react-jsx"...`.

edemaine (approved):
`"preserve"` is a good argument for `ts` over `js`. And I don't have a strong feeling on `dts` vs. `emitDeclaration`. @STRd6 do you want to make a decision here? Or I'd be happy to brainstorm more, e.g. in Discord.

Otherwise I think this is good to go. Regarding the esbuild sourcemap issue, maybe it can be done by modifying the sourcemap to change the filename (even via a regex replacement?). Sorcery is interesting but is maybe overkill for this purpose, as we already have lots of sourcemap processing code. But yeah, separate PR seems reasonable.

Mokshit06:
I don't have any strong opinions on the `dts` flag. Any one of `dts`, `emitDeclaration`, or `declaration` do convey the intent clearly. 

We can change the filename, but that would lead to incorrect sourcemaps because we are doing 2 transformations: civet -> ts -> esbuild. So we'll need to merge the sourcemaps produced by civet->ts step, and the one by ts->esbuild step for correct mapping.

PR #815, opened 2023-11-22
Fix Promise<void> in non-async function

### LINKED ISSUE #814: Promise<void> in non-async function
I have a [real-world example](https://github.com/edemaine/character-layout/blob/35fe01da44dee49f968fe71ef3f3dd7ac991a254/src/App.civet#L121-L138) where I have a function with return value `Promise<void>`, yet I want implicit return, because I construct the `Promise` explicitly via `new Promise`. This code breaks with modern Civet which turns off implicit returns when the return value is `Promise<void>`.

I think a `Promise<void>` return value should turn off implicit returns only for functions that are `async`, either explicitly or implicitly because of `await`. This function isn't `async` and that's causing a problem.

edemaine:
Fixes #814

We were setting `modifier.async` wrong in two places. 😅

### DIFF
--- a/source/parser.hera
+++ b/source/parser.hera
@@ -1603,8 +1603,8 @@ FunctionSignature
       async,
       generator,
       modifier: {
-        async: !async.length,
-        generator: !generator.length,
+        async: !!async.length,
+        generator: !!generator.length,
       },
       block: null,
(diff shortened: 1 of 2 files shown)


STRd6 (approved):
Thanks!

PR #817, opened 2023-11-23
Assignments in & functions

edemaine:
As discussed in Discord, it's nice to be able to assign to properties in a pipeline, especially fat, as an alternative to `..` cascade notation.  The DOM manipulation example at the top of reference.md is particularly compelling.

### DIFF
--- a/test/pipe.civet
+++ b/test/pipe.civet
@@ -289,6 +289,16 @@ describe "pipe", ->
     """
 
     testCase """
+      multi-line with assignments
+      ---
+      document.createElement('div')
+      ||> .className = 'civet'
+      ||> .appendChild document.createTextNode 'Civet'
+      ---
+      let ref;(ref = document.createElement('div')).className = 'civet',ref.appendChild(document.createTextNode('Civet')),ref
+    """
+
+    testCase """
       ref with implicit return
       ---
       ->
--- a/civet.dev/reference.md
+++ b/civet.dev/reference.md
@@ -524,6 +524,12 @@ url |> fetch |> await
 |> callback
 </Playground>
 
+<Playground>
+document.createElement('div')
+||> .className = 'civet'
+||> .appendChild document.createTextNode 'Civet'
+</Playground>
+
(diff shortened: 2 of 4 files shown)


test/pipe.civet
```diff
@@ -288,6 +288,16 @@ describe "pipe", ->
...
     """
 
+    testCase """
+      multi-line with assignments
+      ---
+      document.createElement('div')
+      ||> .className = 'civet'
+      ||> .appendChild document.createTextNode 'Civet'
+      ---
+      let ref;(ref = document.createElement('div')).className = 'civet',ref.appendChild(document.createTextNode('Civet')),ref
```
bbrk24, line 298: Intuitively I would expect something more like

```js
const ref = document.createElement('div');
ref.className = 'civet';
ref.appendChild(document.createTextNode('Civet'));
```

which would also be more TypeScript-friendly. I can't imagine TS likes the isolated `let ref`.

Is there a reason for it being this instead?

edemaine: [TypeScript is happy with it](https://www.typescriptlang.org/play?#code/DYUwLgBATiBmDcAoAFDWEC8EAmB7AxgK4C2IAdmAHT4wCGYIAoqKRcgOTYCWAbuwJT9qwWgGdRAOVqlMEdvl7h2AGkRpKtAA6by2AMIALLsGzI8RVlRoh6IACogAHmAm5sIDnsVgB-VWkQAekCAPQB+IA).

We use this technique a bunch, and this isn't related to this PR, but here's the motivation:

```js
if bool or (document.createElement('div') ||> ...)
↓↓↓
let ref
if(bool || (ref = document.createElement('div'), ..., ref)) {
```
```

We want to short-circuit before doing the `ref` assignment.

STRd6 (approved):
Looks good, I’ve been meaning to add something like this for a while!

edemaine:
Happy Thanksgiving! And thanks for the reviews 😊

Issue #824, opened 2023-11-30
x<y>z


edemaine:
`x<y>z` currently fails to parse, because `x<y>` parses as a unary expression (type arguments), preventing the `<` from being recognized as a binary operator.  By contrast, `x<y>z` [parses in TypeScript](https://www.typescriptlang.org/play?#code/B4HgngfAXkA) as `x < y > z` (not that that means anything in JavaScript, but it does in Civet).

I think we should add some kind of assertion after type arguments that allows for tight comparisons like this. I'm not sure exactly what the right assertion is though... maybe there's one in the TypeScript source?

STRd6:
Even simpler `x<y` is currently a parse error... 🙀

PR #830, opened 2023-12-05
Fix snug `x<y`

STRd6:


source/parser.hera
```diff
@@ -2970,10 +2969,10 @@ PrivateIdentifier
 # NOTE: Merged into MethodDefinition
 
 # NOTE: Allow arbitrary whitespace before regular assignment,
-# but require nonnewline whitespace before operator assignment.
+# but only allow non-newline whitespace before operator assignment.
 WAssignmentOp
   __ AssignmentOp
-  _ OperatorAssignmentOp
+  _? OperatorAssignmentOp
```
edemaine, line 2975: I think this change allows something like `x[1]xor= true` or `x[1]min= x[0]` which doesn't look great... but I guess it's OK, looking at the new test case. Certainly can't mean anything else.

source/parser.hera
```diff
@@ -2991,23 +2990,23 @@ AssignmentOp
 # This is separate from AssignmentOp because it only works in certain contexts
 # (in particular, not at the beginning of a line).
 OperatorAssignmentOp
-  Xor "=" &Whitespace _? ->
+  Xor "=" _? ->
```
edemaine, line 2993: This allows `x xor=y` which... is probably fine, just pointing it out.

source/parser.hera
```diff
@@ -3110,10 +3109,7 @@ BinaryOpSymbol
     }
   "<<"
   "«" -> "<<"
-  # NOTE: Avoid matching JSX opening tag by requiring non-identifier character
-  # (e.g. whitespace) after "<".  This does forbid 1<2 or x<y.
-  /<(?!\p{ID_Start}|[_$])/ ->
-    return "<"
+  "<"
```
edemaine, line 3112: Kind of surprised this works! But must be an old way of avoiding ambiguity.

edemaine (approved):
Looks good!

I don't understand the Coveralls error... It's says a line in e.g. `xnor=` is now uncovered, but it should be in tests. Unless it's `a xnor=b` that's uncovered; could add a test.

source/parser.hera
STRd6: It originally caused some trouble with `class X extends A<B>` but switching to enforcing `ExtendsTarget` as a `LeftHandSideExpression` fixes it by requiring parens around expressions.

test/binary-op.civet
```diff
@@ -79,6 +79,14 @@ describe "binary operations", ->
...
   """
 
+  testCase """
+    snug <
+    ---
+    x<y
+    ---
+    x<y
+  """
+
```
edemaine, line 89: Perhaps should also add `x<y>z` from #824?

PR #834, opened 2023-12-06
Generalize pin expressions to allow x.y and ^x.y

### LINKED ISSUE #821: Pattern matching with an enum value
```ts
enum A {
  B=0,
  C=1
}

const x:A = A.B

switch x
  B
    => console.log "B"
  C
    => console.log "C"
```

I though the code above should be parsed correctly, but in playground, it shows error like this:

```ts
enum A {
  B=0,
  C=1
}

const x:A = A.B

switch x
  B
  ^ ParseError
```

edemaine:
Fixes #821 (pattern matching on enum values) by allowing two new patterns:

* General pinned expressions: `^x.y`, `^x()`, and more generally `SingleLineExpressionWithIndentedApplicationForbidden` expressions in pins. I can't see why pins need to be identifiers; they can be any expression. I tried to make something that resembles one-liners, though not totally sure I got it perfect.
* `x.y` (a new `ActualMemberExpression` that has at least one member access) acts as the pinned expression `^x.y`. Currently no calls are allowed, which is maybe not ideal. Suggestions for generalization welcome.

### DIFF
--- a/source/lib.civet
+++ b/source/lib.civet
@@ -2746,7 +2746,7 @@ function getPatternConditions(pattern, ref, conditions): void {
       conditions.push([
         ref,
         " === ",
-        pattern.identifier,
+        pattern.expression,
       ])
       break
     case "Literal":
--- a/source/parser.hera
+++ b/source/parser.hera
@@ -986,6 +986,13 @@ MemberExpression
     }
     return $1
 
+ActualMemberExpression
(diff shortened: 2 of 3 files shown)


STRd6 (approved):
It looks like this includes some of the completion fixes as well.

Update to pins looks good!

edemaine:
Oops, thanks for catching that! Removed the accidental `lsp` changes.

Issue #805, opened 2023-11-15
Inserting brackets around selection doesn't work with the vscode extension


unlessgames:
Normally when you select something in VSCode and press `(` the selection gets surrounded by `()` instead of being replaced with `(`. The same thing happens with curly braces and square brackets as well. 

This feature seems to work for every kind of file except for .civet when the extension is installed so I assume the extension overrides the default behaviour in some way it shouldn't.

As a sidenote for those looking for a somewhat clunky workaround, you can configure this snippet for civet files and assign it to a separate shortcut, you'd need to do this for each type of bracket which is not ideal but which is still better than nothing...

``` 
"surroundParens": {
    "prefix": "(",
    "scope":"civet",
    "language":"civet",
    "body": ["($SELECTION)$0"],
    "description": "Surround with parens"
  }
```

and the keybinding 

```
{
  "key": "alt+9",
  "command": "editor.action.insertSnippet",
  "when": "editorTextFocus && editorHasSelection",
  "args": {
    "name": "surroundParens",
    "langId": "civet"
  }
}
```

bbrk24:
Same thing with quote marks as well.

edemaine:
Should be fixed in extension version 0.3.10!

unlessgames:
Maybe it would be nice to also enable this functionality for angle brackets `< >` too as they are common brackets in both typescript and JSX.

I found it is much more common that I want to fix a forgotten `Promise<_>` or something similar instead of replacing the selection with `<`.

PR #835, opened 2023-12-06
Add bracket/comments matching

bkuri:
Should help towards fixing #805. Haven't tested it extensively, but it should be a solid start.


STRd6:
Thanks!

PR #843, opened 2023-12-10
Fix special relational operator precedence

### LINKED ISSUE #842: `objectIs` disturbs operator precedence?
```
Math.floor(a / b) % 2 is 0
Math.floor(a / b) % 2 === 0
```
generates
``` ts
Math.floor(a / b) % 2 === 0;
Math.floor(a / b) % 2 === 0;
```
but,
```
"civet objectIs"
Math.floor(a / b) % 2 is 0
Math.floor(a / b) % 2 === 0
```
becomes
``` ts
var is: {
  <B, A extends B>(a: A, b: B): b is A;
  <A, B>(a: A, b: B): a is A & B;
} = Object.is as any;
Math.floor(a / b) % is(2 as const, 0 as const);
Math.floor(a / b) % 2 === 0;
```

which results in unexpected order of operation. Is this an intended behavior?

edemaine:
Fixes #842

I need to add some tests, but I think this should fix precedence of operators like `is in` and `is` in objectIs mode (#842), placing them at the same level as equality comparisons as in [this table](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Operator_precedence#table). I'm not totally sure this is the best approach, but seems like the easiest fix for now...

Perhaps we should more generally have a precedence table that this can refer to, which would help fix the precedence of `xor`.


STRd6:
Is this missing a test?

edemaine:
Yes, sorry, had run out of time yesterday. Added tests and fixed some bugs.

STRd6 (approved):

PR #855, opened 2023-12-15
Fix #850: Wrap parens around thick pipes with refs in declarations

### LINKED ISSUE #850: Thick pipe needs parens when used in declaration
```civet
x := y * 2 ||> console.log
```

Actual: 

```js
let ref;
const x = console.log((ref = y * 2)),
  ref;
```

Expected:

```js
let ref;
const x = (console.log((ref = y * 2)),
  ref);
```

STRd6:


source/lib.civet
```diff
@@ -2401,6 +2418,26 @@ function processDeclarationConditionStatement(s: IfStatement | IterationStatemen
...
+  // @ts-ignore
+  .forEach ({ bindings }: DeclarationStatement) =>
+    bindings?.forEach (binding) =>
+      { initializer } := binding
+      return unless initializer
+
+      exp := initializer[2]
+      // Wrap thick pipeline 'x ||> f' => 'f(x),x' with parentheses when inside an initializer
+      if exp?.type is "PipelineExpression"
+        if exp.children.at(-2) is ","
```
edemaine, line 2434: I'm not totally sure this condition is the right one... `exp.children.includes(",")` would be more intuitive to me. But perhaps it's guaranteed that each thing between commas is a single object, in which case this is correct.

edemaine (approved):

source/lib.civet
STRd6: It's related to this line in `processPipelineExpression` where we push the comma then the return ref. https://github.com/DanielXMoore/Civet/blob/3115ff3fc694077bd8d6b3f5ffc4f56b8a0f26bf/source/lib.civet#L3221

PR #859, opened 2023-12-16
Support implicit .civet importing unless implicitExtension: false

edemaine:
Most bundlers support implicit `.ts`/`.js`/etc. extensions. It's reasonable to expect that `.civet` will be implicit as well, so this PR adds it as default behavior.  @Mokshit06 suggested there be an option to disable this behavior, though, so you can force explicit extensions via `implicitExtension: false`.

I've tested this in esbuild, Vite, Rollup, and Webpack.

### DIFF
--- a/integration/unplugin/src/index.ts
+++ b/integration/unplugin/src/index.ts
@@ -64,12 +64,14 @@ function tryFsResolve(file: string): string | undefined {
   return undefined;
 }
 
-function resolveAbsolutePath(rootDir: string, id: string) {
-  const resolved = tryFsResolve(path.join(rootDir, id));
-
-  if (!resolved) return tryFsResolve(id);
-
-  return resolved;
+function resolveAbsolutePath(rootDir: string, id: string, implicitExtension: boolean) {
+  const file = path.join(rootDir, id);
+  // Check for existence of resolved file and unsolved id,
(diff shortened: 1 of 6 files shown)


STRd6 (approved):
Looks like a good and more convenient default. I also agree that the option to disable implicit extensions is valuable especially in very large projects where performance could be impacted.

integration/unplugin/src/index.ts
```diff
@@ -215,15 +223,22 @@ const civetUnplugin = createUnplugin((options: PluginOptions = {}) => {
     },
     resolveId(id, importer) {
       if (/\0/.test(id)) return null;
-      if (!isCivet(id) && !isCivetTranspiled(id)) return null;
 
       id = cleanCivetId(id);
       const absolutePath = path.isAbsolute(id)
         ? resolveAbsolutePath(rootDir, id)
```
Mokshit06, line 229: I think it would fail here for absolute imports with implicit civet extension. I'm not sure if we want to account for that case though?

edemaine: Nice catch! I've pushed a patch, which I tested and seems to work for both types of absolute paths (at least in Vite). Could you review?

Mokshit06: Yep this looks good!

PR #860, opened 2023-12-16
handleHotUpdate to fix Vite HMR (watch in dev mode)

edemaine:
Fix Vite dev mode file watching (#773).

This was a challenge to get working. I confirmed that it works on Windows when modifying `main.civet` or the dependent `module.civet` in `integration/unplugin/examples/vite`.


STRd6 (approved):
Awesome!

PR #871, opened 2023-12-18
Update docs style

STRd6:
- Removed border radius
- Added shadows
- Fixed light mode code legibility


STRd6:
image

image

image

edemaine:
Sorry for the delay. The new style looks great! I feel kind of sad being in dark mode and not seeing the shadows though. 😄 (except on the three buttons on the front page — I can see the shadows there because of a lighter background)

I find these links a little low contrast:

image

So I might tweak those colors to be a bit brighter.

PR #878, opened 2023-12-18
svg backgrounds

STRd6:


edemaine:
Why palm trees? (Also 85kB.)

I find the square dots a little weird, but maybe better than the current blank space. Oh, I think it's mainly because this is my usual browser width:

image

It looks better when the browser is wider.

Related: Why do we force the width of the page, instead of being responsive?  A minimum width seems reasonable, but I'm not sure why we should have a background at all...

STRd6:
> Why palm trees? (Also 85kB.)

I'm going for a sort of vaporwave / art deco jungle aesthetic.

final (2)

We could optimize our non-svg images and save ~1MB before we worry about the svg size :)

> Related: Why do we force the width of the page, instead of being responsive? A minimum width seems reasonable, but I'm not sure why we should have a background at all...

I'm not sure what this means, the square pattern is centered in the content so remains symmetric for all widths. The backgrounds disappear for screens <960px though we could improve the responsive design eventually to add more brand to it as well.

edemaine:
Ooh, pretty. I'm definitely a fan of the colors. For non-geometric imagery (palm trees), I guess I was hoping for something Civet-related. Some ideas:
* Instead of dots, we could have coffee beans?
* Instead of palm trees, we could have a cute Civet?

I was asking why we cap the width, instead of letting it fill the browser width, like so:

image

But now that I hacked that in, I can see why — it's pretty hard to read.

STRd6:
We should have a Civet and coffee cups or beans in addition to rather than instead of. Eventually we can have a rich jungle backdrop. Maybe have a Civet with a cup of coffee on the 'Tip' card and say "Civet sez:" :P

edemaine (approved):

unlessgames:
Opinion alert!

While the language itself aims for simplicity, the theme of "vaporwave/art deco/jungle/cosy cute civet" tries to be many things at once which I find a bit contradictory, as an obvious comparison the [page for coffeescript](https://coffeescript.org/#overview) communicates the concise and simplistic nature of the language very well. 

Maybe it would be better to go with a less saturated and more earthy color palette that resembles the shades of a civet's fur, coffee and the jungle better, I can see how art deco could be brought into this to aim for a kind of 20s exotic coffee label vibe but I can't find the vaporwave and loud pink and purple relevant. Overall, style is quite subjective so don't take it the wrong way, just sharing my opinion.

Civets and coffee are obvious elements that can't be separated from the language and palm trees come *naturally* with civets (as they live around and feed on the juice of palm trees afaik). Also, the pattern of civets' fur is rather interesting visually so it might be a good choice to capitalize on that somehow before reaching for something unrelated like vaporwave or art deco. Personally I find the civet mascot much more approachable and inviting (much like civet the language) than the somewhat cold nature of the vaporwave palette and the luxury art deco feels.

PR #880, opened 2023-12-18
Update font; tagline

STRd6:
image

Tagline could still use some work.


civet.dev/public/fonts/Breamcatcher.woff2
edemaine, line 1: Should we delete Bangers?

edemaine (approved):
Maybe "Code More With Less in a Superset of TypeScript" would be a bit better for the tagline?

I'm not sure about "new millennium" given that TS was introduced in 2012. 🙂 But I wasn't sure about the previous subtitle either.

unlessgames:
I find the new font to not be very legible and its stretched shape makes it feel like it doesn't have enough weight compared to the tagline below. Personally I liked the comic book vibes of the old one as it went better with the logo and it was easier to read. If art deco is the new direction here are some fonts that might be considered

* [engebrechtre](https://www.1001fonts.com/engebrechtre-font.html?text=Civet%20Playground%20Reference%20TypeScript) as a less quirky art deco font that is more readable
* [righteous](https://www.1001fonts.com/righteous-font.html) also more readable and less common
* [gurindam](https://www.1001fonts.com/gurindam-font.html?text=Civet%20Playground%20Reference%20TypeScript) is more art nouveau than art deco and quite quirky but "Civet" in particular looks pretty nice
* [baloo](https://www.1001fonts.com/baloo-font.html?text=Civet%20Playground%20Reference%20TypeScript) a clean and friendly modern font, not very art deco but nice

Also, just having some fun with taglines

>Type less, code more

>Take the typing out of TypeScript

>TypeScript digested

>A more digestable TypeScript

>Natural script processing reimagined

>A naturally sweet flavor of TypeScript

>A smooth superset of TypeScript

>A familiar language for the web

STRd6:
> More types, less typing

PR #886, opened 2023-12-19
Stricter unary operators (before &)

### LINKED ISSUE #885: Strange input allowed in single-argument function shorthand
After looking at [the fun uses of `await` and `delete`](https://civet.dev/reference#single-argument-function-shorthand) that we get after #883, I found that civet accepts this kind of input
```js
x.filter not.condition
x.forEach delete.old

↓↓↓

x.filter(($) => !$.condition);
x.forEach(($1) => delete$1.old);
```
In my opinion either this should be invalid syntax or there should be a space after delete

This trick doesn't work with `await`:
```js
x.map await.json()
                 ^ ParseError
```

edemaine:
Fixes #885

Also some more docs examples

### DIFF
--- a/civet.dev/reference.md
+++ b/civet.dev/reference.md
@@ -748,7 +748,8 @@ x.map &.name
 x.map &.profile?.name[0...3]
 x.map &.callback a, b
 x.map +&
-x.map await &.json()
+x.map typeof &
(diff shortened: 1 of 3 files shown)


STRd6 (approved):
Thanks!

PR #896, opened 2023-12-27
as! T

STRd6:


edemaine (approved):
I confirmed that this is not valid TS syntax; it says `'!' at the end of a type is not valid TypeScript syntax.`.

I still think `as!` makes a bit more sense, thinking of it as a new operator (and then it's just `as!` → `as unknown as`), but I'm also OK with this.

STRd6:
I agree that `as!` is probably better since it is better localized to the operator and will conserve the suffix `T!` in case of future TS changes or as a possible separate language feature in the future.

source/parser.hera
```diff
@@ -386,7 +386,15 @@ UnaryWithoutParenthesizedAssignmentBody
 
 UnaryPostfix
   QuestionMark
-  ( __ ( As / Satisfies ) Type )+ -> { ts: true, children: $0 }
+  TypePostfix+ -> { ts: true, children: $0 }
+
+TypePostfix
+  _:ws As:as ExclamationPoint?:ex Type:type ->
+    if (ex) {
+      return [" as unknown", ws, as, type]
```
edemaine, line 394: Not sure it's worth it, but `" as unknown"` could take `$loc` from the ExclamationPoint.

edemaine (approved):

PR #898, opened 2023-12-28
++ concat operator

edemaine:
As discussed in Discord. `++` is concat in Haskell and Elixir.

Also Unicode version ⧺


STRd6 (approved):
Looks good, thanks!

PR #899, opened 2023-12-29
++= concat assignment

edemaine:
Add `++=` assignment operator that acts like `concat`: multi-push if given an array, single push if not. Thanks to Beberka for pointing out the [isConcatSpreadable](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol/isConcatSpreadable) interface.

Here's [verification of the `concatAssign` typing](https://www.typescriptlang.org/play?#code/G4QwTgBAxg9gdlEAXAggZzQSwOZwFwQA8AQgDQQoQCmAHklXACZoQDeADgK5oAWBAFEh6Y0BFOQBGBYgEoIAXgB8EYDEyMAvhAA+EfsWp0GzCJzgBrODADucANoBdCAH42XXgKEix5AHT+pCFkFZVV1LQI4KmAqMBlFfgAbHlEKcjAU6TklCgU9ZLR0lOzlfn4MlhBKuABPGWdfOwBlGoBbCRhE3xEAYXhEJCb2MCoQRhAJRKonVySUiCqF2plfdx5fEHZ2RJq5wogKheq6iAECo6W61e4ecuLyApkAbgAoF9gEZHQsXH47AEZSAAmUgAZgc5AALDJ3v0vhgcHA-oCQeDyHZIaQAKwOGEfAbfRHI4FgiEQABEEjG5LxcNQCN+AJJaIgdkp1NxQA).

I imagine the TypeScript could use less `any`s, but it's tricky because there are two cases.

### DIFF
--- a/civet.dev/reference.md
+++ b/civet.dev/reference.md
@@ -388,6 +388,19 @@ a ++ b ++ c
 [1,2,3] ⧺ rest
 </Playground>
 
+You can use `++` to concatenate arrays or strings,
+or your own types by providing a `concat` method.
+Remember that
+[`Array.prototype.concat`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/concat)
+appends a single item unless it is an array
+(or has the [`Symbol.isConcatSpreadable` property](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol/isConcatSpreadable)),
+in which case it flattens it into the target array.
+Civet's assignment operator behaves the same:
+
+<Playground>
+a ++= b
+</Playground>
+
 ### Assignment Operators
 
 <Playground>
--- a/source/lib.civet
+++ b/source/lib.civet
@@ -1235,24 +1235,34 @@ function replaceBlockExpression(node: BlockStatement, child: ASTNodeBase, replac
  * which is useful for working with e.g. the `expressions` property.
  * Returns -1 if `child` cannot be found.
  */
-function findChildIndex(parent, child) {
-  const children = Array.isArray(parent) ? parent : parent.children
-  const len = children.length
-  for (let i = 0; i < len; i++) {
-    const c = children[i]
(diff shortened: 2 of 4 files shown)


bbrk24:
There's a reason I wrote `Array::push.apply(lhs, rhs)` rather than `lhs.push(...rhs)`. iinm, only the former works for code like this:

```civet
arr ++= 0: foo,
  1: bar,
  length: 2,
  [Symbol.isConcatSpreadable]: true
```

The spread operator also requires a `Symbol.iterator`, I believe, while `Function::apply` works with any array-like. `Array::concat` also works with any array-like that has `Symbol.isConcatSpreadable` set to true.

bbrk24:
If we want this to work for any `push` method rather than enforcing `Array::push`, I guess it would have to be `lhs.push.apply(lhs, rhs)`, right?

edemaine:
The current solution seems advantageous in some cases: it *only* requires the `rhs` to offer `Symbol.iterator`. It's true that `apply` works for array-likes (offering keys of `length` and `0`, `1`, ...), but it doesn't work for straight iterators. Which is better?

[NodeList](https://developer.mozilla.org/en-US/docs/Web/API/NodeList), for example, supports both, so this doesn't help narrow things down.

edemaine:
Ah, I see that `Array.prototype.concat` expects an array-like, so if we're aiming to be compatible with that, `lhs.push.apply(lhs, rhs)` probably makes sense...

edemaine:
On further testing, `array.push.apply` does work with an iterator argument, so this seems superior. (My confusion stems from `console.log.apply` not working with an iterator argument... but presumably that's because `console` is special?)

bbrk24:
> My confusion stems from console.log.apply not working with an iterator argument... but presumably that's because console is special?

More likely Array is special. Many methods on `Array.prototype`, such as `indexOf`, are intentionally designed to work on more types. (This is why CoffeeScript does `[].indexOf.call` for its `in` operator.) Regardless, it doesn't matter as long as this function works.

STRd6 (approved):
👏

PR #904, opened 2023-12-31
Optional function return types

edemaine:
Follow-up to #902 for function return values, plus some refactoring

### DIFF
--- a/source/parser.hera
+++ b/source/parser.hera
@@ -6520,14 +6520,14 @@ TypeIndex
 # parse both (?!).  For simplicity, we allow either in all cases.
 # In some cases (e.g. let/const), we transpile them away later.
 TypeSuffix
-  QuestionMark?:optional _? Colon Type:t -> {
+  _? QuestionMark?:optional _? Colon Type:t -> {
     type: "TypeSuffix",
     ts: true,
     optional,
     t,
     children: $0,
   }
-  QuestionMark:optional _? -> {
+  _? QuestionMark:optional _? -> {
     type: "TypeSuffix",
     ts: true,
     optional,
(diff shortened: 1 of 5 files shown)


source/parser.hera
```diff
@@ -6536,18 +6536,22 @@ TypeSuffix
   # TypeScript has a special error for ! without : ("Declarations with definite
   # assignment assertions must also have type annotations.") but we parse it
   # so that the user can get this more useful error message.
-  NonNullAssertion _? (Colon Type)?:ct ->
+  _? NonNullAssertion _? (Colon Type)?:ct ->
```
STRd6, line 6539: I don't necessarily want to allow whitespace before the non-null assertion since it will crowd out a possible future `!` binary op.

test/types/function.civet
```diff
@@ -74,6 +74,17 @@ describe "[TS] function", ->
     }
   """
 
+  testCase """
+    longhand optional return
+    ---
+    function x(a: number, b?: number) ?: number
```
STRd6, line 80: I'm not as happy with this syntax as with the `let x?: T` syntax.

`function x(a: number, b?: number): number?` seems preferable. TypeScript even seems to be indirectly pushing in that direction as well:
image

I'm happy to merge this to gain more practical usage and experimentation with the expectation that we may change or deprecate as we learn more and consolidate with #144

STRd6 (approved):

test/types/function.civet
edemaine: I think I agree with you. But I do think this syntax is symmetric, and maybe the user would prefer one consistent style or another. 

#144 offers an alternative syntax for all optional types, so presumably someone could switch wholesale to `?` next to the type. Before or after depends on what we think is best. Also this depends on whether we think `null` should be part of `?` as a type operator.

My thought was, while we figure out #144, this syntax seems unambiguous so we can do it now. (That said, eager to hear your thoughts on which way #144 should go.)

PR #905, opened 2024-01-01
Conditional types via if/unless

edemaine:
As suggested by @esthedebeste on Discord, this PR allows for conditional (ternary) types to be given by `if`/`else` and `unless`/`else` expressions, with the usual support for then conditions (with condition delimited by parentheses, or then clause indented, or explicit `then`).

The `else` clause is optional, with a default value of `never`, which I find to be very common when writing conditional types.

Also added support for indenting a type specification in `let`.

### DIFF
--- a/civet.dev/reference.md
+++ b/civet.dev/reference.md
@@ -1458,6 +1458,19 @@ function f(x?: string)?: string
   x
 </Playground>
 
+### Conditional Types
+
+TypeScript's ternary types can be written using `if`/`unless` expressions,
+with optional `else` blocks:
+
+<Playground>
+let verb:
+  if Civet extends Animal
+    if Civet extends Cat then "meow"
+  else
+    string
+</Playground>
+
 ### Import
 
 <Playground>
--- a/civet.dev/reference.md
+++ b/civet.dev/reference.md
@@ -1469,6 +1469,8 @@ let verb:
     if Civet extends Cat then "meow"
   else
     string
+let breed: unless Civet extends Animal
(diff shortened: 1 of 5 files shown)


STRd6 (approved):
Thanks for this! Glad to see it was relatively simple to implement.

esthedebeste:
Does typescript support the unless output; `!(a extends b)`? I haven't seen a `!` unary in type context before.

bbrk24:
> Does typescript support the unless output; `!(a extends b)`?

[It does not](https://www.typescriptlang.org/play?target=8#code/C4TwDgpgBAYg9nAPAFQHxQLxQIQAplQQAewEAdgCYDOUVwATgJZkDmAlFAPxQDkAhjygAuXgCMeAbgCwAKCA).

edemaine:
Thanks for catching this! I fixed `unless` to instead swap the then/else clauses.

STRd6:
> Thanks for catching this! I fixed `unless` to instead swap the then/else clauses.

There's probably slightly more that needs to be done to support unless since extends is a `<=` relationship.

For example

```

```

STRd6:
ty

edemaine:
I don't think so; `if !x then y else z` should be equivalent to `if x then z else y` in all cases.

In particular, `A extends B ? B extends A ? false : false : true` is equivalent to `A extends B ? false : true` because the first two cases are both `false`.

STRd6:
Makes sense, I overthought things and confused myself!

PR #908, opened 2024-01-02
`T?` → `T | undefined`; `T??` → `T | undefined | null`

edemaine:
Fixes #144 according to recent proposal in Discord. Do we like it? `T?` → `T | undefined` is nicely consistent with TypeScript's other optional types, but `T??` → `T | undefined | null` corresponds to our runtime `x?` operator. Maybe the closest we can get, though.

Also remove some excess parentheses (finally) in this type manipulation.

I notice that our rules are of the form `TypeUnary` but our AST node `type` strings are of the form `UnaryType`. Not sure whether we should fix this...

### DIFF
--- a/civet.dev/reference.md
+++ b/civet.dev/reference.md
@@ -1458,6 +1458,14 @@ function f(x?: string)?: string
   x
 </Playground>
 
+More generally, `T?` allows for `undefined` and
+`T??` additionally allows for `null`:
+
+<Playground>
+let i: number?
+let x: string??
+</Playground>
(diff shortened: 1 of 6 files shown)


STRd6 (approved):

STRd6:
The naming convention on the rules is that all TS/Type rules begin with `Type...` to group them easier. I don't necessarily think they need to be the same as the naming convention for AST nodes but I'm happy to consolidate them to a better system if we figure one out.

PR #912, opened 2024-01-04
Added (+) binary op to function shorthand

STRd6:
This was pretty simple to add but currently `(&)` as the `$=>$` shorthand is shadowing `(&)` as the binary operator.

It's probably fine to keep it for now but we may want to revisit the identity function shorthand or allow it to be the binary `&`.

### DIFF
--- a/source/parser.hera
+++ b/source/parser.hera
@@ -1777,6 +1777,7 @@ FunctionExpression
       block,
     }
   # Identity function shorthand
+  # NOTE: Currently shadows (&) binary & operator, we might want to revisit
   "(&)" ->
     const ref = makeRef("$"), body = [ref]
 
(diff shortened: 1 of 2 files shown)


source/parser.hera
```diff
@@ -1794,14 +1795,42 @@ FunctionExpression
...
 
+  # BinaryOp function shorthand
+  OpenParen:open !Identifier !Not BinaryOp:op CloseParen:close ->
+    const refA = makeRef("a")
+      refB = makeRef("b"),
+      body = [refA, op, refB]
+
+    const parameters = {
+      type: "Parameters",
+      children: [open, refA, ",", refB, close],
```
edemaine, line 1814: I don't think we should use `open` and `close` here; rather, either the whole expression or body should be wrapped in these parentheses. Otherwise e.g. `(+)+5` will bring the `+5` into the lambda which seems weird (not that this example is especially meaningful, but something like `(+) compose (*)` for a custom `operator compose` might be).

source/parser.hera
```diff
@@ -1794,14 +1795,42 @@ FunctionExpression
...
+      children: [parameters, " => ", body],
       ref,
       body,
       ampersandBlock: true,
       block,
       parameters,
     }
 
+  # BinaryOp function shorthand
+  OpenParen:open !Identifier !Not BinaryOp:op CloseParen:close ->
```
edemaine, line 1807: This seems like a good start. But I think we'd also like to include the `Identifier` and `Not Identifier` cases of `BinaryOp`. We'd just need to call `processBinaryOpExpression` in the body. I can work on this in a future PR though.

edemaine (approved):

Issue #791, opened 2023-11-01
Operators as functions via parens


edemaine:
The general idea is that `(x)` should be a way to transform an operator `x` into a function. This turns out to [match Haskell](https://wiki.haskell.org/Section_of_an_infix_operator) and has some other nice benefits.

## Level 1: Indented Operators

For long complicated expressions, it's nice to be able to switch into a LISP-like prefix notation for writing operators with indentation. For example:

```js
(and)
  (or)
    cond1
    cond2
  (||)
    cond3
    cond4
// becomes
(cond1 || cond2) && (cond3 || cond4)

(*)
  (+)
    part1
    part2
  factor
// becomes
(part1 + part2) * factor
```

This is morally equivalent to indentation-based function application that we have already, plus Level 2 below, but this needs to be handled as a special case anyway to get desired shortcutting behavior.

Of course, this should work with regular application too:

```js
(and) a, b → a && b
(and)(a, b) → a && b
```

But again note that this is not *actually* regular function application, in order to get shortcutting. I think this is preferable, but we could also skip Level 1 if we think it's better not to shortcut...?

## Level 2: Operators as Functions [DONE]

DONE: Without an immediate application, `(x)` builds a binary function. For example:

```js
array.reduce (and) → array.reduce((a, b) => a && b)
array.reduce (+) → array.reduce((a, b) => a + b)
```

Note that it might also be natural to define `(+)` as a function that takes any number of arguments and sums them, but this seems less useful because it doesn't work with [`Array.prototype.reduce`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/reduce) which provides extra arguments.

DONE: Given a custom `operator foo`, `(foo)` is just a way to write the function name `foo` without it being treated as an infix operator.

## Level 3: Sections

Inspired further by [Haskell](https://wiki.haskell.org/Section_of_an_infix_operator), we could get some nice partial applications via `(arg op)` and in some cases `(op arg)`:

```js
(1+) → (b) => 1 + b
(+1) → (+1)
(2**) → (b) => 2 ** b
(**2) → (a) => a ** 2
```

---

(I mentioned this on Discord back in September, but never wrote it up.)

PR #914, opened 2024-01-05
`(foo)` for custom operators `foo`

edemaine:
Extension of #912 as mentioned in https://github.com/DanielXMoore/Civet/pull/912#discussion_r1441907448

Also fixes complex operators like `is in` which currently crash the compiler.

This completes Level 2 of #791.


STRd6 (approved):
Thanks!

PR #921, opened 2024-01-16
Support TypeScript `paths` alias

### LINKED ISSUE #920: VSCode typescript checker doesn't see path aliases and won't work with implicit .civet extensions
As we discussed in [this discord channel](https://discord.com/channels/933472021310996512/1061703668279291924/1196747039376146584)
VSCode shows error if import file without `.civet` extension or when using path alias: `Cannot find module '@/<some path>' or its corresponding type declarations`

Example tsconfig:

```json
{
   "compilerOptions": {
     "baseUrl": ".",
     "target": "ESNext",
     "useDefineForClassFields": true,
     "module": "ESNext",
     "lib": ["ESNext", "DOM"],
     "moduleResolution": "Bundler",
     "strict": true,
     "resolveJsonModule": true,
     "isolatedModules": true,
     "esModuleInterop": true,
     "noEmit": true,
     "noUnusedLocals": true,
     "noUnusedParameters": true,
     "noImplicitReturns": true,
     // "skipLibCheck": true,
     "forceConsistentCasingInFileNames": true,
     // "jsx": "preserve",
     "composite": true,
     "downlevelIteration": true,
     "allowSyntheticDefaultImports": true,
     "allowImportingTsExtensions": true,
     "moduleDetection": "force",
     // "types": [
     // "bun-types"
     // ]
     "moduleSuffixes": [".civet", ""],
     "jsx": "react-jsx",
     "jsxImportSource": "preact",
     "skipLibCheck": true,
     "paths": {
       "@/*": ["./src/*"],
       "react": ["./node_modules/preact/compat/"],
       "react-dom": ["./node_modules/preact/compat/"]
     }
   },
   "ts-node": {
     "transpileOnly": true,
     "compilerOptions": {
       "module": "ES2020"
     }
   },
   // "include": ["./src/**/*"]
   "include": ["node_modules/vite/client.d.ts", "src/**/*", "./src/*"]
}
```

edemaine:
Fixes #920, at least in my simple testing.

There's still work to do in the resolution code, e.g. support for `rootDirs`. I've left TODOs accordingly.

Mixty also reports testing this and that it works.

### DIFF
--- a/lsp/source/lib/typescript-service.mts
+++ b/lsp/source/lib/typescript-service.mts
@@ -8,7 +8,7 @@ import type {
 import BundledCivetModule from "@danielx/civet"
 import BundledCivetConfigModule from "@danielx/civet/config"
 
-import ts, { JsxEmit } from "typescript"
+import ts, { JsxEmit, isExternalModuleNameRelative } from "typescript"
 const { version: typescriptVersion } = ts
 import type {
   CompilerHost,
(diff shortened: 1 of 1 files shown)


STRd6 (approved):
Thanks! TypeScript module resolution scares me 🐱

edemaine:
Indeed, it is quite a rabbit hole... At least I have a sketch of the full approach now.

Could you do an LSP release? (Or teach me how and give me credentials.)

STRd6:
Published LSP as v0.3.11

@edemaine if you send me your VS Marketplace user id I should be able to add you to the project. Then you need to create an access token then run these commands:

```
cd lsp
vsce login # only the first time
yarn vsce-publish
```

There's also ovsx publishing which is similar but I'll need to add you to that separately.

edemaine:
Thanks! I think I'm now `edemaine@mit.edu` on VS Marketplace, and just signed up on osvx with my GitHub account (edemaine).

Issue #922, opened 2024-01-18
LSP: Auto import


itsMapleLeaf:
The VSCode extension doesn't seem to suggest auto-imports at the moment. I rely on them heavily, and would love to see support for them.

mixty1:
And possibility to configure it, for example:
import like this:
```coffee
import { prop } from 'lib'
```

or like this:
```coffee
{ prop } from 'lib'
```

and if it type only:
```coffee
type { SomeType } from 'lib'
```
```coffee
import type { SomeType } from 'lib'
```
```coffee
import { type SomeType } from 'lib'
```
```coffee
{ type SomeType } from 'lib'
```

crystalthoughts:
Another vote for this from me . It seems that VSCode can spot a civet file but, `ctrl + space`ing it doesn't add the import declaration at the top. When manually writing it there's a 'cannot find module' error annotation.

PR #924, opened 2024-01-21
LSP pass on more completion info

edemaine:
I ported over more of [TypeScript's completion code](https://github.com/microsoft/vscode/blob/main/extensions/typescript-language-features/src/languageFeatures/completions.ts), resulting in these minor improvements:

* Show source of imports in completion description
* Optional completions get `?` suffix (e.g. completing `x.y` when `x` has type `{ y?: ... }`
* Deprecated completions get some kind of deprecation flag
* Respect `filterText`, `sortText`, `isRecommended`, `replacementSpan` (not sure if these are used though)

My original goal was #922, but I got kind of lost how to create a custom completion command to execute a follow-up query (`completionEntryDetails`) and get the additional edits to do. Do you know how to create a command ([`registerCommand`](https://github.com/microsoft/vscode/blob/e244acbb172c428cb219717a07bf55d2737492ca/extensions/typescript-language-features/src/commands/commandManager.ts#L26)) in this context? Can it even be done in a service, or is it supposed to be in `extension.civet`?

### DIFF
--- a/lsp/source/lib/util.mts
+++ b/lsp/source/lib/util.mts
@@ -197,7 +197,7 @@ function shouldIncludeEntry(item: NavigationTree | NavigationBarItem): boolean {
   return !!(item.text && item.text !== '<function>' && item.text !== '<class>');
 }
 
-function parseKindModifier(kindModifiers: string): Set<string> {
+export function parseKindModifier(kindModifiers: string): Set<string> {
   return new Set(kindModifiers.split(/,|\s+/g));
 }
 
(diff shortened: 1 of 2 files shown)


STRd6 (approved):

STRd6:
I'm pretty sure `registerCommand` needs to be called from `extension.civet` and we may need to add a message handler to respond and message the LSP.

edemaine:
Sounds right! I'll merge this for now, and give the above a try if I find the time... (alternatively, feel free to take it on)

PR #931, opened 2024-01-30
Fix optional property access shorthand

### LINKED ISSUE #930: Optional chaining with computed property access shorthand
Optional chaining if used together with shorthand computed property access has an invalid output.
Civet `0.6.64`:
```civet
foo?.1
foo?.`bar`
```
Result:
```ts
foo?[1]
foo?[`bar`]
```
Expected:
```ts
foo?.[1]
foo?.[`bar`]
```

edemaine:
Fixes #930

### DIFF
--- a/source/parser.hera
+++ b/source/parser.hera
@@ -18,6 +18,7 @@ const {
   expressionizeTypeIf,
   forRange,
   gatherBindingCode,
+  gatherRecursive,
   getIndentLevel,
   getTrimmingSpace,
   hasAwait,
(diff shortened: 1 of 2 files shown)


edemaine:
By the way, you'll see some existing test cases were already wrong (incorrect JS). I wonder if we should add to `testCase` some kind of js validity testing, e.g. prettier parse phase or perhaps esbuild has something faster?

STRd6 (approved):

STRd6:
Adding a validation pass on the output could be useful if it is fast enough.

PR #932, opened 2024-01-30
Test validity of JS/TS outputs via esbuild parsing

edemaine:
I gave a try at running output JS/TS through esbuild to check for validity.  It seems to work, quickly — I measured a slowdown from 8.724 seconds to 8.837 seconds, so about a 1% slowdown.

However, perhaps as expected, a lot of our code snippets are invalid, mainly because they lack context. For example:
* Lack of context: Using `yield` outside a generator function, `await` outside an `async` function, `return` outside a function, `continue`/`break` outside a loop, `super` or private fields outside a class, etc.
* Intentionally bad code like `let foo!` (which TS generates helpful error messages for)
* Doubly defined variables (easy to fix)

On the other hand, I think this found some actual bugs in our output:
* Handling of anonymous classes is incorrect; `class {}` is invalid whereas `(class {})` is valid.
* `catch (e) console.error(e)` is incorrect generated output.

So I think there's something to this testing! It just needs more work.

At this point, I wanted to ask: what do you think we should do about out-of-context examples? Some options I see:
1. Add the appropriate context, like surrounding `class`, `function`, etc., to all/most examples.
2. Add `testCase.partial` or something to indicate these examples are out of context and shouldn't be tested via esbuild (but then we lose the benefit for these tests).
3. Automatically add context. E.g., if we detect `yield` but no `function*`, add an automatic wrapper. Probably we could do reasonably well with some regular expressions. Plausibly, we could even find a universal wrapper that works in all cases, e.g. an async generator method inside a class, and always add that.


STRd6:
I think we need to update our build script to esbuild `0.20.x`, `watch` is no longer a valid option we can likely just remove it.

For the context we could add a `withContext` test helper that will wrap the tests inside it with a context block. I think that should explicitly cover groups of test cases without much indirection.

STRd6:
Nice! Looks like it found several legitimate bugs.

edemaine:
Yep! So seems worth doing. We probably need to fix these bugs on this branch before merging, though, so we don't have a failing build.

edemaine:
OK, I got all the bugs worked out! This now ready for review.

source/parser.hera
```diff
@@ -5940,7 +5940,13 @@ JSXAttributes
           classValue = ["{[", ...exprs, "].filter(Boolean).join(\" \")}"]
         }
       } else { // strings only
-        classValue = JSON.stringify(stringPart)
+        if (!stringPart.includes('&') && !stringPart.includes('"')) {
+          classValue = `"${stringPart}"`
+        } else if (!stringPart.includes('&') && !stringPart.includes("'")) {
+          classValue = `'${stringPart}'`
+        } else {
+          classValue = `{${JSON.stringify(stringPart)}}`
```
STRd6, line 5958: Should we be concerned about ampersands in the string here?

STRd6 (approved):

source/parser.hera
STRd6: Never mind, I see that it is inside the braces so it should be fine.

PR #936, opened 2024-02-01
Fix array of objects type

### LINKED ISSUE #934: ParseError with array-of-objects type
```ts
let x: {}[]
```
is a ParseError currently.

edemaine:
Fixes #934

### DIFF
--- a/source/parser.hera
+++ b/source/parser.hera
@@ -6612,7 +6612,8 @@ TypeSuffix
 
 MaybeIndentedType
   # NOTE: Let InterfaceBlock take first crack at an indented block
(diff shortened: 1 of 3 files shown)


STRd6 (changes requested):
I only see the test, not the fix.

edemaine:
Oops, forgot to `git stash pop`. Should be ready for review now!

STRd6 (approved):
Might be good to also add a test case for:

```coffee
  testCase """
    let type array of empty object
    ---
    let x: {}[]
    ---
    let x: {}[]
  """
```

PR #940, opened 2024-02-03
Support TS instantiation expressions

### LINKED ISSUE #939: Typescript Instantiation Expressions
Currently `0.6.66` typescript instantiation expressions do not work.
They either throw a parser error or are confused with a chained comparison.
Error:
```civet
bar := foo<number>
```
Unexpected output:
```civet
bar := foo<number>
baz
```
```ts
const bar = foo < number && number > baz;
```

edemaine:
Fixes #939

This actually cleaned up the grammar some; before, we had to check for `TypeArguments` in every call scenario. Now it's part of `CallExpression`.

There's some fun ambiguity with tight comparison chains. But note that `x<y>(z)` was already treated like a function call, not a comparison chain. I preserved the working of `x<y>z` and `x<y>0` but there might be other more subtle cases that break...

### DIFF
--- a/source/parser.hera
+++ b/source/parser.hera
@@ -193,17 +193,17 @@ Arguments
     return $skip
 
 ImplicitArguments
-  ( TypeArguments !ImplementsToken )?:ta ApplicationStart InsertOpenParen:open _?:ws NonPipelineArgumentList:args InsertCloseParen:close ->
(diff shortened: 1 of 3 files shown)


STRd6 (approved):
Looks good!

I also added a test for an identifier on the following line to make sure it doesn't get parsed as a chained comparison.

PR #941, opened 2024-02-04
Playground eval

### LINKED ISSUE #938: civet.dev/playground - Displaying console output from async functions
On the [playground](https://civet.dev/playground) looks like some `console.log` calls inside async functions are lost. This doesn't seem to affect immediately resolved promises.

Example `Civet 0.6.66`:
```civet
dog := 'https://dog.ceo/api/breeds/image/random'

main()

function main
  await show_foo()
  await fetch_dog();

function fetch_dog
  fetch dog |> await |> .json() |> await |> console.log;

async function show_foo
  console.log 'foo';
```
[playground](https://civet.dev/playground?code=ZG9nIDo9ICdodHRwczovL2RvZy5jZW8vYXBpL2JyZWVkcy9pbWFnZS9yYW5kb20nCgptYWluKCkKCmZ1bmN0aW9uIG1haW4KICBhd2FpdCBzaG93X2ZvbygpCiAgYXdhaXQgZmV0Y2hfZG9nKCk7CgpmdW5jdGlvbiBmZXRjaF9kb2cKICBmZXRjaCBkb2cgfD4gYXdhaXQgfD4gLmpzb24oKSB8PiBhd2FpdCB8PiBjb25zb2xlLmxvZzsKCmFzeW5jIGZ1bmN0aW9uIHNob3dfZm9vCiAgY29uc29sZS5sb2cgJ2Zvbyc7)

Output is just: `foo`
Expected:
```
foo
{ ...responseJSON }
```

edemaine:
Playground "Run" button now supports:

* Running code like `1+2` now outputs `[EVAL] 3`, without the need for wrapping in `console.log`.  Like most REPLs, outputs nothing if the eval'd value is `undefined`.
* `console.error`, `console.warn`, etc. (prefixing output with `[ERROR] `, `[WARN] `, etc.). `console.timeEnd` etc. don't currently work well (but they still work in the actual console); this could be fixed later.
* `console` commands now work from `async` functions too (fixes #938). This is done by modifying the output AST to replace all `console` identifiers with `civetconsole`. (Note: if you run async code twice in rapid succession, you may get output from the first run in the second run.)
* Top-level `await`, by wrapping the Civet code in an `async do` block (or `do ->` block in CoffeeScript mode). I did this at the Civet level so that we still get implicit return of the last value, which (once resolved) gets printed with `[EVAL]` prefix like the sync case.


edemaine:
Ugh, `yarn test` fails with my attempt to export `lib` from the main Civet bundle. The issue is:
* `parser.hera` and `lib.civet` are in CJS. I think this is necessary until Hera is [fully ESM](https://github.com/DanielXMoore/Hera/pull/13)?
* `main.civet` is ESM, and wants to export from `lib.civet`

esbuild can build these fine, but running in actual ESM mode as in `yarn test` fails, because ESM isn't supposed to import CJS. Any ideas for a workaround?

STRd6:
I think [this branch](https://github.com/DanielXMoore/Hera/pull/13/files) in Hera is maybe 90% of the way to emitting ESM. I recall leaving it though because I encountered an issue with the latest NPM loaders API and I couldn't find a way to actually get Node to load ESM using the new API. Maybe the examples and documentation are better now and I can continue it soon.

edemaine:
#942 fixed the build here! So this is ready for review.

STRd6 (approved):

PR #942, opened 2024-02-05
Hera ESM support

STRd6:
I worked around the Node loader issue by using `shortCircuit`. It's not ideal because I could imagine it might be useful to chain loaders like Babel/etc. but this got it working for now.

@edemaine This might help you on your https://github.com/DanielXMoore/Civet/pull/941 branch.


edemaine (approved):
Wow, this is great! Should be exactly what I need for #941.

PR #945, opened 2024-02-06
Shorthand for type indexed access

edemaine:
Mainly this adds `T."x"` and related access to types, paralleling our existing support for this notation in expressions.

For consideration, I also added `T::x` shorthand for `T["x"]` when `coffeePrototype` is enabled. I think this intuitive for those who have this flag on (think of `T` as a class), though of course there isn't an actual notion of prototype here. I could also remove this if you think it's not ideal. (It still leaves our options for `::` open when `coffeePrototype` is off. I also don't think this conflicts with any of our current intentions for `::`.)

### DIFF
--- a/civet.dev/reference.md
+++ b/civet.dev/reference.md
@@ -1568,6 +1568,29 @@ enum Direction
   Right = 2 * Left
 </Playground>
 
+### Indexing Types
+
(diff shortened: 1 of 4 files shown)


STRd6 (approved):

STRd6:
Looks good and makes sense to only enable `::` with `coffeePrototype` for now so we don't close of other options for `::` too early.

PR #948, opened 2024-02-07
Operators as functions via parens

### LINKED ISSUE #791: Operators as functions via parens

edemaine:
Fixes #791.

* Previously, #912 implemented Level 2: `(+)` and other binary operators make binary functions.
* However, there was a bug: `(+) 1, 2` or `(+) |> wrap` didn't parse at the start of the line (including inside a function body) because `(+)` got treated as a function declaration, so the parser treated that as a complete line (ignoring the application/pipe that followed). There was a similar issue with `&+1 |> wrap`.
  * This is fixed by not treating `(+)` or `&+1` as function declarations; they're arrow functions anyway.
  * In fact, it would make sense to move these rules from `FunctionExpression` into `ArrowFunction`. However, this didn't work on a simple try... not sure why.
* Implemented Level 1: immediate application of e.g. `(&&)` magically unwraps, which enables its use with any positive number of arguments and enables short-circuit behavior.
* Implemented Level 3: sections like `(+ 1)` and `(1+)`. `(+1)` remains a unary operator.


STRd6 (approved):
Nice, I'm glad it wasn't too complicated to add this!

PR #950, opened 2024-02-07
Port CLI to more modern Civet

edemaine:
I ported the CLI code to more modern Civet, and added types so that TypeScript is happy with it. I tried testing all the options (but it wouldn't hurt to double-check – in particular, I've never used a config file, so I didn't test that).

There's still a bit of a mix of `require` and `import`, which works fine in practice. Currently, dynamic import of a JSON file generates a warning, which I'd like to avoid.

(This is in preparation for adding a `--typecheck` flag.)

### DIFF
--- a/source/cli.civet
+++ b/source/cli.civet
@@ -53,12 +53,13 @@ encoding .= "utf8" as BufferEncoding
 fs from "fs/promises"
 type { Stats } from fs
 path from "path"
+type { CompileOptions } from "@danielx/civet"
 
 interface Options
   run?: boolean
   compile?: boolean
   output?: string
-  config?: string | boolean | ???
+  config?: string | boolean | CompileOptions
   ast?: boolean
   noCache?: boolean
   inlineMap?: boolean
@@ -233,11 +234,11 @@ function cli
   {filenames, scriptArgs, options} .= parseArgs argv[2..]
 
   if options.config is not false // --no-config
-    options.config ?= await findConfig(process.cwd()) as Options
+    options.config ?= await findConfig process.cwd()
 
   if options.config
     options = {
-      ...(await loadConfig options.config as string) as Options,
+      ...await loadConfig options.config as string
       ...options
     }
 
--- a/source/cli.civet
+++ b/source/cli.civet
@@ -1,9 +1,9 @@
-"civet coffeeCompat"
-
 { compile } from ./main.civet
 { findConfig, loadConfig } from ./config.civet
 
-version = -> require("../package.json").version
+function version: string
+  require("../package.json").version
+
 if process.argv.includes "--version"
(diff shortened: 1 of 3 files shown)


source/config.civet
```diff
@@ -9,7 +9,7 @@ configFileNames := new Set [
   "civetconfig.civet"
 ]
 
-function findInDir(dirPath: string): Promise<string | undefined>
+function findInDir(dirPath: string): Promise<string | ???>
```
esthedebeste, line 12: `???` means `unknown`, this isn't what findInDir is returning. The `undefined` return value indicates not having found a config file.

source/config.civet
```diff
@@ -26,7 +26,7 @@ function findInDir(dirPath: string): Promise<string | undefined>
 
   return
 
-export function findConfig(startDir: string)
+export function findConfig(startDir: string): Promise<string | ???>
```
esthedebeste, line 29: Same as `findInDir`

source/config.civet
```diff
@@ -40,7 +40,7 @@ export function findConfig(startDir: string)
 
   return
 
-export function loadConfig(path: string)
+export function loadConfig(path: string): Promise<???>
```
esthedebeste, line 43: This should be a `Promise<CompileOptions>`, with `CompileOptions` from `types/types.d.ts`

esthedebeste (changes requested):

source/config.civet
esthedebeste: well i suppose it isnt checked to be a CompileOptions object, but that is how it's used.

source/cli.civet
```diff
@@ -51,16 +51,36 @@ if process.argv.includes "--help"
 encoding .= "utf8" as BufferEncoding
 
 fs from "fs/promises"
+type { Stats } from fs
 path from "path"
+type { CompileOptions } from "@danielx/civet"
```
STRd6, line 56: We should use the relative path here so we get any updated types without having to update the dev dependency.

edemaine: Sure, I can point to the `.d.ts` for now. I think ideally we move the type definitions into an actual `.civet` file, and the `.d.ts` is generated (via unplugin say), but that can come later.

edemaine: Actually, I don't see how to import a type from a `.d.ts` file, because `declare module { ... }` is ambient, not importable/module compatible. ☹️ 

If you know a way, I can implement it (or feel free to). Otherwise, I think this should wait until the `.d.ts` has a build step.

STRd6: Just add a TODO comment in the meantime then.

STRd6 (approved):

PR #955, opened 2024-02-08
Cleanup Call arguments AST, fix (+) processing

edemaine:
#948 recursively processed all commas, which leads to some weird behavior:
```js
(+) f(x, y), z
↓↓↓
f(x) + y + z
```

This PR bites the bullet and cleans up all forms of `Arguments` so that you get one array with all the arguments and commas at the top level (similar to processing we do with top-level statements, IIRC). This lets us process commas without recursion.

Currently, whitespace is combined with some arguments and commas, so that the array really alternates between arguments and commas. Alternatively, we could allow whitespace at the top level of the array too. I guess this can be decided later.

I also changed the behavior of `(+)(1)` to expand to just `(1)`. This seems beneficial when you're working with complex nested prefix formulas, and you remove one item, expecting `(and)(cond)` to give `cond` and not crash.

### DIFF
--- a/source/lib.civet
+++ b/source/lib.civet
@@ -819,31 +819,26 @@ function processCallMemberExpression(node)
   // use of the operator. This is useful for e.g. shortcutting (&&).
   if children[0]?.parenthesizedOp?.token and children[1]?.type is "Call"
     op := children[0].parenthesizedOp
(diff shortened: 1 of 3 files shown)


STRd6 (approved):
Thanks!

Issue #744, opened 2023-09-20
Oddities/pitfalls in rules for preparing return value


gwhitney:
The reference says

> you can prepare [a function's return value] ahead of time using return.value (or its shorthand, assigning to return).

So I can do `return := a: 3, b: 7` and that works fine. Now what if I want, before the end of the function, to modify the `a` property of the return value or add a property `c`? I have to write `return.value.a += 3` or `return.value.c = 11`. Then two statements stacked like
```
return := a: 3, b: 7
return.value.b -= 3
```
don't visually look like they refer to the same property `b` on the return value, as there's that "extra" property extraction `.value` sandwiched into that second line.

I recognize that `return.b -= 3` would not technically be assigning to return, it's assigning to a _property_ of return, but it feels very analogous to assigning to return, and it's a parse error so it seems to be available for being interpreted as modifying the `b` property of the return value being prepared. And as it feels as though one of the design guides for Civet is to produce pretty code, I'll just say
```
return := a: 3, b:7
return.b -= 3
return.c = 'Something new'
```
looks prettier at least to my eyes.

However, I recognize the ambiguity that a convention like this would produce for:
```
return := name: 'Benjamin'
return.value = 100
```
I.e., did you mean to be replacing the return value of the function altogether with 100 (violating its constness)  or setting its `value` property to 100 so that the function would ultimately return `{name: 'Benjamin', value: 100}` ? And I recognize the crisis of
```
return := thing: 'one', n: 7
return.n = 10*return.value.n + 1
```
now looking really bad -- unless it's feasible to always interpret `return.prop` as a property of the value being prepared. But then presumably an entirely new notation would be needed to refer to the entire return value being prepared, because in such a world  surely `return.value` should refer to its `value` property.

[...]

PR #961, opened 2024-02-08
Allow return.value in pipeline

### LINKED ISSUE #958: ParseError when piping to `return.method`
```civet
f := ->
  return := foo()
  bar()
    ||> baz
    |> return.method
```

gives a ParseError pointing at the `.`

edemaine:
Fixes #958

### DIFF
--- a/source/parser.hera
+++ b/source/parser.hera
@@ -675,7 +675,7 @@ PipelineHeadItem
 PipelineTailItem
   Await
   Yield
-  Return
+  Return !"." -> $1
   AmpersandFunctionExpression
   !Ampersand PipelineHeadItem -> $2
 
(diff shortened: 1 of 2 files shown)


source/parser.hera
```diff
@@ -675,7 +675,7 @@ PipelineHeadItem
 PipelineTailItem
   Await
   Yield
-  Return
+  Return !"." -> $1
```
STRd6, line 678: Mabye want to be AccessStart so we can pick up `?.`, etc.

edemaine: Not technically possible currently, because we currently explicitly check only for `return.value`, but `!AccessStart` seems safer against the future, also depending on how #744 works out.

STRd6 (approved):

PR #972, opened 2024-02-10
Fix concatAssign for arrays

bbrk24:
I made a small mistake when originally writing `concatAssign`: the value of `Array.prototype[Symbol.isConcatSpreadable]` is `undefined`, not `true`. Setting the value to `false` does make it no longer concat-spreadable, but if it's `undefined` we need to *also* check if it's an Array.

[Relevant section of the ECMAScript specification](https://tc39.es/ecma262/multipage/indexed-collections.html#sec-isconcatspreadable):

> 1. If *O* [is not an Object](https://tc39.es/ecma262/multipage/ecmascript-data-types-and-values.html#sec-object-type), return **false**.
> 2. Let *spreadable* be ? [Get](https://tc39.es/ecma262/multipage/abstract-operations.html#sec-get-o-p)(*O*, [@@isConcatSpreadable](https://tc39.es/ecma262/multipage/ecmascript-data-types-and-values.html#sec-well-known-symbols)).
> 3. If *spreadable* is not **undefined**, return [ToBoolean](https://tc39.es/ecma262/multipage/abstract-operations.html#sec-toboolean)(*spreadable*).
> 4. Return ? [IsArray](https://tc39.es/ecma262/multipage/abstract-operations.html#sec-isarray)(*O*).

### DIFF
--- a/source/parser.hera
+++ b/source/parser.hera
@@ -7510,7 +7510,7 @@ Reset
         module.prelude.push({
           children: [
             preludeVar, ref, typeSuffix,
-            " = (lhs, rhs) => ((rhs", asAny, ")?.[Symbol.isConcatSpreadable] ? (lhs", asAny, ").push.apply(lhs, rhs", asAny, ") : (lhs", asAny, ").push(rhs), lhs);\n"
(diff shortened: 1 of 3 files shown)


edemaine (approved):
Thanks for catching and fixing this!

PR #975, opened 2024-02-10
TypeScript's `import ... = require(...)` and `export = ...`

### LINKED ISSUE #971: Compiler does not respect `"module": "CommonJS"`
Even if I set `"module": "CommonJS"` in my tsconfig and `"type": "commonjs"` in my package.json, Civet does not translate `import` and `export` statements.

edemaine:
This is necessary for working with TypeScript + CommonJS.  See https://www.typescriptlang.org/docs/handbook/modules/reference.html#export--and-import--require

Fixes #971 in a certain sense, in particular https://github.com/DanielXMoore/Civet/issues/971#issuecomment-1936853962

### DIFF
--- a/source/parser.hera
+++ b/source/parser.hera
@@ -4703,6 +4703,18 @@ MaybeNestedExpression
 
 # https://262.ecma-international.org/#prod-ImportDeclaration
 ImportDeclaration
+  # NOTE: TypeScript's CommonJS import syntax
+  # https://www.typescriptlang.org/docs/handbook/modules/reference.html#export--and-import--require
+  Import _ Identifier _? Equals __ "require" NonIdContinue Arguments ->
+    // Replace `import` with `const` in JS output
+    const imp = [
+      { ...$1, ts: true },
+      { ...$1, token: "const", js: true },
+    ]
+    return {
+      type: "ImportDeclaration",
+      children: [imp, $0.slice(1)],
+    }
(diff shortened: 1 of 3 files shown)


STRd6:
I think this also works with destructured imports to some extent: https://www.typescriptlang.org/play?experimentalDecorators=true&module=1#code/JYWwDg9gTgLgBAbygUwIYBMBiwA2yDKAngHYDGcqAznChgDRyEQC+cAvDcgI4CuwKACgBEAM0pCAlACgpTKbXRSgA

edemaine:
I agree it generates some code, but it's wrong; notice the `require()` call has no argument. Also the Errors say "'from' expected" (after `import {...}`).

What I could get to work is `import X = Y` where `Y` is an existing namespace; also `Y.Z.W`. That kind of aliasing doesn't seem too useful though, and in any case it would be implemented differently (the whole line is TypeScript only). I can add it though.

STRd6 (approved):

bbrk24:
> I think this also works with destructured imports to some extent

This is https://github.com/microsoft/TypeScript/issues/46752.

edemaine:
Cool. It is natural, so we could add this feature at the Civet level. But given that I only just learned about `import ... =`, I'm not sure it's used very much, as CommonJS slowly migrates to ESM, so maybe not a priority?

PR #976, opened 2024-02-10
Nested vs. implicit vs. inline object literals

### LINKED ISSUE #968: Object indentation bug
Input:
```coffee
{
  signature: modifier: {}
  children: [parameters, " => ", body]
}
```

[Currently transpiles to](https://civet.dev/playground?code=ewogIHNpZ25hdHVyZTogbW9kaWZpZXI6IHt9CiAgY2hpbGRyZW46IFtwYXJhbWV0ZXJzLCAiID0%2BICIsIGJvZHldCn0%3D):
```js
({
  signature: {
    modifier: {},
    children: [parameters, " => ", body],
  },
});
```

Expected: (like [CoffeeScript](https://coffeescript.org/#try:%7B%0A%20%20signature%3A%20modifier%3A%20%7B%7D%0A%20%20children%3A%20%5Bparameters%2C%20%22%20%3D%3E%20%22%2C%20body%5D%0A%7D))

```js
({
  signature: {
    modifier: {},
  },
  children: [parameters, " => ", body],
});
```

Unbraced input has the same issue:

```coffee
signature: modifier: {}
children: [parameters, " => ", body]
```

edemaine:
Fixes #968 as well as related [weird behaviors](https://civet.dev/playground?code=ZnVuYyBmb286IDEKYmFyOiAy):
```coffee
func foo: 1
bar: 2
↓↓↓
func({ foo: 1, bar: 2 });
```

The issue is that the following parses as an `InlineObjectLiteral` thus `ObjectLiteral` thus `PrimaryExpression`:

```coffee
foo: 1
bar: 2
```

---

This PR splits **implicit object literals** from two to three notions:
* `NestedImplicitObjectLiteral` (AS BEFORE): starts with an indent, arbitrary spacing
* `ImplicitObjectLiteral` (NEW): indentation already consumed, but can have multiple lines at same indentation level — assumed to be used only at beginning of lines
* `InlineObjectLiteral` (MODIFIED): can appear anywhere in an expression, so generally spans only one line, except that explicit trailing commas let you go to next line (as in CoffeeScript)

Previously, `InlineObjectLiteral` acted like `ImplicitObjectLiteral` and that was a problem because it could be used anywhere, not just at beginnings of lines. The main challenge was figuring out where we wanted to add `ImplicitObjectLiteral` in addition to general expressions which included `InlineObjectLiteral`.

Also fixed bugs like the following which [didn't parse](https://civet.dev/playground?code=bG9vcAogIGNvbnRpbnVlOiB4) because there was no check for object literals first, so `continue` matched all by itself:

```coffee
loop
  continue: 1
```

### DIFF
--- a/source/parser.hera
+++ b/source/parser.hera
@@ -2325,12 +2325,11 @@ NonSingleBracedBlock
   ImplicitNestedBlock
 
   # Immediate nested object literal
-  InsertOpenBrace &EOS ObjectLiteral:s InsertCloseBrace ->
+  InsertOpenBrace NestedImplicitObjectLiteral:s InsertCloseBrace ->
     return {
       type: "BlockStatement",
       expressions: [s],
-      // Remove &EOS assertion
-      children: [$1, s, $3],
+      children: $0,
     }
 
 DeclarationOrStatement
@@ -2687,6 +2686,7 @@ ElementListRest
 ArrayElementExpression
   # NOTE: Prevent multiple JSX tags from combining implicitly into a fragment
   JSXTag
+  ImplicitObjectLiteral &ArrayElementDelimiter -> $1
   # NOTE: Allow for postfix splat like CoffeeScript
   # NOTE: Allow empty exp spread for destructuring
   ExtendedExpression?:exp __:ws DotDotDot:dots &ArrayElementDelimiter ->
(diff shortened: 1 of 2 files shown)


STRd6 (approved):

STRd6:
Thanks!

PR #977, opened 2024-02-10
Fix private field glob getters

### LINKED ISSUE #967: Extra `this` in glob getters of private fields
```civet
class Foo
  #x
  #y

  get @{#x,#y}
```
Compiles to:
```js
class Foo {
  #x;
  #y;

  get x() {
    return this.this.#x;
  }
  get y() {
    return this.this.#y;
  }
}
```

edemaine:
Fixes #967


STRd6 (approved):
👏

PR #980, opened 2024-02-11
Allow enums on one line

### LINKED ISSUE #978: Enums require line breaks
```ts
// allowed
enum E1
  A,
  B,
  C,

// not allowed
enum E2
  A, B, C
```

bbrk24:
Fixes #978

I made the `Nested` part of `NestedEnumProperty` optional, allowing it to be on the same line as the previous enum property. I also had to add `_?` or else it didn't allow whitespace after the commas.

I tried to make sure this doesn't inadvertently allow anything it shouldn't, like

```civet
enum E
  A B // look ma, no comma!
```

still causes a syntax error. But there may have been something I forgot to check.

### DIFF
--- a/source/parser.hera
+++ b/source/parser.hera
@@ -6707,15 +6707,14 @@ EnumBlock
     }
 
 NestedEnumProperties
-  PushIndent NestedEnumProperty*:props PopIndent ->
-    const flatProps = props.flat()
-    if (!flatProps.length) return $skip
+  PushIndent NestedEnumPropertyLine*:props PopIndent ->
+    if (!props.length) return $skip
     return {
-      properties: flatProps.map((p) => p.property),
+      properties: props.flat().map((p) => p.property),
       children: $0,
     }
 
-NestedEnumProperty
+NestedEnumPropertyLine
   ( Nested EnumProperty ) ( _? EnumProperty )* ->
     return [$1, ...$2].map(pair => ({
       property: pair[1],
--- a/source/parser.hera
+++ b/source/parser.hera
@@ -6708,18 +6708,19 @@ EnumBlock
 
 NestedEnumProperties
   PushIndent NestedEnumProperty*:props PopIndent ->
-    if (!props.length) return $skip
+    const flatProps = props.flat()
+    if (!flatProps.length) return $skip
     return {
-      properties: props.map((p) => p.property),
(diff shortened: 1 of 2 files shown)


edemaine (approved):
Nice fix! This makes me wonder whether we could simplify other similar rules. Often we write a rule for a comma-separated single line, and then a rule for assembling multiple lines. It's now how this combines both into one rule.

source/parser.hera
```diff
@@ -6715,7 +6715,7 @@ NestedEnumProperties
     }
 
 NestedEnumProperty
-  Nested EnumProperty ->
+  Nested? EnumProperty _? ->
```
STRd6, line 6718: I'd do this slightly differently, similar to how we do lines of array items (see https://github.com/DanielXMoore/Civet/blob/d2a6b8b9e3e8b4d42b4f1bef8a4c2350625572b8/source/parser.hera#L2670)

edemaine: I've been trying to figure out whether they are equivalent. One issue with the current code is that it does not enforce a leading `Nested`. But I think because of the prior `PushIndent`, the first `NestedEnumProperty` must in fact have a leading `Nested`. And `Nested` is the only way to get to the next line.

That said, this is a lot of thought, so maybe it is better to require `Nested` and explicitly consider a full line? Something like `Nested EnumProperty ( _? EnumProperty )*` might suffice.

bbrk24: I did some testing on the CLI to make sure that none of these were accepted:
```civet
enum E1 A

enum E2/**/A

enum E3
A
```
I didn't add them as test cases because I wasn't sure how, but now that I've seen 518d041556f60d5fad2244691db07294b25689e2 I can add these three as negative test cases.

edemaine: I agree that none of them work, but it takes a lot of thought to convince ourselves that it's correct, so the more explicit rule seems better?

source/parser.hera
```diff
@@ -6708,18 +6708,19 @@ EnumBlock
 
 NestedEnumProperties
   PushIndent NestedEnumProperty*:props PopIndent ->
-    if (!props.length) return $skip
```
edemaine, line 6711: I think this should have stayed. The point is to skip if nothing was matched, not if there end up being no properties (though these are equivalent).

PR #992, opened 2024-02-14
Report error nodes in LSP

STRd6:
The LSP now reports parse errors. Unfortunately they don't have location information yet so they will only appear at the top of the file. Eventually we can add location info to error nodes, apply the sourcemapping, update the LSP to check for this location, and then they will appear in the proper place.

### DIFF
--- a/lsp/source/lib/typescript-service.mts
+++ b/lsp/source/lib/typescript-service.mts
@@ -77,6 +77,7 @@ interface Transpiler {
   compile(path: string, source: string): {
     code: string,
     sourceMap?: SourceMap
+    errors?: Error[]
   } | undefined
 }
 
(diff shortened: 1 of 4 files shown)


lsp/source/lib/typescript-service.mts
```diff
@@ -542,11 +543,15 @@ function TSService(projectURL = "./") {
...
-    })
+    const errors: Error[] = [],
+      result = Civet.compile(source, {
+        ...civetConfig,
+        filename: path,
+        sourceMap: true,
+        errors,
+      })
+
+    return Object.assign(result, { errors })
```
edemaine, line 554: Seeing this interface makes me wonder whether `Civet.compile` should directly return `errors` along with code and sourceMap, instead of (or in addition to?) modifying options. Is the issue that `compile` doesn't return an object when not requesting sourceMap?

edemaine (approved):
Seems like a good start!

I agree we should eventually do the source mapping. Ideally within main.civet (during gen?) so we can use it in all contexts.

lsp/source/lib/typescript-service.mts
STRd6: Yes, I think in the long run the errors can be returned alongside the sourceMap. When not requesting a sourceMap we should still throw because it only returns the compiled code as a string.

PR #1000, opened 2024-02-15
Don't create empty var decs with auto-var

### LINKED ISSUE #943: Extra comma when `autoVar` is used with pattern matching
```civet
'civet autoVar'
worker.onmessage = (event) ->
  switch event.data
    [count]
      threadCount = count
```
Becomes:
```ts
worker.onmessage = function(event) {
  var , threadCount;
  let m;
  if(m = event.data,Array.isArray(m) && m.length === 1) {
      const [count] = m;
      return threadCount = count}
}
```
The extra comma in the `var` statement breaks things.

STRd6:
Fixes #943

### DIFF
--- a/source/lib.civet
+++ b/source/lib.civet
@@ -3600,7 +3600,7 @@ function populateRefs(statements: ASTNode): void {
 }
 
 // CoffeeScript compatible automatic var insertion
-function createVarDecs(statements, scopes, pushVar) {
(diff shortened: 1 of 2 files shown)


edemaine (approved):
Do you need to do a similar change for autoLet?

STRd6:
`autoLet` and `autoConst` use a different code path because they are block scoped and not function scoped. This specific example worked in those cases.

PR #1001, opened 2024-02-15
Fix cached node mutation when removing trailing comma from rest property

STRd6:


source/lib.civet
```diff
@@ -3990,7 +3990,9 @@ function reorderBindingRestProperty(props) {
 
     // JS forbids trailing comma after rest property
     if Array.isArray(rest.delim) and rest.delim.-1?.token is ","
-      rest.delim.pop()
+      // Can't mutate delim in case of cached nodes
+      rest.delim = rest.delim.slice(0, -1)
+      rest.children = rest.children.slice(0, -1).concat(rest.delim)
```
edemaine, line 3995:
```suggestion
      rest.children = [ ...rest.children.slice(0, -1), rest.delim ] 
```

concat will inline the delim array, which is annoying if we ever need to find/modify it again.

edemaine (approved):
Thanks for fixing this!

PR #1008, opened 2024-02-16
Pipe new

STRd6:
Allow piping to `new X`


edemaine (approved):
I'm surprised this was this easy!

I can't think of any issues with this, though it feels a little weird that new expressions never need to be wrapped in parens.

STRd6:
NewExpressions are apparently valid LeftHandSideExpressions: https://262.ecma-international.org/#prod-LeftHandSideExpression

angelhdzmultimedia:
🎉😍🔥👏🤯 I'm impressed. You guys are cool!

PR #1009, opened 2024-02-16
Fix #947 - Better const function semantics

### LINKED ISSUE #947: Function transformation loses `const` semantics
As discussed on the Discord this morning:

> ```civet
> // this will crash JS
> x := 5
> x := 10
> // this won't :-(
> y := -> 5
> y := -> 10
> ```

STRd6:
Removing our transform of const declarations to function to keep them consistent with const semantics.

If people really want function semantics they can use the `function` declaration.

### DIFF
--- a/source/parser.hera
+++ b/source/parser.hera
@@ -1951,16 +1951,6 @@ FunctionExpression
     }
 
   AmpersandFunctionExpression
-  Identifier:id __ ConstAssignment _*:ws ThinArrowFunction:fn ->
-    return {
-      ...fn,
-      id,
-      children: [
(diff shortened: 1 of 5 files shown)


edemaine (approved):
I was surprised to find that even CoffeeScript doesn't transform `x = ->` into a hoistable function. Of course, it hoists the *name* of the function via automatic var, but you can't use it until after you assign it. So yeah, I agree, I think we should remove this magic and stay closer to the source. If we break code that assumes hoisting, it should be clear because you'll try to use a function before it's declared.

PR #1014, opened 2024-02-16
Unary prefix and postfix in braced literal shorthand

### LINKED ISSUE #1012: Braced literals with `await`
We can do like this (from [the reference](https://civet.dev/reference#braced-literals)):
```js
another := {person.name, obj?.c?.x}
computed := {foo(), bar()}
named := {lookup[x+y]}
```
But not this:
```js
asyncComputed := {await asyncFoo()}
```
We can only explicitly write like this:
```js
asyncComputed := {asyncFoo: await asyncFoo()}
```
Which is not cool.

It would be great to be able to use asynchronous functions and properties along with synchronous ones.

edemaine:
Fixes #1012 by allowing unary prefix and postfix notation in braced literals shorthand properties in most cases. This allows for negation, type casting, and await.

Also add missing support for positive integers as property names.

### DIFF
--- a/civet.dev/reference.md
+++ b/civet.dev/reference.md
@@ -124,8 +124,9 @@ Both braced and unbraced literals support shorthand for
 computed property names:
 
 <Playground>
-negate := {-1: 1, 1: -1}
-templated := {`${prefix}${suffix}`: result}
+negate := {-1: +1, +1: -1}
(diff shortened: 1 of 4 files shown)


STRd6 (approved):
👍 I can see this coming in handy

PR #1015, opened 2024-02-16
Support extends shorthand in type parameters

### LINKED ISSUE #1013: `extends` shorthand in type declaration
`<` for `extends` seems to be allowed in the body of a type, but not its declaration. None of these work:

```ts
type Foo<T < string> = T
type Foo<T: string> = T
type Foo<T <: string> = T
```

Coming from languages like Swift and C# I'm more used to the second one, but the first is probably most consistent with how Civet already abbreviates `extends`.

edemaine:
Fixes #1013

However, fixing this requires:
* There are some ambiguous situations like `class X<Y<Z>`; it could mean `class X extends Y<Z>` or `class X<Y extends Z>`. But I guess you're asking for trouble in this situation. Currently it means the latter because greedy.
* Forbidding space between a type and its parameters: `T<X>` is valid, but `T <X>` is invalid. This is contrary to TypeScript, but in line with our usual rules about whitespace I think? Potentially controversial, though TS documentation seems to consistently have no space.
* Alternatively, we could still allow spaces, but then `class X < Y<Z>` does not parse how we'd expect.

### DIFF
--- a/source/parser.hera
+++ b/source/parser.hera
@@ -7166,7 +7166,7 @@ TypeArgumentDelimiter
   TypeParameterDelimiter
 
 TypeParameters
-  _? OpenAngleBracket TypeParameter+:parameters __ CloseAngleBracket ->
+  OpenAngleBracket TypeParameter+:parameters __ CloseAngleBracket ->
     return {
       type: "TypeParameters",
       parameters,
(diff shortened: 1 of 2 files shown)


STRd6 (approved):
I think disallowing `A <T>` is acceptable.

edemaine:
I realize this restriction also removes a different ambiguity: `A <T>` could be a function call with a JSX argument. I think it now is.

edemaine:
By the way, I noticed that type arguments already forbid spaces before the `<`. So `func<T>` is valid but `func <T>` means `func(<T/>)`. So this change for type parameters actually seems to make things more symmetric.

PR #1017, opened 2024-02-17
Fix #1002 - Allow postfix loops, etc. in declarations

### LINKED ISSUE #1002: Postfix in declaration
I'd like to do
```
stuff := n*2 for n of [0..9]
```
(or with a newline after the const assignment if need be) but I can't seem to get anything like this to work. Am I missing anything?

STRd6:


source/parser.hera
```diff
@@ -5049,11 +5049,19 @@ LexicalDeclaration
     }
   # NOTE: Added const shorthand
   # NOTE: Using ExtendedExpression to allow for If/SwitchExpressions
-  InsertConst ( BindingPattern / BindingIdentifier ) TypeSuffix? __ ConstAssignment ExtendedExpression ->
+  InsertConst ( BindingPattern / BindingIdentifier ) TypeSuffix? __ ConstAssignment ExtendedExpression:expression ( _? PostfixStatement )?:post ->
```
edemaine, line 5052: Is there a reason you can't just replace `ExtendedExpression` with `PostfixedExpression`?

STRd6: Just didn't think of it, it's much better that way, thanks!

edemaine (approved):

PR #1018, opened 2024-02-17
Add type assignment shorthand

### LINKED ISSUE #176: `::=` type declaration shorthand
```typescript
Id ::= string | number
// 👇 ts
type Id = string | number

User ::=
  id: number
  name?: string

// 👇 ts
type User = {
  id: number
  name?: string
}
```

Suggested by @sultan99 with some modifications.

bbrk24:
Fixes #176 

Fairly straightforwards, adds `X ::= Y` as shorthand for `type X = Y`. For the tests I just picked a few existing tests using the `type` keyword and copied them but with `::=` instead.

### DIFF
--- a/source/parser.hera
+++ b/source/parser.hera
@@ -5064,6 +5064,11 @@ LetAssignment
   ".=" ->
     return { $loc, token: "=" }
 
+
+TypeAssignment
+  "::=" ->
+    return { $loc, token: "=" }
+
 # https://262.ecma-international.org/#prod-LexicalBinding
 # merged with https://262.ecma-international.org/#prod-VariableDeclaration
 LexicalBinding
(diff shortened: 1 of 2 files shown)


STRd6 (approved):
Looks great, thanks!

source/parser.hera
```diff
@@ -6599,6 +6604,14 @@ TypeAliasDeclaration
       ts: true,
     }
 
+  InsertType IdentifierName:id TypeParameters? __ TypeAssignment ( ( _? Type ) / ( __ Type ) ) ->
```
edemaine, line 6607: I think we'd ideally check for a nested tyoe here (with PushIndent Type? PopIndent, if there isn't a rule already). This is necessary to set indentation correctly. I think other assignments have a similar rule?

STRd6: The don't currently but they should.

STRd6: Oh, actually they do from `ExtendedExpression` -> ... -> `NestedNonAssignmentExtendedExpression`

STRd6: I think for this we could use `MaybeIndentedType` instead of `( ( _? Type ) / ( __ Type ) )`.

bbrk24: I copied `( ( _? Type ) / ( __ Type ) )` from the other formation of `TypeAliasDeclaration`. It's definitely not what I would have come up with on my own.

PR #1020, opened 2024-02-17
Fix #998 - Properly handle void async generators and iterators

### LINKED ISSUE #998: Extra return in async generator functions
Async equivalent of #862

Reproducer:
```civet
f1 := :AsyncGenerator<number, void> ->
  await Promise.resolve()
  yield 5

f2 := :AsyncIterator<number, void> ->
  await Promise.resolve()
  yield 5
```

STRd6:


source/lib.civet
```diff
@@ -2671,7 +2706,10 @@ function processReturn(f: FunctionNode, implicitReturns: boolean): void
     isMethod := f.type is "MethodDefinition"
     isConstructor := isMethod and name is "constructor"
     isVoid := isVoidType(returnType?.t) or
-      (async and isPromiseVoidType(returnType?.t)) or
+      (async and (
+        isPromiseVoidType(returnType?.t) or
+        isAsyncGeneratorVoidType(returnType?.t))
```
edemaine, line 2711: I think `isAsyncGeneratorVoidType` needs to be `and`ed with `generator`; otherwise, a function that calls another generator won't implicitly return it, as it should. For example, maybe add this test:

```js
function f(): AsyncGenerator<number, void>
  makeGenerator() |> await |> &()
---
async function f(): AsyncGenerator<number, void> {
  return (await makeGenerator())()
}
```

This might be a good candidate for our new indentation operators:
```js
isVoid := (or)
  isVoidType returnType?.t
  async and isPromiseVoidType returnType?.t
  async and generator and isAsyncGeneratorVoidType returnType?.t
  not async and generator and isGeneratorVoidType returnType.t
```

PR #1022, opened 2024-02-17
Fix #959 - Don't duplicate comments when hoisting refs

### LINKED ISSUE #959: Displaced `ts-expect-error` prevents error suppression
Let's say that `foo` and the return type of `bar` are nominally incompatible but I know what I'm doing is safe. So I write

```civet
// @ts-expect-error
foo ?=
  bar()
  ||> baz
  ||> qux
```

Civet inserts a statement between the `ts-expect-error` comment and the `foo ??= ...` statement, so the error suppression fails. In fact, it gives a second error, because I have an unused expect-error directive.

STRd6:


source/lib.civet
```diff
@@ -1491,21 +1491,26 @@ function hoistRefDecs(statements: StatementTuple[]): void {
...
-  }
+  { expressions } := block
+  index := expressions.findIndex (exp) =>
+    exp is node or (Array.isArray(exp) and exp[1] is node)
+
+  if index < 0
+    throw new Error("Couldn't find expression in block for hoistable declaration.")
+
+  let statement
+  if index > 0
```
edemaine, line 1504: I'm not sure why this condition changed from `if expressions[index] is node`. It seems like the code inside the `if` only works when `expressions[index]` is already an `[indent, exp, delimiter]` array, which I'm guessing `node` is not supposed to be? Whereas the `else` case is when there is no indentation to find. There's no use of `index-1` so `index > 0` doesn't seem relevant.

edemaine (approved):
OK, the intent makes sense now; we're either given the statement tuple or the expression within a statement tuple.

PR #1023, opened 2024-02-17
Typechecking CLI

### LINKED ISSUE #61: CLI for type checking
Vue has a cli utility called [vue-tsc](https://www.npmjs.com/package/vue-tsc), which can be used for type-checking in CI env etc. It would be good to have something similar for civet.

edemaine:
Following up on https://github.com/DanielXMoore/Civet/issues/61#issuecomment-1921882118, I realized that we don't need to duplicate the functionality of the unplugin: we can call the unplugin directly from the CLI!  (with a small amount of unplugin emulation)

image

* New `--typecheck` option runs the TypeScript compiler on all specified source files and outputs all diagnostics.
  * No sourcemapping to the original files; this is a missing feature in the unplugin. Added a TODO.
  * You need to specify all `.civet` files in one command line to check everything. So something like `civet --typecheck src/**/*.civet`.
  * Should `-t` be an alias for this? It seems important, but we might also eventually have a `--ts` feature to mirror the unplugin `ts` option, so maybe better to save it.
* New `--emit-declaration` option generates `.d.ts` files for all input `.civet` files. Actually only emits if `--typecheck` doesn't produce error diagnostics.
* Note that `-c`/`--compile` is orthogonal to the above options. If specified, we compile to `.tsx`/`.jsx` files, and this happens even if `--typecheck` emits error diagnostics. But you can also run typechecking without outputting any files, via `civet --typecheck`.
* I guess this fixes #61, though it's unresolved whether there should be a separate `bin` called `civetsc`/`tscivet`. Perhaps more crucially, we're missing a "typecheck everything" command that respects an `include`/`exclude` feature like `tsc` and `tsconfig.json`. That said, I imagine anyone using the CLI would be explicitly listing `.civet` files anyway for compilation, so maybe this is sufficient?

### DIFF
--- a/source/cli.civet
+++ b/source/cli.civet
@@ -420,7 +420,7 @@ function cli
           console.error error
           process.exit 1
 
-  process.exitCode = errors
+  process.exitCode = Math.min 255, errors
   if unplugin?
     try
       await unplugin.buildEnd.call {
@@ -429,7 +429,7 @@ function cli
       }
     catch error
       if match := (error as Error).message.match /Aborting build because of (\d+) TypeScript diagnostic/
-        process.exitCode = +match[1]
+        process.exitCode = Math.min 255, errors + +match[1]
       else
         process.exitCode = 1
         throw error
--- a/source/cli.civet
+++ b/source/cli.civet
@@ -50,7 +50,7 @@ if process.argv.includes "--help"
       --inline-map     Generate a sourcemap
       --no-cache       Disable compiler caching (slow, for debugging)
       --typecheck      Run TypeScript and output diagnostics
-      --emit-declaration  Run TypeScript and emit .d.ts declaration files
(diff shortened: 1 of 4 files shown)


source/cli.civet
```diff
@@ -387,4 +420,18 @@ function cli
           console.error error
           process.exit 1
 
+  process.exitCode = errors
```
STRd6, line 423: Problably want to add `Math.min(errors, 255)`. I think most shells wrap at 256.

bbrk24: Bash and (most?) other POSIX shells do, but notably PowerShell doesn't -- it uses 32-bit exit codes. Microsoft likes to put useful information in the high bytes of the exit code which can be annoying when invoking a .exe file from bash.

STRd6 (approved):

PR #1025, opened 2024-02-18
Made some more progress on internal typings

STRd6:


source/lib.civet
```diff
@@ -2509,37 +2544,36 @@ function processCoffeeInterpolation(s, parts, e, $loc) {
...
+    children: [pattern, suffix, initializer]
   }
 
-  const children = [decl, binding]
+  children := [decl, binding]
 
   makeNode {
     type: "Declaration",
-    names: id.names,
+    names: pattern.names,
```
edemaine, line 2576: This can be simplified to `pattern.names`

edemaine (approved):
Great progress!

PR #1026, opened 2024-02-18
Split lib.civet into many files

STRd6:
Hopefully it will be easier to move things around now that the biggest split is done.


source/lib.civet
```diff
@@ -6,7 +6,8 @@
  * here eventually.
  */
 
-type ASTNode = ASTNodeBase | ASTError | ASTRef | ASTLeaf | ASTNode[] | string | undefined
+type ASTNode = ASTNodeBase | ASTError | ASTRef | ASTLeaf | ASTNode[] | ASTString | undefined
+type ASTString = string & ASTNodeBase
```
edemaine, line 10: I don't understand this type definition... Strings don't have type etc. right? I could see adding `type: never, children: never` though.

source/lib.civet
```diff
@@ -4272,71 +4270,47 @@ function wrapIIFE(expressions: StatementTuple[], async): ASTNode[] {
 }
 
 export {
```
edemaine, line 4272: While we're discussing cleanup, one thing that would be nice I think is to just move `export` to the function declarations themselves. Then we will have this list in only one place, where we use it in lib.

edemaine (approved):

source/lib.civet
STRd6: It's a hack to fix the type errors on `gather .type is "SomeNode"` but it causes errors when inserting strings so probably a wash.

source/lib.civet
STRd6: It's a hack to fix the type errors on gather .type is "SomeNode" but it causes errors when inserting strings so probably a wash.

edemaine: Ah, I see. Maybe `type?: never` would be a better hack for strings? I think this would enable checking but without actually allowing it to exist...

source/lib.civet
STRd6: I don't necessarily want to export all of them, only the ones we use. Exporting them all would make dead code elimination slightly harder.

Issue #1007, opened 2024-02-15
Precedence of custom operators?


gwhitney:
My main intention here is to file a documentation bug; there doesn't seem to be any discussion of the precedence of custom operators in the Reference.

My concern here is that to deal with #1006 I am currently doing

`operator at<T>(A: T[], i: number) A.at(i) as T`

but this `at` has the wrong precedence: `[1,10,5] at 2-1` is 4, not 10 as it "should" be. Anything I can do? I.e., a secondary part of this issue would be ability to influence the precedence, at least among a few options, like maybe default (=medium), tight, and loose.

gwhitney:
P.S. Rather than opening another issue (sorry I have been piling on lately), I notice that unlike function definitions which are hoisted (if that's the right term), the operator declaration is only in effect subsequent to its occurrence. That's a pain for defining a bunch of utility operators at the bottom of a source file.  (But totally understandable if you are trying to keep Civet to one-pass parsing -- is that the case?)

edemaine:
Agreed that operators have a particular precedence, and I'm not sure it's the "right" one. And even better if we can override precedence. I'm not sure how other languages specify this; that's one problem. Another issue is that Civet doesn't really process precedence of built-in binary operators, leaving that to JS. But it would be relatively easy (I think) to make custom operators have lower precedence than arithmetic etc. operators, which seems more likely "right". Is this your sense? And plausibly also to specify custom operators' relative precedence.

Yes, operators' infix ability only apply to the future, and this is necessary for the one-pass parse. I don't think this is likely to change. Note that you can import operators from other files though, so consider a utility file? (The import will need to come at the top.)

edemaine:
I suspected `xor` had the wrong precedence too. I'm working on fixing that (along with the bug). The current plan is to put them at the bottom of everything that Civet considers a binary operator, i.e., just below `||`. Not totally sure whether `xor` should be above, below, or equivalent to `||`, but that seems roughly the correct location, and below everything is easiest to implement.

As I do this, I start to understand how we could implement custom operator precedence even without building a binary operator parse tree. Though I'm not sure which approach is better.

Would this bottom level be a good default level for custom operator precedence? Or is it more naturally above relational operators?

I'd definitely like to work out / see the right notation for precedence specification. Maybe something like this:

```js
operator foo > (+)
operator foo tighter (+)
operator bar == (*)
operator bar equiv (*)
operator baz looser (**)
operator baz < (**)
```

[...]

PR #1029, opened 2024-02-18
xor precedence above ||

edemaine:
This should probably wait until after #1026 gets merged, if this code is moving.

* Put `xor` at a precedence "at the bottom of what Civet considers a binary operator", i.e., just below `||`. Not sure whether it should instead be the same level as `||`.
* Fix a core issue in precedence mangling where a low-precedence operator wouldn't actually process the binary operators in the right-hand side. This should make it possible to implement any precedence in this way, in particular custom operator precedence.
* Fix a bug in precedence mangling where `i` would skip over binary operators if `start` moved back a bunch.


STRd6:
Big merge is done.

edemaine:
I realized that `^` has precedence between `|` and `&` in JavaScript, so I placed `^^` between `||` and `&&` instead of below `||`.

Given that the `^^` precedence was no longer extreme, I bit the bullet and implemented a general binary operator precedence system, which this was already close to. This should really make it easy to customize operator precedence, once we figure out a syntax. For now, I've also moved the default custom operator precedence to between relational and arithmetic. Previously, it was tighter than everything, which is pretty counterintuitive (see #1007).

Also a new `BinaryOp` type which is nice.

edemaine:
@STRd6 Do you want to review the changes? In particular, changing custom operator precedence is breaking, but probably in a good direction.

STRd6 (approved):
Looks good! This should help us when we add custom operator precedence at some point.

PR #1031, opened 2024-02-18
Custom operator precedence

### LINKED ISSUE #1007: Precedence of custom operators?

edemaine:
I'm not sure this syntax is perfect, and happy to have more back-and-forth here. But the cool thing is that the precedence mechanism works!

Fixes #1007


test/binary-op.civet
```diff
@@ -827,3 +827,79 @@ describe "custom identifier infix operators", ->
...
+      same as symbol
+      ---
+      operator mul same (*) (a, b)
+        a*b
+      a+b*c mul d*e+f
+      ---
+      function mul (a, b) {
+        return a*b
+      }
+      a+mul(b*c, d)*e+f
```
STRd6, line 842: If `mul` is the same precedence as `*` shouldn't it be:

`a+b*mul(c, d)*e+f`

edemaine: No, it's acting left-associative with `*`, so `b*c` goes first.

bbrk24:
Is there a way to specify associativity, or are custom operators always left-associative?

STRd6 (approved):
Happy to get this out experimentally while we can think about nicer syntax.

edemaine:
@bbrk24 Currently the engine only supports left associativity, so that's all for now. It should be easy to add right associativity (and invalid associativity) in the future.

PR #1033, opened 2024-02-19
Working towards discriminated union for nodes and type predicates for traversal

STRd6:


edemaine (approved):
I'm on mobile so am not seeing the TypeScript feedback, but overall these changes seem good. Definitely some nice type improvements! 

Is it still possible to use the gather functions if the predicate does not have an `is` return type? Or is that required now?

STRd6:
We could add an overload to allow for that. Currently you can use them without but it displays a TS error.

PR #1038, opened 2024-02-19
More custom operator precedence

edemaine:
Followup to #1031 to address https://github.com/DanielXMoore/Civet/issues/1007#issuecomment-1952386874

* `import { operator foo same (+) }` to specify precedence while `import`ing a custom `operator`
* `left` and `right` and `nonassoc` custom associativity as normal/requested (where `nonassoc` generated an error if you try to use associativity)
* Inspired by [Raku](https://docs.raku.org/language/functions#Associativity) two more:
  * `relational` folds in with our chained relational operators like `<` and `instanceof`
  * `multiarg` combines all operands into arguments: `a x b x c` → `x(a, b, c)`
  
I noticed that my operator syntax design has an ambiguity: `operator left` could mean "the following operators are left-associative" or it could mean "declare an operator `left`". Currently the former is preferred. Similarly with the words above, and `same`/`tighter`/`looser`. One approach, taken by Raku, is to add an `is` before `same`/`tighter`/`looser`/`left`/`right`/etc. Another approach would be to always put the modifiers after the identifier. Except maybe for declaration-style `operator {min} := Math` perhaps?


STRd6:
Putting the modifiers after the identifier would make it more consistent with the import format. We could eventually allow for `{x, operator y} := z` and similar.

For associativity maybe `left right none` (I don't like the Python style of all lowercase with arbitrary words shortened). `multiarg` could be `...` unless that might class with something later.

Looks great overall and I'm glad it was able to fit in with our existing capabilities in a reasonably straightforward way!

STRd6 (approved):

edemaine:
I was a bit uncomfortable with the arbitrary abbreviations. I thought `none` was a bit ambiguous because it's not clearly about associativity.  Perhaps `non` as in Raku would be a bit better?

Raku uses `list` instead of `multiarg`. We could use that, but it doesn't feel very JS (which has no "lists"). Maybe `arguments` would be a good JSism? (referring to the old variable-length arguments mechanism)

edemaine:
I've removed the capability to put operator behaviors before the operator name, removing the ambiguity.

The one exception is declarations, where I didn't want to write all the rules to implement `operator {min right, max same (+)} := Math` etc., so there's still just `operator right same (+) {min, max} := Math`. It's a bit inconsistent, but at least it's functional and not ambiguous.

STRd6 (approved):
Looks good, we can keep iterating as we test it out.

Issue #925, opened 2024-01-22
VSCode Extension CPU load 100%


mixty1:
I don't know why, but today i have 100% load cpu
Ext version: `v0.3.11`
![CleanShot 2024-01-22 at 18 41 39@2x](https://github.com/DanielXMoore/Civet/assets/9297404/16c3f2f9-3704-4692-8026-4b5c3a41884d)

mixty1:
I think it's because #918 issue happens. When i save file with a lot of object then extension is loads CPU to 100%

PR #1042, opened 2024-02-19
Snapshots similar to how Vue language tools does

STRd6:
I thought this might help with https://github.com/DanielXMoore/Civet/issues/925 and it might help a little but doesn't seem to be a big improvement.

If we do want better incremental parsing support in the future then we can build off of this.

### DIFF
--- a/lsp/source/lib/typescript-service.mts
+++ b/lsp/source/lib/typescript-service.mts
@@ -18,7 +18,6 @@ import type {
 } from "typescript"
 
 const {
-  ScriptSnapshot,
   createCompilerHost,
   createLanguageService,
   parseJsonConfigFileContent,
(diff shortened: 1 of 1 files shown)


edemaine (approved):
Looks interesting. Who calls getChangeRange? I assume VSCode does so and does something, like keep annotations from outside the changed range?

Someday we should look at rerunning the compiler only starting at the first changed byte, clearing cache that involves any later characters. I think this requires Hera changes though, to keep track of which bytes are examined to e.g. determine a negative match.

STRd6:
I don't think the bottleneck is in our compiler. Even plain non-transpiled TS files take ~1 second per 3k lines when getting semantic diagnostics:

```
Updating diagnostics for doc: file:///home/daniel/apps/civet/source/parser.ts
cached range { span: { start: 181724, length: 1 }, newLength: 0 }
 service.getSyntacticDiagnostics   47.39ms
  service.getSemanticDiagnostics 4175.97ms
service.getSuggestionDiagnostics   27.08ms
Updating diagnostics for doc: file:///home/daniel/apps/civet/lsp/source/lib/typescript-service.mts
 service.getSyntacticDiagnostics    0.48ms
  service.getSemanticDiagnostics    0.04ms
service.getSuggestionDiagnostics    0.69ms
Updating diagnostics for doc: file:///home/daniel/apps/civet/source/parser.hera
 service.getSyntacticDiagnostics    0.52ms
  service.getSemanticDiagnostics 3792.82ms
service.getSuggestionDiagnostics   29.53ms
```

Either we're misconfiguring something with our TS server or maybe there is a different way that the native TS extension handles it. We can dig around in their source to find out.

STRd6:
Just tested the builtin TS, it also takes the same amount of time (maybe longer, I think they have a lengthy debounce). So they're probably not doing anything clever.

PR #1044, opened 2024-02-20
Non-transpiled files weren't being added to the path map causing them…

STRd6:
… to not update


edemaine (approved):
Oh, that's great! I noticed things falling behind sometimes and having to restart server, but hadn't narrowed it down to .ts files. In hindsight I suspect this is the exact issue.

PR #1049, opened 2024-02-21
Fix #1047 - Properly escape newlines in multi-line strings

### LINKED ISSUE #1047: Problems with multi-line strings
According to the documentation:

> Strings can span multiple lines:
> ```civet
> console.log "Hello,
> world!"
> ```

I have found a number of cases where this doesn't work. These will be illustrated on the CLI using `printf`, to prevent errors such as `\r` being transformed to `\n` in the playground.

1. With two newlines in the same string:
   ```
   $ printf 'x := "a\nb\nc"\n' | npx civet
   const x = "a\nb
   c"
   ```
2. With Windows-style `\r\n` newlines. Interestingly, the mode of failure differs depending on the surrounding text:
   ```
   $ printf 'x := "\r\n"\r\n' | npx civet
   const x = "
   "
   $ printf 'x := "a\r\n"\r\n' | npx civet
   const x = "a
   \n"
   ```
3. With "Classic Mac OS" `\r` newlines:
   ```
   $ printf 'x := "\r"\r' | npx civet
   const x = "
   "
   ```
   This is probably the case we need to care about the least of the three.

STRd6:
I converted `\r\n|\n|\r` all to `\n`. It seems like the right thing to do.

### DIFF
--- a/source/parser/string.civet
+++ b/source/parser/string.civet
@@ -137,12 +137,14 @@ function processCoffeeInterpolation(s, parts, e, $loc)
   }
 
 /**
-Adjust a parsed string by escaping newlines
+Adjust a parsed string by escaping unescaped newlines.
 */
 function modifyString(str: string)
-  // Replace non-escaped newlines with escaped newlines
-  // taking into account the possibility of a preceding escaped backslash
-  return str.replace(/(^.?|[^\\]{2})(\\\\)*\n/g, '$1$2\\n')
+  str.replace /((?:\\.|[^\r\n])*)(\r\n|\n|\r)?/gsu, (_, chars, nl) ->
+    if nl
+      chars + '\\n'
+    else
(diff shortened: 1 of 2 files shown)


edemaine (approved):
I think this looks good.

Just checking: do we definitely want to normalize line endings to `"\n"`? Or should we preserve the input?

I checked CoffeeScript for comparison and it replaces newlines (including `\r\n`) with a single space! Unless you end the oine with a backslash, and then there's no space (or newline) in the output. I kind of understand their motivation: writing long single-line strings (e.g. Error message) while keeping your file nice and narrow. Should we consider this behavior?

STRd6:
We should consider the CoffeeScript behavior for a compat flag eventually. Preserving the input might be wanted but I'll wait until someone actually asks for it.

PR #1051, opened 2024-02-21
Suppress ESLint `no-cond-assign` with `if const` etc

bbrk24:
If the condition to an `if` or `while` statement is just an assignment and nothing else, this trips ESLint's `no-cond-assign` rule. While this should continue to trigger for Civet code like `if x = y`, `if x := y` feels intentional enough that it shouldn't trigger that error. ESLint allows you to suppress the error by doubling up the parens, so this PR doubles the parens in these cases.

The typings -- or, more frequently, lack thereof -- inside the compiler made this a little difficult to navigate at first, but I got there in the end.

### DIFF
--- a/source/parser/pattern-matching.civet
+++ b/source/parser/pattern-matching.civet
@@ -450,8 +450,7 @@ function processDeclarationCondition(condition, rootCondition, parent): void
     // More complex conditions (triggered by pattern matching or `until`/`unless`) don't need double parens
     // @ts-ignore Just because pattern might not have a type at runtime doesn't mean it's unsafe
     if pattern.type is "Identifier" and (grandparent?.type is "IfStatement" or grandparent?.type is "WhileStatement")
-      // steal the parens from the parent, forcing them to double up
-      [condition.parent.children.0, ref, initializer, condition.parent.children.-1]
+      ["(", ref, initializer, ")"]
     else
       [ref, initializer]
 
--- a/source/parser/pattern-matching.civet
+++ b/source/parser/pattern-matching.civet
@@ -448,22 +448,24 @@ function processDeclarationCondition(condition, rootCondition, parent): void
   children :=
     // Check that the declaration is a plain assignment (no pattern-matching) and the immediate grandchild of an `if` or `while`
     // More complex conditions (triggered by pattern matching or `until`/`unless`) don't need double parens
-    // @ts-ignore Just because binding.pattern might not have a type at runtime doesn't mean it's unsafe
-    if binding.pattern.type is "Identifier" and (grandparent?.type is "IfStatement" or grandparent?.type is "WhileStatement")
+    // @ts-ignore Just because pattern might not have a type at runtime doesn't mean it's unsafe
+    if pattern.type is "Identifier" and (grandparent?.type is "IfStatement" or grandparent?.type is "WhileStatement")
       // steal the parens from the parent, forcing them to double up
       [condition.parent.children.0, ref, initializer, condition.parent.children.-1]
     else
       [ref, initializer]
 
-  Object.assign condition,
+  Object.assign condition, {
     type: "AssignmentExpression"
-    children: children
+    children
     hoistDec:
       type: "Declaration"
       children: ["let ", ref, suffix]
       names: []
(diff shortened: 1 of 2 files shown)


source/parser/pattern-matching.civet
```diff
@@ -438,14 +439,25 @@ function processDeclarationCondition(condition, rootCondition, parent): void
...
   { pattern, suffix, initializer, splices, thisAssignments } := binding
 
+  grandparent := condition.parent?.parent
+  children :=
+    // Check that the declaration is a plain assignment (no pattern-matching) and the immediate grandchild of an `if` or `while`
+    // More complex conditions (triggered by pattern matching or `until`/`unless`) don't need double parens
+    // @ts-ignore Just because binding.pattern might not have a type at runtime doesn't mean it's unsafe
+    if binding.pattern.type is "Identifier" and (grandparent?.type is "IfStatement" or grandparent?.type is "WhileStatement")
+      // steal the parens from the parent, forcing them to double up
+      [condition.parent.children.0, ref, initializer, condition.parent.children.-1]
```
edemaine, line 454: I think it would be better to use `"("` and `")"` for the newly generated parens here. It feels dicey to re-use tokens which have specific `$loc`ations. Also nicer not to rely on the structure of `condition.parent`.

source/parser/pattern-matching.civet
```diff
@@ -438,14 +439,25 @@ function processDeclarationCondition(condition, rootCondition, parent): void
...
+    else
+      [ref, initializer]
+
   Object.assign condition,
     type: "AssignmentExpression"
-    children: [ref, initializer]
+    children: children
```
edemaine, line 460: Maybe worth wrapping the object in braces to cleanup `children`, `pattern`, and `ref`.

test/if.civet
```diff
@@ -859,8 +859,8 @@ describe "if", ->
       else
         b
       ---
-      let ref, y;x = (ref = f())?
-        {y} = ref,a
+      let ref, y;x = ((ref = f()))?
+        ({y} = ref,a)
```
edemaine, line 863: Hmm, maybe these are two settings where we should discriminate based on parent or grandparent?
* if expressions already wrap the condition in parens when writing `?:`. Ideally we could detect this situation and not double wrap here...
* I didn't expect the `then` clause to be wrapped. Does eslint want it to be?

source/parser/pattern-matching.civet
```diff
@@ -438,14 +439,25 @@ function processDeclarationCondition(condition, rootCondition, parent): void
...
   // TODO: Add support for `let` and `const` declarations with multiple bindings in conditions
   binding := bindings[0]
   { pattern, suffix, initializer, splices, thisAssignments } := binding
 
+  grandparent := condition.parent?.parent
+  children :=
+    // Check that the declaration is a plain assignment (no pattern-matching) and the immediate grandchild of an `if` or `while`
+    // More complex conditions (triggered by pattern matching or `until`/`unless`) don't need double parens
+    // @ts-ignore Just because binding.pattern might not have a type at runtime doesn't mean it's unsafe
+    if binding.pattern.type is "Identifier" and (grandparent?.type is "IfStatement" or grandparent?.type is "WhileStatement")
```
edemaine, line 452: Are there grandparent settings where we don't want to do this, given that we're in a declaration condition? Mostly I'm curious, but we could potentially also remove the use of `grandparent` altogether. (See also comment below.)

Also, `binding.pattern` is already available as `pattern`.

test/if.civet
bbrk24: If the `then` clause isn't wrapped, it's actually not valid JS, because of the precedence of the comma operator relative to the ternary.

bbrk24: To answer the first question though, ESLint requires double-wrapping with `?:` as well as `if` and `while`.

source/parser/pattern-matching.civet
bbrk24: Ah, okay. I wasn't sure if strings would be accepted and I didn't know how to fake the location data.

edemaine (approved):
Nice fixes, thanks!

PR #1060, opened 2024-02-24
Fix Playground around top-level await

edemaine:
* Fix preamble detection to not e.g. eat `"x"` in `"x" |> console.log`
* Fix detection of coffeeCompat mode
* Fix coffeeCompat `do` (needed to be wrapped in parens)
* Catch errors during promise and output them
* Format JS parse errors for output (escape \n)

### DIFF
--- a/civet.dev/public/playground.worker.js
+++ b/civet.dev/public/playground.worker.js
@@ -8,73 +8,14 @@ onmessage = async (e) => {
   const highlighter = await getHighlighter();
   const inputHtml = highlighter.codeToHtml(code, { lang: 'coffee' });
 
+  let tsCode, ast, errors = []
   try {
-    let ast = Civet.compile(code, { ast: true });
-    const errors = [];
-    let tsCode = Civet.generate(ast, { errors });
+    ast = Civet.compile(code, { ast: true });
+    tsCode = Civet.generate(ast, { errors });
     if (errors.length) {
       // TODO: Better error display; copied from main.civet
       throw new Error(`Parse errors: ${errors.map($ => $.message).join("\n")}`)
     }
-    let jsCode = '';
-
-    if (jsOutput) {
-      // Wrap in IIFE if there's a top-level await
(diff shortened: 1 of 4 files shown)


civet.dev/public/playground.worker.js
```diff
@@ -8,31 +8,55 @@ onmessage = async (e) => {
...
+    // Wrap in IIFE if there's a top-level await
+    // Use Civet's async do, so Civet does implicit return of last value
+    try {
       const topLevelAwait = Civet.lib.gatherRecursive(ast,
         (n) => n.type === 'Await',
         Civet.lib.isFunction
       ).length > 0
       if (topLevelAwait) {
-        const prefix = /^(\s*|;|\'([^'\\]|\\.)*\'|\"([^"\\]|\\.)*\"|\/\/.*|\/\*[^]*?\*\/)*/.exec(code)[0]
+        // Eat prologue strings and comments
```
STRd6, line 46: It's not necessary to make this change but you may be able to use the actual parser grammar to split out the prologue strings similar to how we use `ReservedWord` in https://github.com/DanielXMoore/Civet/blob/0423ac91ec9ddd68ad7d8c2c1fb6c41cb93f9538/source/parser/pattern-matching.civet#L412

STRd6 (approved):

civet.dev/public/playground.worker.js
edemaine: Nice idea! To avoid crazy imports, I ended up doing it more like the CLI did, with `parse` and `startRule`:

https://github.com/DanielXMoore/Civet/blob/0423ac91ec9ddd68ad7d8c2c1fb6c41cb93f9538/source/cli.civet#L127-L130

But lacking a system for really matching a prefix (`fail` seems pretty weak... we don't want to include failed partial matches), I added a new rule to make this easy. Let me know what you think!

STRd6 (approved):

Issue #909, opened 2024-01-03
array access helpers


unlessgames:
It's very common that you want to wrap around arrays and it's also quite common (at least in games and creative coding) that you want a random element from an array. It would be nice to have some concise syntax to do these kinda things. 

We already have `%%` which is very nice but how about:

```coffee
array[x%] ~> array[x %% array.length]

array[%] ~> array[Math.floor(Math.random() * array.length)]

array[?] ~> array.length > 0

array[#] ~> array.length
```

STRd6:
What about:

```
array[#?] // array.length > 0
// indexing out of array bounds can cause perf deoptimizations
array[?] // let l = array.length; l ? array[Math.floor(Math.random() * l)] : undefined 
```

`array[x%]` looks good as is.

unlessgames:
I'd guess asking for the length would be more used than wanting a random element so I'd give the shorter version for that, but if `array[#]` was included as length shorthand then it would be consistent to have `array[#?]` be the length check.

Or if it's going to be 2 characters anyway, how about allowing comparing the length like
```coffee
array[>0] ~> array.length > 0
array[<4] ~> array.length < 4
array[==12] ~> array.length == 12
```

STRd6:
What about `#` as a general shorthand for `.length` anywhere? That way it will work with strings, arrays, and anything else. It also plays nicely with `arary#<4`. A standalone `#` will be similar to how we resolve private identifiers with an implicit `this` becoming so `#` becomes `this.length`.

```
array# // array.length
"a string also"# // "a string also".length
function push(item)
  @[#] = item // return this[this.length] = item
function wrap(index)
  @[index %% #] // return this[mod(index, this.length)]
```

edemaine:
This looks fun! In particular, this makes it easy to work with the array length, e.g. `array# > 1` means `array.length > 1`.

Perhaps we also want `#?` to mean `.length > 0`, so that we can easily convert lengths to booleans?

`#` as `this.length` is nice in the examples you showed where working with `this`, but maybe that's a bit rate. I also kind of expected the following:
* `array[#-1]` to be `array[array.length-1]`
* `array[i % #]` to not be `array[i % array.length]`

So an alternative would be for a bare `#` to mean the length of the thing left of the current enclosing `[`. Or is this too magical...?

STRd6:
I think having `#` be the length of the object to the left of the current enclosing `[` is just magical enough ✨ 

It might not get in in the first version I'm working on today, however.

[...]

PR #1062, opened 2024-02-24
Add # length shorthand #909

STRd6:
This add `x.#` as a shorthand for `x.length`. It doesn't support without the `.` because our current implementation of private identifiers doesn't support that yet either. We can add it for both at some point in the future.

See #909

### DIFF
--- a/source/parser.hera
+++ b/source/parser.hera
@@ -975,14 +975,27 @@ ThisLiteral
       thisShorthand: true,
     }
   AtThis
-  PrivateThis
+  HashThis
 
 # Adds a #id -> this.#id shorthand as a "PrivateThis" literal in most cases
 # with the exception of keeping `#a in b` as is for branded in checks.
 # https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/in#using_the_in_operator_to_implement_branded_checks
 # This is slightly different from how https://262.ecma-international.org/#sec-relational-operators handles relational operators
 # grammatically but should be equivalent in practice
-PrivateThis
+HashThis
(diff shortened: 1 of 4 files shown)


source/parser.hera
```diff
@@ -975,14 +975,27 @@ ThisLiteral
...
 
-# Adds a #id -> this.#id shorthand as a "PrivateThis" literal in most cases
+# Adds a #id -> this.#id shorthand as a "HashThis" literal in most cases
 # with the exception of keeping `#a in b` as is for branded in checks.
 # https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/in#using_the_in_operator_to_implement_branded_checks
 # This is slightly different from how https://262.ecma-international.org/#sec-relational-operators handles relational operators
 # grammatically but should be equivalent in practice
-PrivateThis
+HashThis
+  LengthShorthand:id ( &( _+ ( ( Not __ In ) / In ) ) "" )?:beforeIn ->
```
edemaine, line 986: Is there a reason you had to add `""` here?

edemaine (approved):
Could you add documentation to `reference.md`? Or I can add it.

source/parser.hera
STRd6: https://github.com/DanielXMoore/Hera/issues/16

edemaine: Ah, I see. Fun. I guess alternatively we could split into two rules, like `PrivateIdentifier`. No strong feeling either way.

STRd6: I generally prefer it as a single rule to prevent backtracking.

PR #1064, opened 2024-02-26
Allow arguments to property bind

### LINKED ISSUE #954: Arguments to bind shorthand
`x@y z` becomes `x.y.bind(x)(z)`. I think `x.y.bind(x, z)` would be a more useful interpretation of that statement.

bbrk24:
Title. A weird side effect is that `x@y()` is now equivalent to `x@y`.

Closes #954


STRd6:
Looks great, thanks!

PR #1067, opened 2024-02-27
Sourcemap improvement

### LINKED ISSUE #1053: Tooltip error with `if const`
Minimum reproducer: https://github.com/bbrk24/Civet-LSP-Tooltip-Bug

```civet
if signal := signals.get funcName
```

If I mouse over the first three characters of `signals`, I get the tooltip for the generated ref. If I mouse over the right half is `signals`, or over `get` or `funcName`, I get no tooltip at all.

STRd6:
Improved forward sourcemapping by resetting the best column when advancing the best line.

Also fixed lsp tests by removing reference to .coffee file in integration test.

Fixes #1053

### DIFF
--- a/lsp/source/lib/util.mts
+++ b/lsp/source/lib/util.mts
@@ -315,7 +315,7 @@ export function forwardMap(sourcemapLines: SourcemapLines, position: Position) {
         // find best line without going beyond
         const [_p, _f, srcLine, srcOffset] = mapping
         if (srcLine <= origLine) {
-          if (srcLine >= bestLine && (srcOffset <= origOffset) && (srcOffset >= bestOffset)) {
+          if (srcLine > bestLine && (srcOffset <= origOffset) || srcLine === bestLine && (srcOffset <= origOffset) && (srcOffset >= bestOffset)) {
             bestLine = srcLine
             bestOffset = srcOffset
             foundLine = i
--- a/lsp/integration/project-test/a.civet
+++ b/lsp/integration/project-test/a.civet
@@ -1,5 +1,4 @@
 b, {b as b2} from ./b.civet
-c from ./c.coffee
 h from ./h.hera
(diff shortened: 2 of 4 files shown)


lsp/source/lib/util.mts
```diff
@@ -315,7 +315,7 @@ export function forwardMap(sourcemapLines: SourcemapLines, position: Position) {
         // find best line without going beyond
         const [_p, _f, srcLine, srcOffset] = mapping
         if (srcLine <= origLine) {
-          if (srcLine >= bestLine && (srcOffset <= origOffset) && (srcOffset >= bestOffset)) {
+          if (srcLine > bestLine && (srcOffset <= origOffset) || srcLine === bestLine && (srcOffset <= origOffset) && (srcOffset >= bestOffset)) {
```
edemaine, line 318: Should `srcOffset <= origOffset` be moved to the previous `if`, given that it needs to hold in all cases?

I admit I don't understand why `srcLine <= origLine` or `srcOffset <= origOffset` are requirements...

STRd6: The normal direction for sourcemapping is reverse, given a position in the generated file it points to a position in the source file.

To do the opposite and find the position in the generated file from the source file ("forward map") we want to find the "best" mapping. The best mapping is the closest mapping with line <= original line (ideally it will be equal to the original line), and column <= (original column).

To find that mapping we check every reverse mapping holding on to the best one so far. If we're in the middle of an identifier and the best one begins a few characters before the original column that is probably fine since we don't map into the middle of identifiers. We linearly advance the found line and offset by the difference.

It's all slightly complicated but hopefully that explains it adequately.

edemaine: Thanks, that helps. Perhaps this could be added as a comment?

PR #1069, opened 2024-02-27
Fix #1065 - pipe to unaray word ops; await ops

### LINKED ISSUE #1065: Inconsistency of pipe assignment with unspaced prefix word operator and optional-chained method
Apologies for the long title, it's the best way I could think to describe this.

This compiles as expected:
```js
foo.bar |>= not?.baz()
// vvv
foo.bar = !foo.bar?.baz()
```
This is a ParseError:
```js
foo.bar |>= await?.baz()
```
And these invent new identifiers:
```js
foo.bar |>= void?.baz()
// vvv
foo.bar = voidfoo.bar?.baz()
```
```js
foo.bar |>= typeof?.baz()
// vvv
foo.bar = typeoffoo.bar?.baz()
```

(Of the four, `await` is the only one that does not generate the expected compilation when a space is inserted before the `?`.)

STRd6:


source/parser/unary.civet
```diff
@@ -53,19 +53,24 @@ function processUnaryExpression(pre, exp, post)
...
+    last := pre[l - 1]
+    if last.type is "Await"
+      if last.op
+        if exp.type !== "ParenthesizedExpression"
+          exp = ["(", exp, ")"]
+
+        exp =
+          type: "CallExpression",
+          children: [...last.children, "Promise", last.op, exp]
+        pre = pre.slice(0, -1)
```
edemaine, line 66: `pre.pop()` (in place) would probably be more efficient. Or are we not allowed to modify here?

(also 5 lines down)

test/pipe.civet
```diff
@@ -504,6 +512,24 @@ describe "pipe", ->
...
+      foo.bar = await foo.bar?.baz()
+      foo.bar = void foo.bar?.baz()
+    """
+
+    testCase """
+      snug unary ops with await op
+      ---
+      foo.bar |>= await.all?.baz()
+      ---
+      foo.bar = await Promise.all(foo.bar?.baz())
```
edemaine, line 530: These snug cases look a little crazy to me... But at least they work.

edemaine (approved):

source/parser/unary.civet
STRd6: We shouldn't modify here since this is called from the parser sometimes. There's usually only zero or one await ops per expression as well.

edemaine: Perhaps we should start adding `readonly` type annotations for functions we call from the compiler. I believe this can implement checks for this sort of thing.

PR #1074, opened 2024-02-29
#202 - Don't wrap StatementExpressions in IIFE in declaration

STRd6:
This is about 50% of #202 

Most of the work was refactoring `pushRef` to be more general `assignResults` to handle pushing to a collection or assigning to a ref.

Things remaining:

- iteration statement expressions
- if statement expression
- single assignment expression rhs

Much later:

- arbitrarily nested expressions (need to unwrap and maintain refs to keep execution order the same)
- arbitrarily chained assignments (ditto)

### DIFF
--- a/source/parser.hera
+++ b/source/parser.hera
@@ -49,7 +49,6 @@ import {
   reorderBindingRestProperty,
   replaceNodes,
   typeOfJSX,
-  wrapIIFE,
 } from "./parser/lib.civet"
 
 /**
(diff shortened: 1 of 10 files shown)


edemaine (approved):
I was wondering how you were going to deal with the precedence of something like `x := y++ + switch y ...` but now I see those aren't unwrapped, and compiled as before.

Overall, this definitely seems like a nice first step in this direction! The whole process is a little daunting, but maybe we can chip away at most important cases like this until it's all done.

Maybe it's worth adding a test that demos the #202 feature, like:

```coffee
x := switch y
  <? 'number'
    y+1
  else
    return 'bad'
```

Note that `if ... return` of #202 still isn't supported, because we don't seem to allow expressionizing `if` statements... Personally I'd find that more useful than `switch`, but hopefully it's not too much harder than what's in this PR. (I guess we should try to parse an IfExpression first, and on failure, parse an IfStatement that worst-case gets IIFE wrapped?)

PR #1089, opened 2024-03-06
Fix #1075 - Pattern matching array length type guard

### LINKED ISSUE #1075: `switch` over tuple doesn't type-narrow `else` branch
```
declare const m: [0, null] | [1, number]

switch m
  [0, null]
    ;
  [1, x]
    console.log x
  else
    m satisfies never
    // Error: Type '[0, null] | [1, number]' does not satisfy the expected type 'never'.
```

STRd6:


edemaine (approved):
Overall this seems good, and a nice workaround of a feature that *should* be in TS.

I worry a bit about the inefficiency of an added function call in pattern matching. I wonder whether we should use a guard function only in TS, and use the old compilation in JS?

source/parser.hera
```diff
@@ -7523,6 +7523,19 @@ Reset
         // [indent, statement]
         module.prelude.push(["", [preludeVar, isRef, typeSuffix, " = Object.is", asAny, ";\n"]])
       },
+      /**
+       * Array length check with type guard.
+       * From bbrk24 https://github.com/DanielXMoore/Civet/issues/1075#issuecomment-1971897483
```
bbrk24, line 7528: As much as I appreciate credit, notice the Discord link in that comment -- I didn't come up with this.

STRd6:
The function is as inline-able as it gets so it probably won't have much of an impact.

PR #1099, opened 2024-03-16
Fix #1097 - for loop over character range

### LINKED ISSUE #1097: for loop over letters
The following two examples compile to the same incorrect loop:

```js
for letter of ['a'..'z']
for each letter of ['a'..'z']
↓↓↓
for (let i = 'a'; i <= 'z'; i++) {
```

Sadly `i++` doesn't work here. (TypeScript helps discover this.)

On the other hand, `['a'..'z']` and longer ranges like `['0'..'z']` have the code we need for these cases. (I may have actually broken these cases when I optimized `for..of` over ranges and `for each..of`. If so, oops!) Similarly, we should compile to something like:

```js
for (let i = 97; i <= 122; i++) {
  const letter = String.fromCharCode(i);
```

Workaround for now: `for each letter of (['a'..'z'])`

STRd6:


edemaine (approved):
Looks great, thanks for the quick fix!

Maybe also worth adding a test for `for each..of`? Output should be identical.

PR #1104, opened 2024-03-17
Fix #1100 - pipe inside StatementExpression

### LINKED ISSUE #1100: Can't use `||>` inside `try` expression
MWE:

```
fooResult := try
  foo()
    ||> bar
```

> <pre>Error: Couldn't find block to hoist declaration into.
>     at .../node_modules/@danielx/civet/dist/main.js:670:13
>     at Array.forEach (&lt;anonymous&gt;)
>     at hoistRefDecs (.../node_modules/@danielx/civet/dist/main.js:661:53)
>     at processProgram (.../node_modules/@danielx/civet/dist/main.js:3913:3)</pre>

STRd6:


edemaine:
I wonder if `parent` assignment should go in `assignResults` instead?

STRd6:
Inside `assignResults` is probably a little bit better.

STRd6:
Even though it's better it's more work than I'm going to do right now so I'll merge this and hopefully clean it up more later.

PR #1109, opened 2024-03-21
Fix #1102 - negative index assignment lhs

### LINKED ISSUE #1102: `x.-1` on left-hand side
`x.-1 = 5` compiles to invalid `x.at(-1) = 5`. We probably want to compile to `x[x.length-1] = 5` (with appropriate ref if needed).

STRd6:


source/parser.hera
```diff
@@ -1273,31 +1274,36 @@ PropertyAccess
...
       type: "Index",
       children: [
-        dot,
+        start,
         literal,
         "]",
       ]
     }
 
   # NOTE: Added shorthand x.-1 -> x.at(-1)
```
edemaine, line 1287: This comment should probably change.

source/parser/lib.civet
```diff
@@ -843,7 +858,37 @@ function processStatementExpressions(statements: StatementTuple[]): void
...
+    .forEach (exp) =>
+      { parent } := exp
+      index := parent.children.indexOf(exp)
+      let ref, subexp
+      if index is 1
+        child := parent.children[0]
+        ref = maybeRef(child)
+        if ref !== child
+          subexp = parent.children.splice 0, 1
+      else
```
edemaine, line 872: Maybe worth checking for and throwing if index < 0

test/property-access.civet
```diff
@@ -279,31 +279,39 @@ describe "property access", ->
...
+      x[x.length-1]
     """
 
     testCase """
       optional negative
       ---
       x?.-1
       ---
-      x?.at(-1)
+      x?.[x.length-1]
```
edemaine, line 306: This test seems to be a duplicate of above. (Probably my bad.) Maybe it's supposed to be testing `x?'s -1`?

edemaine (approved):
Seems a bit excessive to have a separate pass for this, but I imagine we can optimize that in the future.

PR #1111, opened 2024-03-21
Fix #1101 - reserved word object globs

### LINKED ISSUE #1101: Keyword in object property glob
`state.{swap,insert,delete}` and `{state.{swap,insert,delete}}` fail to parse because `delete` is a keyword, while `{state.delete}` parses fine.

STRd6:


edemaine (approved):
Thank you!

PR #1128, opened 2024-03-31
Modern NodeJS register and CLI require/import fixes

edemaine:
* `@danielx/civet/register` now calls Node 20.6.0+'s `module.register` method to register ESM, in addition to CJS registration.
* New recommended method for ESM and CJS registration (for Node 20.6.0+):
  `node --import @danielx/civet/register`
  (avoids new warning ``ExperimentalWarning: `--experimental-loader` may be removed in the future; instead use `register()` ``)
* Removed duplicate CJS registration code from ESM loader; should use `@danielx/civet/register` for this
* CLI runs `@danielx/civet/register` even in REPL mode, so you can `require` or `import` Civet files interactively (one catch: I couldn't get `import` to work on Node <20.6.0)
* CLI provides `importModuleDynamically` to enable dynamic `import()`
* CLI supports top-level `await` (like Node CLI), based on previous code written for the playground

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,10 +22,9 @@ The modern way to write TypeScript.
   [esbuild](source/esbuild-plugin.civet),
   [Vite](https://github.com/edemaine/vite-plugin-civet),
   -->
+  [ESM/CJS loader](source/esm.civet),
   [Babel](source/babel-plugin.mjs),
   [Gulp](integration/gulp),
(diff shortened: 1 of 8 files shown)


STRd6 (approved):
Thanks! Maybe the nightmare of ESM is finally behind us.

PR #1130, opened 2024-03-31
Dynamic import declarations and expressions

### LINKED ISSUE #727: non-toplevel `import`ing
```ts
async function lazyUseLibrary
  import { getX } from library
  return getX()
// to
async function lazyUseLibrary() {
  const { getX } = await import("library");
  return getX();
}
```
note that things like `assert { type: "json" }` still work: `await import('./foo.json', { assert: { type: 'json' } })`

edemaine:
Fixes #727:
* `import { getX } from library` → `const { getX } = await import('library')` if not at the top level
* `import lib, {getX} from library` → `const lib = await import('library'), {getX} = lib` if not at the top level
* `stuff := import { getX, getY } from library` → `const stuff = (await import('library')).{getX, getY}`

This complements existing functionality of `stuff := import 'library'` → `stuff := import('library')`. Note that `stuff := import library` has a different meaning than above (use the variable `library` instead of the string).

~~I haven't implemented `assert`/`with` yet. I probably should...~~ Import assertions via `assert`/`with` should work now.

Also fixes a bug where `{x: y} := obj` got implicitly returned as `{x}`, but it should be `{y}` (!).

### DIFF
--- a/civet.dev/reference.md
+++ b/civet.dev/reference.md
@@ -1823,10 +1823,37 @@ import {X: LocalX, Y: LocalY} from "./util"
 ### Dynamic Import
 
 If it's not the start of a statement,
-dynamic `import` does not require parentheses.
+dynamic `import` does not require parentheses:
 
 <Playground>
 {x} = await import url
+</Playground>
+
+### Dynamic Import Declarations
+
+If you're not at the top level, `import` declarations get transformed
+into dynamic imports:
+
+<Playground>
+function load
+  * as utils from ./utils
+  { version: nodeVer,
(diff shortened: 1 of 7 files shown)


bbrk24:
> `import lib, {getX} from library` → `const lib = await import('library'), {getX} = lib` if not at the top level

Shouldn't this be
```js
const ref = await import('library'),
  { getX, default: lib } = ref;
```
?

STRd6 (approved):
Thanks, this looks like a very useful feature that makes dynamic imports much easier to use.

PR #1133, opened 2024-04-01
Hoist declarations out of async wrapper and other cleanup

edemaine:
Previously, top-level `await` in the CLI or Playground caused code to get wrapped in a wrapper that hid all declared variables. This PR hoists those declarations into top-level `var`s so that they can get accessed in future operations. (It does lose the `const`/`let`ness of the variables, but this seems a far better loss, and `const` is mostly annoying in a REPL environment.)

As a bonus, top-level `import` declarations also get wrapped to use the new dynamic import declarations, so they appear to just work as usual. You can now `fs from fs` in the CLI and then access `fs.readFileSync` in the next operation. Take that, Node REPL! (`import fs from 'fs'` fails in Node, with an error message to suggest instead writing `const { default: fs } = await import("fs");`)

Along the way, I did some cleanup that ended up not being relevant, but seems worth including:
* Blocks are more consistent about keeping `expressions` in the `children`
* Removed `insertReturn` that sometimes caused a double `return` at the end of an `async do` block, once `do` blocks made a correct call to `wrapIIFE`.
* Added `ast: "raw"` option to return raw AST. Took me a long time to understand why length-1 arrays were getting unarrayed until I remembered that the returned AST was pruned. I didn't end up using an unpruned AST, but it could be useful for dealing with `expressions` for example.
* Added more compiler options to `types.d.ts`

### DIFF
--- a/source/cli.civet
+++ b/source/cli.civet
@@ -282,10 +282,7 @@ export function repl(options: Options)
         // Wrap in IIFE if there's a top-level await
         // Use Civet's async do, so Civet does implicit return of last value
         // (copied from playground.worker.js)
-        topLevelAwait := lib.gatherRecursive(ast,
-          .type is 'Await',
-          lib.isFunction
-        ).length > 0
+        topLevelAwait := lib.hasAwait ast
         if topLevelAwait
           [prologue, rest] := parse input, startRule: 'ProloguePrefix'
           prefix := input.slice 0, -rest.length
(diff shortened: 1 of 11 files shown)


STRd6 (approved):
Thanks!

source/cli.civet
```diff
@@ -322,11 +327,13 @@ export function repl(options: Options)
...
           // If there was a top-level await, the result is a promise
           // that we need to await before returning it.
+          let threw = false
           try
             result = await result
           catch error
+            threw = true
             callback error as Error, undefined
-          callback null, result
+          callback null, result unless threw
```
vendethiel, line 336: isn't that the same as doing `callback null,result` inside the `try`?

STRd6: If the callback throws then the catch would also trigger if it were inside the `try`.

PR #1144, opened 2024-04-03
Fix #1043 - Add missing unary numeric and undefined pin patterns in switch

### LINKED ISSUE #1043: Missing patterns
`switch` pattern doesn't support the following patterns (or subpatterns):

* [ ] `-1`
* [ ] `+1`
* [ ] `undefined` (but `^undefined` works...)

STRd6:
There may be a better way to handle the unification of unary numeric literals but this seems good for now.

### DIFF
--- a/source/parser.hera
+++ b/source/parser.hera
@@ -1630,6 +1630,20 @@ PinPattern
       children: [expression],
       expression,
     }
+  # Handle unary numeric in switch patterns
+  ([+-] NumericLiteral):expression ->
+    return {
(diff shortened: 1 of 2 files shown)


edemaine (approved):
Looks good! I could imagine that a signed literal could even be in the literal category, but I guess literal vs. pin doesn't make much difference.

In the future, I could imagine supporting an arbitrary math expression over literals (e.g. 2**16 could be useful) but not nearly as important.

PR #1151, opened 2024-04-05
Partial application placeholders

STRd6:
`.` placeholder for partial function application.

Works for the common cases but `...` is a bit strange so I'll save that for a future iteration.

### DIFF
--- a/test/partial-placeholder.civet
+++ b/test/partial-placeholder.civet
@@ -24,3 +24,82 @@ describe "partial placeholder", ->
     ---
     $ => f($, $)
   """
+
+  testCase """
+    null chained call
+    ---
+    f()?.g?.().h ., x
+    ---
+    $ => f()?.g?.().h($, x)
+  """
+
+  testCase """
+    object value
+    ---
+    f a: .
+    ---
+    $ => f({a: $})
+  """
+
+  testCase """
+    object key
+    ---
+    f .: a
+    f {.: a}
+    ---
+    $ => f({[$]: a});
+    $1 => f({[$1]: a})
+  """
+
+  testCase """
+    complex nested object and array
+    ---
+    f a: [ x, b: . ]
+    ---
+    $ => f({a: [ x, {b: $} ]})
+  """
+
+  testCase """
+    nested call
+    ---
+    f .(a, b)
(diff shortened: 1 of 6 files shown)


edemaine:
Very cool! Looking at the last example, I wonder if `f ., .` should in fact mean `(a, b) => f(a, b)` instead of `(a) => f(a, a)`. I could be wrong, but I feel like the multiple arguments is a more common/useful interpretation of a partial. Originally I had thought we could use `.0` and `.1` (by analogy to `$0` and `$1`) but I realize those already have a meaning (number literal).

test/partial-placeholder.civet
```diff
@@ -0,0 +1,142 @@
...
+    $ => f(a, $)
+  """
+
+  testCase """
+    both args
+    ---
+    f ., .
+    ---
+    $ => f($, $)
+  """
```
edemaine, line 26: As I mentioned above, this is the only example I question. Though I believe I originally suggested this behavior, I wonder whether `($0, $1) => f($0, $1)` would be more useful. Granted, it's no longer an ampersand function this way, so we'd need to use different code when there's more than one argument.

Hmm, thinking more, maybe there are applications for repeated first argument. We're always wondering how to re-use an `&`. You could imagine something like:
```js
f
  key: .
  value: .
```

But I'm not totally sure which is more useful. The behavior you've implemented is more consistent with the future plans for `&`, which is both good (consistency) and bad (different behaviors might collectively cover more cases).

edemaine (approved):
Looks great! Just unsure what the best behavior is for repeated `.`s on the same expression...

test/partial-placeholder.civet
STRd6: I'm leaning towards every `.` reference being the single argument because it combines better with pipelines. Also once we get multiple arguments it is more complex and the benefit is reduced compared to an explicit arrow function (ex. how would one reverse the first and second arguments?). There's definitely room to explore further in the future though.

edemaine: Sounds good. I guess we'll have `...[3]` if we really want. And perhaps more practically, we'll hopefully have `&3`.

PR #1157, opened 2024-04-08
Preserve Vite's default extensions

edemaine:
It turns out the default behavior of our Vite unplugin disables all of Vite's default resolve extensions, including e.g. `.jsx` which was breaking a project I have that uses external JSX libraries (written in Solid, so distributes `.jsx` files).

### DIFF
--- a/integration/unplugin/src/index.ts
+++ b/integration/unplugin/src/index.ts
@@ -14,6 +14,7 @@ import * as tsvfs from '@typescript/vfs';
 import type { UserConfig } from 'vite';
 import type { BuildOptions } from 'esbuild';
 import os from 'os';
+import { DEFAULT_EXTENSIONS } from './constants.js';
 
 // Copied from typescript to avoid importing the whole package
 enum DiagnosticCategory {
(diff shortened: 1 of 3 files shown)


integration/unplugin/src/index.ts
```diff
@@ -458,7 +458,11 @@ export const rawPlugin: Parameters<typeof createUnplugin<PluginOptions>>[0] =
 
         if (implicitExtension) {
           config.resolve ??= {};
-          config.resolve.extensions ??= [];
+          config.resolve.extensions ??= [
+            // Vite's DEFAULT_EXTENSIONs from
+            // https://github.com/vitejs/vite/blob/main/packages/vite/src/node/constants.ts#L29
+            '.mjs', '.js', '.mts', '.ts', '.jsx', '.tsx', '.json',
```
STRd6, line 464: Could we import this from Vite if present or would it be too much of a hassle?

edemaine: It's not exported from Vite. I was thinking of submitting a Vite PR to expose it.

But there's also an issue of adding Vite as a dependency. Is there a nice way to copy the constant over with just a dev dependency? I guess I could do it in a build script without much trouble.

edemaine: I succeeded in getting this to work via `import { DEFAULT_EXTENSIONS } from '../../../node_modules/vite/dist/node/constants.js'`, but sadly esbuild wouldn't treeshake that module, so it ended up adding a bunch of unnecessary code to the bundle.

To fix this, I've added a Node step to `build.sh` that extracts this to an also-checked-in file. Does that seem reasonable?

STRd6: Yes, that seems like a good solution.

edemaine: Another idea I had: we could put this in a `prepare` script so it runs when getting dependencies but not when building, which would slightly speed up the dev cycle (I yarn build more often than yarn install). Not sure the time penalty is enough to bother though.

Issue #1070, opened 2024-02-27
& function shorthand revamp, and ./... argument placeholders


edemaine:
Based on Discord discussion, kicked off with a big idea from @STRd6:

## `&` precedence

> I think the rule would be lift the `&` function to be either at statement level or within a function application, whichever is encountered first

Alternative wording:

> The scope for the function wrapping `&` is up to but not including function calls, stopping at beginning of expression [so not including e.g. the left part of a declaration]

Thus:

```js
[&, x]        →  $ => [$, x]
[&, &]        →  $ => [$, $]
& + &         →  $ => $ + $
[...&]        →  $ => [...$]
[[&]]         →  $ => [[$]]
{text: &}     →  $ => {text: $}
{&: "value"}  →  $ => {[$]: "value"}
x > &         →  $ => x > &
f ...& // invalid, at least for now
```

Note that the rule reasonably solves #85, allowing for things like `&+&` and `[&, &]`. If you want two different functions, you need to separate them into different statements or calls. In the array example, we could write `[(do &), (do &)]`. Or perhaps `[&x, &y]`.

Some of these examples are from #480, which I would view this as replacing. But some of those examples don't work:

```js
f(a, &, b)
console.log &
x.filter !Array.isArray &
```

These don't have a clear meaning with the new rule.
* Perhaps `&` that can't be lifted is just `($) => $`. We can't escape a function call this way. But we'll propose another way next!
* Or perhaps `&` does get lifted one level when it's lonely like this, as in #480. Feels a bit asymmetric. And wait for the next part!

## `.` and `...` argument placeholders

At the top level of function arguments and array literals, we propose that `.` adds a magic function wrapper one level up:

```js
f(a, ., b)    →  $ => f(a, $, b)
console.log . →  $ => console.log($)
x.filter !Array.isArray .   // leading unary operators get pulled in???
              →  x.filter($ => !Array.isArray($))
[a, ., b]     →  $ => [a, $, b]
[., .]        →  $ => [$, $]
```

This gives us a form of partials and placeholders (#75):

```js
last |> f a, b, .     → f(a, b, last)
array |> [x].concat .   →  [x].concat(array)
```

In contexts where `...x` makes sense (function arguments, array literals, and object literals), `...` also adds a magic function wrapper one level up, but with multiple arguments:

```js
f(a, ..., b)    →  (...$) => f(a, ...$, b)
console.log ... →  (...$) => console.log(...$)
[...]

PR #1159, opened 2024-04-09
New ampersand proposal (#1070)

edemaine:
This is a large PR resulting from several days. Sorry for all the stuff at once!

Mainly this implements a variation of the proposed `&` behavior from #1070:

> Lift the function wrapper of any `&`s to the nearest:
> * statement
> * just before function application (either an argument or function itself)
> * right-hand side of assignment
> * initializer of declaration
> * pipeline tail (non-head) item
> * pipeline (`&` in pipeline head wraps the whole pipeline)

Hopefully that's everything! We may find we need to add more stopping rules, but it's fairly straightforward to do so in `processPlaceholders`, which shares a bunch of code with existing `.` placeholders.

This is a BREAKING CHANGE, so I suggest we finally move to 0.7.0. For example:
* `&` is now an identity function is many situations, like `x.filter &`
* `(&)` is no longer an identity function special case; it's now `(a,b) => a&b` like other binary operators
* `&` functions are generally wrapped higher than before
* For example, `1+&` and `&+1` are now treated similarly, whereas before `1+&` added 1 to a function (assuming something to the right) or was invalid (if nothing to the right)
* No longer one function wrapper per `&`

You're right that one of the hardest parts was getting `ForbiddenImplicitCalls` correct. We need to forbid some `&` here when they correspond to a binary `&` operator. After several iterations, I hope that I have it right.

Other changes:
* Pipelines were buggy in some situations. I made `x = y |> z` parse as `x = (y |> z)` (as usual, and I *think* what we want) but `x |> y = z |> w` as `x |> (y = z) |> w` (which currently parses as `x |> y = (z |> w)` which seems bad). I had to modify `NonPipelineAssignmentExpression` to avoid pipelines more thoroughly, which involved forking a few more rules.
* `"Initializer"` AST nodes
* `ASTNodeBase` was from an older TypeScript model of nodes; made new `ASTNodeObject` that uses the new node types, and we should generally phase over from `ASTNodeBase` to `ASTNodeObject`. Also helper function `isASTNodeObject`.
* Other helper types: `IsToken`, `IsParent`, `ASTNodeParent` (nonleaf), and corresponding functions `isToken` and `isParent`. Slowly getting to cleaner cases for different node types.
* Changed all `parent` typings to `parent?: Parent`. Often we don't want to add `parent`s right away, e.g. when passing to `makeNode`, or if we're still in the first pass of the compiler.
* `prepend` helper does the smooshing of whitespace that we do so often. I may have missed some occurrences but a bunch of them now use it.
* Inplace versions of `prepend` and `insertTrimmingSpace` which are helpful because we're modifying an existing AST; making a new node with the modification would make it painful to adjust all the pointers.
* `findChildIndex` had a major bug (missing `return false`, so `arrayRecurse` *always* returned a true value!)
* `maybeWrap` and `maybeUnwrap` are core for introducing an object node at any place in the AST, so that it has a parent and we can do a `replaceNode` on it. The AST is a little ugly in the middle, but it works, and wrappers get removed before final AST output.
* Cleanup some test cases to use `testCase.js`

~~TODO: DOCUMENTATION~~


STRd6 (approved):
Awesome, looks great! Glad to see how this simplifies the parser a little.

It might be nice to add a couple tests mixing both placeholders in semi-complex situations to make sure there isn't any possibility of them hitting the same ancestor and getting weird.

It'll be great to get this in as 0.7.0. Thank you for seeing it through.

PR #1165, opened 2024-04-11
Infinite loop detection and no return

### LINKED ISSUE #1161: Eliminating `return` after `loop` in some cases
When writing an infinite generator, such as the below, Civet emits an unreachable return statement:

```civet
f := (a, b) ->
 yield a
 loop
   yield b
   [a, b] = [b, a + b]
```

```js
const f = function* (a, b) {
  yield a;
  const results = [];
  while (true) {
    yield b;
    results.push(([a, b] = [b, a + b]));
  }
  return results;
};
```

While this could be removed, there are some things we have to be careful of:
- If there's a `break` inside the loop, the return statement becomes reachable
- Only return statements in the same scope as the `loop` itself may be omitted -- conditionals may not trigger, and `try` blocks may terminate due to an error.

edemaine:
Fixes #1161 by extending `isExit` and using it to omit implicit `return`/`push`.

* Basic infinite loop detection in `isExit`: `loop`/`while(true)`/`do..while(true)` without contained `break`.
  * Currently do not distinguish between `break` of this loop vs. other loops
  * Do correctly ignore `break` inside nested functions
* Extend `isExit` to handle `if/else` statements.
  * There might be other work to do here, e.g. `do` statements.
* Avoid implicit `return` of last statement when it is `isExit`.

We can continue to refine `isExit` to improve this mechanism.

One case I wasn't sure of: currently a block is considered `isExit` if its last statement is. So e.g. a block with a `break` in the middle but not end of the block is considered non-exiting. I don't *think* this case matters much, because then the author write inaccessible code, so more inaccessible code is probably fine? But it might be nicer to change the definition to "any line of the block `isExit`". I'm not sure.

### DIFF
--- a/source/parser.hera
+++ b/source/parser.hera
@@ -3958,25 +3958,42 @@ IterationExpression
 LoopStatement
   LoopClause:clause Block:block ->
     return {
+      ...clause,
       type: "IterationStatement",
       children: [...clause.children, block],
       block,
     }
 
 LoopClause
-  Loop -> {
-    type: "IterationStatement",
-    children: [$1],
-  }
+  Loop:kind ->
+    const expression = {
+      type: "Literal",
+      children: ["true"],
+      raw: "true",
+    }
+    const condition = {
+      type: "ParenthesizedExpression",
+      children: [ "(", expression, ")" ],
+      expression,
+    }
+    return {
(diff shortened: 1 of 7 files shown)


STRd6 (approved):
Thanks!

PR #1167, opened 2024-04-15
Pipe assignment to index

STRd6:
Allow for pipe assignment with index properties like:

```
body.children[1] |>= .filter [, e] => !exps.includes e
body.children.-1 |>= .filter [, e] => !exps.includes e
```


edemaine (approved):
Nice catch!

PR #1170, opened 2024-04-18
Improve ASI algorithm

### LINKED ISSUE #1169: ASI fails before symbol-typed property
```civet
class Foo
  [Symbol.toStringTag] = 'Foo'
  [Symbol.isConcatSpreadable] = true
```
Current compilation:
```js
class Foo {
  [Symbol.toStringTag] = ("Foo"[Symbol.isConcatSpreadable] = true);
}
```
Expected compilation:
```js
class Foo {
  [Symbol.toStringTag] = "Foo";
  [Symbol.isConcatSpreadable] = true;
}
```

edemaine:
* Run ASI algorithm in class bodies; previously they were `type: "ClassBody"` so completely ignored. Now they're `type: "BlockStatement"` with `subtype: "ClassBody"` (in case we need to know someday).
* Recurse into first non-null child/array element instead of just first, and add a string base case. This seems like a better-behaved recursion. Probably some cases of the `switch` could now be eliminated, but I didn't try.

Fixes #1169

### DIFF
--- a/source/parser.hera
+++ b/source/parser.hera
@@ -870,17 +870,21 @@ ImplementsTarget
 # https://262.ecma-international.org/#prod-ClassBody
 # NOTE: Nesting and indentation sensitive
 ClassBody
-  __ OpenBrace NestedClassElements?:elements __ CloseBrace  ->
+  __ OpenBrace NestedClassElements?:expressions __ CloseBrace  ->
+    if (!expressions) expressions = $0[2] = []
     return {
-      type: "ClassBody",
+      type: "BlockStatement",
+      subtype: "ClassBody",
       children: $0,
-      elements,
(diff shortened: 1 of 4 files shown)


STRd6 (approved):
I'm a little bit worried about making "ClassBody" a "BlockStatement" but since it doesn't seem to be causing any problems it's probably fine.

Thanks for the fix!

edemaine:
It is a special kind of block, with e.g. assignments. I actually don't know what happens if you pit function calls in there. Certainly methods are special. If it ends up being a problem, we can certainly complicate the ASI code to consider multiple block-like things. But e.g. the new ampersand code in working on probably also wants class bodies to be considered blocks.

PR #1177, opened 2024-04-21
Limit & to return/yield; support indented yield argument

edemaine:
* Limit `&` lifting to stop at `return` and `yield`, as it's not useful to `return` or `yield` inside a generated wrapper.
* Add support for `yield`ing an indented expression, like we already do with `return`.
* AST node for `"YieldExpression"`

As this PR makes clear: Some AST nodes use property `exp` (e.g. `"AssignmentExpression"`, `"Initializer"`) while some use property `expression` (e.g. `"ReturnStatement"` and new `"YieldExpression"`). Could we unify on one or the other? Preference which?


STRd6 (approved):
Looks good! If we're going to converge on `exp` vs `expression` I prefer `expression`.

PR #1179, opened 2024-04-22
Brace block improvements, including `&` statements

### LINKED ISSUE #1176: `&` doesn't always work in `if` statement
For some reason, these generate invalid JS:
```civet
if & then x         // $ => if ($) x
return if & then x  // $ => return ($ ? x : void 0)
```
While these do not:
```civet
foo if & then x   // foo(($) => ($ ? x : void 0))
f := if & then x  // const f = ($) => ($ ? x : void 0)
```

edemaine:
Fixes #1176 

I unified a place where we were manually bracing a block into a call to `braceBlock`. But they differed in the closing brace: the new use ended with ` }` while other instances ended with `}`. I'm somewhat tempted to switch to the spaced version throughout, which would lead to changes like this:

```diff
switch x
  {a: y, b: y}
    y
↓↓↓
   if(typeof x === 'object' && x != null && 'a' in x && 'b' in x) {
     const {a: y1, b: y2} = x;
     const y = [y1, y2];
-    y }
+    y}
```

But this would involve changing ~70 tests. Do you think it's worth it?

### DIFF
--- a/source/parser/auto-dec.civet
+++ b/source/parser/auto-dec.civet
@@ -1,5 +1,6 @@
 import type {
   ASTNode
+  BlockStatement
 } from ./types.civet
 
 import {
@@ -12,6 +13,7 @@ import {
 } from ./util.civet
 
 import {
+  braceBlock
   getIndent
 } from ./block.civet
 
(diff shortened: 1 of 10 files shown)


STRd6 (approved):
Looks good!

As for the space before the closing brace I've actually been leaning towards not inserting any spaces/newlines to keep the generated code as close as possible to the original source. There are fewer `$loc` tokens to keep track of and it usually doesn't look too bad visually since it is closer to the original.

edemaine:
Good point; makes sense. I looked at unifying `braceBlock` in the other direction and I get only two failing tests:

```diff
-(x): void => { console.log(x)}
+(x): void => { console.log(x) }

-const ignore = <T, U>(a: T, b: U) : void => { null}
+const ignore = <T, U>(a: T, b: U) : void => { null }
```

It's just a bit uglier that I hesitate to do it... but it would simplify the code slightly (no second argument to `braceBlock`). I guess I'll go for it.

PR #1180, opened 2024-04-23
First version of comptime (synchronous, no outer scope)

edemaine:
This is a first version of `comptime` blocks based on [Rust's comptime crate](https://docs.rs/comptime/latest/comptime/). Limitations:

* Block must be run synchronously. To allow for async code, I realize we need to change the compilation API to allow async at some stage.
* Output must `JSON.stringify`able, so must have no reference cycles.
* No access to outer scopes. We could add this some day, but we don't have any variable tracking whatsoever, so it's probably a longer term thing.

The elephant in the room is security. `eval` always feels dangerous. On the other hand, I can't immediately think of a threat model where you're running an NPM script that runs `civet` for compilation and it couldn't run any other command. Still, if feels a little weird that compiling with the `civet` CLI could run arbitrary code. But this is also inherent to macros; see e.g. how [Rust's comptime crate is implemented](https://docs.rs/comptime/latest/src/comptime/lib.rs.html): it runs an external copy of Rust.

Another example is editing in VSCode. As you're typing code, it will *immediately run the comptime blocks on every keystroke* which could take down the language server (infinite loop) or other "fun" accidents.

I feel like, at the least, we need a flag that controls `comptime`, and maybe the default should be off. So maybe `civet --comptime` to enable, or in the unplugin there's a `{comptime: true}` option. I haven't added this yet because I wanted to get input first on what you think is best, but it's my current thoughts.

Alternatively, we could add a `comptime` phase separate in between `parse`/`prune` and `generate`. This would have the potential advantage of keeping those phases synchronous, so only the `comptime` phase needs to be asynchronous. But it would make it harder to use from an API perspective. Or maybe we offer multiple top-level APIs, synchronous which forbid comptime, and asynchronous which allows it?

Also:
* Fix two significant bugs in `replaceNode` (premature `return` and bad resetting of `parent`)
* Get the `Predicate` stuff actually working. Previously was assuming that the first argument has a relevant type, but it can really be anything.

### DIFF
--- a/source/parser/comptime.civet
+++ b/source/parser/comptime.civet
@@ -76,8 +76,6 @@ function serialize(value: ???): string
       <? "object"
         if stack.has val
           throw new Error "circular reference detected"
-        if Array.isArray val
-          return `[${val.map(serialize).join ","}]`
         stack.add val
         str :=
           switch Object.getPrototypeOf val
@@ -89,22 +87,24 @@ function serialize(value: ???): string
             when Set::
               "new Set([" + (
                 for item of val as Set<???>
-                  serialize item
+                  recurse item
               ).join(",") + "])"
             when Map::
               "new Map([" + (
                 for [key, value] of val as Map<???, ???>
-                  `[${serialize key},${serialize value}]`
+                  `[${recurse key},${recurse value}]`
               ).join(",") + "])"
+            when Array::
+              `[${(val as ???[]).map(recurse).join ","}]`
             when Object::
               "{" + (
                 for own key, value in val as {[key: string]: ???}
-                  `${JSON.stringify key}:${serialize value}`
+                  `${JSON.stringify key}:${recurse value}`
               ).join(",") + "}"
             else
               // One day we may handle other classes like so:
               // str += `__proto__:${val.constructor.name}`
-              throw new TypeError `cannot serialize value with prototype ${Object.getPrototypeOf val}`
+              throw new TypeError `cannot serialize object with prototype ${Object.getPrototypeOf val}`
         stack.delete val
         str
       else
--- a/source/parser/comptime.civet
+++ b/source/parser/comptime.civet
@@ -1,0 +1,42 @@
+import type {
+  ComptimeExpression
+  ComptimeStatement
+  StatementTuple
+} from ./types.civet
+
+import {
+  gatherRecursive
+} from "./traversal.civet"
+
+generate, { prune, type Options } from "../generate.civet"
+
+function processComptime(statements: StatementTuple[]): void
+  gatherRecursive statements,
+    (node): node is (ComptimeStatement | ComptimeExpression) =>
+      node.type is "ComptimeStatement" or node.type is "ComptimeExpression"
+  .forEach (exp) =>
+    content := exp.type is "ComptimeStatement" ? exp.block : exp.expression
+    // Convert the comptime block into JS code
+    options: Options := js: true
+    js := generate prune(content), options
+    // If there are any errors, leave the comptime subtree alone
+    // (so there's still an error).
+    return if options.errors?
+    output := eval?.(`"use strict";${js}`)
+    if exp.type is "ComptimeExpression"
+      let string
+      try
+        string = JSON.stringify output
+      catch e
+        exp.children = [
+          type: "Error"
+          message: `comptime result ${output} not JSON serializable: ${e}`
+        ]
+        return
+      exp.children = [string]
+    else
+      exp.children = []
+
+export {
+  processComptime
+}
--- a/source/parser/comptime.civet
+++ b/source/parser/comptime.civet
@@ -81,6 +81,11 @@ function serialize(value: ???): string
         stack.add val
         str :=
           switch Object.getPrototypeOf val
+            when RegExp::
+              re := val as RegExp
+              `/${re.source}/${re.flags}`
+            when Date::
+              `new Date(${(val as Date).getTime()})`
             when Set::
               "new Set([" + (
                 for item of val as Set<???>
--- a/source/parser/comptime.civet
+++ b/source/parser/comptime.civet
@@ -1,3 +1,5 @@
+"civet coffeePrototype"
+
 import type {
   ComptimeExpression
   ComptimeStatement
@@ -26,7 +28,7 @@ function processComptime(statements: StatementTuple[]): void
     if exp.type is "ComptimeExpression"
       let string
       try
-        string = JSON.stringify output
+        string = serialize output
       catch e
         exp.children = [
           type: "Error"
@@ -37,6 +39,74 @@ function processComptime(statements: StatementTuple[]): void
     else
       exp.children = []
 
+function serialize(value: ???): string
+  stack := new Set<object>
+  function recurse(val: ???): string
+    switch val
+      <? "string"
+        // Using JSON.stringify to handle escape sequences
+        JSON.stringify val
+      <? "number", <? "boolean", == null
+        String val // not toString to accomodate null/undefined
+      <? "bigint"
+        `${val}n`
+      <? "function"
+        // Some functions can actually be serialized
+        hasNoProps := (and)
+          Object.getOwnPropertyNames(val).every & is in ["length", "name", "arguments", "caller", "prototype"]
+          Object.getOwnPropertySymbols(val)# is 0
+          (or)
+            val:: is undefined
+            (and)
+              Object:: is Object.getPrototypeOf val::
+              Object.getOwnPropertyNames(val::)# is 1 // constructor
+              Object.getOwnPropertySymbols(val::)# is 0
+              val::constructor is val
+        unless hasNoProps
+          throw new TypeError "cannot serialize function with properties"
+        string := Function::toString.call val
+        if /\{\s+\[native code]\s+\}$/.test string
+          // builtin, or returned from Function::bind
+          throw new TypeError "cannot serialize native function"
+        string
+      <? "symbol"
+        if key? := Symbol.keyFor val
+          return `Symbol.for(${JSON.stringify key})`
+        throw new TypeError "cannot serialize unique symbol"
+      <? "object"
+        if stack.has val
+          throw new Error "circular reference detected"
+        if Array.isArray val
+          return `[${val.map(serialize).join ","}]`
+        stack.add val
+        str :=
+          switch Object.getPrototypeOf val
+            when Set::
+              "new Set([" + (
+                for item of val as Set<???>
+                  serialize item
+              ).join(",") + "])"
+            when Map::
+              "new Map([" + (
+                for [key, value] of val as Map<???, ???>
+                  `[${serialize key},${serialize value}]`
+              ).join(",") + "])"
+            when Object::
+              "{" + (
+                for own key, value in val as {[key: string]: ???}
+                  `${JSON.stringify key}:${serialize value}`
+              ).join(",") + "}"
+            else
+              // One day we may handle other classes like so:
+              // str += `__proto__:${val.constructor.name}`
+              throw new TypeError `cannot serialize value with prototype ${Object.getPrototypeOf val}`
+        stack.delete val
+        str
+      else
+        throw new TypeError `cannot serialize ${typeof val} value`
+  recurse value
+
 export {
   processComptime
+  serialize
 }
--- a/source/parser/comptime.civet
+++ b/source/parser/comptime.civet
@@ -10,9 +10,15 @@ import {
   gatherRecursive
 } from "./traversal.civet"
 
+{ getInitialConfig } from "../parser.hera"
+
 generate, { prune, type Options } from "../generate.civet"
 
 function processComptime(statements: StatementTuple[]): void
+  // Prevent comptime setting from being overridden by directives
+  // by looking at initialConfig instead of config
+  return unless getInitialConfig()?.comptime
+
   gatherRecursive statements,
     (node): node is (ComptimeStatement | ComptimeExpression) =>
       node.type is "ComptimeStatement" or node.type is "ComptimeExpression"
--- a/test/comptime.civet
+++ b/test/comptime.civet
@@ -1,0 +1,45 @@
+{ testCase, throws } from ./helper.civet
+{ compile } from ../source/main.civet
+assert from assert
+
+describe "comptime", ->
+  testCase """
+    statement with indented block
+    ---
+    comptime
+      x := 5
+      x * x
+    ---
+
+  """
+
+  testCase """
+    one-line statement
+    ---
+    comptime x := 5
+    ---
+
+  """
+
+  it "statement has side effect", =>
+    global.outside = 0
+    compile "comptime global.outside = 7"
+    assert.equal global.outside, 7
+
+  testCase """
+    one-line expression
+    ---
+    value := comptime 1+2
+    ---
+    const value = 3
+  """
+
+  testCase """
+    expression with indented block
+    ---
+    value := comptime
+      x := 5
+      x * x
+    ---
+    const value = 25
+  """
--- a/test/comptime.civet
+++ b/test/comptime.civet
@@ -55,6 +55,8 @@ describe "serialize", ->
     assert.equal serialize(Infinity), "Infinity"
     assert.equal serialize(-Infinity), "-Infinity"
     assert.equal serialize(NaN), "NaN"
+  it "big numbers", =>
+    assert.equal serialize(1000000000000000n), "1000000000000000n"
   it "arrays and nulls", =>
     assert.equal serialize([1, undefined, 3, null]), "[1,undefined,3,null]"
   it "strings", =>
@@ -62,7 +64,7 @@ describe "serialize", ->
   it "symbols", =>
     assert.equal serialize(Symbol.for 'MyKeyedSymbol'),
       'Symbol.for("MyKeyedSymbol")'
-    assert.throws => serialize Symbol()
+    assert.throws (=> serialize Symbol()), /cannot serialize unique symbol/
   it "regexps", =>
     assert.equal serialize(/a[b\s]c/), "/a[b\\s]c/"
     assert.equal serialize(/abc/gi), "/abc/gi"
@@ -78,7 +80,7 @@ describe "serialize", ->
       'new Map([["a",1],["b",2],["c",3]])'
   it "custom classes", =>
     class Foo {}
-    assert.throws => serialize new Foo()
+    assert.throws (=> serialize new Foo()), /cannot serialize object with prototype/
   it "arrow functions", =>
     assert.equal serialize((x: ???) => x), "(x) => x"
   it "regular functions", =>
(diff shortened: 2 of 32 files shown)


bbrk24:
> Output must `JSON.stringify`able

This is a more serious limitation than just reference cycles and custom classes. `JSON.stringify` silently translates `[1, undefined, 3]` into `[1, null, 3]`.

edemaine:
Ah, I didn't know that one. A recent robust and fairly small serializer library I know of is [seroval](https://github.com/lxsmnsyc/seroval) (created for use in Solid SSR I believe). Are we OK adding a dependency for this? Alternatively, we could probably make our own pretty easily with pattern matching, especially if we don't care about cycles and custom classes. Hmm, maybe our own is cool because we could plausibly fix custom classes one day, as we have influence over the compile and run time...

bbrk24:
Would [this](https://civet.dev/playground?code=J2NpdmV0IGNvZmZlZVByb3RvdHlwZScKCmNpcmN1bGFyUmVmRGV0ZWN0aW9uOiBvYmplY3RbXSA6PSBbXQoKZnVuY3Rpb24gc2VyaWFsaXplKHZhbDogPz8%2FKTogc3RyaW5nCiAgc3dpdGNoIHZhbAogICAgPD8gJ3N0cmluZycKICAgICAgLy8gVXNpbmcgSlNPTi5zdHJpbmdpZnkgdG8gaGFuZGxlIGVzY2FwZSBzZXF1ZW5jZXMKICAgICAgSlNPTi5zdHJpbmdpZnkgdmFsCiAgICA8PyAnbnVtYmVyJywgPD8gJ2Jvb2xlYW4nLCA9PSBudWxsCiAgICAgIFN0cmluZyB2YWwgLy8gbm90IHRvU3RyaW5nIHRvIGFjY29tb2RhdGUgbnVsbC91bmRlZmluZWQKICAgIDw%2FICdiaWdpbnQnCiAgICAgIGAke3ZhbH1uYAogICAgPD8gJ2Z1bmN0aW9uJwogICAgICAvLyBTb21lIGZ1bmN0aW9ucyBjYW4gYWN0dWFsbHkgYmUgc2VyaWFsaXplZAogICAgICBoYXNOb1Byb3BzIDo9IChhbmQpCiAgICAgICAgT2JqZWN0LmdldE93blByb3BlcnR5TmFtZXModmFsKSMgaXMgMiAvLyAnbGVuZ3RoJyBhbmQgJ25hbWUnCiAgICAgICAgT2JqZWN0LmdldE93blByb3BlcnR5U3ltYm9scyh2YWwpIyBpcyAwCiAgICAgIHVubGVzcyBoYXNOb1Byb3BzCiAgICAgICAgdGhyb3cgbmV3IFR5cGVFcnJvciAnY2Fubm90IHNlcmlhbGl6ZSBmdW5jdGlvbiB3aXRoIHByb3BlcnRpZXMnCiAgICAgIHN0cmluZyA6PSBGdW5jdGlvbjo6dG9TdHJpbmcuY2FsbCB2YWwKICAgICAgaWYgL1x7W1xuXHNdK1xbbmF0aXZlIGNvZGVdW1xuXHNdK1x9JC8udGVzdCBzdHJpbmcKICAgICAgICAvLyBidWlsdGluLCBvciByZXR1cm5lZCBmcm9tIEZ1bmN0aW9uOjpiaW5kCiAgICAgICAgdGhyb3cgbmV3IFR5cGVFcnJvciAnY2Fubm90IHNlcmlhbGl6ZSBuYXRpdmUgZnVuY3Rpb24nCiAgICAgIHN0cmluZwogICAgPD8gJ3N5bWJvbCcKICAgICAgaWYga2V5PyA6PSBTeW1ib2wua2V5Rm9yIHZhbAogICAgICAgIHJldHVybiBgU3ltYm9sLmZvcigke0pTT04uc3RyaW5naWZ5IGtleX0pYAogICAgICB0aHJvdyBuZXcgVHlwZUVycm9yICdjYW5ub3Qgc2VyaWFsaXplIHVuaXF1ZSBzeW1ib2wnCiAgICA8PyAnb2JqZWN0JwogICAgICBpZiB2YWwgaXMgaW4gY2lyY3VsYXJSZWZEZXRlY3Rpb24KICAgICAgICB0aHJvdyBuZXcgRXJyb3IgJ2NpcmN1bGFyIHJlZmVyZW5jZSBkZXRlY3RlZCcKICAgICAgY2lyY3VsYXJSZWZEZXRlY3Rpb24ucHVzaCB2YWwKICAgICAgdHJ5CiAgICAgICAgaWYgQXJyYXkuaXNBcnJheSB2YWwKICAgICAgICAgIHJldHVybiBgWyR7dmFsLm1hcChzZXJpYWxpemUpLmpvaW4gJywnfV1gCiAgICAgICAgc3RyIC49ICd7JwogICAgICAgIGZvciBvd24ga2V5LCB2YWx1ZSBpbiB2YWwgYXMgYW55CiAgICAgICAgICBzdHIgKz0gYCR7SlNPTi5zdHJpbmdpZnkga2V5fToke3NlcmlhbGl6ZSB2YWx1ZX0sYAogICAgICAgIC8vIENvZGUgZm9yIGhhbmRsaW5nIGNsYXNzZXM6CiAgICAgICAgLy9pZiBPYmplY3Q6OiAhPT0gT2JqZWN0LmdldFByb3RvdHlwZU9mIHZhbAogICAgICAgIC8vICBzdHIgKz0gYF9fcHJvdG9fXzoke3ZhbC5jb25zdHJ1Y3Rvci5uYW1lfWAKICAgICAgICByZXR1cm4gc3RyICsgJ30nCiAgICAgIGZpbmFsbHkKICAgICAgICBjaXJjdWxhclJlZkRldGVjdGlvbi5wb3AoKQoKY29uc29sZS5sb2cgc2VyaWFsaXplIFsxLCB1bmRlZmluZWQsIDNdCmNvbnNvbGUubG9nIHNlcmlhbGl6ZSAnSGVsbG8sIFdvcmxkIScKY29uc29sZS5sb2cgc2VyaWFsaXplIGZvbzogdHJ1ZSwgYmFyOiBiYXo6IGZhbHNlCmNvbnNvbGUubG9nIHNlcmlhbGl6ZSAmCmNvbnNvbGUubG9nIHNlcmlhbGl6ZSBTeW1ib2wuZm9yICdNeUtleWVkU3ltYm9sJwoKdHJ5CiAgY2lyY3VsYXIgOj0ge30KICBjaXJjdWxhci5jaXJjdWxhciA9IGNpcmN1bGFyCiAgY29uc29sZS5sb2cgc2VyaWFsaXplIGNpcmN1bGFyCmNhdGNoIGUKICBjb25zb2xlLmVycm9yIGUudG9TdHJpbmcoKQoKdHJ5CiAgY29uc29sZS5sb2cgc2VyaWFsaXplIFN5bWJvbCgpCmNhdGNoIGUKICBjb25zb2xlLmVycm9yIGUudG9TdHJpbmcoKQoKdHJ5CiAgZnVuYyA6PSAmCiAgZnVuYy5hID0gMQogIGNvbnNvbGUubG9nIHNlcmlhbGl6ZSBmdW5jCmNhdGNoIGUKICBjb25zb2xlLmVycm9yIGUudG9TdHJpbmcoKQoKdHJ5CiAgY29uc29sZS5sb2cgc2VyaWFsaXplIHBhcnNlSW50CmNhdGNoIGUKICBjb25zb2xlLmVycm9yIGUudG9TdHJpbmcoKQ%3D%3D) work as a starting point for rolling our own serializer?

STRd6:
This looks like a good start. I think it will take a bit of experimentation to figure out the best way to handle comptime but we should be able to get there incrementally.

Having `comptime` off by default with an option or flag to enable seems best to me, especially considering the potential LSP hanging issues. Eventually we may switch to a separate thread that can be cached/killed. This also mitigates some risk from `civet ...` shell commands having more capabilities than someone might expect. Being explicit lets people opt in and choose their trust level based on the code they are working with.

I looked briefly at https://github.com/lxsmnsyc/seroval but it still seems a bit large to include for this one feature, at least for now. Having our own simple one or even just documenting and living with the JSON limitations while we experiment is fine.

STRd6 (approved):

STRd6:
One other security note: if we enable `comptime` on the web playground we need to be sure to use a sandbox otherwise the sharing links can be used for XSS.

edemaine:
Yeah, requiring interaction vs. just following a link is a significant difference. On the Playground, maybe we add an additional checkbox to enable `comptime` which is only visible when the word `comptime` appears in the source code. I think we can continue to enable it in the Reference (maybe just on the `comptime` examples), but anyone editing those examples live would need to be a bit careful to not write an infinite loop. I'll look at this today.

bbrk24:
> Would [this](https://civet.dev/playground?code=J2NpdmV0IGNvZmZlZVByb3RvdHlwZScKCmNpcmN1bGFyUmVmRGV0ZWN0aW9uOiBvYmplY3RbXSA6PSBbXQoKZnVuY3Rpb24gc2VyaWFsaXplKHZhbDogPz8%2FKTogc3RyaW5nCiAgc3dpdGNoIHZhbAogICAgPD8gJ3N0cmluZycKICAgICAgLy8gVXNpbmcgSlNPTi5zdHJpbmdpZnkgdG8gaGFuZGxlIGVzY2FwZSBzZXF1ZW5jZXMKICAgICAgSlNPTi5zdHJpbmdpZnkgdmFsCiAgICA8PyAnbnVtYmVyJywgPD8gJ2Jvb2xlYW4nLCA9PSBudWxsCiAgICAgIFN0cmluZyB2YWwgLy8gbm90IHRvU3RyaW5nIHRvIGFjY29tb2RhdGUgbnVsbC91bmRlZmluZWQKICAgIDw%2FICdiaWdpbnQnCiAgICAgIGAke3ZhbH1uYAogICAgPD8gJ2Z1bmN0aW9uJwogICAgICAvLyBTb21lIGZ1bmN0aW9ucyBjYW4gYWN0dWFsbHkgYmUgc2VyaWFsaXplZAogICAgICBoYXNOb1Byb3BzIDo9IChhbmQpCiAgICAgICAgT2JqZWN0LmdldE93blByb3BlcnR5TmFtZXModmFsKS5ldmVyeSAmIGlzIGluIFsnbGVuZ3RoJywgJ25hbWUnLCAncHJvdG90eXBlJ10KICAgICAgICBPYmplY3QuZ2V0T3duUHJvcGVydHlTeW1ib2xzKHZhbCkjIGlzIDAKICAgICAgICAob3IpCiAgICAgICAgICB2YWwucHJvdG90eXBlIGlzIHVuZGVmaW5lZAogICAgICAgICAgKGFuZCkKICAgICAgICAgICAgT2JqZWN0OjogaXMgT2JqZWN0LmdldFByb3RvdHlwZU9mIHZhbC5wcm90b3R5cGUKICAgICAgICAgICAgT2JqZWN0LmdldE93blByb3BlcnR5TmFtZXModmFsLnByb3RvdHlwZSkjIGlzIDEgLy8gY29uc3RydWN0b3IKICAgICAgICAgICAgT2JqZWN0LmdldE93blByb3BlcnR5U3ltYm9scyh2YWwucHJvdG90eXBlKSMgaXMgMAogICAgICAgICAgICB2YWwucHJvdG90eXBlLmNvbnN0cnVjdG9yIGlzIHZhbAogICAgICB1bmxlc3MgaGFzTm9Qcm9wcwogICAgICAgIHRocm93IG5ldyBUeXBlRXJyb3IgJ2Nhbm5vdCBzZXJpYWxpemUgZnVuY3Rpb24gd2l0aCBwcm9wZXJ0aWVzJwogICAgICBzdHJpbmcgOj0gRnVuY3Rpb246OnRvU3RyaW5nLmNhbGwgdmFsCiAgICAgIGlmIC9ce1tcblxzXStcW25hdGl2ZSBjb2RlXVtcblxzXStcfSQvLnRlc3Qgc3RyaW5nCiAgICAgICAgLy8gYnVpbHRpbiwgb3IgcmV0dXJuZWQgZnJvbSBGdW5jdGlvbjo6YmluZAogICAgICAgIHRocm93IG5ldyBUeXBlRXJyb3IgJ2Nhbm5vdCBzZXJpYWxpemUgbmF0aXZlIGZ1bmN0aW9uJwogICAgICBzdHJpbmcKICAgIDw%2FICdzeW1ib2wnCiAgICAgIGlmIGtleT8gOj0gU3ltYm9sLmtleUZvciB2YWwKICAgICAgICByZXR1cm4gYFN5bWJvbC5mb3IoJHtKU09OLnN0cmluZ2lmeSBrZXl9KWAKICAgICAgdGhyb3cgbmV3IFR5cGVFcnJvciAnY2Fubm90IHNlcmlhbGl6ZSB1bmlxdWUgc3ltYm9sJwogICAgPD8gJ29iamVjdCcKICAgICAgaWYgdmFsIGlzIGluIGNpcmN1bGFyUmVmRGV0ZWN0aW9uCiAgICAgICAgdGhyb3cgbmV3IEVycm9yICdjaXJjdWxhciByZWZlcmVuY2UgZGV0ZWN0ZWQnCiAgICAgIGNpcmN1bGFyUmVmRGV0ZWN0aW9uLnB1c2ggdmFsCiAgICAgIHRyeQogICAgICAgIGlmIEFycmF5LmlzQXJyYXkgdmFsCiAgICAgICAgICByZXR1cm4gYFske3ZhbC5tYXAoc2VyaWFsaXplKS5qb2luICcsJ31dYAogICAgICAgIHN0ciAuPSAneycKICAgICAgICBmb3Igb3duIGtleSwgdmFsdWUgaW4gdmFsIGFzIGFueQogICAgICAgICAgc3RyICs9IGAke0pTT04uc3RyaW5naWZ5IGtleX06JHtzZXJpYWxpemUgdmFsdWV9LGAKICAgICAgICAvLyBDb2RlIGZvciBoYW5kbGluZyBjbGFzc2VzOgogICAgICAgIC8vaWYgT2JqZWN0OjogIT09IE9iamVjdC5nZXRQcm90b3R5cGVPZiB2YWwKICAgICAgICAvLyAgc3RyICs9IGBfX3Byb3RvX186JHt2YWwuY29uc3RydWN0b3IubmFtZX1gCiAgICAgICAgcmV0dXJuIHN0ciArICd9JwogICAgICBmaW5hbGx5CiAgICAgICAgY2lyY3VsYXJSZWZEZXRlY3Rpb24ucG9wKCkKCmNvbnNvbGUubG9nIHNlcmlhbGl6ZSBbMSwgdW5kZWZpbmVkLCAzXQpjb25zb2xlLmxvZyBzZXJpYWxpemUgJ0hlbGxvLCBXb3JsZCEnCmNvbnNvbGUubG9nIHNlcmlhbGl6ZSBmb286IHRydWUsIGJhcjogYmF6OiBmYWxzZQpjb25zb2xlLmxvZyBzZXJpYWxpemUgJgpjb25zb2xlLmxvZyBzZXJpYWxpemUgU3ltYm9sLmZvciAnTXlLZXllZFN5bWJvbCcKCnRyeQogIGNpcmN1bGFyIDo9IHt9CiAgY2lyY3VsYXIuY2lyY3VsYXIgPSBjaXJjdWxhcgogIGNvbnNvbGUubG9nIHNlcmlhbGl6ZSBjaXJjdWxhcgpjYXRjaCBlCiAgY29uc29sZS5lcnJvciBlLnRvU3RyaW5nKCkKCnRyeQogIGNvbnNvbGUubG9nIHNlcmlhbGl6ZSBTeW1ib2woKQpjYXRjaCBlCiAgY29uc29sZS5lcnJvciBlLnRvU3RyaW5nKCkKCnRyeQogIGZ1bmMgOj0gJgogIGZ1bmMuYSA9IDEKICBjb25zb2xlLmxvZyBzZXJpYWxpemUgZnVuYwpjYXRjaCBlCiAgY29uc29sZS5lcnJvciBlLnRvU3RyaW5nKCkKCnRyeQogIGNvbnNvbGUubG9nIHNlcmlhbGl6ZSBwYXJzZUludApjYXRjaCBlCiAgY29uc29sZS5lcnJvciBlLnRvU3RyaW5nKCkKCmNvbnNvbGUubG9nIHNlcmlhbGl6ZSBzZXJpYWxpemU%3D) work as a starting point for rolling our own serializer?

This sort of demonstrates its own flaw with allowing the serialization of functions: the captured `hasOwn` and `circularRefDetection` are not declared/initialized. As convenient as it would be to be able to have comptime functions in simple cases like `(1+)`, I'm not fully convinced it's worth the potential errors from invalid captures.

edemaine:
I think it's still potentially useful to serialize functions (though seroval doesn't even try). We can state the limitation in the documentation.

I've pushed a version (including @bbrk24 as coauthor) that supports `Map`s and `Set`s too, which seems like it would be helpful. Security next!

edemaine:
Security updates:
* `comptime` is part of `parseOptions` but it can't be set in directives via `"civet comptime"` (to avoid surprises). This is done by looking at the *original* `parseOptions` instead of the one modified by the code.
* CLI can enable it via `--comptime` (or `--civet comptime`).
* Docs examples don't support `comptime` by default; only the comptime examples do
* Playground has `comptime` turned off by default, but if you load or type an example with `comptime`, a checkbox appears to enable it.

**Question:** Should `comptime` be settable via config files? Currently it's allowed. This seems like it might be helpful, and is maybe less surprising, but I'm not certain.

I also did some major restructuring that we've always (?) wanted to do for `parser.hera`, which moves a lot of code out of there into new files. @STRd6 Do you want to review these? It wouldn't hurt to have a second set of eyes.

Modulo potentially disabling `comptime` in config files, I think this is ready for a first merge, though I'll want to fix `require` soon if possible.

source/parser/comptime.civet
```diff
@@ -0,0 +1,123 @@
...
+} from "./traversal.civet"
+
+{ getInitialConfig } from "../parser.hera"
+
+generate, { prune, type Options } from "../generate.civet"
+
+function processComptime(statements: StatementTuple[]): void
+  // Prevent comptime setting from being overridden by directives
+  // by looking at initialConfig instead of config
+  return unless getInitialConfig()?.comptime
```
STRd6, line 20: If `comptime` is disabled does that mean that the expression is evaluated at runtime?

test/comptime.civet
```diff
@@ -0,0 +1,141 @@
...
+
+  testCase """
+    one-line statement
+    ---
+    comptime x := 5
+    ---
+
+  """, options
+
+  it "statement has side effect", =>
```
STRd6, line 30: Eventually we wouldn't want `comptime` to have side effects within the same node process but we can handle that later.

STRd6 (approved):
Thanks! Looks like a great start to comptime and I'm also glad you got a lot of cleanup in as well.

source/parser/comptime.civet
edemaine: Yes. I figured that was a reasonable fallback. In particular, it gives a chance for correct automatic typing in VSCode. I should probably add mention of this to the documentation though.

I also wondered whether a warning (or maybe a comment in the code output?) might be appropriate. I've seen some comptime systems add `/* compiled on ...date... */`. Could also plausibly add `/* comptime disabled during compilation */` or something to help debug.

test/comptime.civet
edemaine: Perhaps. I was thinking of switching to `import('javascript:...')` (I feel like I saw this somewhere recently but not sure where?) instead of `eval` so it'd at least be in a separate module. Here I just want to measure that the code actually executes. If it ends up being out of process, we could write a file or something, but we'll cross that bridge when it comes.

edemaine:
> **Question:** Should `comptime` be settable via config files? Currently it's allowed. This seems like it might be helpful, and is maybe less surprising, but I'm not certain.

What do you think about this one? The safer side would be to disable it, I guess, so it needs to be specified in the build step (CLI or unplugin or esbuild plugin etc.). Thinking about it more, this feels more robust, because we look for config files all the way up the parent, so I vaguely worry about an attack (or accident) where someone places a config file up the tree and then gets arbitrary code execution while compiling. More consistent to disable it I guess.

PR #1184, opened 2024-04-25
Insert "function" into ES6 methods

### LINKED ISSUE #1182: `comptime` object literal with ES6 method generates invalid JS
e.g.
```civet
x := comptime ({ foo() {} })
```
Compiles to:
```js
const x = {"foo":foo() {}};
```

bbrk24:
Fixes #1182 

I just realized while typing this that my first regex doesn't handle default values in arguments to arrow functions, and since the default value can be an IIFE, I would basically have to fully parse JS in order to handle that. That's not a task that regex is sufficient for, so I'm marking this as a draft pending advice on what to do in that scenario.

### DIFF
--- a/source/parser/comptime.civet
+++ b/source/parser/comptime.civet
@@ -74,7 +74,10 @@ function serialize(value: ???): string
         if /\{\s+\[native code]\s+\}$/.test string
           // builtin, or returned from Function::bind
           throw new TypeError "cannot serialize native function"
-        // Check for ES6 methods that have been removed from the object, and add "function"
+        // Check for ES6 methods that have been removed from the object, and add "function" if possible
+        if string.0 is '['
+          // In general, the method name could be an arbitrary JS expression. This is not currently handled.
+          throw new Error "cannot serialize method with computed name"
         unless /^(?:async\s+)?(?:function(?!\p{ID_Continue})|\(|(?:\p{ID_Start}|[_$])(?:\p{ID_Continue}|[\u200C\u200D$])*\s*=>)/u.test string
           // Slightly more annoying than just `'function ' + string`: need to insert it before `async`
           return string.replace /^(async\s+)?/u, (_, maybeAsync = '') => maybeAsync + 'function '
--- a/source/parser/comptime.civet
+++ b/source/parser/comptime.civet
@@ -74,6 +74,12 @@ function serialize(value: ???): string
         if /\{\s+\[native code]\s+\}$/.test string
           // builtin, or returned from Function::bind
           throw new TypeError "cannot serialize native function"
+        // Check for ES6 methods that have been removed from the object, and add "function"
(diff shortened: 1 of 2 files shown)


edemaine:
I believe arrow functions always start with `(` while functions and methods do not. Is that enough to detect arrow functions?

bbrk24:
No, `x => x` is a valid arrow function. I believe `&` generates arrow functions without parens.

edemaine:
Ah, so maybe `\(|\w+\s*=>` matches all arrow functions?

bbrk24:
I think this still fails to deal with objects like `{ ["foo"]() {} }`, actually, but there may not be much we can do about that in the general case.

source/parser/comptime.civet
```diff
@@ -74,6 +74,12 @@ function serialize(value: ???): string
         if /\{\s+\[native code]\s+\}$/.test string
           // builtin, or returned from Function::bind
           throw new TypeError "cannot serialize native function"
+        // Check for ES6 methods that have been removed from the object, and add "function"
+        unless /^(?:async[\s\n]+)?(?:function|\(|\w+[\n\s]*=>)/u.test string
```
edemaine, line 78:
```suggestion
        unless /^(?:async[\s\n]+)?(?:function(?!\p{ID_Continue})|\(|(?:\p{ID_Start}|[_$])(?:\p{ID_Continue}|[\u200C\u200D$])*\s*=>)/u.test string
```

* Don't want to match `functionfoo()`
* Copy regexp from `IdentifierName` to properly handle other leading identifiers. Particularly crucial is `$` which is generated by `&`. (Alternatively, we could probably just use `[^()]*=>`...)
* `\n` is included in `\s` so not sure why you're including it explicitly.

test/comptime.civet
```diff
@@ -139,3 +139,6 @@ describe "serialize", ->
     inner: never[] := []
     outer := [inner, inner]
     assert.equal serialize(outer), "[[],[]]"
+  it "ES6 method", =>
+    assert.equal serialize({ foo() { } }.foo), "function foo() { }"
+    assert.equal serialize({ async bar() { } }.bar), "async function bar() { }"
```
edemaine, line 144: Could you add a test for an `&` function? (probably with arrows)

source/parser/comptime.civet
bbrk24: My implementation already handles `$ => $` correctly which is probably a result of me making the same mistake twice: `replace` is called but no replacement is performed. I seem to remember some regex implementation not counting linebreaks in `\s`, but I can't find any where that's the case, so I was probably just mistaken. I'll update this and the other regex accordingly.

bbrk24: Actually, I think this can be done with a single regex -- the second one is unnecessary, and the first one can be passed directly to `replace` after removing the first `?:`.

edemaine (approved):
Nice!

PR #1186, opened 2024-04-26
Async compile API with comptime, support require

edemaine:
This changes the main `compile` interface to return a promise ~~when `comptime: true`~~ unless `sync: true` is specified. It remains as synchronous as possible, even implementing basic `comptime`, but not advanced features like `require` and awaiting an asynchronous comptime block.

`parse` remains synchronous in all cases, and we still use it for parsing the occasional extra rule.

A new `parseProgram` in `parser.hera` is now where `filename` and `initialConfig` get set (cleaning up a bit of a hack in `main.civet`), and where the interface becomes potentially asynchronous (when `comptime: true`), via a new `processProgramAsync` second round.

~~Soon~~ This now enables async code with `await` in `comptime` blocks! I've also added support for `require` in `comptime` blocks, by calling `createRequire` and setting the global `require` (along with `__filename` and `__dirname`). Of course, this only works on Node or similar backend platform; in the browser build, these dynamic `import`s fail and no `require` gets built.

Unfortunately, I realize the reference documentation is somewhere in between: it has examples with `comptime` but that don't need `require`, and because of Markdown limitations I believe it has to be done synchronously. So I need to go back to the drawing board here... Some options I see:

* `comptime: 'sync'` is a special `comptime` mode that processes `comptime` blocks but keeps it synchronous (so can't handle `require`). Starting to get a little weird API-wise though...
* `sync: true` options to specify the intended interface. [CHOSEN INTERFACE]
* `compileSync` as an explicitly synchronous interface alternative to `compile`, just more limited than `compile` (no `require` in `comptime` blocks). I wanted to avoid this because `compile` has quite a bit of code, but I could refactor it. Perhaps this is cleanest?
* Alternatively, `compile` could remain synchronous but limited, while `compileAsync` becomes the async interface. This is good for backwards compatibility, but really users should migrate to the async interface when possible.


STRd6:
I think long term we'll need `compile` to be async, especially once we have more TS integration and mutli-file interactions. In the meantime we can add `compileAsync` and plan to shift it to the default at some point.

edemaine:
OK, I'm going to try to build two APIs that can co-exist with as little code duplication as possible. Is it OK to use `compile` (async) and `compileSync`? Or would you prefer `compile` (sync) and `compileAsync`?

edemaine:
Actually, thinking more about avoiding code duplication, I think it would be easier to have `sync` or `async` be an option passed into the `compile` interface. Do you think this would be OK too?

STRd6:
It's fine to switch `compile` to async now since we currently don't have many dependents. If it breaks for someone then they can adjust their code to `compileSync`.

STRd6:
Adding a `sync` option is also fine 👍

edemaine:
OK, I implemented the `sync: true` interface, and updated what I believe to be all uses of `compile` throughout the codebase. We'll need to release new versions of:

* Gulp plugin (I can do this)
* VSCode plugin / LSP (Daniel needs to do this)
* Discord bot (I can do this, still need to edit this one)

I think all of these should be done *before* we release the next version of Civet, to avoid them crashing with the latest Civet.

STRd6 (approved):

PR #1187, opened 2024-04-26
Allow empty body in loops, if/else, do, comptime

### LINKED ISSUE #1162: `yield x; loop` has strange behavior
This is obviously a ParseError:

```
f := (x) ->
  yield x
  loop
```

However, this creates an array and pushes an infinite number of generator functions to it:

```
f := (x) ->
  yield x; loop
```

Shouldn't they both be errors?

edemaine:
I propose allowing for empty blocks in most settings. We already support them in `function`s, and `case`/`when`/pattern matching. This PR adds them to all loops, if/else, and for symmetry, do blocks and comptime.

CoffeeScript supports this in `if/else` expressions that have both `if` and `else`: [both empty](https://coffeescript.org/#try:if%20x%0Aelse), [empty if](https://coffeescript.org/#try:if%20x%0Aelse%0A%20%20y). I think this hints at why it is useful:

* Suppose you have code with an if/else construct, and you comment out the then or else block. Currently, that makes a parse error. Commenting out the else block is not too bad: just comment out the `else` keyword too. But commenting out the if block is hard: you need to comment out the `if` line, the `else` keyword, *and dedent the else block*. This makes for a lot of code changes if you want to debug removing some code (e.g. to isolate a bug).
* This is exacerbated by comments not counting as a block. So you can't do e.g.
  ```coffee
  if empty
    // don't need to do anything in this case
  else
    doStuff()
  ```
  I've wanted to write code like this at times. Currently it's a parse error in Civet.
* Related, as you're typing lines in VSCode, I feel like it's less likely that you have code with parse errors, so you keep the useful feedback in more situations.

My guess is that CoffeeScript couldn't handle empty `if` blocks without `else`, and loop blocks, because they are somewhat ambiguous with postfix if/loops. But a little work on the "ExpressionizedStatement" code made it easy to not expressionize in this case of empty blocks.

For example, this fixes #1162: the surprising behavior was the `loop` becoming a postfix after a semicolon, but when it can be an infinite loop by itself, the behavior is intuitive (though not very useful). Similarly, @bbrk24 mentioned `x := comptime loop` being counterintuitive: it [used to](https://civet.dev/playground?code=eCA6PSBjb21wdGltZSBsb29w) give an infinite array of `comptime` identifiers, but now it gives the more natural but not very useful behavior of adding an infinite loop to a comptime function.

### DIFF
--- a/source/parser.hera
+++ b/source/parser.hera
@@ -3959,7 +3959,7 @@ LabelledItem
 IfStatement
   # NOTE: Added paren-less condition
   # NOTE: Block isn't Statement so we can handle implied braces by nesting
-  ( IfClause / UnlessClause ):clause Block:block ElseClause?:e ->
+  ( IfClause / UnlessClause ):clause BlockOrEmpty:block ElseClause?:e ->
     return {
       type: "IfStatement",
       children: [...clause.children, block, e],
@@ -3969,8 +3969,8 @@ IfStatement
     }
 
 ElseClause
-  Nested Else Block
-  _? Else Block
+  Nested Else BlockOrEmpty
+  _? Else BlockOrEmpty
 
 IfClause
   If Condition:condition ->
--- a/source/parser.hera
+++ b/source/parser.hera
@@ -303,6 +303,7 @@ ForbiddenImplicitCalls
   OmittedNegation _? Identifier:id ->
     if (state.operators.has(id.name)) return $0
     return $skip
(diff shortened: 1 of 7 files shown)


bbrk24:
Funny enough, I encountered #1162 for the exact reason you mentioned about commenting out code. Frankly I don't find either interpretation of `comptime loop` "useful", but the new way is more intuitive.

STRd6 (approved):

PR #1188, opened 2024-04-27
Serialize typed arrays and well-known symbols

bbrk24:
Typed arrays were briefly discussed on the discord last night. While I didn't mention well-known symbols, they seem useful and were also pretty straightforward to implement.

I believe symbol-keyed properties in objects are still not serialized, but that can be a separate issue.

### DIFF
--- a/source/parser/comptime.civet
+++ b/source/parser/comptime.civet
@@ -125,7 +125,8 @@ function serialize(value: ???): string
               `new ${val.constructor.name}([${(val as any).join ','}])`
             when BigInt64Array::, BigUint64Array::
               `new ${val.constructor.name}([${Array.from(val as ArrayLike<bigint>, `${&}n`).join ','}])`
-            when Buffer::
+            // Spelled differently for browsers, where `Buffer` doesn't exist
+            when globalThis.Buffer?.prototype
               `Buffer.from([${(val as Buffer).join ','}])`
             else
               // One day we may handle other classes like so:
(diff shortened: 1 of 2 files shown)


source/parser/comptime.civet
```diff
@@ -114,6 +120,13 @@ function serialize(value: ???): string
                 for own key, value in val as {[key: string]: ???}
                   `${JSON.stringify key}:${recurse value}`
               ).join(",") + "}"
+            when Int8Array::, Uint8Array::, Int16Array::, Uint16Array::, Int32Array::, Uint32Array::, Float32Array::, Float64Array::, Uint8ClampedArray::
+              // There's no "TypedArray" interface in TS
+              `new ${val.constructor.name}([${(val as any).join ','}])`
+            when BigInt64Array::, BigUint64Array::
+              `new ${val.constructor.name}([${Array.from(val as ArrayLike<bigint>, `${&}n`).join ','}])`
+            when Buffer::
```
edemaine, line 128: Can you test this on the browser build? (yarn docs:dev, click on Playground) I'm worried about what happens when this global isn't defined. Maybe `global.Buffer` works better?

bbrk24: It seems to work fine, but I'm surprised that it does. I'll change this to `global.Buffer?::` or whatever the correct syntax is.

bbrk24: So not only does that spelling not work (#1191), `global` doesn't exist in the browser, either!

edemaine: Hmm. `globalThis`?

bbrk24: Yep, `globalThis` seemed to work. Thank you JavaScript very cool

edemaine (approved):

PR #1190, opened 2024-04-27
Fix comptime negative zero

bbrk24:


edemaine (commented):
1. Maybe we should separate the number case from the others to avoid an unnecessary comparison in the others?
2. Perhaps we should use `.toPrecision()` on numbers? I think this guarantees full reconstruction for all numbers except `-0` (but I haven't checked Infinity, NaN, etc.)

bbrk24:
1 is a good point, I'll do that. Without an argument to either one, `toString()` and `toPrecision()` seem to do the same thing:
image

edemaine:
Ah, this seems to be [documented](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_objects/Number/toPrecision) too:

> If the precision argument is omitted, behaves as Number.prototype.toString().

edemaine (approved):

PR #1194, opened 2024-04-27
Comptime classes and generators

### LINKED ISSUE #1189: Comptime generator functions
Currently, generator functions are rejected by the comptime serializer. There's a few reasons for this. Given a `function* f()`:

- `Object.getPrototypeOf f::` is not `Object::`, but rather a special generator prototype. I added this check as an attempt to detect and reject classes, but this [does not work](https://civet.dev/playground?code=eCA6PSBjb21wdGltZSBjbGFzcyBY) anyways.
- `f::` is an object with no constructor. The check currently expects either `undefined`, or an object whose constructor is the function itself.
- If it's a method on an object literal, i.e. `{ *f() {} }` the current check for computed names wouldn't work anyways. It would have to be changed from `string.0 is '['` to `/^(\*\s*)?\[/.test string`, since the `*` for a generator method with a computed name goes before the open bracket.

bbrk24:
Fixes #1189 

Technically, this doesn't detect the case where a class's prototype is modified after initialization, i.e:

```civet
C := comptime do
  class C {}
  C.prototype.toString = -> 'C'
  C
```

but I'm not 100% sure this is possible in the general case. In any case, the current behavior of creating `function class C {}` is definitely wrong.

### DIFF
--- a/test/comptime.civet
+++ b/test/comptime.civet
@@ -148,6 +148,10 @@ describe "serialize", ->
     func: any := (x: ???) => x
     func.a = 1
     assert.throws (=> serialize func), /cannot serialize function with properties/
+
+    func2: any := (@: object) -> @
+    func2.prototype.a = 1
+    assert.throws (=> serialize func2), /cannot serialize function with modified prototype/
   it "native functions", =>
     assert.throws (=> serialize parseInt), /cannot serialize native function/
   it "circular", =>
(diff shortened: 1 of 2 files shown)


test/comptime.civet
```diff
@@ -165,6 +169,8 @@ describe "serialize", ->
     assert.equal serialize({ functionF() { } }.functionF), "function functionF() { }"
     assert.equal serialize({ async bar() { } }.bar), "async function bar() { }"
     assert.throws (=> serialize { [Math.sqrt 5]() {} }), /cannot serialize method with computed name/
+    assert.throws (=> serialize { *[Math.PI]() {} }), /cannot serialize method with computed name/
```
edemaine, line 172: Maybe add a positive test for a generator method that does work?

edemaine (approved):

PR #1196, opened 2024-04-28
Comptime function properties

bbrk24:
Serialize function properties for `comptime` using `Object.defineProperties`. We could afford to do this for properties that aren't `{ +configurable, +writable, +enumerable }` on `Object`s too, but that isn't covered in this PR. This PR also fixes a couple things with circular reference detection, now that functions can be circular.

### DIFF
--- a/package.json
+++ b/package.json
@@ -128,7 +128,6 @@
       "coffee"
     ],
     "loader": [
-      "ts-node/esm",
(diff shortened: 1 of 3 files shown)


edemaine (approved):

edemaine:
No more ts-node for tests! Performance boost!! 🥳

PR #1198, opened 2024-04-28
Handle property descriptors (incl. getters/setters) in comptime

bbrk24:


source/parser/comptime.civet
```diff
@@ -139,10 +139,31 @@ function serialize(value: ???): string
...
-              "{" + (
-                for own key, value in val as {[key: string]: ???}
-                  `${JSON.stringify key}:${recurse value}`
-              ).join(",") + "}"
+              // The string representing the object's own enumerable, writable properties
+              objStr .= '{'
+              // The string representing other property descriptors, if applicable, not including the outermost braces
+              descStr .= ''
+              descriptors := Object.getOwnPropertyDescriptors(val) as Record<string | symbol, PropertyDescriptor>
+              for prop of (Object.getOwnPropertyNames(val) as (string|symbol)[]) ++ Object.getOwnPropertySymbols(val)
```
edemaine, line 147:
```suggestion
              for each prop of (Object.getOwnPropertyNames(val) as (string|symbol)[]) ++ Object.getOwnPropertySymbols(val)
```

Wow, enumeration is hard. I suppose we could call `Object.getOwnPropertyDescriptor` inside the loop instead of the plural form outside the loop, but this way may be more time-efficient (if a bit less memory efficient).

source/parser/comptime.civet
```diff
@@ -139,10 +139,31 @@ function serialize(value: ???): string
...
+                  else
+                    objStr += JSON.stringify prop
+                  objStr += `:${recurse desc.value},`
+                else
+                  if prop <? 'symbol'
+                    descStr += `[${recurse prop}]`
+                  else
+                    descStr += JSON.stringify prop
+                  descStr += `:${recurse desc},`
+              objStr += '}'
```
edemaine, line 161: Could we remove the trailing commas from `objStr` and `descStr`? Something like `if (objStr# > 1) objStr = objStr[...-1]` and `descStr = descStr[...-1]` inside the next `if`.

source/parser/comptime.civet
```diff
@@ -139,10 +139,31 @@ function serialize(value: ???): string
...
-                  `${JSON.stringify key}:${recurse value}`
-              ).join(",") + "}"
+              // The string representing the object's own enumerable, writable properties
+              objStr .= '{'
+              // The string representing other property descriptors, if applicable, not including the outermost braces
+              descStr .= ''
+              descriptors := Object.getOwnPropertyDescriptors(val) as Record<string | symbol, PropertyDescriptor>
+              for prop of (Object.getOwnPropertyNames(val) as (string|symbol)[]) ++ Object.getOwnPropertySymbols(val)
+                desc := descriptors[prop]
+                if desc.enumerable and desc.configurable and desc.writable
```
edemaine, line 149: Do you need to check for `get` and `set` for custom getters and setters? Or I guess that makes an accessor descriptor which won't have `writable` set?

bbrk24: A descriptor will never have both `writable` and `get` defined, and `undefined` is falsy.

source/parser/comptime.civet
bbrk24: Good point, I think I will do `getOwnPropertyDescriptor`, if only to make this a little easier to read.

edemaine (approved):
Nice! In the future, feel free to update docs in the same PR. I can do it this time.

PR #1199, opened 2024-04-28
"is like" pattern matching operator

### LINKED ISSUE #391: Pattern matching operator
Proposal: `=~` (as in Perl and Ruby) and `is like` operator to do the equivalent of a single [pattern match `switch`](https://civet.dev/cheatsheet#pattern-matching). For example:

```coffee
match := x =~ [/\s+/, ...rest]
---
const match = Array.isArray(x) && x.length >= 1 && /\s+/.test(x[0])
const [, ...rest] = match ? x : [undefined]
```

```coffee
if x =~ [/\s+/, ...rest]
  console.log rest
---
if Array.isArray(x) && x.length >= 1 && /\s+/.test(x[0])
  const [, ...rest] = x
  console.log(rest)
```

(based on discussions in Discord)

edemaine:
Fixes #391

Most of this PR is code cleanup:
* `getPatternConditions` no longer requires a third `conditions` argument; it starts as an empty array that gets returned
* This lets the condition combination code get simplified a lot (also thanks to the recent `.` feature)
* New `source/parser/ref.civet` to house ref helpers
* New `makeRefAssignment` to ease creation of `hoistDec` and `refAssignment`
* New `maybeRefAssignment` = `maybeRef` + `makeRefAssignment`
* New `flatJoin` to ease combining expressions with ` && ` or ` || `
* More types (but still incomplete)

### DIFF
--- a/civet.dev/reference.md
+++ b/civet.dev/reference.md
@@ -66,6 +66,9 @@ if [, dir, base] := /^(.*\/)?([^/]*)$/.exec file
   console.log dir, base
 </Playground>
 
+Note that array lengths must match exactly because this is a form of
+[pattern matching](#pattern-matching).
+
 <Playground>
 if {x, y} := getLocation()
   console.log `At ${x}, ${y}`
@@ -1197,7 +1200,7 @@ switch x
 
 <Playground>
 switch x
-  [{type: "text", content: /\s+/}, ...rest]
+  [{type: "text", content: /^\s+$/}, ...rest]
     console.log "leading whitespace"
   [{type: "text", content}, ...rest]
     console.log "leading text:", content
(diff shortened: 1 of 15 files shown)


STRd6 (approved):
Thanks, looks great!

I was thinking about chaining `is like` with other conditionals but couldn't really think of anything that made sense so limiting it to the last position seems reasonable.

edemaine:
I think we can add support for other positions in the future, but my thought was that everything else would end up being consumed by the pattern, so last position makes sense.

But we could have conditionals *before* `is like`. For example, `x? is like ...` could make sense as `x? and x is like ...`, similar to how `x? < y?` works now. Currently it matches the Boolean `x?` which isn't terrible useful. Similarly, `x < y is like ...` could make sense as `x < y && y is like ...`.

PR #1200, opened 2024-04-29
Pattern matching fixes

### LINKED ISSUE #1147: Pattern matching with literals generates incorrect code
```coffee
if {type: "Identifier", name: "Infinity"} := end
  true
```
[generates](https://civet.dev/playground?code=aWYge3R5cGU6ICJJZGVudGlmaWVyIiwgbmFtZTogIkluZmluaXR5In0gOj0gZW5kCiAgdHJ1ZQ%3D%3D) this invalid code:
```js
let ref;if ((ref = end) && 'type' in ref && ref.type === "Identifier" && 'name' in ref && ref.name === "Infinity") {
  const {type: "Identifier", name: "Infinity"} = ref;
  true
}
```

### LINKED ISSUE #1201: Empty array matching still broken
```coffee
switch x
  {children: []}
    console.log "empty"
```
[generates](https://civet.dev/playground?code=c3dpdGNoIHgKICB7Y2hpbGRyZW46IFtdfQogICAgY29uc29sZS5sb2cgImVtcHR5Igo%3D) the incorrect code
```js
if(typeof x === 'object' && x != null && 'children' in x && Array.isArray(x.children) && len(x.children, 0)) {
    const {children: } = x;
    console.log("empty")}
```

edemaine:
Generally share more code between pattern matching and declaration conditions to fix the following:

* Fix declaration conditions with literals (fix #1147)
* Fix duplicate bindings in declaration conditions
* Bonus: Avoid refs in simple declaration conditions
* Added: Fix empty arrays in object pattern matching (fix #1201)

### DIFF
--- a/source/parser/types.civet
+++ b/source/parser/types.civet
@@ -433,7 +433,7 @@ export type TryStatement
 
 export type BindingIdentifier = AtBinding | Identifier | ReturnValue
 
-export type BindingPattern = BindingRestElement | ObjectBindingPattern | ArrayBindingPattern | PinPattern | Literal | RegularExpressionLiteral | RegularExpressionLiteral
+export type BindingPattern = BindingRestElement | ObjectBindingPattern | ArrayBindingPattern | PinPattern | Literal | RegularExpressionLiteral
 
 export type PatternExpression = BindingPattern | ConditionFragment
 
--- a/source/parser/types.civet
+++ b/source/parser/types.civet
@@ -433,7 +433,7 @@ export type TryStatement
 
 export type BindingIdentifier = AtBinding | Identifier | ReturnValue
 
-export type BindingPattern = BindingRestElement | ObjectBindingPattern | ArrayBindingPattern | PinPattern | Literal | RegularExpressionLiteral
+export type BindingPattern = BindingRestElement | ObjectBindingPattern | ArrayBindingPattern | PinPattern | Literal | RegularExpressionLiteral | RegularExpressionLiteral
 
 export type PatternExpression = BindingPattern | ConditionFragment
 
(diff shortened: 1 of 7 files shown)


source/parser/types.civet
```diff
@@ -433,7 +433,7 @@ export type TryStatement
 
 export type BindingIdentifier = AtBinding | Identifier | ReturnValue
 
-export type BindingPattern = BindingRestElement | ObjectBindingPattern | ArrayBindingPattern | PinPattern | Literal | RegularExpressionLiteral
+export type BindingPattern = BindingRestElement | ObjectBindingPattern | ArrayBindingPattern | PinPattern | Literal | RegularExpressionLiteral | RegularExpressionLiteral
```
STRd6, line 436: It looks like `RegularExpressionLiteral` is duplicated here.

STRd6 (approved):

test/switch.civet
```diff
@@ -776,6 +777,31 @@ describe "switch", ->
...
+    testCase """
+      empty array in object
+      ---
+      switch x
+        {children: []}
+          console.log "empty"
+      ---
+      function len<T extends readonly unknown[], N extends number>(arr: T, length: N): arr is T & { length: N } { return arr.length === length }
+      if(typeof x === 'object' && x != null && 'children' in x && Array.isArray(x.children) && len(x.children, 0)) {
+          const {children} = x;
```
STRd6, line 789: Oops, I missed this earlier but it should probably compile to `const {children: []}` otherwise it will shadow a possible `children` declared in the outer scope. Same for empty objects as well.

edemaine: Oh, whoops, sorry. My bad adding commits. I hadn't realized the difference but makes sense.

PR #1202, opened 2024-04-30
Fix handling of empty arrays and objects in patterns

edemaine:
Fix the issue mentioned here: https://github.com/DanielXMoore/Civet/pull/1200#discussion_r1583864309

This turned out to be tricky to figure out, as it had to do with the generator!  ArrayBindingPattern nodes have a `length` field which was zero in this case. We could instead remove `length` fields from AST node objects, but it seems like a useful field name (I'd rather have it *defined* for ObjectBindingPattern too so that code could be more symmetric), and maybe used a fair amount around array patterns, so changing the generator seemed better.  Note that empty arrays are still also removed, using the code down below.

### DIFF
--- a/source/generate.civet
+++ b/source/generate.civet
@@ -69,18 +69,16 @@ export default gen
 // Useful for debugging so I don't need to step though tons of empty nodes
 // Also remove parent pointers so we can JSON.stringify the tree
 export function prune(node: any): any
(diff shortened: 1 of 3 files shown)


STRd6 (approved):
Good catch, thanks!

PR #1208, opened 2024-05-02
Use `is like` in some cases

edemaine:
I think these show some nice examples of `is like`.

On the other hand, I currently can't get VSCode to parse these files with these changes (because it's refusing to use updated Civet in `node_modules`), so can't get TS feedback, so we might want to fix that first...

### DIFF
--- a/source/parser/op.civet
+++ b/source/parser/op.civet
@@ -139,8 +139,10 @@ function processBinaryOpExpression($0)
           b = recurse b
 
         // typeof shorthand: x instanceof "String" -> typeof x === "string"
-        if op.token === "instanceof" and b.type === "Literal" and
-           b.children is like [ {type: "StringLiteral"}, ... ]
+        if op.token === "instanceof" and b is like {
+          type: "Literal"
+          children: [ {type: "StringLiteral"}, ... ]
+        }
           a = ["typeof ", makeLeftHandSideExpression(a)]
           if op.negated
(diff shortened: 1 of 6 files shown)


STRd6:
> On the other hand, I currently can't get VSCode to parse these files with these changes (because it's refusing to use updated Civet in `node_modules`), so can't get TS feedback, so we might want to fix that first...

I had a similar problem recently and had to quit and restart VSCode which seemed to fix it... 🤷

source/parser/op.civet
```diff
@@ -140,7 +140,7 @@ function processBinaryOpExpression($0)
 
         // typeof shorthand: x instanceof "String" -> typeof x === "string"
         if op.token === "instanceof" and b.type === "Literal" and
-           b.children?.[0]?.type === "StringLiteral"
+           b.children is like [ {type: "StringLiteral"}, ... ]
```
STRd6, line 143: This one isn't that much clearer with 'is like'.

Maybe:

```
b is like {type: "Literal", children: [{type: "StringLiteral"}, ...]}
```

to include more of the structure of `b` in the check.

STRd6 (approved):

edemaine:
I figured out the issue: when running the VSCode *on the Civet repo* (at least when that repo is `npm link`ed?), I get the live version of Civet from that repo in `dist`. I hadn't run `yarn build` in a while, so I wasn't getting the updates, despite the version number appearing correct.

PR #1213, opened 2024-05-04
CLI avoid errors when output pipe gets closed

edemaine:
Tested on Windows and Linux, with
```sh
civet -c - <source/main.civet | head -5
```
Previously we got error messages like
```
const o=[Object.preventExtensions({"o":unode:events:496
      throw er; // Unhandled 'error' event
      ^

Error: write EPIPE
    at WriteWrap.onWriteComplete [as oncomplete] (node:internal/stream_base_commons:94:16)
Emitted 'error' event on Socket instance at:
    at emitErrorNT (node:internal/streams/destroy:169:8)
    at emitErrorCloseNT (node:internal/streams/destroy:128:3)
    at process.processTicksAndRejections (node:internal/process/task_queues:82:21) {
  errno: -32,
  code: 'EPIPE',
  syscall: 'write'
}

Node.js v20.12.2
```

or

```
node:events:497
      throw er; // Unhandled 'error' event
      ^

Error: EOF: end of file, write
    at Socket._write (node:internal/net:61:18)
    at writeOrBuffer (node:internal/streams/writable:564:12)
    at _write (node:internal/streams/writable:493:10)
    at Writable.write (node:internal/streams/writable:502:10)
    at cli (C:\Users\edemaine\Projects\Civet\dist\civet:508:26)
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
Emitted 'error' event on Socket instance at:
    at emitErrorNT (node:internal/streams/destroy:169:8)
    at emitErrorCloseNT (node:internal/streams/destroy:128:3)
    at process.processTicksAndRejections (node:internal/process/task_queues:82:21) {
  errno: -4095,
  syscall: 'write',
  code: 'EOF'
}

Node.js v21.7.3
```


STRd6 (approved):
Is there a different error that would trigger when writing to a file and running out of disk space?

My only concern is that we should error in that case but if it is too difficult to distinguish then we shouldn't worry about it since this seems to be more common.

edemaine:
Currently I'm just hiding two error messages that I've seen in this piping situation: EOF and EPIPE. So any other error messages should still be passed through. In particular ENOSPC seems to be what would happen when the disk is full (but I haven't tested, as it's not very easy to.)

bbrk24:
Shouldn't `-o /dev/full` produce the "drive full" error?

PR #1214, opened 2024-05-04
Better require error messages in comptime

edemaine:


STRd6 (approved):

edemaine:
By the way, I'm thinking `throws` should allow for specifying what error message we get. What do you think would be the ideal format (presumably something after `---`)? Perhaps the exact contents of `error.toString()`, which seems to be of the form `type: message`. Alternatively, we might need regexes in some contexts, if the error message varies, but I can't think of a setting where that's the case.

STRd6:
Adding an optional exact error message to match to throws sounds good, something like:

````
throws ```
test name
---
...
---
exact error message
```
````

And allow omitting the final `---` to accept any error message.

PR #1216, opened 2024-05-04
Don't emit newlines before block prefixes

### LINKED ISSUE #1113: @ts-ignore ignored before for..in loop
```coffee
obj :=
  a : 0
  b : 1
  c : 2

// @ts-ignore
for key, value in obj
//       ^^^^------ typescript still complains (with "noImplicitAny" enabled)
  console.log value
```

bbrk24:
Fixes #1113 

90% of the effort here was fixing all the tests.


STRd6 (approved):
Thanks for all the test cleanup!

PR #1217, opened 2024-05-04
throws can specify exception; ParseErrors type; bug fixes

edemaine:
* As discussed in #1214, `throws` can now take a third string separated by `---`. It can specify either:
  * Just the error `name`
  * `name: message`, but skipping the verbose part of `ParseError`s
* Add new `ParseErrors` class to represent errors generated from `"Error"` nodes
* Fix a few errors that were crashing instead of generating `ParseError(s)`. This was actually a useful exercise!

### DIFF
--- a/source/parser/lib.civet
+++ b/source/parser/lib.civet
@@ -371,10 +371,19 @@ function processCallMemberExpression(node: ASTNodeParent): ASTNode
       prefix = prefix.concat(glob.dot)
 
       for part of glob.object.properties
+        if part.type is "Error"
+          parts.push part
+          continue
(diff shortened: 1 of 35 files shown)


source/parser/lib.civet
```diff
@@ -517,7 +526,10 @@ function lastAccessInCallExpression(exp)
   return exp if exp.type is "Identifier"
   let children, i
   do
+    // Leaf node occurs e.g. for import.meta
+    return unless exp.children?
     { children } = exp
+    console.log exp unless children?
```
STRd6, line 532: Should remove this `console.log`

PR #1220, opened 2024-05-05
Use context's globals instead of serializing there

edemaine:
#1219 was kind of ugly: it serialized the `serialize` function, parsed it in the comptime context, and then ran it there, so that it could have access to the context's globals.

This PR takes what I think is a cleaner approach: run the `serialize` in our usual context, but give it access to the comptime context's `globalThis` and check for matching objects from either `globalThis`.

In both approaches we have to copy globals over to comptime context, because new contexts don't get the usual Node builtins, such as `URL`. (See https://github.com/nodejs/node/issues/28823) I'm now doing this in the way that `node:repl` does, which is better: it gives the comptime context access to important Node globals like `fetch`, `atob`, `process`, etc. This also restores Node's `global` which seems to be global to all modules, whereas `globalThis` is local to the module, which makes sense and enables some magic interactions.

### DIFF
--- a/source/parser/comptime.civet
+++ b/source/parser/comptime.civet
@@ -57,25 +57,23 @@ function runComptime(statements: StatementTuple[]): unknown[]
     return if options.errors?
 
     // Run the JS code at compile time (now)
-    let output: unknown, context: object
+    let output: unknown, context: NodeJS.Dict<unknown>, contextGlobal?: NodeJS.Dict<unknown>
     try
       context = vm.createContext?() ?? globalThis
       filename := context.__filename = resolve getFilename() ?? ""
       context.__dirname = dirname filename
       context.require = createRequire filename
       if vm.runInContext?
-        // Copy serializable globals over to vm context
-        for global of [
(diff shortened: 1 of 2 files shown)


source/parser/comptime.civet
```diff
@@ -203,22 +194,22 @@ function serialize(value: ???): string
         stack.add val
         str :=
           switch Object.getPrototypeOf val
-            when RegExp::
+            when RegExp::, context?.RegExp?::
```
bbrk24, line 197: Are both `?`s necessary? If `context` is `undefined` the rest of the chain is skipped.

edemaine: Fair point. I think in the future we might want to add *more* `?`s before `::`, for both local and remote checks, but given that we currently require that `RegExp` (say) is present, we can assume that it's also in `context`, given how we copy over all globals. (I wonder if there are JS runtimes e.g. with no `TypedArray` support. But we can wait for that to arise.)

STRd6 (approved):

PR #1222, opened 2024-05-05
comptime else blocks, else refactor

edemaine:
Most of the work here is refactoring `ElseClause` into its own AST object node, instead of the previous use of a tuple. I think this is worth doing; a little less magic working when working with these clauses. Also added missing `BlockStatement`s when generating these in pattern matching. Also cleaned up that pattern matching code a bit, which took a while.

Added feature: `comptime` blocks can specify an `else` block to specify a different thing to do when `comptime` mode is off. This can be useful for typing, especially `Promise` vs. non-`Promise`. It can also help with things like linting when `comptime` is off: you get to specify the effective code that should be linted.

### DIFF
--- a/source/parser.hera
+++ b/source/parser.hera
@@ -4025,12 +4025,7 @@ IfStatement
     }
 
 ElseClause
-  Nested Else BlockOrEmpty:block -> {
-    type: "ElseClause",
-    children: $0,
-    block,
-  }
-  _? Else BlockOrEmpty:block -> {
+  ( Nested / _? ) Else BlockOrEmpty:block -> {
     type: "ElseClause",
     children: $0,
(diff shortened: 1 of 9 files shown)


source/parser.hera
```diff
@@ -4025,8 +4025,16 @@ IfStatement
     }
 
 ElseClause
-  Nested Else BlockOrEmpty
-  _? Else BlockOrEmpty
+  Nested Else BlockOrEmpty:block -> {
```
STRd6, line 4028: The could be combined into `( Nested / _? ) Else BlockOrEmpty:block` to prevent a little backtracking.

STRd6 (approved):

PR #1228, opened 2024-05-06
Declaration inside unless/until adds declaration after block

### LINKED ISSUE #756: `unless const` is a bit useless
Comparing Swift and Civet here

```swift
if let foo = getFoo() {
  // foo is in scope and not nil
}
// foo is out of scope

guard let foo = getFoo() else {
  // foo is out of scope
}
// foo is in scope and not nil
```

```civet
if foo := getFoo()
  // foo is in scope and not null
// foo is out of scope

unless foo := getFoo()
  // foo is in scope but guaranteed to be null
// foo is out of scope so the value is unusable
```

### LINKED ISSUE #1215: `guard`-`else`
Now that we have `isExit` for other reasons, Swift-style `guard` seems implementable. At first glance, it seems identical to `unless`:
```js
guard CONDITION else {
  STATEMENTS
}
// vvv
if (!CONDITION) {
  STATEMENTS
}
```
However, there are two key differences:
1. If the condition creates any bindings (such as `guard x? := getX() else { ... }`), the bindings are only available *outside* the `else` block.
2. `STATEMENTS` must be an "exit" of some sort, in order for `1` to be possible.

The primary use case is to prevent "pyramids of doom" when there are many bindings to be done:
```
if x? := getX()
  if y? := getY()
    if z? := getZ()
      setPoint x, y, z
// vvv
guard x? := getX() else return
guard y? := getY() else return
guard z? := getZ() else return
setPoint x, y, z
```

edemaine:
So I implemented the idea that declarations of `unless` and `until` go after the block instead of inside:

```js
unless x := f()
  return
console.log x
↓↓↓
let ref
if ((ref = f()) {
  return
}
const x = ref
console.log(x)
```

However, I'm no longer sure this is a good idea. For example, what if we have an `else` block?

```js
unless x := f()
  return
else
  // should x be bound here?
// should x be bound here?
```

I'm pretty sure `x` should be bound inside the `else`, and for symmetry with the other situations, I think also afterward? So my temptation would be to put the declaration of `x` above the `unless` instead of after it, so that it's accessible everywhere. It doesn't seem particularly bad to expose it to the "then" clause — this is actually more symmetric with `if` — what's important is that it's exposed in the "else clause and outside block, because it can be useful there.  Whereas `if` keeps the protection of the binding to the "then" clause, because that's the only place it is helpful. What do you think?

Fixes #1215, fixes #756

### DIFF
--- a/source/parser/declaration.civet
+++ b/source/parser/declaration.civet
@@ -18,18 +18,20 @@ import type {
 
 import {
   blockWithPrefix
+  braceBlock
   makeEmptyBlock
   replaceBlockExpression
 } from ./block.civet
 
 import {
+  findAncestor
+  findChildIndex
   gatherRecursiveAll
 } from ./traversal.civet
 
 import {
   getPatternBlockPrefix
   getPatternConditions
-  nonMatcherBindings
 } from ./pattern-matching.civet
 
 import {
@@ -264,10 +266,13 @@ function processDeclarationConditionStatement(s: IfStatement | IterationStatemen
   { condition } := s
   return unless condition?.expression
   { expression } .= condition
+
   // Support for negated conditions built by unless/until
-  switch expression
-    {type: 'UnaryExpression', children: ['!', {type: 'ParenthesizedExpression', expression: expression2}]}
-      expression = expression2
+  negated .= false
+  if {type: 'UnaryExpression', children: ['!', {type: 'ParenthesizedExpression', expression: expression2}]} := expression
+    expression = expression2
+    negated = true
+
   processDeclarationCondition expression, condition.expression, s
 
   { ref, pattern } := expression
@@ -290,6 +295,24 @@ function processDeclarationConditionStatement(s: IfStatement | IterationStatemen
       conditions.forEach (c) =>
         condition.children.push " && ", c
       condition.children.push ")"
+
+  if negated and condition.expression.blockPrefix
+    { ancestor, child } := findAncestor s,
+      (a): a is BlockStatement => a.type is "BlockStatement"
+    unless ancestor?
+      throw new Error "Couldn't find block for postfix declaration"
+    index := findChildIndex ancestor.expressions, child
+    if index < 0
+      throw new Error "Couldn't find where in block to put postfix declaration"
+    ancestor.expressions.splice index + 1, 0, ...condition.expression.blockPrefix
+    updateParentPointers ancestor
+    braceBlock ancestor
+    switch s.type
+      when "IfStatement"
+        block := s.else?.block ?? s.then
+        if block.bare and not block.semicolon
+          block.children.push block.semicolon = ";"
+    return
 
   switch s.type
     when "IfStatement"
--- a/source/parser/declaration.civet
+++ b/source/parser/declaration.civet
@@ -5,11 +5,11 @@ import type {
   ASTRef
   Binding
   BlockStatement
-  Children
   DeclarationStatement
   IfStatement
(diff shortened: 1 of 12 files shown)


source/parser/declaration.civet
```diff
@@ -264,10 +266,13 @@ function processDeclarationConditionStatement(s: IfStatement | IterationStatemen
   { condition } := s
   return unless condition?.expression
   { expression } .= condition
+
   // Support for negated conditions built by unless/until
-  switch expression
-    {type: 'UnaryExpression', children: ['!', {type: 'ParenthesizedExpression', expression: expression2}]}
-      expression = expression2
+  negated .= false
+  if {type: 'UnaryExpression', children: ['!', {type: 'ParenthesizedExpression', expression: expression2}]} := expression
```
STRd6, line 272: We may want to add the negated flag as a property on the `IfStatement` node to distinguish between

`if !(x)` and `unless x`

STRd6:
I think moving the declaration above the block sounds good. Having it available inside the `then`, `else`, and following blocks seems to make sense in the case of `unless`. It might be a little strange but we can try it and see how it goes.

edemaine:
Thinking about this more: Moving the declaration above the block works well for a simple assignment:

```js
unless x := f()
  return
else
  console.log 'good', x
process(x)
↓↓↓
const x = f()
if (!x) {
  return
} else {
  console.log('good', x)
}
process(x)
```

But it's trickier for pattern matching assignment:

```js
unless [x, {type}] := f()
  return
else
  console.log 'good', x, type
process x, type
↓↓↓
const ref = f()
const [x, {type}] = ref
if (!(Array.isArray(ref) && ref.length === 2 && typeof ref[1] === "object" && ref[1] != null && "type" in ref[1])) {
  return
} else {
  console.log('good', x, type)
}
process(x, type)
```

We can't do the destructuring until after the `if` check. Then it's difficult to it (especially with `const`) for both `else` and after the `else`. Even worse, it doesn't really make to do it after the `else`, because in the `then` situation (unless the `then` clause has a guaranteed exit) there's no meaning for `x` and `type`. I see now that #1215 mentions this about Swift `guard`s:

> 1. If the condition creates any bindings (such as `guard x? := getX() else { STATEMENTS }`), the bindings are only available outside the else block.
> 2. `STATEMENTS` must be an "exit" of some sort, in order for 1 to be possible.

This is a little more conditional than I had in mind. Here's one possible behavior:

* Declaration in `unless` goes to `else` block if there is one
* If there's no `else` block *and* the `then` block has an exit according to `isExit`, then the declaration goes after the `unless` block (as currently implemented).
* Otherwise, the declaration doesn't go anywhere.
* One weird case if you have an `else` block *and* the `then` block has an exit. In this case, you didn't need an `else` block at all. In this case, perhaps the `else` block just becomes a braced block (in case it has other declarations) without a leading `else`, and we can put the declaration before it.

I think this makes some sense: the declaration goes in the place(s) where it makes sense. But wouldn't hurt to get more input on this before I go ahead and implement it.

edemaine:
I implemented the above plan, and fixed the missing negation of pattern matching mentioned in https://github.com/DanielXMoore/Civet/issues/756#issuecomment-2097279047
This should be ready to go!

STRd6 (approved):
Thanks! This is a great extension to the usefulness of declarations inside `unless`. I really like how you handled the cases involving `else` blocks and maintaining the declaration in the proceeding code if there is a guaranteed exit.

PR #1229, opened 2024-05-06
try else blocks

### LINKED ISSUE #1134: try...else
Python allows an `else` block for `try`, which is convenient; see [one recent example](https://github.com/DanielXMoore/Civet/blob/8dd331d0cd4347c405d71893aefaf5e10590ffa2/source/cli.civet#L330-L335). I propose:

```js
try
  ...
catch e
  ...
else
  ...
↓↓↓
let caught: true | undefined
try {
  ...
} catch (e) {
  ...
  caught = true
}
if (!caught) {
  ...
}
```

Special case with missing `catch` block:

```js
try
  ...
else
  ...
↓↓↓
let caught: true | undefined
try {
  ...
} catch (e) {
  caught = true
}
if (!caught) {
  ...
}
```

edemaine:
Fixes #1134

### DIFF
--- a/test/try.civet
+++ b/test/try.civet
@@ -238,3 +238,84 @@ describe "try", ->
       ---
       let ref;try { ref = await foo() } catch(e) {ref = void 0;};thing = ref
     """
+
+  describe "try else", ->
+    testCase """
+      just else
+      ---
+      try
+        foo()
+      else
+        console.log 'no exception'
+      ---
+      let ok = true;try {
+        foo()
+      } catch(e) {ok = false} finally {if(caught) {
+        console.log('no exception')
+      }}
+    """
+
+    testCase """
+      catch and else
+      ---
+      try
+        foo()
+      catch e
+        console.log 'exception', e
+      else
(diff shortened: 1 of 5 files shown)


test/try.civet
```diff
@@ -238,3 +238,84 @@ describe "try", ->
...
+      just else
+      ---
+      try
+        foo()
+      else
+        console.log 'no exception'
+      ---
+      let ok = true;try {
+        foo()
+      } catch(e) {ok = false} finally {if(caught) {
```
STRd6, line 253: Looks like this should be `ok` instead of `caught`.

STRd6 (approved):

edemaine:
I forgot to mention: the fun innovation here relative to #1134 is the idea of putting the `else` block inside the `finally` block, which keeps everything "contained" within the `try` block. Also, we don't need to double-wrap when we have both `else` and `finally` blocks. We still achieve the goal of the `else` block not wrapped in `try..catch`.

PR #1232, opened 2024-05-07
Fix duplicate property merging for props not actually bound

### LINKED ISSUE #1230: Matching nested property with same name generates invalid JS
```civet
switch m
  { foo: { foo: ^bar } }
    ;
```
```js
if(typeof m === 'object' && m != null && 'foo' in m && typeof m.foo === 'object' && m.foo != null && 'foo' in m.foo && m.foo.foo === bar) {const { foo: foo1: { foo: foo2 } } = m;const foo = [foo2, foo1];
    ;}
```

edemaine:
Fixes #1230

I think these are all the cases where things aren't actually bound, but let me know if you think of something else. We currently don't seem to have a computation anywhere of the names that are bound to.

### DIFF
--- a/source/parser/pattern-matching.civet
+++ b/source/parser/pattern-matching.civet
@@ -424,6 +424,11 @@ function aggregateDuplicateBindings(bindings) {
   for p of props
     { name, value } := p
 
+    // We do not bind a property that gets matched against
+    // an array or object pattern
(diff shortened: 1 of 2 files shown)


STRd6 (approved):
Looks good!

PR #1233, opened 2024-05-08
Ampersand fixes

### LINKED ISSUE #1231: Ampersand bugs
## if/then

```js
e.children = e.children.map
  if & is e.block then block else &
// ↓↓↓
e.children = e.children.map(($) =>
  $ === e.block ? block : ($1) => $1
);
```

I expected this to be a single function, but each `&` gets treated as its own function wrapper. By contrast, `?:` works fine:

```js
e.children = e.children.map(($) =>
  $ === e.block ? block : $
);
```

## Assignment

The [docs](https://civet.dev/reference#single-argument-function-shorthand) say that we don't lift beyond an assignment, but I find this not to be the case:

```js
f = & + 1
// ↓↓↓
($) => (f = $ + 1);
```

edemaine:
Fixes #1231 

* `&` with binary operators in assignments (e.g. `x = &+1`) previously didn't stop at the assignment
* `&` in if/else previously stopped because the then and else clauses are wrapped in blocks; now we continue up if the clause is a one-liner
* `&` in ternary: `cond ? & : 0` previously treated `: 0` as typing. Now we require no space before `: Type` to disambiguate from ternary
* `.: x` was previously treated as `[.]: x`. While this is cool, it prevented typing of `.` placeholders. I've gone ahead and made `.` consistent with `&`: it can now be typed via `: Type`.

I'm a little worried about if/else. I'm sure how to feel about this example:

```js
=>
  if cond
    &+1
  else
    &+2
// ↓↓↓ (currently)
() => {
  if (cond) {
    return $ => $ + 1;
  } else {
    return $1 => $1 + 2;
  }
};
// OR (this PR)
() => {
  return $ => { if (cond) {
    return $+1
  }
  else {
    return $+2
  }}
}
```

If you can think of a better rule for if/else, please let me know. (We could require there to be no indentation, i.e. `bare`, but that feels a bit limiting; see https://github.com/DanielXMoore/Civet/issues/1231#issuecomment-2098783437 )

### DIFF
--- a/source/parser.hera
+++ b/source/parser.hera
@@ -902,7 +902,7 @@ Placeholder
     }
 
 AmpersandTypeSuffix
-  &( _? QuestionMark? _? Colon ) TypeSuffix -> $2
+  &( QuestionMark? Colon ) TypeSuffix -> $2
 
 # https://262.ecma-international.org/#prod-ClassDeclaration
 ClassDeclaration
--- a/source/parser.hera
+++ b/source/parser.hera
@@ -878,14 +878,15 @@ ParenthesizedExpression
 
 Placeholder
   # Partial function application: f(., x) -> $ => f($, x)
-  Dot:dot !/(?:\p{ID_Continue}|[\u200C\u200D$.])/ ->
+  Dot:dot !/(?:\p{ID_Continue}|[\u200C\u200D$.])/ PlaceholderTypeSuffix?:typeSuffix ->
     return {
       type: "Placeholder",
       subtype: ".",
+      typeSuffix,
       children: [ dot ],
     }
   # Ruby/Crystal style block shorthand: &+1 -> $ => $+1
-  Ampersand:amp ![&=] AmpersandTypeSuffix?:typeSuffix ->
+  Ampersand:amp ![&=] PlaceholderTypeSuffix?:typeSuffix ->
     return {
       type: "Placeholder",
       subtype: "&",
@@ -901,7 +902,7 @@ Placeholder
       children: [ { token: "&" } ],
     }
 
-AmpersandTypeSuffix
+PlaceholderTypeSuffix
   &( QuestionMark? Colon ) TypeSuffix -> $2
 
 # https://262.ecma-international.org/#prod-ClassDeclaration
@@ -3269,15 +3270,6 @@ ComputedPropertyName
       type: "ComputedPropertyName",
       expression,
       children: [ $1, expression, $4 ],
-      implicit: true,
-    }
-  # `.: value` is shorthand for `[.]: value`
-  # (while `&: T` is a type annotation)
-  &"." InsertOpenBracket:open Placeholder:expression InsertCloseBracket:close ->
-    return {
-      type: "ComputedPropertyName",
-      expression,
-      children: [ open, expression, close ],
       implicit: true,
     }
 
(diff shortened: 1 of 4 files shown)


STRd6:
I think most of this is pretty good but I'm not sure that `&` should lift above statements even when inside the blocks of an `if/else`.

When the `if/else` block is inside an expression like the example from #1231:

```
e.children = e.children.map
  if & is e.block then block else &
```

Then it may make sense to me to lift them but still seems to be a bit ambiguous.

One example illustrating some of the strangeness:

```
if f &
  &
```

The `&` in the block would be lifted above the `f(&)`

edemaine:
The Civet source code already uses this helpful pattern a bunch (from some of my PRs within the last couple of days):

https://github.com/DanielXMoore/Civet/blob/7d3554cc5b27f65e72bafffd5b31ddd7f7becdc7/source/parser/declaration.civet#L363

Do you think that `&` should be lifted out of `?:`, but not out of `if/then/else`? This feels asymmetric, which is why I originally added lifting out of `if/then/else`. But `if/then/else` is way more powerful than `?:` (in particular, multiple indented lines isn't really what we imagined lifting `&` over), so I'm not sure. Some options for restrictions:

1. Lift out of `then`/`else` clause that's one expression (what's currently implemented by this PR)
2. Lift out of `then`/`else` clause that's a bare one-liner (`if x then y else z` without newlines or indentation)
3. Lift out of `then`/`else` clause when the `if` is expressionized (I believe it would have a `StatementExpression` parent at this stage, so we could detect it). This would include the example with `map`, but I believe it would not include the example in my first message (that's an implicit return, not expressionization).
4. Never lift out of `then`/else` clause (current behavior). In this case, we should consider not lifting out of `?:` either...

We can combine these. E.g., 1+2+3 would be the most restrictive, other than the full restriction of 4.

STRd6:
I think lifting out of ternary `? : ` is fine because it is an expression. Even lifting out of the `if &` condition makes sense. But lifting out of a full statement within a containing block seems like too much.

Lifting out of `if/else` when it is in a `StatementExpression` might be ok but I think for now I'd lean towards Option 4. Keeping ternary lifting seems fine because it is an expression and not a statement.

Also interested in exploring further if we can find a consistent rule but for now "never lifts above a statement" seems relatively clear to understand.

edemaine:
Agreed.  However, your point about `if &` makes me think of another option:

5. Lift out of `then`/`else` clause only if the `if` condition also has a lifted `&`.

What do you think?

I think this can be implemented because we process `&` left to right, so we can check whether this `if` has already been "activated" by a lifted condition.

edemaine:
Yet another idea: if we wanted to enable something like `& + if cond then &/2 else &/3`, we could generalize to:

6. Lift out of `then`/`else` clause if this would end up lifting to an ancestor to which another `&` has already been lifted.

The idea for both Options 5 and 6 is, if you have other `&`s already lifting to an upper level, then it's less surprising for other `&`s to refer to the same `&`.

edemaine:
I went ahead and implemented the more conservative Option 5. (Not sure why there's an unnecessary IIFE wrapper there though...) Also happy to revert both `if`/`else` changes if you'd rather play it safe.

STRd6 (approved):
This seems good for now but we may still tweak the `if` blocks later. I guess we can see how they feel and refine further.

PR #1256, opened 2024-05-19
Upgrade Playground to modern Prettier

edemaine:
I did this originally to fix some of the Prettier bugs we've been seeing. It didn't end up helping, but will be useful in the future to upgrade the Playground worker to the new Prettier 3 async interface.

### DIFF
--- a/package.json
+++ b/package.json
@@ -76,7 +76,7 @@
   "devDependencies": {
     "@danielx/civet": "0.7.4",
     "@danielx/hera": "^0.8.13",
-    "@prettier/sync": "^0.3.0",
+    "@prettier/sync": "^0.5.2",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -76,15 +76,15 @@
   "devDependencies": {
     "@danielx/civet": "0.7.4",
     "@danielx/hera": "^0.8.13",
-    "@prettier/sync": "^0.3.0",
```
STRd6, line 79: Do we still need prettier/sync?

edemaine: Yes, we still use it when doing Markdown processing of `<Playground>` at build time, because that interface needs to be synchronous.

STRd6 (approved):

PR #1262, opened 2024-05-21
Error nodes, ParseErrors get source-mapped location

### LINKED ISSUE #1251: Add source location to error nodes
Primarily for ease of development and to make errors more obvious with the LSP, so that errors for things like `for each ... in` point to the correct line.

edemaine:
Fixes #1251 (modulo TODOs below)

The basic idea is, during `generate`, to translate an `"Error"` node into a `ParseError` object with the current source-mapped location. To get accurate column numbers, I had to tweak the source mapping code, so this requires careful checking that I didn't mess anything up.

Now `options.errors` becomes a list of `ParseError` objects instead of `"Error"` nodes.  So far I've just consumed these in `main.civet` when building `ParseErrors`, which now has the array of `ParseError` objects, and gives line/column feedback in the CLI or `testCase`.

TODO (possibly in future PRs):
* [ ] Playground line/column highlighting should use these line/column numbers
* [ ] VSCode plugin should use these line/column numbers

Relies on https://github.com/DanielXMoore/Hera/pull/18 so tests won't work without `npm link` until that gets released.

Other changes:

* Migrate from `source/util.civet` to Hera's `source/sourcemap.civet`. This was left as a TODO.
* `SourceMap` now directly exported
* `ParseError` now exported
* `isCompileError` checks for `ParseError` and `ParseErrors`


STRd6 (approved):
Thanks, this is a great cleanup of a many sourcemapping TODOs! 

Maybe eventually Hera could import the sourcemap class from Civet so it's no longer duplicated.

edemaine:
Or a separate `@danielx/sourcemap` package? But yeah, Hera already depends on Civet, so makes sense to just use in that way.

PR #1263, opened 2024-05-23
Fix complex negative property access

edemaine:
## Input:
```js
lastChild := node.children.-1
```
## Output without this PR:
```js
let ref;const lastChild = (ref = node.children[ref.length-1])
```
## Output with this PR:
```js
let ref;const lastChild = (ref = node.children)[ref.length-1]
```

## Context

Civet doesn't work if built with current Civet because this line is compiling incorrectly:

https://github.com/DanielXMoore/Civet/blob/41ba011b4a506af1c0a08b22a838a46e9f637fd2/source/parser/binding.civet#L140

Hopefully this will allow us to upgrade Civet and fix the CLI according to #1261.

### DIFF
--- a/source/parser/lib.civet
+++ b/source/parser/lib.civet
@@ -1099,8 +1099,8 @@ function processStatementExpressions(statements: StatementTuple[]): void
 
 function processNegativeIndexAccess(statements: StatementTuple[]): void
   gatherRecursiveAll(statements, (n) => n.type is "NegativeIndex")
-    .forEach (exp) =>
-      { children } := exp.parent
+    .forEach (exp): void =>
+      { children } := exp.parent!
 
       let start = 0
       start++ while start < children# and isWhitespaceOrEmpty children[start]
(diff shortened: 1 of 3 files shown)


edemaine:
P.S. I wonder about adding another test mode that runs `yarn test` but where it uses the current code to build itself (after building itself with the current release). Essentially `yarn build` + `yarn test` using the current build instead of NPM. I wouldn't run this all the time, probably, but it would be especially good for CI testing to avoid surprises like this in the future. Civet is of course the largest example of Civet code, so it's a good real-world test.

Alternatively or in addition, an interesting idea could be to run `yarn build` with the current Civet dep, and then run `yarn build` again with the resulting build, and output a diff of the output (and ideally show that in CI, maybe as an artifact). Sometimes diffs are fine/good, but it might be good to be aware of them and make it part of code review?

STRd6 (approved):

PR #1265, opened 2024-05-27
Fix source-map-support import with new NodeJS register

edemaine:
Our rewrite in ESM support to NodeJS's new `module` `register` function (#1128) messed up `@cspotcode/source-map-support`. It seems that the `register`ed module (`esm.civet`) runs in a separate Worker or other context, so it doesn't enable sourcemap support in the main thread. We had a second call to source-map-support for CJS, but it was broken because it didn't specify `retrieveFile`.

A result of this is we can't (easily) use a cache to share the compiled source code between the build for loading and the build for sourcemapping. I think this is fine, as we only need to recompile when dealing with an exception, which is presumably exceptional. But I kept a cache in the sourcemapping side, in case the user uses exceptions for a lot of logic or something.

Also fixed a bug in CJS `require.extensions` mode (which still seems important for `node -r @danielx/civet/register`) that didn't use `sync` mode.

Also removed some claims that this ESM/CJS functionality needs `ts-node`. This hasn't been true for a while: we compile in `js: true` mode.

### DIFF
--- a/README.md
+++ b/README.md
@@ -240,8 +240,6 @@ to set up your environment to get productive right away and have a Good Time℠.
 Code coverage with [c8](https://github.com/bcoe/c8) "just works" thanks to their source map
 integration and Civet's source maps.
 
-Currently Civet's ESM loader depends on [ts-node](https://www.npmjs.com/package/ts-node)
-
 #### c8 + Mocha
 
 `package.json`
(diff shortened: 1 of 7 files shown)


STRd6 (approved):
Thanks! Node ESM updates are a civet's natural enemy. 😿

PR #1268, opened 2024-05-27
Fix empty block in pattern matching with implicit return

### LINKED ISSUE #1264: Empty block following pattern in switch expression generates invalid JS
Input:
```civet
f := (x) => switch x
  [_]
    // nothing
  else
    foo()
```
Output (with `--js` to make the `len` function shorter):
```js
function len(arr, length) { return arr.length === length }
const f = (x) => { if(Array.isArray(x) && len(x, 1)) {return const [_] = x;}
else  {
    return foo()
  } }
```
`return const` isn't a valid type of statement.

edemaine:
Fixes #1264

Also fix function declaration at end of loop

### DIFF
--- a/source/parser/function.civet
+++ b/source/parser/function.civet
@@ -259,29 +259,35 @@ function assignResults(node: StatementTuple[] | ASTNode, collect: (node: ASTNode
     {type} = exp
 
   switch exp.type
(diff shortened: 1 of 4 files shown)


STRd6 (approved):
Great fix, thanks!

PR #1271, opened 2024-05-27
Revamp automatic async and generator, operator support

### LINKED ISSUE #1248: Generator operators
```civet
operator foo (a)
  yield a
```

Actual output:
```js
function foo(a) {
  return yield a;
}
```

Expected output:
```js
function* foo(a) {
  return yield a;
}
```

`operator*` also doesn't work

Workaround:
```civet
operator foo (a)
  (->
    yield a
  )()
```

edemaine:
* `async`/`*` now added in a postprocessing stage
* More code sharing
* Errors in `=>` now appear at `await` statements
* Fix support in `operator` (fixes #1248)
* Allow explicit `async` and `*` in `operator` definition (also #1248)


STRd6 (approved):
Thanks!

PR #1276, opened 2024-06-03
Added a cache based on mtime for watch/serve mode in esbuild

STRd6:
This speeds up esbuild serve quite a bit on larger projects.

### DIFF
--- a/integration/unplugin/src/index.ts
+++ b/integration/unplugin/src/index.ts
@@ -367,13 +367,17 @@ export const rawPlugin: Parameters<typeof createUnplugin<PluginOptions>>[0] =
       const basename = id.slice(0, match.index + match[1].length);
 
       const filename = path.resolve(rootDir, basename);
-      const mtime = fs.statSync(filename).mtimeMs;
-      const cached = cache?.get(filename);
-      if (cached && cached.mtime === mtime) {
-        return cached.result;
+
+      let mtime;
+      if (cache) {
+        mtime = (await fs.promises.stat(filename)).mtimeMs;
+        const cached = cache?.get(filename);
+        if (cached && cached.mtime === mtime) {
+          return cached.result;
+        }
       }
 
-      const rawCivetSource = fs.readFileSync(filename, 'utf-8');
+      const rawCivetSource = await fs.promises.readFile(filename, 'utf-8');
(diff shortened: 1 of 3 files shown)


integration/unplugin/src/index.ts
```diff
@@ -363,7 +367,13 @@ export const rawPlugin: Parameters<typeof createUnplugin<PluginOptions>>[0] =
       const basename = id.slice(0, match.index + match[1].length);
 
       const filename = path.resolve(rootDir, basename);
-      const rawCivetSource = await fs.promises.readFile(filename, 'utf-8');
+      const mtime = fs.statSync(filename).mtimeMs;
```
edemaine, line 370: Perhaps this should be a avoided if `cache` is null?

integration/unplugin/src/index.ts
```diff
@@ -363,7 +367,13 @@ export const rawPlugin: Parameters<typeof createUnplugin<PluginOptions>>[0] =
...
+      const cached = cache?.get(filename);
+      if (cached && cached.mtime === mtime) {
+        return cached.result;
+      }
+
+      const rawCivetSource = fs.readFileSync(filename, 'utf-8');
```
edemaine, line 376: Is there a reason to switch to sync interface? I wonder if promises interface (also stat) might support multitasking slightly better.

edemaine (approved):
Nice idea! Can you also document this? I think in a couple of READMEs.

integration/unplugin/src/index.ts
STRd6: My intuition is that opening a lot of files and reading a little bit of each is slower than opening one and reading all of it but I haven't done a full benchmark yet.

edemaine: Ah, fair point - I'm not sure what's best. Probably fine either way.

edemaine (approved):
Looks good! I think `integrations/unplugin/README.md` also has a section.

integration/unplugin/src/index.ts
STRd6: I did a brief unscientific test and the promises version seemed to finish faster. The promises version probably has better throughput where node/the os can buffer multiple files in parallel.

Issue #918, opened 2024-01-11
Maximum call stack size exceeded


mixty1:
[...]

PR #1279, opened 2024-06-09
Replacing some instances of __ with stricter whitespace checking

STRd6:
I was experimenting with #918 by trying to reduce recursion opportunities by avoiding rules that consume zero tokens. It didn't solve the issue but I think it's a good direction.

### DIFF
--- a/source/parser.hera
+++ b/source/parser.hera
@@ -213,7 +213,7 @@ NonPipelineExtendedExpression
 NonAssignmentExtendedExpression
   # Check for nested expressionized statements first
   NestedNonAssignmentExtendedExpression
-  __ ExpressionizedStatementWithTrailingCallExpressions ->
+  _? ExpressionizedStatementWithTrailingCallExpressions ->
     return prepend($1, $2)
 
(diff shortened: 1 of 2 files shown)


source/parser.hera
```diff
@@ -621,13 +621,13 @@ AssignmentExpression
...
   # TODO: Ideally this wouldn't be needed.
-  __ AssignmentExpressionTail
+  ( EOS _? ) AssignmentExpressionTail
   # NonPipelineAssignmentExpression
 
 NonPipelineAssignmentExpression
   # NOTE: Try to match a single line assignment expression before matching newline then assignment expression
   NonPipelineSingleLineAssignmentExpression
-  __ NonPipelineAssignmentExpressionTail
+  ( EOS _? ) NonPipelineAssignmentExpressionTail
```
edemaine, line 630: I wonder if `NotDedented` would make sense here? I think maybe not, the idea is to break indentation here. But thought I'd mention the possibility.

Overall, looks like a good approach!

PR #1282, opened 2024-06-13
Binary op fixes

### LINKED ISSUE #1281: Trailing `or` on newline after callback function in implicit application parses as nested implicit function call instead of method body
```civet
x.filter ([a, b]) ->
  a or 
  b
```

```javascript
x.filter(function([a, b]) {})(
  a || 
  b)
```

edemaine:
A few binary op fixes:
* Handle trailing spaces before newlines (fixes #1281)
* `is !like` to match other `!` shorthand for `not`
* Fix support for `⩵` and `⩶` (Unicode double and triple equals). These were supported to work, but weren't in the lookahead.

### DIFF
--- a/source/parser.hera
+++ b/source/parser.hera
@@ -469,7 +469,7 @@ BinaryOpExpression
     return processBinaryOpExpression($0)
 
 BinaryOpRHS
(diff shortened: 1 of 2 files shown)


STRd6 (approved):
Thanks!

PR #1283, opened 2024-06-19
Fix some TS errors and increase code coverage

STRd6:
I think we can hit 100% coverage and remove all the type errors if we slowly chip away at it.

### DIFF
--- a/source/parser/util.civet
+++ b/source/parser/util.civet
@@ -123,8 +123,6 @@ function isFunction(node: ASTNode): node is FunctionNode
       type is "FunctionExpression"
       type is "ArrowFunction"
       type is "MethodDefinition"
-      "async" in node and !!node.async
-      // do blocks can be marked async to prevent automatic await
 
   return false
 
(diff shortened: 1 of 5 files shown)


STRd6:
Thanks for the feedback. I’m camping this weekend but I’ll incorporate it when I get back.

source/parser/util.civet
```diff
@@ -123,8 +123,6 @@ function isFunction(node: ASTNode): node is FunctionNode
       type is "FunctionExpression"
       type is "ArrowFunction"
       type is "MethodDefinition"
-      "async" in node and !!node.async
-      // do blocks can be marked async to prevent automatic await
```
edemaine: Interesting - I guess we don't need to mark `async do` blocks as functions after all? Presumably they get wrapped in IIFE by this point.

STRd6: Yeah, I think we're creating proper function nodes everywhere now.

edemaine (approved):

PR #1292, opened 2024-06-22
Self-test

edemaine:
Implementing https://github.com/DanielXMoore/Civet/pull/1263#issuecomment-2127573885 :

> I wonder about adding another test mode that runs yarn test but where it uses the current code to build itself (after building itself with the current release). Essentially yarn build + yarn test using the current build instead of NPM. I wouldn't run this all the time, probably, but it would be especially good for CI testing to avoid surprises like this in the future. Civet is of course the largest example of Civet code, so it's a good real-world test.

I came across this again in https://effectivetypescript.com/2024/04/16/inferring-a-type-predicate/ (text and video), which calls this TypeScript's "self-check", so I've adopted the "self-test" terminology here.
* `yarn test:self` is a variation on `yarn test` that runs Mocha with the current built version, after running `yarn build`. (Open to better naming convention. Building first seems important here.)
* CI includes a parallel job to run this self-test. So wait time should be the same, just using more minutes.

While implementing this, I found a bug where a line of `source/parser/util.civet` wouldn't parse. So even more evidence that this is an important test to have! I fixed that bug too.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -39,7 +39,7 @@ jobs:
       with:
         file: coverage/lcov.info
 
-  build:
+  self-test:
     name: Self-Test
     runs-on: ubuntu-latest
     steps:
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -55,7 +55,6 @@ jobs:
     - name: Install and Test
       run: |
         yarn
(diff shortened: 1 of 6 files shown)


STRd6 (approved):
Thanks! It's good to automate this so we can catch errors earlier.

PR #1293, opened 2024-06-25
Fix argument parsing

edemaine:
Fixes #918 by preventing accidental double matching of different cases of arguments.

This does not fix one missing `PushIndent` mentioned in https://github.com/DanielXMoore/Civet/issues/918#issuecomment-2189107922, which I'll probably leave for another PR: We don't consider indented expressions that are normal expressions, only those that are expressionized statements. So `x =\n  if ...` is considered an indented expression, but `x =\n  y` isn't. I think this is broken. For example, `a =\n  x\n  y` probably shouldn't [parse as a function call](https://civet.dev/playground?code=YSA9CiAgeAogIHk%3D).

PR #1295, opened 2024-06-25
MaybeNested expressions

edemaine:
Follow up to https://github.com/DanielXMoore/Civet/pull/1293#issue-2372895605

> We don't consider indented expressions that are normal expressions, only those that are expressionized statements. So `x =\n  if ...` is considered an indented expression, but `x =\n  y` isn't. I think this is broken.

Basically, I added a few `MaybeNested...Expression`s that check for a properly indented expression. And I reviewed all uses of `ExtendedExpression` and a few others and decided whether to add this prefix. Mostly right-hand sides of assignments get this new treatment.

However, because we fallback to not-properly-nested parsing, it's still the case that `a =\n  x\n  y` [parses as a function call](https://civet.dev/playground?code=YSA9CiAgeAogIHk%3D). As a result of this, I can't actually think up a situation where this PR affects compilation. But it makes me feel more confident that `PushIndent` will have the correct indentation in some future situations... but given that it doesn't matter yet, it's also plausible that this part of the PR should wait. What do you think? (Alternatively, we could try to be stricter: when the expression starts `FurtherIndented`, it must be properly nested. I'm not sure whether this would work well in practice though... we're maybe used to having weirdly (non)indented expressions.)

Also added missing feature of `throw`ing a nested expression, using the (existing but renamed) `MaybeParenNestedExtendedExpression`. This parallels existing features for `return` and `yield`.

Also renamed `MaybeIndentedType` to `MaybeNestedType` for consistency.

### DIFF
--- a/source/parser.hera
+++ b/source/parser.hera
@@ -660,7 +660,7 @@ ActualAssignment
   # NOTE: Consolidated assignment ops
   # NOTE: UpdateExpression instead of LeftHandSideExpression to allow
   # e.g. ++x *= 2 which we later convert to ++x, x *= 2
-  ( NotDedented UpdateExpression WAssignmentOp )+ ExtendedExpression ->
+  ( NotDedented UpdateExpression WAssignmentOp )+ MaybeNestedExtendedExpression ->
     $1 = $1.map(x => [x[0], x[1], ...x[2]])
     $0 = [$1, $2]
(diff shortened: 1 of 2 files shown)


STRd6 (approved):
This is worth getting in at least for the `throw` capability.

PR #1299, opened 2024-06-29
Add class with A, B mixin notation

STRd6:
Added the usage half of mixins.

```
      class A with B,C
        x = 3
      ---
      class A extends C(B(Object)) {
        x = 3
      }
```

### DIFF
--- a/source/parser.hera
+++ b/source/parser.hera
@@ -13,6 +13,7 @@ import {
   blockWithPrefix,
   convertNamedImportsToObject,
   convertObjectToJSXAttributes,
+  convertWithClause,
   dedentBlockString,
   dedentBlockSubstitutions,
   deepCopy,
@@ -57,6 +58,7 @@ import {
   reorderBindingRestProperty,
   replaceNodes,
   skipImplicitArguments,
+  trimFirstSpace,
   typeOfJSX,
 } from "./parser/lib.civet"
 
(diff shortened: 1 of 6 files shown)


source/parser.hera
```diff
@@ -6108,6 +6128,10 @@ While
   "while" NonIdContinue ->
     return { $loc, token: $1 }
 
+With
+  "with" NonIdContinue ->
+    return { $loc, token: $1 }
```
edemaine, line 6133: This might be usable in `import...with` syntax too.

source/parser/types.civet
```diff
@@ -675,3 +675,8 @@ export type VoidType = ASTLeaf & type: "VoidType"
 export type TypeLiteralNode = ASTLeaf | VoidType
 
 export type ThisAssignments = [string, ASTRef][]
+
+export type WithClause =
+  type: "WithClause"
+  children: Children
+  targets: [WSNode, ExpressionNode][]
```
edemaine, line 682: I think this needs to be added to something... `OtherNode`?

source/parser/util.civet
```diff
@@ -229,9 +229,18 @@ function insertTrimmingSpace(target: ASTNode, c: string): ASTNode
...
+    target.replace(/^ ?/, c)
   else
     target
 
+/**
+ * Trims the first single space from the spacing array or node's children if present
+ * maintains $loc for source maps
+ */
+function trimFirstSpace(target: ASTNode): ASTNode
+  insertTrimmingSpace target, ""
```
edemaine, line 242: We use this a zillion times. I wonder if just giving `insertTrimmingSpace` an optional second argument (defaulting to empty string) would be better? Possibly also renaming it to `trimFirstSpace`...

edemaine (approved):
Off to a good start!

source/parser/types.civet
STRd6: Currently it's consumed in the parse phase so doesn't end up in the final tree.

source/parser/util.civet
STRd6: Maybe `trimFirstSpaceAndInsert` and keep the `trimFirstSpace` alias.

edemaine: `replaceFirstSpace`?

Issue #1239, opened 2024-05-09
Finishing civetconfig 


edemaine:
* [x] VSCode extension uses default civetconfig
* [x] CLI uses default civetconfig
* [ ] unplugin should support default civetconfig, along with `config` override
* [ ] eslint plugin should probably do the same
* [ ] Babel plugin should probably do the same
* [ ] Gulp plugin should probably do the same
* [ ] What about ESM and Bun plugins?
* [ ] Document in https://civet.dev/config

danielbayley:
There are also the proposed potential config files, as @STRd6 [mentioned here](https://github.com/DanielXMoore/Civet/issues/647#issuecomment-1693542267).

> Allow for parsing different config file extensions based on what modules exist in the local user project

Plus I threw together a quick PR to allow `civetConfig` in `package.json` also in #1300.

Issue #647, opened 2023-08-23
JSONC/JSON5, yaml, toml ...


zolomatok:
... you know where this is going 😄

Do you see a future where Civet might support the type-checked imports of "JSON with Comments" files, or maybe even YAML or TOML? Like how TS already supports direct, type-checked JSON imports.

I'm one of those peeps that are heartbroken, _heartbroken_ I tell ya, by the fact that JSON can't have comments.
I'd use other data formats, but those can't be imported nicely.

... pwetty pwease?

STRd6:
Are there any links to issues in the TypeScript repo about this? Before implementing such a feature it is important for us to know where TS stands so we can ensure long term compatibility.

zolomatok:
Good thinking, but I haven't found any mention of this in the roadmap.

There are two relevant entries in the Issues, [this one](https://github.com/microsoft/TypeScript/issues/12242), which requests support for JSON5 and [this one](https://github.com/microsoft/TypeScript/issues/40027), but that concerns itself with the format of _specifically_ `tsconfig.json` and not anything else.

Neither have any satisfying resolution. The former is closed with a comment that I think is missing the point, the latter hasn't seem to have received any official response.

**Bun** purportedly have support for [toml imports](https://bun.sh/guides/runtime/import-toml), but it doesn't really seem to work as stated in the docs. I played around with it and most I could do is import the toml file, but it's `unknown`. That does work and if you print it, you do see the parsed object tree. But since it's an `unknown`, you don't have autocomplete and type information, you don't have access to the properties like shown in the docs. At least not a way VSCode won't complain about.

For that to work, you need to create a `.d.ts` file, according to [this](https://discord.com/channels/876711213126520882/1113107656882782309) discord thread.
But I haven't tried that, cause I don't like polluting my project with .d.ts stuph.

STRd6:
Ok, after reading through the TS issues I think this is a good idea. Here's my thoughts:

- Allow for parsing different config file extensions based on what modules exist in the local user project
  - If json5 is installed treat `.json` as JSON5
  - If yaml is installed also look for `.yml` and `.yaml` configs
  - If toml is installed look for `.toml` configs
  - Easy to extend for additional extensions
  - Doesn't bloat the base civet install because these are all optional
  - Update the docs describing that if you want `json5` configs all you need to do is `npm install --save-dev json5` or `yarn add --dev json5`, etc.
    - Strongly recommend setting an explicit config path for speed
[...]

PR #1300, opened 2024-06-30
Allow `civetConfig` in `package.`[`json`|`yaml`]

danielbayley:
Allow reading config from `civetConfig` property in `package.json`.

Includes support for `pnpm` [`package.yaml`](https://github.com/pnpm/pnpm/issues/1100), and maybe a starting point for the implementation of support for different config file formats, as @STRd6 mentioned in https://github.com/DanielXMoore/Civet/issues/647#issuecomment-1693542267.

> Allow for parsing different config file extensions based on what modules exist in the local user project

Related to #1239.


STRd6:
This looks helpful, thanks!

PR #1302, opened 2024-07-01
Typecheck fixes

edemaine:
3 fixes for `civet --typecheck`, and `typecheck: true` in the unplugin:

1. Fix default JSX setting for TypeString engine. Not sure if this was always broken, or the lines got re-ordered, but we need to give a default value via an enum instead of a string here.
2. Include errors/warnings from `tsconfig.json` parsing. I noticed that errors such as `No inputs were found in config file 'tsconfig.json'.` were lying around in the parsed config output when debugging. Now we emit them along with (before) all the other errors, to better match `tsc` behavior.
3. Mangle [`imports` field](https://nodejs.org/api/packages.html#subpath-imports) in `package.json` for TypeScript. As reported by @danielbayley on Discord, it's reasonable to expect something like `"#types": "./src/types.civet"` to work, and it does when running code via the Civet CLI for example, but TypeScript needs it to read `"#types": "./src/types.civet.tsx"` so that it understands the extension. Now we modify the `package.json` as TypeScript reads it. (There might be other fields that need to similarly mogrified, such as `exports`, but I'm not certain whether it's necessary...)

### DIFF
--- a/integration/unplugin/src/index.ts
+++ b/integration/unplugin/src/index.ts
@@ -202,7 +202,34 @@ export const rawPlugin: Parameters<typeof createUnplugin<PluginOptions>>[0] =
           if (fsMap.has(filename)) return true
           return systemFileExists(filename.slice(0, -4))
         };
+
         system.readFile = (filename: string, encoding: string = 'utf-8'): string | undefined => {
+          // Mogrify package.json imports field to use .civet.tsx
+          if (path.basename(filename) === 'package.json') {
+            const json = systemReadFile(filename)
+            if (!json) return json
+            const parsed: Record<string, unknown> = JSON.parse(json)
+            let modified = false
+            function recurse(node: unknown) {
+              if (node && typeof node === 'object') {
+                for (const key in node) {
+                  const value = (node as Record<string, unknown>)[key]
+                  if (typeof value === 'string') {
+                    if (value.endsWith('.civet')) {
+                      (node as Record<string, unknown>)[key] = value + '.tsx'
+                      modified = true
+                    }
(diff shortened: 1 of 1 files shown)


STRd6 (approved):
Thanks!

danielbayley:
Nice work, thanks again @edemaine!

> TypeScript needs it to read `"#types": "./src/types.civet.tsx"`

Just:
~~~ json
"#types": "./src/types.civet"
~~~
or even:
~~~ json
"#*": "./src/*.civet"
~~~
also seem to work, which is nice 👍🏼

> Include errors/warnings from `tsconfig.json` parsing. I noticed that errors such as `No inputs were found in config file 'tsconfig.json'.` were lying around in the parsed config output when debugging. Now we emit them along with (before) all the other errors, to better match `tsc` behavior.

@edemaine Is this desirable though, in the case of `civet --typecheck`ing? There may well be no existing `.ts*` files before either `--compile` or `--emit-declaration` steps—only `*.civet` files at the point `--typecheck`ing is run as the first step, right?

So this will always produce the `TS18003: No inputs were found in config file 'tsconfig.json'.` error, meaning even if all the types are sound, this will still fail in a `package.json` script for example…

edemaine:
Oh, good point. Perhaps we need to mogrify `tsconfig.json` so that `includes: '**/*.civet'` works too. (I take it doesn't currently?)

danielbayley:
> so that `include: '**/*.civet'` works too. (I take it doesn't currently?) 

Nope, TypeScript is extremely xenophobic with file extensions!

PR #1304, opened 2024-07-03
More CLI/typechecking fixes

### LINKED ISSUE #1303: `civet *.civet` file execution bug when reading config from any file
There seems to be a bug where running a `civet *.civet` file directly does not execute anything if reading config from any file.

Meanwhile, `node --import @danielx/civet/register *.civet` does work on the same file (relinquishing the ability to read config from a file), if that helps narrow it down?

I initially thought it might have something to do with my changes in #1300, but rolled back several versions, and ran into the same behaviour…

edemaine:
CLI config files:
* Avoid using `package.json` as config file when it doesn't have `civetConfig` key.
* Enforce config filename checks to occur in our desired priority order, instead of being filesystem-specific. So e.g. `civetconfig.json` takes priority over `package.json`.
* Allow `civetconfig.js` (and all other config files can use `.js` — even `package.js` though this isn't documented)
* Allow `civetconfig.yaml` (and all other config files can use `.yaml`/`.yml` — previously was just `package.yaml`/`package.yml`)
* Fix `options.run` getting clobbered when any config file got read (which was particularly bad with `package.json`). Fixes #1303

Typechecking:
* Types in `node_modules/@types` are now automatically read, via a workaround until https://github.com/microsoft/TypeScript-Website/pull/3165
* `includes`/`excludes`/`files` now supported, including `.civet` extensions
* If you don't specify `includes`/`excludes`/`files`, TypeScript's default behavior of finding all files in the project directory will now include `.civet` files. No more `TS18003: No inputs were found in config file 'tsconfig.json'.` as reported in https://github.com/DanielXMoore/Civet/pull/1302#issuecomment-2204849270
* `civet --typecheck` and `civet --emitDeclaration` without any filename arguments now typechecks/emits declarations all files specified by `include`/`exclude`/`files`, just like `tsc`

### DIFF
--- a/civet.dev/config.md
+++ b/civet.dev/config.md
@@ -111,17 +111,17 @@ In the root directory of your project, or in a `.config` subdirectory,
 you can add one of the following files:
 
 * `🐈.json`
-* `🐈.civet`
 * `civetconfig.json`
-* `civetconfig.civet`
 * `civet.config.json`
-* `civet.config.civet`
 * `package.json` with a `"civetConfig"` property
-* [`package.yaml`](https://github.com/pnpm/pnpm/issues/1100) with a `"civetConfig"`
-property (and [yaml](https://eemeli.org/yaml) `install`ed as optional
-[`peerDependencies`](https://docs.npmjs.com/cli/v10/configuring-npm/package-json#peerdependencies))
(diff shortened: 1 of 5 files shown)


STRd6 (approved):
Lots of great fixes, thanks!

PR #1305, opened 2024-07-06
Fix pipe mode in CLI

edemaine:
Fix `civet < foo.civet > bar.tsx` accidentally turning on `js: true` and `inlineSourceMaps: true`

Introduced in #1304, as reported by @STRd6 on Discord


STRd6 (approved):
Thanks!

PR #1308, opened 2024-07-08
CLI tests

### LINKED ISSUE #47: Add testing for Civet CLI
There should be some basic tests of the CLI and its options.

edemaine:
Fixes #47

* Get CLI to actually load in ESM mode:
  * Avoid `.ts` extension for Vite constants JavaScript
  * Avoid `require`
  * Move `cli()` call to `civet` binary via build script, instead of `cli.civet` trying to detect whether we're the main module across both ESM and CJS
* Move more CLI code into `parseArgs`
* Lots of tests for `parseArgs` (though I imagine there could be more)

## Outstanding issue

I'm getting a weird "race condition" when testing multiple instances of `--civet` argument parsing, which calls Hera's `parse`. Somehow the subsequent calls continue where the previous ones left off, instead of resetting `pos` to 0. I don't understand how this is happening, and I can't reproduce this in the CLI (`civet --civet tab=4 --civet autoLet` works fine). It doesn't seem to be `async` related (I tried switching everything to sync and got the same error). It seems to be something specific to Mocha or ESM mode...


STRd6:
It could do with how we are passing state from the parser module to main/cli. `Reset` should reset the state but maybe it's not updating the state at the module level.

edemaine:
Your comment made me investigate caching, which turned out to be the issue. The cache from the last Civet compilation test was remaining active, and without resetting, the cache was getting re-used from one test to the other.  I'm not sure `events: {enter: undefined, exit: undefined}` is the nicest way to reset the cache, and I feel like this should have been done at the end of the Civet `compile` step instead of when parsing the `--civet` command-line option, but it works for now.

STRd6 (approved):

PR #1314, opened 2024-07-09
Hoist array binding and rest ref declarations. Fixes #1139. Fixes #1312

### LINKED ISSUE #1139: `[a, ..., b] = c` doesn't declare `ref`
`[a, ..., b] = c` transpiles to `[a, ...ref] = c, [b] = ref.splice(-1)`, with no declaration of `ref`.

The path being followed here:
* `BindingElement`
* `BindingRestElement`
* `EmptyBindingPattern` creates a ref, but no `process` function declares these refs. (It works in the context of pattern matching.)

### LINKED ISSUE #1312: Array destructuring into @vars doesn't declare temporary variables
```civet
[ @x, @y ] = z
{ @x, @y } = z
```

Object destructuring works because we can use `({x: this.x} = y)` but arrays need a temporary variable and it looks like we're failing to hoist the ref declarations.

https://civet.dev/playground?code=WyBAeCwgQHkgXSA9IHoKeyBAeCwgQHkgfSA9IHo%3D

STRd6:


edemaine (approved):
Nice fix!

PR #1323, opened 2024-07-18
support for 'unique symbol' type in TS

STRd6:
Fixes:

```civet
const callbacks: unique symbol = Symbol("eventCallbacks")
```

Where it fails to detect that `unique symbol` is a type.

https://civet.dev/playground?code=Y29uc3QgY2FsbGJhY2tzOiB1bmlxdWUgc3ltYm9sID0gU3ltYm9sKCJldmVudENhbGxiYWNrcyIp

### DIFF
--- a/source/parser.hera
+++ b/source/parser.hera
@@ -7366,8 +7366,8 @@ TypeLiteral
     return $0
   "void" NonIdContinue ->
     return { type: "VoidType", $loc, token: $1 }
-  "unique symbol" NonIdContinue ->
-    return { type: "UniqueSymbolType", $loc, token: $1 }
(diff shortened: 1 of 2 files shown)


source/parser.hera
```diff
@@ -7366,6 +7366,8 @@ TypeLiteral
     return $0
   "void" NonIdContinue ->
     return { type: "VoidType", $loc, token: $1 }
+  "unique symbol" NonIdContinue ->
```
edemaine, line 7369: I think `"unique" _ "symbol"` would be better; it seems TypeScript allows comments and newlines in between the two words.

edemaine (approved):
Nice catch!

PR #1324, opened 2024-07-18
Use register to avoid experimental loaders deprecation warning

STRd6:
Depends up updated Hera: https://github.com/DanielXMoore/Hera/pull/19


edemaine (approved):
I haven't tested this directly myself, but the changes look good and CI is working so I trust this is good to go!

PR #1331, opened 2024-08-03
Add `%/`/`÷` integer division, fix `%%` precedence

edemaine:
Civet has `%%` (nice modulo) from CoffeeScript, but lacked `//` (integer division) from CoffeeScript/Python/etc. because it conflicts with JavaScript comments.

In Discord we came up with `%/` and `÷` for integer division, because:
* `%/` is parallel to `%%` (both are like adding a `%` prefix)
* `%/` doesn't mess up regexes (a previous proposal of `/%` would prevent the regular expression `/% /`)
* `%/` is different from Nim's different operator `/%` (unsigned division).
* `%/` looks somewhat similar to Unicode `÷` (which is how Julia represents integer division, and seems nice).
* Only downside is that `%/` is maybe hard to remember... but often used with `%%` so it will probably look reasonable.

I also added CoffeeScript `//` via `coffeeDiv` compatibility flag. This might break code that uses a blanket `coffeeCompat`, if it also used JS `//` comments. But arguably this was a bug, and now we're more CoffeeScript-compatible. Anyone running into this issue can use `coffeeCompat -coffeeDiv`.

### DIFF
--- a/civet.dev/config.md
+++ b/civet.dev/config.md
@@ -62,6 +62,7 @@ For now, we have the following related options:
 | [`coffeeBooleans`](reference#coffeescript-booleans) | `yes`, `no`, `on`, `off` |
 | [`coffeeClasses`](reference#coffeescript-classes) | CoffeeScript-style `class` methods via `->` functions |
 | [`coffeeComment`](reference#coffeescript-comments) | `# single line comments` |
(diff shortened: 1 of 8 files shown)


STRd6 (approved):
Looks great, thanks!

PR #1353, opened 2024-08-09
changelog generation: fetch PRs mostly in parallel

edemaine:
This is a quick performance boost for changelog generation.

It now takes 1-2 seconds on my machine to generate a changelog. I'd be curious what it is for you (and whether you have a different bottleneck).


STRd6 (approved):
Takes about 5s for me which is reasonable.

Issue #1168, opened 2024-04-16
Trailing pipe or property access should expressionize statement


edemaine:
Consider:

```js
switch @value
  Value.Ace then 'A'
  Value.Jack then 'J'
  Value.Queen then 'Q'
  Value.King then 'K'
  else @value
.toString()
```

Currently [this compiles to](https://civet.dev/playground?code=c3dpdGNoIEB2YWx1ZQogIFZhbHVlLkFjZSB0aGVuICdBJwogIFZhbHVlLkphY2sgdGhlbiAnSicKICBWYWx1ZS5RdWVlbiB0aGVuICdRJwogIFZhbHVlLktpbmcgdGhlbiAnSycKICBlbHNlIEB2YWx1ZQoudG9TdHJpbmcoKQ%3D%3D):

```js
let m;if(m = this.value,m === Value.Ace) { 'A'}
else if(m === Value.Jack) { 'J'}
else if(m === Value.Queen) { 'Q'}
else if(m === Value.King) { 'K'}
else  { this.value };
$ => $.toString()
```

It behaves the same within an arrow function too; the `$ => $.toString()` gets implicitly returned but the `switch` value gets dropped. The issue is that the `switch` is a valid `Statement` and that takes priority over expressions.

I'd propose instead that it compiles to:

```js
(()=>{let m;if(m = this.value,m === Value.Ace) { return 'A'}
else if(m === Value.Jack) { return 'J'}
else if(m === Value.Queen) { return 'Q'}
else if(m === Value.King) { return 'K'}
else  { return this.value }})()
.toString()
```

This is what we get [with an explicit `return` added](https://civet.dev/playground?code=cmV0dXJuIHN3aXRjaCBAdmFsdWUKICBWYWx1ZS5BY2UgdGhlbiAnQScKICBWYWx1ZS5KYWNrIHRoZW4gJ0onCiAgVmFsdWUuUXVlZW4gdGhlbiAnUScKICBWYWx1ZS5LaW5nIHRoZW4gJ0snCiAgZWxzZSBAdmFsdWUKLnRvU3RyaW5nKCk%3D), I believe via `ExpressionizedStatementWithTrailingCallExpressions`. Maybe we just need to check for this before regular `Statement`s, but when there's a positive number of trailing expressions?

Similar to postfix `.foo`, it'd be nice to have postfix pipelines. Original motivating example from Discord:

```js
switch @value
  Value.Ace then 'A'
  Value.Jack then 'J'
[...]

PR #1361, opened 2024-08-13
Bulleted arrays with `.` or `•`

### LINKED ISSUE #803: Bulleted arrays
Inspired by [YAML sequences](https://yaml.org/spec/1.2.2/#821-block-sequences) (which use `-`) and [Markdown lists](https://www.markdownguide.org/basic-syntax/#unordered-lists) (which allow `-`, `+`, or `*`), I propose supporting indented implicit lists with some kind of bullet operator:

```js
x :=
  • name: 'hello'
    value: 'world'
  • name: 'goodbye'
    value: 'world'
  • item
↓↓↓
const x = [
  { name: 'hello',
    value: 'world' },
  { name: 'goodbye',
    value: 'world' },
  item
]
```

For Unicode bullets (`•`), this seems like a no-brainer. The question is what ASCII form(s) it should also have.

* `.` looks a lot like a bullet and isn't currently meaningful when followed by a space. I think this is the current winner.
* `~` looks something like a dash and isn't currently meaningful when followed by a space, though we've occasionally discussed it being used for typing (as it was in my typed CoffeeScript branch).
* `-`, `+`, `*` would be great from a compatibility/familiarity perspective. In particular, I dream of a big YAML subset as being a subset of Civet, in the same way that JSON is a subset of JavaScript. We could also plausibly support them in certain contexts, like right after an assignment as in the example above. But in general we should probably give precedence to these acting like regular binary operators. For example:
  ```js
  test
    - item1
    - item2
  ```
  This should probably remain a subtraction, not an implicit function call `test([item1, item2])`. But I think this is mainly an issue when trying to do indented implicit function calls, which is a bit of an edge case (though admittedly a useful one; implicit objects are used this way all the time). When we're at the "beginning" of an expression, I think we could safely use `-`, `+`, `*`.
* `>` is maybe another related option, inspired by the recent JSX ideas, but it has the same limitations as `-`/`+`/`*`.
* `--` is a reasonable option inspired by en-dash (this is how you type it in LaTeX). `++` doesn't make as much sense, and `**` already means exponentiation (and `>>` already means shift-right). The main issue with doubled operators like this is that they look ugly with 2-spaces indentation:
  ```js
  x :=
    -- name: 'hello'
      value: 'world'
  ```

## Context

Civet (and CoffeeScript) currently lack an implicit array mechanism, unlike objects.
They also have an awkward way to write lists of implicit objects — the dedented comma — which this could be a nicer alternative to.
```coffee
[
  name: 'hello'
  value: 'world'
,
  name: 'goodbye'
  value: 'world'
,
  item
]
```

edemaine:
Fixes #803

### DIFF
--- a/civet.dev/reference.md
+++ b/civet.dev/reference.md
@@ -244,6 +244,8 @@ json's "long property"'s `${movie} name`
 
 ## Arrays
 
+### Bracketed
+
 Commas are optional at the ends of lines.
 
 <Playground>
@@ -268,6 +270,30 @@ people := [
   name: "Bob"
   id: 9
 ]
+</Playground>
+
+### Bulleted
+
+Instead of brackets, array items can be specified via `.` or `•` bullets:
+
+<Playground>
+rotate :=
+  . c, -s
+  . s, c
+</Playground>
+
+<Playground>
+func.apply @,
+  • arg1
+  • arg2
+</Playground>
+
+<Playground>
+people :=
+  . name: "Alice"
+    id: 7
+  . name: "Bob"
+    id: 9
 </Playground>
 
 ### Rest
--- a/source/main.civet
+++ b/source/main.civet
@@ -35,6 +35,7 @@ uncacheable := new Set [
   "PushIndent"
   "PopIndent"
   "TrackIndented"
+  "BulletIndent"
 
   // JSX
   "PushJSXOpeningElement"
@@ -247,9 +248,9 @@ makeCache := ({hits, trace}: CacheOptions = {}): CacheEvents ->
 
       // We cache `undefined` when a rule fails to match so we need to use `has` here.
       if stateCache.has(key)
-        if trace
-          logs.push "".padStart(stack.length * 2, " ") + ruleName + ":" + state.pos + "💰"
         result := stateCache.get(key)
+        if trace
+          logs.push "".padStart(stack.length * 2, " ") + ruleName + ":" + state.pos + "💰" + (if result then "✅" else "❌")
         return {
           cache: if result then { ...result }
         }
--- a/source/parser.hera
+++ b/source/parser.hera
@@ -389,7 +389,7 @@ CommaDelimiter
 # (like CommaDelimiter)
 ArgumentList
   # Check for same line arguments then nested arguments
-  ArgumentPart ( CommaDelimiter !EOS _? ArgumentPart )* ( CommaDelimiter ( NestedImplicitObjectLiteral / NestedArgumentList ) )+ ->
+  ArgumentPart ( CommaDelimiter !EOS _? ArgumentPart )* ( CommaDelimiter ( NestedBulletedArray / NestedImplicitObjectLiteral / NestedArgumentList ) )+ ->
     return [
       $1,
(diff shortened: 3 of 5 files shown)


bbrk24:
Can you add `as tuple` or `as const` (or any other type cast, really) to a bulleted array without adding parens?

edemaine:
Good question. No, it's not possible without parens. I think at least `|> as tuple` and `|> as const` should be addable, but even those don't work yet.

I think trailing type casts (`as ...`) in the same places that trailing member/call expressions (and maybe pipes from #1168) would make sense as a separate issue.

STRd6 (approved):
Thanks for implementing this feature. I'm a little bit scared so I think that means we should do it.

edemaine:
I definitely feel like I'm stretching the grammar a bit here. We lack a nice place to put things that work only at the indentation level, not deeper in an expression. For example, I didn't want to allow something like this:

```js
array.map . foo
          . bar
```

(in particular because `array.map . foo` already has a meaning, involving placeholders)

As a result, it seems like I had to add `(Nested)BulletedArray` in a few too many places. Perhaps this could be cleaned up somehow... I'm open to ideas!

Related, I just added support for trailing members/calls and pipelines after top-level bulleted arrays, enabling `|> as const`. I'm not sure I added it to all the right contexts though...

I think this is a reasonable starting point though, and we can extend/fix as we run into further issues.

danielbayley:
@edemaine Surely YAML-like `-` would make more sense here than `.`, which feels a bit too close to…
~~~ js
object
  .method()
  .chaining()
~~~

edemaine:
@danielbayley How would you distinguish from subtraction?

danielbayley:
> @danielbayley How would you distinguish from subtraction?

Iv’e never personally used, or seen any multiline subtraction expressions in the wild. Which is of course, not to say this edge case doesn’t exist…

Maybe the distinction could be that if there is a newline between `-` entries, then it’s an array, else `-`s on the same line is interpreted as subtraction. Then if you did actually want multiline subtraction, you could just wrap that in parentheses? Something like:
~~~ coffee
total := (
  n
  - 1
  - 1
)
~~~

Although, maybe that isn’t even necessary, given that in order to subtract, the first token after the assignment `=` would be a value to subtract from anyway…

@edemaine What do you think?

edemaine:
My worry is examples like this:

```js
func x
  - y
  - z

// parse 1: subtraction (how Civet works now)
func(x
  - y
  - z)

// parse 2: call with array argument (how . works now)
func(x(
  [ y,
    z]))
```

I guess you're saying to prioritize the latter parse, but I don't see how to obtain subtraction behavior without moving the `-` signs onto the previous line (as requierd in CoffeeScript, but something I really don't like about CoffeeScript). Wrapping in parentheses doesn't help.

edemaine:
One possibility that comes to mind: We could add an opt-in YAML compatibility mode via `"civet yaml"` (or some better name) that
* enables `-` and `*` as bullets, but also
* prevents `-` and `*` from being used as binary operators when preceded by a newline.
 
For some, I can imagine this is a worthwhile trade-off.

PR #1364, opened 2024-08-15
Indented if/unless conditions to enable indented function calls (e.g. `(and)`)

### LINKED ISSUE #1090: `if (and)` with indented condition list
This gives a ParseError saying it expected a `::=` where I have `undefined`:
```civet
        return if (and)
            upper.test exactOrAny.toString()
            lower.test exactOrAny.toString()
        then exactOrAny
        else undefined
```
If I oblige and change the last line to `else ::= undefined`, the result is not remotely what I wanted out of this expression (and in fact not valid TS):
```ts
        return (((a,b) => a&&b)?(
            upper.test(exactOrAny.toString()),
            lower.test(exactOrAny.toString())):void 0)
        then(exactOrAny)
        type else = undefined
```
I was hoping to get this:
```ts
return (upper.test(exactOrAny.toString()) && lower.test(exactOrAny.toString())) ? exactOrAny : undefined;
```

edemaine:
This PR explores the proposal from https://github.com/DanielXMoore/Civet/issues/1090#issuecomment-2291837213 :

> I wonder if it would work (and be unambiguous) to allow an *indented condition* to use (further) indented function calls:
>
> ```js
> // if/then/else
> if
>   (and)
>     x
>     y
>   console.log 'yes'
> else
>   console.log 'no'
> 
> // postfix unless
> return unless
>   (and)
>     x
>     y
> ```

It does seem to work quite well! I think this is a natural extension of various settings where we allow indented expressions (e.g. after `return` or `yield`). It looks a little weird when you don't have an indented application, but the expression still has a clear "end" (from the line break):

```js
if
  a + b < 10
  console.log 'small'
```

I'd like to propose that this fixes #1090, but that is perhaps subject to debate.

### DIFF
--- a/source/parser.hera
+++ b/source/parser.hera
@@ -4802,6 +4802,16 @@ Condition
       expression,
     }
 
+  # NOTE: Indented condition allows for further nested function application
+  PushIndent InsertOpenParen:open ( Nested ExtendedExpression )?:expression InsertCloseParen:close PopIndent ->
+    if (!expression) return $skip
+    return {
+      type: "ParenthesizedExpression",
+      children: [open, expression, close],
+      expression,
+    }
+  # NOTE: Unindented condition forbids nested function application
+  # to avoid ambiguity with 'then' clause
   InsertOpenParen:open ExpressionWithObjectApplicationForbidden:expression InsertCloseParen:close ->
     // Don't double wrap parethesized expressions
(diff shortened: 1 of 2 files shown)


STRd6 (approved):
There’s probably more we can do here eventually but this is a good start.

PR #1369, opened 2024-08-18
Indented calls in `if` conditions with explicit `then` clause (#1090)

edemaine:
Inspired by https://github.com/DanielXMoore/Civet/pull/1364#pullrequestreview-2244181171

> There’s probably more we can do here eventually [...]

and specifically https://github.com/DanielXMoore/Civet/issues/1090#issuecomment-1981438104

> Perhaps we could make an exception when there's an explicit `then`.

This PR allows for indented function calls in an `if` condition when there's an explicit `then` to indicate the `then` block. Unlike usual `then`, this one allows for an indented block after it, symmetric to `else`. This lets us write:

```js
if (and)
  cond1
  cond2
then
  // match
else
  // no match
```

as well as #1090's original request:

```js
return if (and)
  upper.test exactOrAny.toString()
  lower.test exactOrAny.toString()
then exactOrAny
else undefined
```

## Performance

One potential issue is that this involves parsing each `if` twice, first with indentation calls enabled, but then failing if there isn't a `then` immediately after, and second with indentation calls disabled.

Running `build/perf-compare.sh` (modified here to parse the entire Civet codebase), the slowdown is a bit inconsistent but I'd say about 5% slowdown (from ~2.25s to ~2.4s parse time). Is this worth it? I'm not sure. (If not, I should cherry-pick some of the small improvements in this PR.)

If you see a way to make the common case more efficient, I'm all ears. Unfortunately, I think we need to check for the `then`-terminated condition before we check the "normal" case; otherwise, we misparse and can't recover.

### DIFF
--- a/build/perf-compare.sh
+++ b/build/perf-compare.sh
@@ -3,7 +3,7 @@
 # Simple performance comparison
 
 count=10
-testFile=source/lib.civet
+testFile=source/**/*.civet
 
 function get_times {
   cmd=$1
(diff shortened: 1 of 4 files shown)


STRd6 (approved):
Looks good! The performance cost seems fine. I think there are things we can do to improve performance across the board in the parser but none of them are very simple. For situations where we parse and then backtrack we may be able to factor out common prefixes to limit the re-parsing.

Issue #661, opened 2023-08-30
Is there any way to fall through in a switch?


gwhitney:
There are cases where falling through in switch statements keeps code DRY; for example, if there is some processing needed when a special condition is met and further processing when a more general condition is met. As a toy example,

```
description .= ''
switch n
   2
      description = 'prime'
      // FALLTHROUGH
   % 2 is 0
      description = 'even ' + description
   else
      description = 'odd'
description += ' number'
```

As far as I can see/understand there isn't anything I can currently write in place of `// FALLTHROUGH`  to have the desired effect that when `n` is 2, the `description` will be `even prime number`. I guess it's not 100% clear what JavaScript this would compile to, maybe

```
let description = ''
if (n === 2) {
   description = 'prime'
   // FALLTHROUGH
   description = 'even ' + description}
else if (n % 2 === 0) {
   description = 'even ' + description}
else {
   description = 'odd'}
description += ' number'
```
where at least Civet is doing the code duplication for me so that I can specify the behavior in a DRY way. Is there any fallthrough mechanism that I missed, or possibly interest in adding one? I certainly agree that breaking should be the default and falling through would require some special notation if allowed at all.

STRd6:
There's not currently any way to fallthrough using Civet's pattern matching. You can fall through using the JS style `switch` and `case` clauses but they won't be able to mix and match with pattern matching.

Having some kind of fallthrough keyword and duplicating the block in pattern matching seems like a reasonable feature to add eventually.

For these kinds of things I often use a pattern like this if I have many checks and am trying to go super DRY:

```civet
checks := [
  [(& is 2), "prime"]
  [(& % 2 is 0), "even"]
]

checks.reduce ([v, descriptions], [check, description]) ->
  [v, [...descriptions, ...(check(v) ? [description] : [])]]
, [n, []]
// then fill in else description if descriptions list is empty
```

... which has some tradeoffs of its own. But I think that the `[(& % 2 is 0), "even"]` predicate trick is pretty handy.

gwhitney:
I really like the syntax `continue switch` to go on to the next test. Very readable/intuitive as to the intended semantics.

As for a use case, I usually use fallthrough when there are two similar cases, one of which just needs a little preprocessing to put it in the form of another. Do I have a specific example with pattern matching in mind right now? No, but I am pretty sure I wanted it at one point, which is why I filed this proposal. Next time I wish I had it, I will try to post as real an example as I can here.

[...]

PR #1373, opened 2024-08-20
`continue switch` to fall through at the end of `when` clause

edemaine:
Part of #661

I realize the original request was for fall through in the pattern matching case. That will require more work... But at least this is a start.

### DIFF
--- a/source/parser.hera
+++ b/source/parser.hera
@@ -5016,6 +5016,14 @@ KeywordStatement
                            // omit colon
     }
 
+  # NOTE: `continue switch` for fallthrough
+  Continue _ Switch ->
+    return {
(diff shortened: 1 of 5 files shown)


source/parser.hera
```diff
@@ -5016,6 +5016,14 @@ KeywordStatement
                            // omit colon
     }
 
+  # NOTE: `continue switch` for fallthrough
+  Continue _ Switch ->
+    return {
+      type: "ContinueStatement",
+      special: $3.token,
+      children: $0,
```
STRd6, line 5024: If we make `children: []` here then we won't need to erase later.

STRd6 (approved):

PR #1374, opened 2024-08-21
Remove Svelte label `$:` exception

STRd6:
Svelte 5 no longer uses `$:` labels.
If you need `$:` labels in Civet, use `:$` instead.


edemaine (approved):
I added to the PR description for link-from-changelog purposes

PR #1375, opened 2024-08-22
Expressionize statements followed by trailing member access/call or pipe

### LINKED ISSUE #1168: Trailing pipe or property access should expressionize statement

edemaine:
Fixes #1168 in a surprisingly simple way: don't parse `if`/`switch`/`try`/iterations as `Statement`s if they're followed by an `AllowedTrailingCallExpression` or `Pipe` operator, leaving them to be parsed as expressions later on.

Also found a simpler way to avoid `async do` and `async for` from being parsed as calls to a function `async`: swapping the order so that `ExpressionizedStatement` gets checked before `UpdateExpression` (which includes `CallExpression`). This brings these into alignment with `async function` and so on which had no special handling.  (This was also a necessary step; the workaround from before didn't work with the rest of this PR.)

### DIFF
--- a/source/parser.hera
+++ b/source/parser.hera
@@ -4125,10 +4125,10 @@ _PostfixStatement
 Statement
   KeywordStatement
   VariableStatement
-  IfStatement
-  IterationStatement
-  SwitchStatement
-  TryStatement
+  IfStatement !ShouldExpressionize -> $1
+  IterationStatement !ShouldExpressionize -> $1
+  SwitchStatement !ShouldExpressionize -> $1
+  TryStatement !ShouldExpressionize -> $1
 
   EmptyStatement
 
@@ -4140,6 +4140,13 @@ Statement
   BlockStatement
 
   # NOTE: no WithStatement
+
+# NOTE: Leave statement with trailing call expressions or pipe operator
+# to be expressionized by
+# ExpressionizedStatementWithTrailingCallExpressions
+ShouldExpressionize
+  AllowedTrailingCallExpressions
+  NotDedented Pipe
 
 # Variant of Statement to forbid , operator (CommaExpression)
 NoCommaStatement
@@ -4983,10 +4990,6 @@ RestoreAll
 CommaExpressionStatement
   # NOTE: semi-colons are being handled elsewhere
   # NOTE: Shouldn't need negative lookahead if shadowed in the proper order
-  # NOTE: Allow for e.g. await forms of IterationExpression that weren't
-  # already parsed as IterationStatement. Must be before AssignmentExpression
-  # to avoid treated `async do ...` like a function call `async(do ...)`.
-  IterationExpression
   # NOTE: CommaExpression allows , operator
   CommaExpression ->
     // Wrap object literal with parens to disambiguate from block statements.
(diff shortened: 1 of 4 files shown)


STRd6 (approved):
Nice one!

PR #1377, opened 2024-08-23
Hyphenated object keys like `Content-Type:`

edemaine:
I was looking back at old Discord messages and noticed this feature suggestion by @esthedebeste. (She also suggested spaces, but that seems harder.)

Hyphenated names are at least a small improvement that's nice for dealing with HTTP headers. Existing support for numbers take priority, so e.g. `-0n:` still turns into a key of `0`, while `-0n-x:` doesn't parse. Basically you need to start with a letter or a hyphen followed by a letter.


STRd6 (approved):
👍

Issue #561, opened 2023-07-02
JSX children as code


texastoland:
I ran into a confusing DX issue experimenting with JSX in the playground.

Working example:

```coffee
const Component = (items: Item[]) =>
  <ul>
    <For each=items()>
      (item) =>
        <li>{
          if item.type === "link"
            <a href=item.url>some link title: {item.title}
          else
            <p>{item.content}
        }
```

Despite being a React instructor (but also CoffeeScript in the past) I intuitively expected Civet to work without braces.

Broken example:

```coffee
const Component = (items: Item[]) =>
  <ul>
    for item of items
      <li>
        if item.type === "link"
          <a href=item.url>some link title: {item.title}
        else
          <p>{item.content}
```

I think code would be a better default for all JSX nodes not only https://civet.dev/cheatsheet#function-children as in:

- https://rescript-lang.org/docs/manual/latest/jsx#departures-from-js-jsx
- facebook/jsx#124 (see `JSXText`)
  - facebook/jsx#35
  - facebook/jsx#8

Proposed example:

```coffee
const Component = (items: Item[]) =>
  <ul>
    for item of items
      <li>
        if item.type === "link"
          <a href=item.url> "some link title: " item.title
        else
          <p> item.content
```

Alternatively signal code children with a marker like `do`:

```coffee
const Component = (items: Item[]) =>
  <ul> do
    for item of items
      <li> do
        if item.type === "link"
          <a href=item.url>some link title: {item.title}
        else
          <p>{item.content}
```

edemaine:
Yes, we've discussed something similar in the `#jsx` channel on Discord. Thanks for putting this proposal together though!

One exception with your proposed example: I don't think `"some link title: " item.title` should work.  (This looks like a potential function call...?)  More intuitive to me would be `"some link title: " + item.title` or `` `some link title: ${item.title}` ``.

## Multiple Children Code Blocks

But I think the bigger issue is how to have multiple code children. Digging up old posts, here's an example I put together with a proposal for treating each top-level line as its own braced block:

```coffee
<div>
  user := getUser()
  if user?
    {name} = user
    <h1> `Welcome ${name}!`
  <h2>Posts</h2>
  for post of posts
    <div .post> post.jsx()
↓↓↓
let user
<div>
  {user = getUser(), void 0}
  {user != null ?
    {name} = user,
    <h1> {`Welcome ${user.name}!`} </h1>
  : void 0
  }
  <h2>Posts</h2>
[...]

PR #1381, opened 2024-08-27
JSX attributes with unbraced indented values, code children with `>`

edemaine:
This PR makes two major advances toward #561, in two commits:

## 1. JSX attributes with unbraces indented values

#561 discussed a few possibilities for a symbol to indicate that the value of an attribute should be an indented block.  I realized that we don't need *any* symbol here; we can just have an indented block (outside `coffeeJSX` mode which maintains full JSX compatibility without respecting indentation). Example:

```coffee
<Show
  when=
    if testing
      timer() > 10
    else
      loading()
  fallback =
    <img src="loading.gif">
    <div>Loading...</div>
>
```

This seems like a clear (uncontroversial) win. It matches our support for indented content in various contexts, most recently condition of an `if` (#1364).

## 2. JSX child code blocks with `>`

I implemented my proposal from https://github.com/DanielXMoore/Civet/issues/561#issuecomment-1806918274 to allow for a JSX code child to be prefixed with `>` instead of wrapping it in braces.

Examples where the content is on the same line as the `>` seem pretty straightforward and nice:

```coffee
<div .greeting>
  Hello
  >' '
  > getUser() |> .first
  !
```

I also added support for indented blocks, with an indentation after the `>`:

```coffee
<div .account>
  >
    user := getUser()
    if user?
      `Logged in as ${user.first}`
    else
      <LoginButton>
```

These get wrapped automatically in `do`, which lets you do the multiple lines of code and local declarations (which are intuitively local given the indentation). This aspect is maybe a little more controversial. It's related to https://github.com/facebook/jsx/issues/39 which we've discussed implementing but haven't: making all JSX braced blocks (probably including attribute values and code blocks) implicitly wrapped in `do`. It's less necessary in Civet where most statements are expressions. But it's relevant exactly when you want to do multiple lines of code.

Alternatively, I could remove the indented block feature, and we could spell this as follows:

```coffee
<div .account>
  >do
    user := getUser()
    if user?
      `Logged in as ${user.first}`
    else
      <LoginButton>
```

On the other hand, maybe `>` is a nice place to introduce divergent behavior like this: a single-line expression (not wrapped in `do`), or a multi-line block implicitly wrapped in `do`. Arguably we should do the same for indented attribute blocks (Part 1 of this PR), if they want to be multiple lines as well?

So potentially a bit of discussion before this should get merged. Let me know your thoughts in particular on whether indented `>` blocks and/or indented attribute values should be wrapped in `do`. (Currently the former is, but the latter isn't.)

And to be clear, this is also meant as a step toward a flag like `"civet jsxCode"` where the `>`s aren't necessary, and instead content needs to be quoted like a string. That would complete #561.

(Also deleted an old `jsx.md` file which seems to have been fully incorporated into `reference.md` long ago.)


STRd6 (approved):
I think this is a good intermediate step on the path towards JSX with natural code.

I'm a little concerned about this stylistically:

```
<div
   ... perhaps many attributes
> !! hard to tell at a glance if this is a close or starting a separate > expression
```

But can't really see a better solution for that at the moment, at least until we get to code as default.

The best way to move forward is probably to test this out and see if new and better ideas come to us.

Thanks for adding this!

PR #1383, opened 2024-08-29
Typing of object/array pattern components via `::`

edemaine:
Fixes #126 by implementing a long-sought syntax for typing individual object properties and array elements in left-hand-side patterns (including function arguments and assignment). Basically, `:: type` can come at the end of any existing object property or array element.  For example:

```js
function Component({
  name: [first:: string, last:: string, ...rest:: string[]],
  counter:: number
  setCounter: sc:: (number) => void
})
↓↓↓
function Component({
  name: [first, last, ...rest],
  counter,
  setCounter: sc
}: {
  name: [ string, string,... string[],],
  counter: number,
  setCounter: (number) => void,}){}
```

Why this notation?
* `name:: type: value` is ambiguous: it could be interpreted as `name:: {type: value}` (typing as object). (I tried to include this as an option at first, but ran into this alternate parse.)
* `name: value: type` is similarly ambiguous: it could be interpreted as `name: {value: type}` (complex destructuring).

I admit I still often type the wrong number of `:`s when writing examples, but I imagine we'll get used to it. (Also open to better proposals for notation.)

I still need to write actual docs, but hopefully the example above and the tests give a reasonable idea.

### DIFF
--- a/source/parser.hera
+++ b/source/parser.hera
@@ -1915,12 +1915,12 @@ BindingPropertyList
 # NOTE: Simplified from spec
 ArrayBindingPattern
   _?:ws1 OpenBracket:open ArrayBindingPatternContent:c __:ws2 CloseBracket:close ->
-    return {
+    return gatherBindingPatternTypeSuffix({
       ...c, // names, blockPrefix, length
       type: "ArrayBindingPattern",
       elements: c.children,
       children: [ws1, open, c.children, ws2, close],
-    }
+    })
 
 ArrayBindingPatternContent
   # NOTE: Added indentation based binding elements
@@ -1979,7 +1979,7 @@ BindingProperty
   BindingRestProperty
 
   # NOTE: Allow ::T type suffix before value
-  _? PropertyName:name _? Colon _? ( BindingIdentifier / BindingPattern ):value BindingPropertyTypeSuffix?:typeSuffix Initializer?:initializer ->
+  _? PropertyName:name _? Colon _? ( BindingIdentifier / BindingPattern ):value BindingTypeSuffix?:typeSuffix Initializer?:initializer ->
     return {
       type: "BindingProperty",
       children: [$1, name, $3, $4, $5, value, initializer],  // omit typeSuffix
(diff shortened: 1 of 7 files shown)


source/parser.hera
```diff
@@ -1975,33 +1978,51 @@ BindingProperty
...
       name,
       value,
+      typeSuffix,
       initializer,
       names: value.names,
     }
 
-  _?:ws Caret?:pin BindingIdentifier:binding Initializer?:initializer ->
+  _?:ws Caret?:pin BindingIdentifier:binding BindingTypeSuffix?:typeSuffix Initializer?:initializer ->
+    $0 = [ws, binding, initializer]  // omit pin and typeSuffix
```
STRd6, line 1994: Probably better use a local variable instead of overwriting `$0` here.

source/parser/binding.civet
```diff
@@ -140,9 +142,71 @@ function arrayElementHasTrailingComma(elementNode)
...
+  return pattern if pattern.typeSuffix?
+  count .= 0
+  switch pattern.type
+    when "ArrayBindingPattern"
+      types: ASTNode[] :=
+        for each elem of pattern.elements
+          typeSuffix .= elem.typeSuffix
+          typeSuffix ??= elem.binding?.typeSuffix
+          count++ if typeSuffix
+          typeElement .= [ typeSuffix?.t, "," ]
```
STRd6, line 157: We should only add trailing commas as authored in the source. It's not necessary for this to merge but ideally trailing commas would always be as authored (some parts of the parser currently do this better than others).

STRd6 (approved):
Looks good, I like the feature! The explanation of the challenges with the other approaches makes sense to me.

I added a few nitpicks but don't let them hold this up if it's too tedious.

source/parser/binding.civet
edemaine: Thanks for encouraging this; it turned out to be pretty easy to improve upon using `delim` properties (just has to add them in one case). It's still not exactly perfect; there's an added comma at the end of an object when a `...rest` part got extracted out so it wouldn't be necessary. But it seems close enough.

Leaving better preservation of whitespace to future work. It's pretty good now, but definitely not perfect.

PR #1386, opened 2024-08-30
`"civet jsxCode"` directive treats JSX children as Civet code

### LINKED ISSUE #561: JSX children as code

edemaine:
This PR should finish the rest of and thus fixes #561.

* `"civet jsxCode"` treats all JSX children as Civet expressions.
* I went with my original proposal where each Civet line of code is treated as a separate expression wrapped in braces. I think this is the most common case, where you have many consecutive `if`s and such, each of which needs to render. And it's easy to make a larger block, using a `do` block. (See example in documentation.)
* I was less sure whether children on the same line should be treated as code, so I made it an option!
  * `"civet jsxCodeNested"` just enables the code mode for nested (indented) children.
  * `"civet jsxCodeSameLine"` just enables the code mode for same-line children (not sure why you'd do this, but I originally had in mind `"civet jsxCode -jsxCodeSameLine"`).
  * `"civet jsxCode"` turns on both of the above.

As expected, #1381 turned out to be a very helpful step in this direction, defining what a code child would look like. Most of that code is re-used here.

As future work, I could see converting strings like `"foo"` into JSX text `foo` instead of the current behavior, `{"foo"}`. But there are lots of subtle differences between strings and text, and I didn't want to get bogged down in that. (I think some JSX preprocessors might already do this anyway.)

### DIFF
--- a/civet.dev/config.md
+++ b/civet.dev/config.md
@@ -39,6 +39,9 @@ and following ESM, require importing files with the correct extension.
 | [`autoLet`](reference#autolet) | automatically declare undeclared variables with `let` |
 | [`autoVar`](reference#autovar) | automatically declare undeclared variables with `var` |
 | [`defaultElement=tag`](reference#implicit-element) | specify default JSX tag: `<.foo>` → `<tag class="foo">` |
+| [`jsxCode`](reference#automatic-code-children) | treat all JSX children as Civet code |
+| [`jsxCodeNested`](reference#automatic-code-children) | treat indented JSX children as Civet code |
+| [`jsxCodeSameLine`](reference#automatic-code-children) | treat same-line JSX children as Civet code |
 | [`objectIs`](reference#object-is) | implement the `is` operator via `Object.is` |
 
 ## ECMAScript Compatibility
--- a/civet.dev/reference.md
+++ b/civet.dev/reference.md
@@ -2507,6 +2507,56 @@ To write an entire block of code, indent it inside `>`
     `Logged in as ${user.first} ${user.last}`
 </Playground>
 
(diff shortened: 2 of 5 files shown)


STRd6 (approved):
Amazing! JSX is almost readable now 😹

Issue #1121, opened 2024-03-26
Binary operators starting on new lines


edemaine:
I *think* this is a bug:
```coffee
extra.split(/\s+/).filter &
++ [state.start, state.goal]
```

Expected compilation:
```js
extra.split(/\s+/).filter($ => $)
.concat([state.start, state.goal])
```

Actual compilation:
```js
extra.split(/\s+/).filter($ =>
  $.concat([state.start, state.goal])
);
```

STRd6:
Related #1280

edemaine:
I notice that the same issue occurs without `&` altogether:

```js
extra.split(/\s+/).filter x
++ [state.start, state.goal]
```

The real issue is that we handle binary operators that start on the next line, and then there's a question of which expression to attach that to.

I guess our intuition is that the implicit function call (on `filter`) should end by the newline, but that's not how things work currently.

edemaine:
Inspired by the bulleted arrays PR (#1361), I have an idea for how to address this: when entering `ImplicitArguments`, push an artificial indentation of the current indentation + 1, so that to remain inside the function call argument, you need to be strictly indented.  This would mean:

```js
filter foo
+ rest
↓↓↓
filter(foo)
+ rest

// but
filter foo
  + rest
↓↓↓
filter(foo
  + rest)
```

The bad news is that chaining multiple implicit function calls on the same line would require additional indentation, like so:

```js
f1 f2 f3 f4 f5 x
     + y
// ^needs 5 spaces because 5 function calls
```

I don't think that's good, but I'm also not quite sure how to fix it.

Maybe instead of increasing the indentation by 1, we add a boolean flag to the indentation saying "> i" where i is the current indentation, instead of the usual meaning of ">= i". And then we won't repeatedly add 1; instead we just repeatedly change the flag to ">" which has no effect beyond the first. Not the simplest solution, but maybe workable...

Let me know if you have better ideas!

edemaine:
I reviewed this idea some more, and realize that we have *almost* this exact functionality, via `ForbidNewlineBinaryOp` (forbid newline followed by nested or indented binary op).  If I just add this to `ImplicitArguments`, then both

```js
extra.split(/\s+/).filter &
++ [state.start, state.goal]
```

```js
extra.split(/\s+/).filter &
  ++ [state.start, state.goal]
```

parse as

```js
extra
  .split(/\s+/)
  .filter(($) => $)
  .concat([state.start, state.goal]);
```

[...]

PR #1390, opened 2024-09-03
Improve indentation handling with union/intersection types

### LINKED ISSUE #1355: Object union type is misparsed
Civet:
```
type WordJSON =
  | type: 'single', input: Input
  | type: 'chord', input: Input[]
```
Expected output:
```ts
type WordJSON = 
  | { type: 'single', input: Input }
  | { type: 'chord', input: Input[] }
  ;
```
Actual output:
```ts
type WordJSON = {
  type: 'single',
  input:
    | Input
    | { type: 'chord', input: Input[] }
    ,
};
```

edemaine:
Fixes #1355

* As expected, some `__`s needed to be `NotDedented`. I just changed the minimal amount here, around binary operators. (Plausibly more should change, but not for this issue.)
* I added a special rule for matching properly indented binary-operator chains like this:
  ```ts
  | T1
  | T2
  ```
* All this still wasn't enough, because the indentation level was already 2 in this case. I needed to add an indentation level of +1 for these items so that `NotDedented` doesn't allow for grabbing the second item.

This feels related to https://github.com/DanielXMoore/Civet/issues/1121#issuecomment-2292132835 as well. (I could use feedback on those ideas too, though doesn't need to happen before this PR lands.)

### DIFF
--- a/source/main.civet
+++ b/source/main.civet
@@ -36,6 +36,7 @@ uncacheable := new Set [
   "PopIndent"
   "TrackIndented"
   "BulletIndent"
+  "PushExtraIndent1"
 
   // JSX
   "PushJSXOpeningElement"
--- a/source/parser.hera
+++ b/source/parser.hera
@@ -7483,6 +7483,9 @@ MaybeNestedType
   # NOTE: Let InterfaceBlock take first crack at an indented block
   # But don't prevent parsing a braced type with unary suffix like {}[]
   NestedInterfaceBlock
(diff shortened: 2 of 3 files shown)


STRd6 (approved):
Looks good! I read over the comment https://github.com/DanielXMoore/Civet/issues/1121#issuecomment-2292132835 and I think that adding the flag that changes the indent checking of some nesting to `>` rather than `>=` seems like it may work well.

PR #1391, opened 2024-09-04
Implicit type arguments, bulleted type tuples, multiple items in indented tuples

edemaine:
This started as an implementation of the idea that `A B` in a type context should parse as `A<B>`. But along the way I noticed that type tuples were woefully lacking compared to array literals.

* Generally revamp type tuple parser to match array literal parser, in particular:
  * Support multiple types on one line of an indented tuple
  * Support bulleted type tuples
* Support implicit type arguments, with space or indentation, similar to function calls

TODO: Documentation


source/parser.hera
```diff
@@ -7679,14 +7691,83 @@ TypeElement
...
+# NOTE: Nested bulleted tuple starts with an indentation.
+NestedTypeBulletedTuple
+  ( InsertSpace InsertOpenBracket ):open PushIndent NestedTypeBullet*:content InsertCloseBracket:close PopIndent ->
+    if (!content.length) return $skip
+    content = content.flat() // combine bullets into one array
+
+    // Remove implicit comma from last element
+    const last = content[content.length - 1]
+    if (Array.isArray(last)) {
+      if (last.at(-1).implicit) last.pop()
```
STRd6, line 7720: This mutation may cause caching issues with backtracking on some parses.

If we move it to a post processing handler then it won’t be an issue or we could make in non-mutative.

STRd6 (approved):
Looks good, just double check that the pop mutation in those 4 places is safe.

Thanks!

PR #1392, opened 2024-09-04
Remove final implicit comma in nested arguments, better matching source

edemaine:
Currently, nested implicit arguments create an extra implicit comma at the end of the arguments.

I had to fix this for #1391 because type arguments can't have a trailing comma in TypeScript.

This led the way to fixing it for function calls too.


STRd6 (approved):
Thanks!

PR #1394, opened 2024-09-08
`for*`, `loop*`, `while*` generator expressions

### LINKED ISSUE #505: `for*` generator expressions
We can make it easy for `for` to generate (`yield`) values instead of accumulating an array and returning it, given that we're already wrapping in an IIFE.

See https://github.com/jashkenas/coffeescript/issues/5379 for this exact idea.  I didn't think `for*` made sense in a CoffeeScript syntax, but it makes sense in Civet which inherits all of ECMAScript; I think it nicely pairs with the recently added `async for`.  I'm also inspired here by [Janet's `generate`](https://janet.guide/control-flow/).

edemaine:
@vendethiel's WIP branch aiming to fix #505

### DIFF
--- a/source/parser.hera
+++ b/source/parser.hera
@@ -4164,7 +4164,9 @@ Statement
   KeywordStatement
   VariableStatement
   IfStatement !ShouldExpressionize -> $1
-  IterationStatement !ShouldExpressionize -> $1
+  IterationStatement !ShouldExpressionize ->
+    if ($1.star) return $skip
+    else return $1
   SwitchStatement !ShouldExpressionize -> $1
   TryStatement !ShouldExpressionize -> $1
 
@@ -4303,6 +4305,7 @@ IterationExpression
       block: statement.block,
       statement,
       async,
+      star: statement.star
     }
 
 # NOTE: Added from CoffeeScript
@@ -4316,7 +4319,7 @@ LoopStatement
     }
 
 LoopClause
-  Loop:kind ->
+  Loop:kind ( _? Star )?:star ->
     const expression = {
       type: "Literal",
       children: ["true"],
@@ -4332,6 +4335,7 @@ LoopClause
       subtype: kind.token,
       children: [kind, condition],
       condition,
+      star
     }
 
 # https://262.ecma-international.org/#prod-DoWhileStatement
@@ -4408,16 +4412,17 @@ ForStatement
     }
 
 ForClause
-  For __ ForStatementControl:c  ->
+  For ( _? Star )?:star __ ForStatementControl:c  ->
     const {children, declaration} = c
 
     return {
       type: "ForStatement",
-      children: [$1, ...$2, ...children],
+      children: [$1, ...$3, ...children],
       declaration: declaration,
       block: null,
       blockPrefix: c.blockPrefix,
       hoistDec: c.hoistDec,
+      star,
     }
 
 ForStatementControl
--- a/source/parser/function.civet
+++ b/source/parser/function.civet
@@ -564,7 +564,7 @@ function processFunctions(statements, config): void
     processReturn(f, config.implicitReturns)
 
 function expressionizeIteration(exp: IterationExpression): void
-  { async, subtype, block, children, statement } := exp
+  { async, star, subtype, block, children, statement } := exp
   i := children.indexOf statement
   if i < 0
     throw new Error("Could not find iteration statement in iteration expression")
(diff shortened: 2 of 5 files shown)


source/parser.hera
```diff
@@ -4164,7 +4164,9 @@ Statement
   KeywordStatement
   VariableStatement
   IfStatement !ShouldExpressionize -> $1
-  IterationStatement !ShouldExpressionize -> $1
+  IterationStatement !ShouldExpressionize ->
+    if ($1.star) return $skip
+    else return $1
```
edemaine, line 4169:
```suggestion
    return $1
```

source/parser.hera
```diff
@@ -4164,7 +4164,9 @@ Statement
   KeywordStatement
   VariableStatement
   IfStatement !ShouldExpressionize -> $1
-  IterationStatement !ShouldExpressionize -> $1
+  IterationStatement !ShouldExpressionize ->
+    if ($1.star) return $skip
```
edemaine, line 4168: Maybe add a comment for this line, like `// Leave starred forms for IterationExpression`

source/parser.hera
```diff
@@ -4332,6 +4335,7 @@ LoopClause
       subtype: kind.token,
       children: [kind, condition],
       condition,
+      star
```
edemaine, line 4338:
```suggestion
      star,
```

source/parser/function.civet
```diff
@@ -575,25 +575,36 @@ function expressionizeIteration(exp: IterationExpression): void
...
-  { resultsRef } := exp
+  if star
+    assignResults block, (node) =>
+      [ "yield ", node ]
+    braceBlock(block)
+    children.splice(i,
+      1,
+      wrapIIFE(children.map((c) => ["", c, undefined]), async, star)
+    )
+    updateParentPointers block // <- is this necessary? I do add `"yield "`
```
edemaine, line 586: Yes, I think so. It could maybe be moved to `assignResults`. But for now probably makes sense to match the old code. (This could be brought outside the `if`.)

source/parser/lib.civet
```diff
@@ -1051,6 +1051,7 @@ function attachPostfixStatementAsExpression(exp, post: [WSNode, IfStatement | It
         children: [statement],
         block: statement.block,
         statement,
+        star: statement.star
```
edemaine, line 1054:
```suggestion
        star: statement.star,
```

source/parser/types.civet
```diff
@@ -698,6 +701,7 @@ export type ArrowFunction =
   parent?: Parent
   name: string
   async: ASTNode[]
+  star?: ASTNode
```
edemaine, line 704: This probably shouldn't be here.

edemaine (commented):
This is looking good! I think tests would go in the various loop files: `test/do.civet`, `test/for.civet`, `test/while.civet`, etc. (A bit spread out, but that's the pattern we've been following.) Each file might deserve a `describe` block for the starred form. Thanks!

source/parser/function.civet
vendethiel: It’s not done on the same AST node however

source/parser/util.civet
```diff
@@ -563,18 +563,33 @@ function wrapIIFE(expressions: StatementTuple[], asyncFlag?: boolean): ASTNode
...
-    returnType: undefined
-    ts: false
-    async
-    block
-    children: [async, parameters, "=>", block]
-  }
+  let fn
+  if generator
+    fn = makeNode {
+      type: "FunctionExpression"
```
STRd6, line 572: One thing we'll want to do in the future (doesn't need to hold up this PR) is to bind `this` and shadow arguments if necessary.

`.bind(this)()`

To detect if we should shadow arguments we can look inside the function to see if there is an `arguments` identifier that isn't declared and if so pass it in like:

```
if needsToShadowArguments
  parameters :=
    type: "Parameters"
    children: ["(arguments)"]
    names: ["arguments"]
  
  ...
  
  .bind(this)(arguments)

```

Thanks for the PR! We're always happy to have more contributors. Again, this doesn't need to hold anything up, just wanted to note it while I was thinking about it and if it doesn't make it in this one then one of us can get to it eventually.

vendethiel:
The `arguments` passing thing is missing, the rest should be good. Please take another look :).

source/parser/util.civet
```diff
@@ -541,7 +541,7 @@ function parenthesizeType(type: ASTNode)
  * uses await, or just adding async if specified.
```
edemaine, line 541: Can you mention the new argument in the comment?

test/loop.civet
```diff
@@ -30,6 +30,17 @@ describe "loop", ->
...
+    loop generator
+    ---
+    loop*
+      1
+    ---
+    (function*(){while(true) {
+      yield 1
+    }})()
+  """
+
```
edemaine, line 43: Can you add a postfix example somewhere, e.g. `1 loop* 1`?

source/parser/function.civet
```diff
@@ -648,7 +648,7 @@ function expressionizeIteration(exp: IterationExpression): void
     braceBlock(block)
     children.splice(i,
       1,
-      wrapIIFE(children.map((c) => ["", c, undefined]), async, star)
+      wrapIIFE([children.map((c) => ["", c, undefined])], async, star)
```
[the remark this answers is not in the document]
vendethiel: I debugged it by copying most of the else again and tweaking back to be what I needed… and that’s what got me in this working state

edemaine:
Replacing with #1398 (or if @vendethiel wants to make their own PR, I could probably push to the branch)

PR #1396, opened 2024-09-11
`break with` and `continue with` for modifying results array in a loop

### LINKED ISSUE #1395: Allow control over the value that the `loop` expression "returns"
I literally just stumbled on this language a few minutes ago, so I do apologize for raising this suggestion while being as inexperienced in the language as possible, and I certainly don't want to disrupt any behavior that the community already enjoys. I did want to raise this as a suggestion, because I've found this to be very enjoyable behavior in Rust.

Basically, in Rust, [`break` can be used to "return" a value from `loop`, and only `loop` (not `while` or `for`)](https://doc.rust-lang.org/std/keyword.break.html).

```rust
const FAVORITE_NUMBER: u8 = 7;

fn main() {
    let mut i = 0;
    let selected = loop {
        if i == FAVORITE_NUMBER {
            break i;
        }
        i += 1;
    };
    println!("{selected}"); // 7
}
```

Once I saw that Civet had the `loop` keyword, I looked to see if this same behavior was possible. I didn't see anything like it in the reference. I wanted to bring this up to ask if
- This is feasible to implement (it might conflict with `break LABEL`)
- This is desirable desirable to the Civet community

edemaine:
Fixes #1395

TODO: Documentation

An issue with TypeScript is clear in the `loop` example: we have `let results = []` but later `results = 'done'`. Perhaps we could detect specifically a `loop` that has no plain `break` and thus will only be exited via `break with`, and then we could just write `let results`.

### DIFF
--- a/test/for.civet
+++ b/test/for.civet
@@ -774,10 +774,23 @@ describe "for", ->
         . "NaN"
       x+1
     ---
-    let results=[];for (const x of y) {
+    let results;results=[];for (const x of y) {
       if (!(x != null)) { results.push(null);continue }
       if(isNaN(x)) { results = [
           "NaN"];break}
       results.push(x+1)
     };const values =results
+  """
+
+  testCase """
+    break/continue with outside expression context
+    ---
+    for x of y
+      continue with null unless x?
+      break with x
+    ---
+    for (const x of y) {
+      if (!(x != null)) { continue }
+      break
+    }
   """
--- a/test/for.civet
+++ b/test/for.civet
@@ -764,3 +764,20 @@ describe "for", ->
         results.push(x ** 2)
       }return results})())
     """
+
+  testCase """
+    break/continue with
+    ---
+    values := for x of y
+      continue with null unless x?
+      if(isNaN x) break with
+        . "NaN"
+      x+1
+    ---
+    let results=[];for (const x of y) {
+      if (!(x != null)) { results.push(null);continue }
+      if(isNaN(x)) { results = [
+          "NaN"];break}
+      results.push(x+1)
+    };const values =results
+  """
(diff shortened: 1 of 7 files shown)


STRd6 (approved):
Looks great!

One solution to help TypeScript is to declare the results as `let results;results=[]`. This gives it the proper union type of all possible exits from the loop.

test/for.civet
```diff
@@ -774,10 +774,23 @@ describe "for", ->
...
       if(isNaN(x)) { results = [
           "NaN"];break}
       results.push(x+1)
     };const values =results
   """
+
+  testCase """
+    break/continue with outside expression context
+    ---
+    for x of y
```
vendethiel, line 788: Maybe this should warn?

edemaine:
> One solution to help TypeScript is to declare the results as `let results;results=[]`. This gives it the proper union type of all possible exits from the loop.

Good idea! Implemented.

I also removed `results=[]` in the special case of `loop` containing `break with` but not `break`.

test/for.civet
edemaine: Hmm, good question. It does feel a little worrying throwing away code (the `with` arguments), though I think it's nicer than failing, in particular as you're transforming code to/from expression mode.

We don't currently have a warning system. @STRd6 Should we implement one? Perhaps `type: "Warning"` nodes, or `warning: true`/`subtype: "Warning"` on `type: "Error"` nodes? This would require some restructuring of current error handling, to report when possible but fail only in certain cases.

STRd6: Seems like something we'd like eventually. We can use a subtype on Error nodes for now.

STRd6 (approved):

test/for.civet
edemaine: Sounds good. If it's OK with you, I'd like to leave this for a future project. Actually checking for `break with`/`continue with` that isn't consumed in a loop is trickier.

I also realize that `break :label with foo` doesn't currently work. (Currently you can only use `break with`/`continue with` to modify the parent loop.) Maybe also good for a future PR.

PR #1397, opened 2024-09-12
`break/continue [label] with` and better error handling

edemaine:


STRd6 (approved):
Looks good but I'd keep it as `LabeledStatement` to be in line with American English.

edemaine:
I prefer American English and "labeled" as well, but I was trying to match this unchanged code:

```js
# https://262.ecma-international.org/#prod-LabelledStatement
LabelledStatement
  Label:label LabelledItem:statement ->
    return {
      type: "LabelledStatement",
```

The types should match the grammar, which this PR achieves. It also seems reasonable for the grammar to match the spelling from the ECMA spec given the overlap. Alternatively, I'm fine with changing everything in the other direction, from Labelled to Labeled.

STRd6:
Good point! Probably best to match the ECMA spec, then default to American English for anything else.

Thanks for the PR!

PR #1398, opened 2024-09-13
`for*`, `loop*`, `while*` generator expressions

### LINKED ISSUE #505: `for*` generator expressions

edemaine:
Fixes #505

I couldn't push to @vendethiel's branch, so created this instead. Hopefully this is now ready to go.

TODO: Docs

### DIFF
--- a/source/parser/util.civet
+++ b/source/parser/util.civet
@@ -600,9 +600,11 @@ function wrapIIFE(expressions: StatementTuple[], asyncFlag?: boolean, generator?
 
   children := [ makeLeftHandSideExpression(fn), "()" ]
 
-  if fn.type is "FunctionExpression" and
-     gatherRecursiveWithinFunction(block, (is like { token: "this" }))#
-    children.splice 1, 0, ".bind(this)"
+  if fn.type is "FunctionExpression"
+    if gatherRecursiveWithinFunction(block, (is like { token: "this" }))#
+      children.splice 1, 0, ".bind(this)"
+    if gatherRecursiveWithinFunction(block, (is like { token: "arguments" }))#
+      children.-1 = parameters.children.-1 = "(arguments)"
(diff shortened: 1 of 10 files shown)


source/parser/util.civet
```diff
@@ -569,22 +570,46 @@ function wrapIIFE(expressions: StatementTuple[], asyncFlag?: boolean): ASTNode
...
+      children: [async, parameters, "=>", block]
+    }
 
-  exp := makeNode
+  children := [ makeLeftHandSideExpression(fn), "()" ]
+
+  if fn.type is "FunctionExpression"
+    if gatherRecursiveWithinFunction(block, (is like { token: "this" }))#
+      children.splice 1, 0, ".bind(this)"
+    if gatherRecursiveWithinFunction(block, (is like { token: "arguments" }))#
```
STRd6, line 606: In the future we'll want to check to see if this `arguments` is declared in a surrounding scope or is the implicit function `arguments` but this is good for now. I don't think our variable scoping machinery makes this very easy yet.

STRd6 (approved):

source/parser/util.civet
edemaine: Oh, I didn't realize `arguments` could be bound to. Fun! Agreed, this should be doable, but probably best left for future work.

PR #1400, opened 2024-09-18
Run `<script type="text/civet">` in browser build

edemaine:
Similar to [CoffeeScript](https://coffeescript.org/#scripts), added support for automatically running `<script type="text/civet">` code that appears in the HTML that loads the browser build of Civet.

The new `source/browser.civet` in general allows us to add code specific to the browser build.

I renamed `source/browser.civet` to `source/browser-shim.civet` but the diff doesn't show it that way.

**Motivation:** Easy to run Civet in online sandboxes without any build tools.

### DIFF
--- a/integration/script/README.md
+++ b/integration/script/README.md
@@ -1,6 +1,14 @@
 # Civet in `<script>` tags
 
-For quick hacks, you can run Civet code directly from your HTML, like so:
+For quick hacks, you can run Civet code directly from your HTML
+using `<script type="text/civet">` tags.
+First, run the browser build of Civet, for example, from CDN:
+
+```html
+<script src="https://cdn.jsdelivr.net/npm/@danielx/civet/dist/browser.js"></script>
+```
+
+Then you can run Civet scripts like so:
 
 ```html
 <script type="text/civet">
@@ -8,7 +16,7 @@ For quick hacks, you can run Civet code directly from your HTML, like so:
 </script>
 ```
 
-You can also run Civet from files, like so:
+You can also run Civet code from files, like so:
 
 ```html
 <script type="text/civet" src="filename.civet"></script>
@@ -16,6 +24,10 @@ You can also run Civet from files, like so:
 
 In the latter case, the HTML needs to be served by a web server
 (not via `file:` URL) to pass CORS.
+
+Note that `<script type="text/civet">` tags get run once,
+when the page is fully loaded.  If you need to run them at a specific time
+(e.g. after dynamically generating content), call `Civet.runScripts()`.
 
 ## Example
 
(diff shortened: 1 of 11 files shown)


source/sourcemap.civet
```diff
@@ -234,7 +234,10 @@ encodeBase64 := (value: number) ->
 
 // Note: currently only works in node
```
STRd6, line 235: Can remove this comment

integration/script/README.md
```diff
@@ -0,0 +1,25 @@
+# Civet in `<script>` tags
+
+For quick hacks, you can run Civet code directly from your HTML, like so:
+
+```html
+<script type="text/civet">
```
STRd6, line 6: Should this example also show adding the Civet browser script to enable this?

STRd6 (approved):

source/browser.civet
```diff
@@ -1,19 +1,28 @@
...
+    if source := (script as HTMLScriptElement).src or script.getAttribute 'data-src'
+      options.filename = source
+      code = fetch source |> await |> .text() |> await
+    else
+      options.filename = script.id or `<script>${i}`
+      code = script.innerHTML
+
+    js := compile code, options |> await
+    window.eval js
+
```
STRd6, line 24: We may also want to add an observer for when scripts are appended to the DOM after content loaded and execute them as well.

edemaine: Interesting idea! This might be a performance hit, so perhaps we should only enable this when calling a function, something like `Civet.autoRunScripts(document.body)`? Or does that beat the point of ease-of-use? Related, it seems that [mutation observers](https://developer.mozilla.org/en-US/docs/Web/API/MutationObserver) need to target a particular node, and we don't necessarily know where `<script>` tags will be added (but `document.head` and `document.body` seem like the most likely places). Perhaps we could make it particularly easy to use, as in `<script src=".../civet/dist/browser.js" data-auto-run-scripts="body">` (where `body` is a query selector).

STRd6: I wouldn't worry too much about the performance and just keep it simple. The direct in browser script is mainly for quick prototyping and should likely be avoided in performance sensitive applications in either case.

edemaine: I agree that `<script type="text/civet">` should only be used for quick prototyping and performance isn't a concern there. My worry is that the browser build might be used in other contexts, too. For example, we use it in our own docs, albeit in a worker thread so none of this DOM code will run. But it seems plausible that someone would want the Civet compiler in browser, without wanting the overhead of mutation observers because they don't use `<script type="text/civet">`. Or do you think the performance impact is small enough? Maybe if we just monitor `document.head`, which I imagine doesn't change often in most applications.

edemaine: OK, I implemented watching for direct children of `head` and `body`, and implemented overrides on the `<script>` tag that includes Civet to turn off all of the functionality of this PR, or selectively just the initial load or the mutation observer. This lets you run scripts when you want, or watch exactly what you want, or turn off everything for performance, if desired.

STRd6 (approved):

PR #1405, opened 2024-10-03
change types order in package.json exports; default config exports

STRd6:
Latest esbuild had a warning about our order when declaring exports so I changed it:

```
▲ [WARNING] The condition "types" here will never be used as it comes after both "import" and "require" [package.json]

    ../package.json:49:6:
      49 │       "types": "./dist/unplugin/rollup.d.ts"
         ╵       ~~~~~~~
```

I also added default exports to config, and fixed up the types for the default export.

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,9 +11,9 @@
   },
   "exports": {
     ".": {
+      "types": "./dist/types.d.ts",
       "import": "./dist/main.mjs",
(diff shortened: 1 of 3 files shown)


edemaine (approved):
Nice — I didn't know this ordering constraint!

PR #1409, opened 2024-10-03
VSCode plugin ignores CoffeeScript files

STRd6:
Don't activate the extension on .coffee files.


edemaine (commented):
I'm OK with this (and if you already released, should probably merge), but isn't the output from Civet potentially helpful here? In the same way that TypeScript can offer useful feedback for `.js` files. I think ideally we'd make this option into a setting for the VSCode plugin.

STRd6:
It could be helpful in theory but we'd need to maintain our own CoffeeScript plugin for the LSP that handles source mapping. I had a prototype one previously but gave it up.

It may be worth exploring in the future especially if someone is trying to migrate a larger CoffeeScript code base.

As it stands currently it wasn't doing anything other than showing that error message at the moment.

edemaine (approved):
Oh, I see! Definitely makes sense then.

PR #1413, opened 2024-10-05
Add syntax highlighting support for /d, /s and /v RegExp flags

bbrk24:
I just replaced all instances of "gimuy" (full list of RegExp flags as of ES2017) with "dgimsuvy" (full list of RegExp flags as of current ES2024), since regex literals with the `d` or `v` flags are not currently highlighted as regexen.


edemaine (approved):
Thanks!

PR #1418, opened 2024-10-05
Type specification in `for..of/in` loops

### LINKED ISSUE #1152: Type specification in `for`
I think we should allow specifying the type of declarations within `for` loops. Examples:

```js
for x: T of y
for x: T in y
for each x: T of y
for let [x, y]: [A, B] of y
```

Some of these currently have a different meaning (e.g. `for {x: T} of y`). An alternative that would preserve this implicit object destructuring meaning would be able to allow type specifiers only when there's an explicit `const`/`let`/`var`:

```js
for const x: T of y
```

### LINKED ISSUE #1415: Type signature gets eaten in destructuring of for loop
```civet
for [type:: TokenType, token] of Object.entries(tokens)
  ;
```

```js
for (const [type, token] of Object.entries(tokens)) {
}
```

Related: #1152, #1402

edemaine:
Fixes #1152
Fixes #1415

### DIFF
--- a/test/for.civet
+++ b/test/for.civet
@@ -236,14 +236,25 @@ describe "for", ->
     }
   """
 
-  describe "", ->
+  describe "binding left-hand side", ->
     testCase """
-      binding lefthandside
+      a.b
       ---
       for a.b of x
         console.log(a)
       ---
       for (a.b of x) {
+        console.log(a)
+      }
+    """
+
+    testCase """
+      array
+      ---
+      for [a.b] of x
+        console.log(a)
+      ---
+      for ([a.b] of x) {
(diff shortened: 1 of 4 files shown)


test/for.civet
```diff
@@ -248,6 +248,17 @@ describe "for", ->
       }
     """
 
+    testCase """
+      array
+      ---
+      for [a.b] of x
+        console.log(a)
+      ---
+      for ([a.b] of x) {
```
STRd6, line 257: Everyday I learn a new and amazing corner of JS syntax.

STRd6 (approved):

PR #1419, opened 2024-10-05
Make `autoVar` skip `declare`d variables

### LINKED ISSUE #1412: `declare` + `autoVar` does not cooperate
`declare var options; options = foo` emits `var options; declare var options; options = foo`

edemaine:
Fixes #1412


STRd6 (approved):
Looks good! May want to add a test.

Thanks!

PR #1423, opened 2024-10-05
`"civet globals"` directive to prevent some auto declarations

### LINKED ISSUE #1411: Feature request: define globals in the compile function for to avoid autoVar-ing them
**Proposal:** Allow developers to add an array of global variable names called `globals` to `CompileOptions` and make `autoVar`, `autoLet` and `autoConst` not shadow the names listed there.

**Use cases:**

* Users write a function body in a game engine and cannot modify the function's descriptor. The function's body is then compiled on game export — function's arguments are lost due to `autoVar`.
  ```coffee
  # Wrapper: function (options?: Sausage)
    if not options
      options = getDefaultSausage() # `options` gets shadowed
  ```
* Civet code may reference and write global variables. For example, written to `window` or coming from a 3rd-party framework. In my engine, people oftentimes define games' global variables in separate top-level scripts so they are visible to all game entities' events.
  ```coffee
  # in a top-level script
  var sausageNumber = 0
  window.debugMode = true
  ```
  ```coffee
  # in a separately compiled script
  function resetAll
    sausageNumber = 0 # `sausageNumber` gets shadowed
    debugMode = false # `debugMode` is shadowed, too
  resetAll()
  ```

**Possible workarounds:**

* Using `window.`
* Compiling functions as whole and stripping away unneeded code
* Compiling whole projects as one big file (not possible in my case, plus goodbye incremental builds and caching)

edemaine:
Fixes #1411

### DIFF
--- a/civet.dev/config.md
+++ b/civet.dev/config.md
@@ -39,6 +39,7 @@ and following ESM, require importing files with the correct extension.
 | [`autoLet`](reference#autolet) | automatically declare undeclared variables with `let` |
 | [`autoVar`](reference#autovar) | automatically declare undeclared variables with `var` |
 | [`defaultElement=tag`](reference#implicit-element) | specify default JSX tag: `<.foo>` → `<tag class="foo">` |
+| [`globals`](reference#globals) | avoid automatically declaring listed global variables |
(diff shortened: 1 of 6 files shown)


STRd6 (approved):
:clap:

PR #1428, opened 2024-10-06
Fix missing semicolon in if/unless followed by else

### LINKED ISSUE #1422: Missing semicolon in `if`/`unless` followed by `then`
```civet
foo |> (a) => unless run? then a else a.option 'run'
```
produces invalid JS (missing semicolon before the `else`). Attempted workarounds don't even parse.

edemaine:
Fixes #1422

We had a bunch of places where we add a semicolon to the `then` block of an `IfStatement` to separate it from an `else` block, but none that applied universally. Now apply the logic when building the AST, instead of when doing implicit returns from functions. There are still a couple of places with the add-semicolon logic, where we add extra `else` clauses.

Also improve more typing. This PR reduces the number of type errors from 718 to 673! (via `civet --typecheck source/*.civet`)

### DIFF
--- a/source/parser.hera
+++ b/source/parser.hera
@@ -4275,6 +4275,15 @@ IfStatement
       kind = { ...kind, token: "if" }
       condition = negateCondition(condition)
     }
+    // Ensure semicolon separating single line from `else`
+    if (block.bare && e) {
+      const semicolon = ";"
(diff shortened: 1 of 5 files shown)


STRd6 (approved):
Thanks!

PR #1430, opened 2024-10-06
Stop implicitly returning final value from generators

edemaine:
This PR turns off implicit `return` at the end of generator functions (`function*`, implicit or explicit).

I think it's extremely common not to want to have implicit `return` in generators, especially when you have loops (very common in generators); it creates a wasteful array. As you can see in this PR, an example is in the docs, where I previously had to add an explicit `return` to prevent a final `return`.

I've not yet used the special final return value of generators (which is what gets set by `return`ing a value). It's pretty hard to use, because the standard way to consume generators (`for x of gen()`) skips it completely. So I think this is the normal case that's desired.

This is a breaking change, though, so worth thinking about. It's also an incompatibility with CoffeeScript (documented here). I guess we could have a compat flag... I just can't imagine many if any people have intentionally used this feature.

### DIFF
--- a/source/parser/function.civet
+++ b/source/parser/function.civet
@@ -118,16 +118,18 @@ function processReturn(f: FunctionNode, implicitReturns: boolean): void
     isMethod := f.type is "MethodDefinition"
     isConstructor := isMethod and name is "constructor"
     isVoid := (or)
-      isVoidType(returnType?.t)
+      generator
+      isVoidType returnType?.t
       (and)
         async
-        (or)
-          isPromiseVoidType(returnType?.t)
-          generator and isAsyncGeneratorVoidType(returnType?.t)
-      (and)
-        not async
-        generator
-        isGeneratorVoidType(returnType?.t)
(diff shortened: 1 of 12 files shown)


source/parser/function.civet
```diff
@@ -118,16 +118,18 @@ function processReturn(f: FunctionNode, implicitReturns: boolean): void
...
         async
-        (or)
-          isPromiseVoidType(returnType?.t)
-          generator and isAsyncGeneratorVoidType(returnType?.t)
-      (and)
-        not async
-        generator
-        isGeneratorVoidType(returnType?.t)
+        isPromiseVoidType returnType?.t
+        //(or)
```
STRd6, line 126: It's fine to delete these commented out lines.

STRd6 (approved):
:+1:

STRd6:
I think since this is such a fringe feature we don't need to worry about breaking changes / coffee compat but if someone comes up with a specific issue we can address it then.

Issue #1429, opened 2024-10-06
await and newlines


edemaine:
## Input

```js
foo:
  foo()
  |> await
```

(smaller version of an example posted by @bbrk24 on Discord)

## Current Output

```js
({foo:await 
  foo()})
```

I didn't realize that this is illegal, because `await` isn't a full keyword. Its argument can't come on a newline:

> `await` is parsed as a [keyword](https://tc39.es/ecma262/multipage/ecmascript-language-lexical-grammar.html#sec-keywords-and-reserved-words) of an [AwaitExpression](https://tc39.es/ecma262/multipage/ecmascript-language-functions-and-classes.html#prod-AwaitExpression) when the [Await] parameter is present. [[ref](https://tc39.es/ecma262/multipage/ecmascript-language-functions-and-classes.html#sec-async-function-definitions)]

## Desired Output

```js
({foo:await foo()})
//or
({foo:await(
  foo())})
```

Related, we don't currently support this, despite supporting similar things with e.g. `return`:

```js
await
  foo()
```

PR #1433, opened 2024-10-07
`await` with indented argument, or multiple arguments as array

edemaine:
As suggested in #1429 and #1431, adds support for

```js
await
  indented expression
↓↓↓
await (
  indented expression
)
```

Then as further suggested by @STRd6 on Discord, you can list multiple expressions too. They get bundled into an array:

```js
await
  e1
  e2
↓↓↓
[
  await e1,
  await e2]
```

Bulleted arrays behave the same, too, in case you think that looks nicer:

```js
await
  . e1
  . e2
```

This is in parallel to a previous way to use `await.op`. You can use those without bullets too:

```js
await.op
  e1
  e2
↓↓↓
await.op([
  e1,
  e2])
```

Note that there's a bit of asymmetry here: with a single argument, it's passed directly into `await.op`, in case it's an array.  More than one argument gets wrapped into an array. I think this is the most useful behavior, but I'm not certain.

### DIFF
--- a/civet.dev/reference.md
+++ b/civet.dev/reference.md
@@ -835,6 +835,20 @@ await.all
       fetch url |> await |> .json() |> await
 </Playground>
 
+Both `await` and `await` operators support an indented application form.
+Multiple arguments automatically get bundled into an array:
+
+<Playground>
+// Sequential
+await
+  first()
+  second()
+// Parallel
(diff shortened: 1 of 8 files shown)


STRd6 (approved):
Great work, thanks!

PR #1435, opened 2024-10-07
Typed `for` loops generate valid TypeScript

edemaine:
It turns out that the code generated by #1418 [generates a TypeScript error](https://www.typescriptlang.org/play/?#code/CYUwxgNghgTiAEA3W8CeAueAKARACxAggHsd4AfeHAd2JgmBwEoBtAXQG4BYAKADM62MMQB2AZwAu8AB6ZJMAJYiA5vGJ80TeAG9e8eMPHEIIAHQllWaUz3679-QHpHAPQD8vAL5A):

> The left-hand side of a 'for...of' statement cannot use a type annotation

TypeScript seems to ignore the annotation, so we need to compile to something else. This PR implements the obvious fix: redeclaring the variable inside the loop. We already had code to do this (for loops like `for {length}, value in foo`, where we needed to keep track of the original key and destructure inside the loop, so that we can find `value`).

Also adds documentation, cleans up a little code, and cleans up some AST to use `typeSuffix` instead of `suffix`.

### DIFF
--- a/source/parser/for.civet
+++ b/source/parser/for.civet
@@ -123,7 +123,7 @@ function processForInOf($0: [
   else if step
     throw new Error("for..of/in cannot use 'by' except with range literals")
 
-  let eachOwnError: ASTError | undefined
+  let eachOwnError: ASTError?
   let hoistDec, blockPrefix: ASTNode[] = []
 
   // for each item[, index] of array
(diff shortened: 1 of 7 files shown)


source/parser/for.civet
```diff
@@ -168,27 +168,55 @@ function processForInOf($0: [
...
+  // so pull such declarations inside the loop:
+  //   for var x: T of y -> for (const x1 of y) {var x: T = x1;
+  // for..in loops need a similar ref-based declaration
+  // for dereferencing to get the associated value.
+  { binding } := declaration
+  pattern .= binding?.pattern
+  if binding?.typeSuffix or (
+    inOf.token is "in" and declaration2 and pattern.type is not "Identifier"
+  )
+    keyRef := makeRef "key"
```
STRd6, line 189: Maybe we should just call it `"ref"` now since it may not actually be the key anymore.

test/types/for.civet
```diff
@@ -5,47 +5,87 @@ describe "[TS] for", ->
...
 
   describe ":: typing of properties", ->
     testCase """
       array
       ---
       for [type:: TokenType, token] of Object.entries(tokens)
+        console.log type, token
       ---
-      for (const [type, token]: [ TokenType,unknown] of Object.entries(tokens));
+      for (const key of Object.entries(tokens)) {const [type, token]: [ TokenType,unknown] = key;
```
STRd6, line 77: Having the `unknown` here may obstruct TS's inference but I don't know a better way off the top of my head. This is fine for now but we can revisit if we find a better way.

STRd6 (approved):

test/types/for.civet
bbrk24: `(typeof key)[1]` should work, right?

edemaine: That would work in this instance, but we're currently using a generic "`::` property types to `:` full type" conversion process. I don't think we generally know what we're assigning (e.g. function parameter), though we clearly do in some cases, and maybe we could adapt the derived type in such cases. I'm going to leave this for future work, though.

source/parser/for.civet
edemaine: I renamed the Civet variable to `itemRef` (to still distinguish from `expRef`, for the thing we're iterating over) and renamed the ref itself to `"key"` or `"item"` depending on whether we're using `for..in` or `for..of` respectively.

PR #1437, opened 2024-10-08
`await` with array literal awaits the items

edemaine:
This PR brings `await [x, y]` into alignment with
```js
await
  . x
  . y
```
and
```js
await
  x
  y
```

(but `await x, y` remains as is, for JS compatibility)

It's mostly a 1-word change (`NestedBulletedArray` → `ArrayExpression`), but I added a little more code to make the spacing come out better.

### DIFF
--- a/civet.dev/reference.md
+++ b/civet.dev/reference.md
@@ -849,6 +849,21 @@ await.all
   second()
 </Playground>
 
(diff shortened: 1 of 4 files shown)


STRd6 (approved):
:clap:

Issue #1280, opened 2024-06-10
Pipe after implicit object is incorrect


STRd6:
```civet
fetch(url)
  |> await
  |> .arrayBuffer()
  |> await 
  |> new Blob [.], type: "application/javascript"
  |> URL.createObjectURL
  |> new Worker
```

This has an unintuitive parse where the value of the implicit object becomes piped instead of the entire line:

```javascript
new Blob([(await (await fetch(url)).arrayBuffer())], {type: new Worker(URL.createObjectURL("application/javascript"))})
```

PR #1438, opened 2024-10-08
Binary operators continue arguments only if strictly indented

### LINKED ISSUE #1121: Binary operators starting on new lines

edemaine:
Fixes #1121 at least, as outlined in https://github.com/DanielXMoore/Civet/issues/1121#issuecomment-2331851911

#1280 shows that we also have the issue with the pipe operators, and this doesn't fix that yet. I think it could, with the same flag, but we'd first need to do `PushIndent`/`PopIndent` to recognize the indentation level of the first `|>` so that the future ones are recognized as being `Nested` and thus not valid continuations of an argument.

There *might* be other places where we need to `ForbidNestedBinaryOp`, but none come to mind yet... at least, this makes it easy to add more if necessary.

### DIFF
--- a/source/main.civet
+++ b/source/main.civet
@@ -49,6 +49,7 @@ uncacheable := new Set [
   "AllowBracedApplication"
   "AllowIndentedApplication"
   "AllowMultiLineImplicitObjectLiteral"
+  "AllowNestedBinaryOp"
   "AllowNewlineBinaryOp"
   "AllowTrailingMemberProperty"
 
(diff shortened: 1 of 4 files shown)


STRd6 (approved):
Nice, glad it ended up being so straightforward!

PR #1441, opened 2024-10-09
Forbid pipes in non-nested implicit arguments

### LINKED ISSUE #1280: Pipe after implicit object is incorrect

edemaine:
Fixes #1280 by adding a new `ForbidPipeline` flag, as discussed in #1440.

This new feature also let us unify `ArgumentList` and `NonPipelineArgumentList`. The former had gotten out-of-date, so I updated to match `NonPipelineArgumentList`, then removed the latter.

## Still some NonPipeline

I tried removing more `NonPipeline...` rules, but it's tricky specifically for parsing pipelines: `PipelineHead` needs to parse everything up to a `Pipe` operator (to avoid left recursion), but it doesn't work to use `ForbidPipeline` for this, because then it breaks code like

```js
=> x |> y
```

With `ForbidPipeline`, we'd end up parsing `=> x` as the head, which isn't what we want. But we also don't want to just `AllowPipeline` when we enter a non-indented function body, because then we'd misparse something like this:

```js
f => x
|> y
```

So it seems we need to keep (at least some of) the remaining 6 `NonPipeline...` rules. Most of them are short, at least. Oh, I could get rid of `NonPipelinePostfixedExpression` at least.

## AllowPipeline cases

I somewhat wonder whether `AllowPipeline` should be automatic with every `PushIndent`. But it's probably safer to control exactly when it's done, and add more cases as we encounter them. Right now, it should be pretty similar to the old behavior: whenever `NonPipelineArgumentList` used an indented rule, it dropped the `NonPipeline` prefix

### DIFF
--- a/source/main.civet
+++ b/source/main.civet
@@ -270,7 +270,6 @@ makeCache := ({hits, trace}: CacheOptions = {}): CacheEvents ->
     exit: (ruleName: string, state: ParseState, result: ParseResult) ->
       if uncacheable.has(ruleName)
         if trace
-          stack.pop()
           logs.push "".padStart(stack.length * 2, " ") + ruleName + ":" + state.pos + "⚠️ " + (if result then "✅" else "❌")
         return
 
--- a/source/main.civet
+++ b/source/main.civet
@@ -52,6 +52,7 @@ uncacheable := new Set [
   "AllowNestedBinaryOp"
   "AllowNewlineBinaryOp"
   "AllowTrailingMemberProperty"
+  "AllowPipeline"
 
   "ForbidClassImplicitCall"
   "ForbidBracedApplication"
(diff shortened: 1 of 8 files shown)


STRd6 (approved):
Thanks for this! The arguments and indentation handling are some of the most challenging parts of the parser. I appreciate the improvements here!

PR #1442, opened 2024-10-09
Assigned expressionized statement followed by pipe, grammar cleanup

### LINKED ISSUE #1439: `for` loop followed by pipe only allowed after newline
This is allowed:
```civet
t =
for [0..1]
  ;
|> foo
```
But this isn't:
```civet
t = for [0..1]
  ;
|> foo
```

edemaine:
Fixes #1439. The problem stemmed from `ExtendedExpression` calling `NonAssignmentExtendedExpression` which calls `ExpressionizedStatementWithTrailingCallExpressions` which doesn't allow for pipes.

What was confusing was that this version worked:
```js
t =
for [0..1]
  ;
|> foo
```
The reason is that the newline prevents `ExpressionizedStatementWithTrailingCallExpressions` from matching, so this rule gets completely ignored. And then everything works!

This suggested a radical solution: remove `NonAssignmentExtendedExpression` and related rules entirely. They're not actually needed, because `UnaryBody` directly calls `NestedNonAssignmentExtendedExpression` and `RHS` directly calls `ExpressionizedStatementWithTrailingCallExpressions`. (It looks like this redundancy began way back in #475.) So there was actually no way to get "unextended expressions" anymore. So I've gone ahead and renamed `*ExtendedExpression` to `*Expression`. And I renamed `NestedNonAssignmentExtendedExpression` to `NestedExpressionizedStatement` which is a more descriptive name. No more `NonAssignment` expressions (which was always confusing to me); now everything (including expressionized statements) is within `AssignmentExpression` as per ECMA spec.

I hope I didn't miss anything, but I think this makes sense: expressionized statements should not be handled at the top level, because they should be within pipelines. Plausibly, the same should be true for trailing call expressions, but I didn't succeed at my first attempt at replacing `ExpressionizedStatementWithTrailingCallExpressions` with `ExpressionizedStatement`, so I'll leave that for another time.

### DIFF
--- a/source/parser.hera
+++ b/source/parser.hera
@@ -230,31 +230,17 @@ TopLevelStatement
     statement = prepend(ws, statement)
     return [statement, delimiter]
 
-# Expressions with comma operator, or expressions with If and Switch
-ExtendedCommaExpression
-  NonAssignmentExtendedExpression
-  CommaExpression
-
 # Expressions with If and Switch, but no comma operator
-ExtendedExpression
-  NonAssignmentExtendedExpression
+Expression
   AssignmentExpression
 
-SingleLineExtendedExpression
-  NonAssignmentExtendedExpression
+SingleLineExpression
   SingleLineAssignmentExpression
 
-NonPipelineExtendedExpression
-  NonAssignmentExtendedExpression
+NonPipelineExpression
   NonPipelineAssignmentExpression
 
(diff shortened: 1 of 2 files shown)


STRd6 (approved):
I remember trying (and failing) to clean some of this up about a year ago. I'm very excited to see that it's doable now!

PR #1444, opened 2024-10-09
Nested function arguments support trailing member access/call

### LINKED ISSUE #869: Unexpected output with indenting function arguments
```coffee
v := (@x, @y) =>
  @length := () => Math.sqrt(@x**2 + @y ** 2)
  @

v
  v 2, 4
    .length()
  0.5
```
the second part compiles to 

```js
v(v(2, 4)).length()(0.5);
```

when I would expect it to become

```js
v(v(2,4).length(), 0.5);
```

Since if I would get that shape if I were to replace the inner `v` with a literal like

```coffee
v 
  2
  4
```
it would just be 

```js
v(2, 4)
```
It would also work if I didn't use indentation like and included parens, like

```coffee
v
  v(2, 4).length()
  0.5
```

or if I used a separate expression like

```coffee
l := v 2, 4
        .length()

v 
  l
  0.5
```

For these reasons it would be nice if the same expression inside or outside would work the same way.

edemaine:
Fixes #869 by re-allowing trailing member/call in *properly nested* arguments (where it's no longer ambiguous whether the trailing applies to the argument or the function result).

A breaking change is that this parse changes:

```js
f
  a
  .b
```

[Previous transpilation](https://civet.dev/playground?code=ZgogIGEKICAuYg%3D%3D):

```js
f(
  a,
  $ => $.b)
```

New transpilation:

```js
f(
  a
  .b)
```

I'm not totally sure which is more intuitive; this is a fundamental ambiguity we've had for a while. But perhaps a good analogy is this:

```js
=>
  a
  .b
```

This returns `a.b`, not `$ => $.b`. This PR matches that behavior in arguments too.

An alternative would be to do different things depending on whether `.b` is indented the same or further than `a`. This would be harder to implement, though, and I guess inconsistent with our not-in-arguments behavior.

### DIFF
--- a/source/parser.hera
+++ b/source/parser.hera
@@ -436,7 +436,7 @@ ArgumentList
     ]
 
 NestedArgumentList
-  PushIndent AllowPipeline NestedArgument*:args RestorePipeline PopIndent ->
+  PushIndent AllowPipeline AllowTrailingMemberProperty NestedArgument*:args RestoreTrailingMemberProperty RestorePipeline PopIndent ->
     if (!args.length) return $skip
     return args.flat()
 
--- a/source/parser/function.civet
+++ b/source/parser/function.civet
@@ -151,7 +151,7 @@ function processReturnValue(func: FunctionNode)
     // Check whether return.value already declared within this function
     { ancestor, child } := findAncestor
       value
-      .type is "Declaration"
+      &.type is "Declaration"
       isFunction
     declaration ??= child if ancestor  // remember binding
 
--- a/test/function-application.civet
+++ b/test/function-application.civet
@@ -572,3 +572,24 @@ describe "function application", ->
     }
       , 100)
   """
+
+  // #869
+  testCase """
+    trailing member in nested argument
+    ---
+    v
+      v 2, 4
+        .length()
+      0.5
+    f
+      x
+      .y
+    ---
+    v(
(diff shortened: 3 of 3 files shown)


edemaine:
It may be worth bumping the minor version number for this PR, because the Civet code itself had an example where the change mattered:

```js
    { ancestor, child } := findAncestor
      value
      .type is "Declaration"
      isFunction
```

I wrote this code, and intended to have three arguments. I don't mind adding a `&` here, but probably also good to signal to users that you might need to update your code...

```js
    { ancestor, child } := findAncestor
      value
      &.type is "Declaration"
      isFunction
```

bbrk24:
Would it be feasible to only do this if it's more indented? i.e.
```js
f
  a
  .b
// vvv
f(a, $ => $.b)



f
  a
    .b
// vvv
f(a.b)
```
?

edemaine:
@bbrk24 That's exactly my "alternative" at the end of the original post. It did come to mind, in particular because the example in #869 was strictly intended. The downside is that this would be harder to implement, and inconsistent with our not-in-arguments behavior... But if we think it's worth doing, we can.

STRd6 (approved):
I think resolving this in favor of consistency with the non-argument version is a good way to do it.

Thanks again!

PR #1446, opened 2024-10-10
Fix `catch` type annotation by forbidding indented type arguments in `catch`

### LINKED ISSUE #1445: Cannot type caught error as `any`
```
try
  foo()
catch e: any
  bar(e)
```
```ts
try {
  foo()
}
catch (e: any<{
  bar(e)
}>) {}
```

edemaine:
Fixes #1445


STRd6 (approved):
👍

Would be nice to add a test case.

edemaine:
Oops, my usual mistake of forgetting to `git add` 😅

Issue #1195, opened 2024-04-27
Top-level `comptime` block is always `undefined` in the REPL


bbrk24:
While testing a local change to comptime serialization:
```
$ ./dist/civet --comptime
Civet 0.7.1 REPL.  Enter a blank line to execute code.
🐱> comptime do
...   f := &
...   f.a = 1
...   f
... 
undefined
🐱> f := comptime do
...   f := &
...   f.a = 1
...   f
... 
undefined
🐱> f
... 
[Function (anonymous)] { a: 1 }
🐱> 
```

I would have expected the first one to also be `[Function (anonymous)] { a: 1 }`, but because top-level comptime blocks are deleted, there was nothing to evaluate and the result was `undefined`.

edemaine:
This is a general question I guess for what the REPL should do for top-level statements that can be expressionized. Currently we do not expressionize, but perhaps we should in the REPL? Compare these examples:
```
🐱> i .= 5
... while i < 10
...   i++
...
9
🐱> do
...   i .= 5
...   while i < 10
...     i++
...
9
🐱> function f
...   i .= 5
...   while i < 10
...     i++
... f()
...
[ 5, 6, 7, 8, 9 ]
```

Those `9` return values kind of magical. I guess it's somehow what `eval` (or the Node equivalents that we use) defines as the last expression, which is different how Civet defines it.

So I think it'd be nice to wrap the top level in an IIFE at the Civet level so the last Civet expression gets returned to the REPL. Kind of an extension of the code I added recently to handle top-level `await` in this way. Amusingly, I think the easy way to wrap in an IIFE is `async do`, because `do` is "too smart" and only uses braces if the block is top-level so won't be returned. So it'd be `async do` in all cases, and `await` the result, even if there aren't any top-level `await`s (but just like we currently do if there were).

edemaine:
Actually, I think this might be an issue worth extending beyond the REPL. There are settings where we write scripts where the last expression is meaningful. `eval` is one example. My [SVG Tiler](https://github.com/edemaine/svgtiler) software is another; though `export`s are encouraged, mapping files can also be the last expression of the file.

[...]

PR #1448, opened 2024-10-10
CLI exposes top-level declarations with top-level `await`

edemaine:
I was thinking about https://github.com/DanielXMoore/Civet/issues/1195#issuecomment-2081269154, and noticed that the CLI had two bugs when running code with top-level `await` (where it builds an IIFE wrapper):
* It no longer correctly found the `BlockStatement` of the IIFE, instead getting distracted by the (relatively new) `BlockStatement` at the very top of the AST and the extra braced block wrapper.
* The search for `BlockStatement` was also broken in `coffee-do` mode.
* All declarations not wrapped in functions were lifted, instead of just the top-level ones (not inside a nested block). Should now match the non-async behavior (except `const` becomes `var`).

Tests:

image

This probably wants to be a separate function, with tests, but at least it's working for now. I'm hoping to add this as a less hacky feature in the future.


STRd6 (approved):
👍

Issue #837, opened 2023-12-07
[Feature request] An argument placeholders for callbacks


mixty1:
It would be nice, if you add to civet something like [this](https://imba.io/docs/basic-syntax/functions#calling-functions-callbacks)

I have this code:
```coffee
export default function throttle(func: (...args: any) => () => void, ms: number)

  isThrottled .= false
  savedArgs: any | null .= null
  savedThis: any | null .= null

  function wrapper(this: any): void
    if isThrottled
      savedArgs = arguments
      savedThis = this
      return

    func.apply this, arguments

    isThrottled = true

    setTimeout =>
      isThrottled = false
      if savedArgs
        wrapper.apply savedThis, savedArgs
        savedArgs = savedThis = null
    ,ms

  wrapper
```

And I would like to have such an opportunity as:
```coffee
export default function throttle(func: (...args: any) => () => void, ms: number)

  isThrottled .= false
  savedArgs: any | null .= null
  savedThis: any | null .= null

  function wrapper(this: any): void
    if isThrottled
      savedArgs = arguments
      savedThis = this
      return

    func.apply this, arguments

    isThrottled = true

    setTimeout(&, ms) =>
      isThrottled = false
      if savedArgs
        wrapper.apply savedThis, savedArgs
        savedArgs = savedThis = null

  wrapper
```

```coffee
someFunc(&, arg1, arg2, &, arg3)
  (argsFromFirstPlaceholder) =>
    // ...some code here
  (argsFromSecondPlaceholder) =>
    // ...some code
```

mixty1:
> It looks like imba also only supports one argument?

Probably yes. By the way, if i'm not mistaken `swift` lang has similar syntax, but without placeholders, just callback after function call or something like that.



> So what this suggests is that `someFunc(&, arg1, arg2, &, arg3)` should map to `($1, $2) => someFunc($1, arg1, arg2, $2, arg3)`

With defined default parameter with callback functions:
```js
($1 = (argsFromFirstPlaceholder) => { /* ...some code here */ }, $2 = (argsFromSecondPlaceholder) => { /* ...some code */ }) => someFunc($1, arg1, arg2, $2, arg3)
```

from this code:
```coffee
someFunc(&, arg1, arg2, &, arg3)
  (argsFromFirstPlaceholder) =>
    // ...some code here
  (argsFromSecondPlaceholder) =>
    // ...some code
```



[...]

PR #1450, opened 2024-10-10
Fix partial placeholder `.` in middle of call chain

### LINKED ISSUE #1449: `.` placeholder lifts outside long string of calls
[The docs](https://civet.dev/reference#partial-function-application) define `.` placeholders as follows:

> if you use `.` within a function call, that call gets wrapped in a one-argument function and `.` gets replaced by that argument

It looks like what actually happens is that `.` lifts outside the entire `CallExpression`, no matter how long it is.  For example:

## Input

```js
f(a)(b)(.)(c)(d)
```

## Current Output

```js
$ => f(a)(b)($)(c)(d)
```

## Expected Output

```js
($ => f(a)(b)($))(c)(d)
```

I think.  The idea is that this expression should be treated as a tree of applications, and this is what you get if you lift the `.` just above its own application.

https://github.com/DanielXMoore/Civet/issues/837#issuecomment-2405726674 gives a [real-world example](https://civet.dev/playground?code=c2V0VGltZW91dCguLCAxMDAwKQogID0%2BIGNvbnNvbGUubG9nICdoZWxsbyB3b3JsZCc%3D).

edemaine:
Fixes #1449

### DIFF
--- a/source/parser/function.civet
+++ b/source/parser/function.civet
@@ -1,6 +1,6 @@
 import type {
+  ArrowFunction
   ASTNode
-  ASTNodeBase
   ASTNodeObject
   BlockStatement
   BreakStatement
@@ -11,11 +11,11 @@ import type {
   IterationExpression
   IterationFamily
   IterationStatement
+  NewExpression
   Parameter
   ParametersNode
   StatementTuple
   SwitchStatement
-  TypeIdentifier
   TypeNode
   Whitespace
 } from ./types.civet
@@ -29,6 +29,10 @@ import {
 import {
   gatherBindingCode
 } from ./binding.civet
+
+import {
+  replaceNode
+} from ./lib.civet
 
 import {
   findAncestor
@@ -806,6 +810,29 @@ function makeAmpersandFunction(rhs: AmpersandBlockBody): ASTNode
 
   fn
 
+/**
+Split a CallExpression node in half, where the inner half ends with child
+number given by `index`.  Returns the inner call.
+*/
+function splitCall(call: CallExpression, index: number): CallExpression | NewExpression
+  subcall: CallExpression := makeNode
+    type: "CallExpression"
+    children: call.children[0..index]
+  call.children[0..index] = [subcall]
+  subcall.parent = call
+  { parent } := call
+  // Pull preceding `new` into inner call to preserve semantics
+  if parent is like {type: "NewExpression", expression: ^call}
+    replaceNode parent, call
+    parent.parent = call
+    parent.expression = subcall
+    parent.children = parent.children.map & is call ? subcall : &
+    subcall.parent = parent
+    call.children[0] = parent
+    parent
+  else
+    subcall
+
 export {
   assignResults
   expressionizeIteration
@@ -815,5 +842,6 @@ export {
   processFunctions
   processReturn
   skipImplicitArguments
+  splitCall
   wrapIterationReturningResults
 }
--- a/source/parser/lib.civet
+++ b/source/parser/lib.civet
@@ -71,6 +71,7 @@ import {
   prepend
   stripTrailingImplicitComma
   trimFirstSpace
+  updateParentPointers
   wrapIIFE
   wrapWithReturn
 } from ./util.civet
@@ -110,6 +111,7 @@ import {
   processCoffeeDo
   processFunctions
   skipImplicitArguments
+  splitCall
 } from ./function.civet
 import { processPatternMatching } from ./pattern-matching.civet
 import {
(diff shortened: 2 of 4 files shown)


STRd6:
I have mixed feelings about this one... I think it could be handy to lift above trailing call expressions much lifting above.

It's easy enough to short circuit the chain by adding parens:

```
(f(.))(b)
```

But I don't know how we'd add in the chain easily if we switch it up. I could be convinced but I'd need to see some more examples.

edemaine:
I can't think of many situations where I want to do multiple calls in a chain. Currying is the obvious example. The current behavior might be useful when currying:

```js
add := (x) => (y) => x+y
leftSpace := add(' ')(.) // works with or without PR
rightSpace := add(.)(' ') // relies on current behavior, not PR
```

On the other hand, I think partial placeholders are designed to replace the need for currying in the first place:

```js
add := (x, y) => x+y
leftSpace := add(' ', .) // works with or without PR
rightSpace := add(., ' ') // works with or without PR
```

Here are some arguments for why I think the PR behavior is better:

* Ability to specify an argument later, as in https://github.com/DanielXMoore/Civet/issues/837#issuecomment-2405726674 : `setTimeout(., 1000) => // here's the callback`
* Intuitively, I think we always envisioned `.` (or `...`) as being placeholders (blanks) *for this specific call*. To do the call, we of course need to grab the function being called (all calls to the left in the chain), but I think it's counterintuitive to also grab things to the right of this call...
* Related, it's easier to explain. To describe the current behavior, the documentation would have to be changed to the following (***added text***), which seems harder to understand:
  > if you use `.` within a ***chain of*** function calls, that call ***chain*** gets wrapped in a one-argument function and `.` gets replaced by that argument

Admittedly, the examples are thin. If you can think of others to help decide, I'd love to see them!

STRd6:
Thanks for the explanation, I'm mostly convinced, however there is one other situation I think needs to be considered:

```
// Currently .normalize() is applied inside the wrapper
y := f(.).normalize()
// Alternative that works with or without this change
x := & |> f |> .normalize()
```

Since there are very few properties of a function that are useful to access it is more generally useful to have the member expressions applied inside.

By a similar argument we may even want to expand to the left further as well:

```
// currently not very useful (await and binary exps are not consumed by the placeholder function)
y := 1 + await f .
---
const y = async $ => 1 + await $
```

We could even extend to handle:

```
// currently needs to use `&` for the placeholder, `.` is a parse error
f := 1 + .
```

If we do `.` and `&` would be synonymous in this case ( `.` could expand up to the enclosing function, statement, or declaration level)

edemaine:
Hmm, nice examples. I'm actually surprised that `await` isn't brought already inside the function, as other unary prefix operators (e.g. `-`) are.

I guess this suggests an alternative interpretation for `.` which might be more useful:
* `&` lifts to *just inside* the nearest function call, assignment, pipeline, `return`, or `yield`
* `.` lifts to *outside* the nearest function call, **and from there, lifts further to *just inside* the nearest function call, assignment, pipeline, `return`, or `yield`**

The bold part is the new part compared to the current definition.
I'm almost tempted to call this new proposal `&&` (with more `&`s to pull outside successive function call levels).
(Related, I've found `.` to always look a little gross when used in some contexts, like `(.).foo` (thankfully `..foo` doesn't work).)

Maybe we could stick to `.` for partial application (just omitted arguments in functions), generalize it to handle multiple arguments, and introduce `&&` to lift further up?

STRd6:
> * `&` lifts to _just inside_ the nearest function call, assignment, pipeline, `return`, or `yield`
> * `.` lifts to _outside_ the nearest function call, **and from there, lifts further to _just inside_ the nearest function call, assignment, pipeline, `return`, or `yield`**

I think this is a good interpretation with the possible addition of _statement_ and _parenthesized expression_ as additional places that the lifting halts.

I'm not very much of a fan of `&&` for lifting additional levels, mainly because at `&&&` it would already become unreasonable to me.

For multiple argument partial application I've been considering `\1, \2, ...`

edemaine:
Ooh, parenthesized expression sounds nice a nice place to stop lifting. Currently the lift just rips through that, which I think is counterintuitive and makes it hard to stop the lift (I usually use a `do` wrapper).

(We already stop at statements, as previously discussed, but it should be documented!)

I'm still uncertain whether `.` is intuitive for going up a "full second level"... maybe. I still learn toward some other notation, and keeping `.` for the original intent of lifting just outside a call... Related:
* Do you think it's worth trying to support "delayed arguments" like `setTimeout(., 1000) => callback` with some notation (`.` or otherwise)?
* I think that, while `f(.).normalize()` is cool and useful, this specific notation isn't especially intuitive... (though maybe it's just my mental model of `.` needs to change)
* I expect `f(.)` to mean `$ => f($)` i.e. a one-argument version of `.` (useful in e.g. `array.map f(.)` to shield `f` from the other arguments). Admittedly it's not that useful to call a method on a function, but it's still surprising...

edemaine:
I just found myself writing the following code (here `&` is of type `{about?: string, code: string}`):
```js
transpileds.map((&.about ?? '')# + &.code#)
```
This wouldn't work if we stop `&` at a parenthesized expression. Which behavior do we want?

STRd6:
Good point, maybe we should continue through parentheses. I can see it coming up in math equations pretty regularly.

PR #1461, opened 2024-10-13
Add missing parentheses with `++` concat operator

### LINKED ISSUE #1456: `++` binds less tightly than `as`
```ts
x := arr as string[] ++ y
// vvv
const x = arr as string[].concat(y)
```

edemaine:
Fixes #1456

Also removed some unnecessary parens around array expressions.

~~I'm not a fan of the extra parens in `x ++ y ++ z`: it previously compiled to `x.concat(y).concat(z)`, and now compiles to `(x.concat(y)).concat(z)`. But I didn't see an easy way to fix, because the binary expression precedence handler doesn't construct "correct" AST nodes.~~  I fixed some of the AST node types to fix this.

### DIFF
--- a/source/parser/util.civet
+++ b/source/parser/util.civet
@@ -441,6 +441,7 @@ function removeHoistDecs(node: ASTNode): void
 
 skipParens := new Set [
   "AmpersandRef"
+  "ArrayExpression"
   "CallExpression"
   "Identifier"
   "JSXElement"
@@ -459,6 +460,8 @@ skipParens := new Set [
  * (Consider renaming and making parentheses depend on context.)
  */
 function makeLeftHandSideExpression(expression: ASTNode)
+  while [ item ] := expression
+    expression = item
   if isASTNodeObject expression
     return expression if expression.parenthesized
     return expression if skipParens.has expression.type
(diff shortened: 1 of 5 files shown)


source/parser/util.civet
```diff
@@ -459,6 +460,8 @@ skipParens := new Set [
  * (Consider renaming and making parentheses depend on context.)
  */
 function makeLeftHandSideExpression(expression: ASTNode)
+  while [ item ] := expression
```
STRd6, line 463: Should we check `isArray` and `length is 1` as well?

STRd6 (approved):

source/parser/util.civet
edemaine: You [already did that](https://civet.dev/playground?code=d2hpbGUgWyBpdGVtIF0gOj0gZXhwcmVzc2lvbg%3D%3D). 😺 

```js
while [ item ] := expression
↓↓↓
function len<
  T extends readonly unknown[],
  N extends number,
>(arr: T, length: N): arr is T & { length: N } {
  return arr.length === length;
}
while (
  expression &&
  Array.isArray(expression) &&
  len(expression, 1)
) {
  const [item] = expression;
}
```

STRd6: Sometimes I spend too much time in JS land and I forget the magic of Civet :cat:

PR #1462, opened 2024-10-13
Pipeline doesn't unwrap in multi-`&` function shorthand

### LINKED ISSUE #1172: Pipeline unwrapping should add refs
```coffee
x() |> &+&
```

currently compiles to

```js
x()+x()
```

We should instead add a `ref` here.

(Previously, it wasn't possible to use the `&` multiple times. Now it is.)

edemaine:
Fixes #1172

I opted to keep the ampersand function a function here, instead of unwrapping and using refs, because:
* it was easier to implement
* it seems slightly closer to the input

### DIFF
--- a/source/parser/function.civet
+++ b/source/parser/function.civet
@@ -1,6 +1,5 @@
 import type {
   ASTNode
-  ASTNodeBase
   ASTNodeObject
   BlockStatement
   BreakStatement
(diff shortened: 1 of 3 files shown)


STRd6 (approved):
Looks good! We'll probably want to use refs for this later but this is a step in the right direction.

PR #1463, opened 2024-10-14
Compiler directives `iife` and `repl` for wrapping program, fix `comptime` implicitly returned from function

### LINKED ISSUE #1195: Top-level `comptime` block is always `undefined` in the REPL

edemaine:
Fixes #1195 as shown here:

image

More generally, I followed the idea in https://github.com/DanielXMoore/Civet/issues/1195#issuecomment-2081269154 to add two general mechanisms for wrapping programs in IIFEs:
* `"civet iife"` adds a wrapper similar to CoffeeScript, except that we also support `import` (via dynamizing `import`). This can be useful when running non-module scripts in the browser (to avoid `var` becoming a global) or when concatenating CJS scripts together (to avoid declarations from earlier scripts from being visible to later scripts). But in particular a step toward:
* `"civet repl"` additionally pulls top-level declarations out, like we had a ton of gross code to do in both the CLI and Playground. Now it's just adding an option and we get exactly the code we need to `eval`!

This approach has an added benefit of correctly returning the last value of a code block from Civet's perspective. For example:

```
🐱> i .= 5
... while i < 10
...   i++
...
[ 5, 6, 7, 8, 9 ]
🐱> do
...   i .= 5
...   while i < 10
...     i++
...
[ 5, 6, 7, 8, 9 ]
```

By contrast, these examples previously returned `9` (see https://github.com/DanielXMoore/Civet/issues/1195#issuecomment-2081268241 ).

To finish #1195, I noticed another problem: `=> comptime 1+2` [crashed](https://civet.dev/playground?code=PT4gY29tcHRpbWUgMSsy) when `comptime: true`. It turns out that `comptime` got expressionized (wrapped in its own IIFE) only when it was used in a true expression context, like `1 + comptime 2+3`. When it was implicitly returned by a function, or added to a loop results array, it didn't work. This was particularly annoying in `"civet repl"` mode because we automatically get a function wrapper. The second commit fixes this bug.

### DIFF
--- a/civet.dev/public/playground.worker.js
+++ b/civet.dev/public/playground.worker.js
@@ -43,36 +43,11 @@ onmessage = async (e) => {
 
   let jsCode = '';
   if (jsOutput) {
-    // Wrap in IIFE if there's a top-level await or import
-    // Use Civet's async do, so Civet does implicit return of last value
     try {
-      const topLevelAwait = Civet.lib.hasAwait(ast) ||
-        Civet.lib.hasImportDeclaration(ast)
-      if (topLevelAwait) {
-        const [prologue, rest] = Civet.parse(code,
-          {startRule: 'ProloguePrefix'})
-        const prefix = code.slice(0, -rest.length)
-        const coffee = prologue.some((p) => p.type === "CivetPrologue" &&
-          (p.config.coffeeCompat || p.config.coffeeDo))
-        ast = await Civet.compile(
-          prefix +
-          (coffee ? '(do ->\n' : 'async do\n') +
-          rest.replace(/^/gm, ' ') +
-          (coffee ? ')' : ''),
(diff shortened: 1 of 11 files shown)


STRd6 (approved):
Nice!

PR #1466, opened 2024-10-14
`repl` directive hoists function and class declarations too, fix one-line declarations in `if`

edemaine:
Follow-up to #1463 to also hoist `function` and `class` declarations like JavaScript does:
* Top-level `function`s gets declaration and value hoisted to outside, by physically moving the declaration
* Inner `function`s get declaration hoisted outside but value remains where it is.
* `class` declarations get hoisted like `let`: only top-level get hoisted.
* Also fixed: inner `var` declarations get hoisted to outside.

As a result, the CLI behaves like this:

```
🐱> f()
... function f() console.log 'hi'
...
hi
[Function: f]
🐱> f()
...
hi
undefined
🐱> if true
...   function g() console.log 'yes'
... else
...   function g() console.log 'no'
... g()
...
yes
undefined
🐱> g()
...
yes
undefined
🐱> class C
...
[class C]
🐱> new C
...
C {}
```

A necessary step to get here was to clean up how declarations and functions get implicitly returned. Before, we added a `ReturnStatement` *inside* the declaration/function. Now it's added to the parent block, which makes more sense. Also, I realized we weren't bracing the block, leading to [the wrong thing getting returned](https://civet.dev/playground?code=PT4gaWYgKHgpIGZ1bmN0aW9uIGYoKQ%3D%3D) in examples like this:

```js
=> if (x) function f()
↓↓↓ old
() => { if (x) function f(){};return f;return }
↓↓↓ new
() => { if (x) { function f(){};return f};return }
```

### DIFF
--- a/source/parser/lib.civet
+++ b/source/parser/lib.civet
@@ -1252,31 +1252,7 @@ function processProgram(root: BlockStatement): void
   adjustAtBindings(statements)
 
   // REPL wants all top-level variables hoisted to outermost scope
-  if config.repl
-    topBlock :=
-      gatherRecursive(rootIIFE!, .type is "BlockStatement")[0]
-    i .= 0
-    // Hoist top-level declarations and all var declarations
-    for each decl of gatherRecursiveWithinFunction topBlock, .type is "Declaration"
-      if decl.parent is topBlock or decl.decl is "var"
-        decl.children.shift() // remove const/let/var
-        root.expressions.splice i++, 0, ["", `var ${decl.names.join ','};`]
-    // Hoist all function declarations, and hoist values when top-level
-    for each func of gatherRecursive topBlock, .type is "FunctionExpression"
-      if func.name and func.parent?.type is "BlockStatement" // declaration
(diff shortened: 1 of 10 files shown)


source/parser/lib.civet
```diff
@@ -1268,8 +1263,34 @@ async function processProgramAsync(root: BlockStatement): Promise<void>
   { expressions: statements } := root
   await processComptime(statements)
 
+function processRepl(root: BlockStatement, rootIIFE: ASTNode): void
+  topBlock :=
+    gatherRecursive(rootIIFE!, .type is "BlockStatement")[0]
+  i .= 0
+  // Hoist top-level declarations and all var declarations
+  for each decl of gatherRecursiveWithinFunction topBlock, .type is "Declaration"
+    if decl.parent is topBlock or decl.decl is "var"
```
STRd6, line 1272: We should add a test to make sure we don't hoist `vars` that are inside function scopes..

STRd6 (approved):

PR #1467, opened 2024-10-14
One-line braced blocks with single-line statements

### LINKED ISSUE #1465: One-line braced blocks
This [doesn't compile](https://civet.dev/playground?code=Y2xhc3MgQyB7IHZhbHVlID0gNSB9Cg%3D%3D):
```js
class C { value = 5 }
```

Neither does this:
```js
if (x) { y = 5 }
```

The reason is that `ExplicitBlock` and the first rule of `ClassBody` require items to be `Nested`. I guess we should allow a single unnested item as an alternative.

edemaine:
Fixes #1465 by allowing for single line of statements inside one-line braces in `class`es and `if`/`else` blocks.

For `if`/`else` blocks, I could actually make the grammar a bit smaller by using some existing rules we had for `BracedContent` (from a different branch of the tree where blocks always need to be braced, such as function bodies).

### DIFF
--- a/civet.dev/comparison.md
+++ b/civet.dev/comparison.md
@@ -168,6 +168,22 @@ if (condition)
 By contrast, in JavaScript, the last indented line would have been
 treated as if it were at the top level.
 
+## Braced Blocks
+
+Civet allows you to wrap your blocks in braces (or not),
+but to avoid conflicts with braced object literals,
+there are some restrictions.
+In particular, a one-line braced block must be on the same line
+as the `if` etc. it's part of:
+
+<Playground>
+// block
+if (x) { y }
+// not a block
+if (x)
+{ y }
+</Playground>
+
 ## Automatic Semicolon Insertion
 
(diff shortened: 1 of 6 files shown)


edemaine:
Looks like this still needs some work: `{ foo } = bar` inside a `do` block is now just parsing the `{ foo }` as a block. I guess we got lucky so far, and now we need to disambiguate object literals from blocks in yet another situation.

edemaine:
Hmm, actually, we don't want to prioritize object literals in this case. Otherwise, `function f() {}` would return an empty object.

edemaine:
I resolved the issue by allowing a one-line braced block like `{ x = 5 }` only when it starts on the same line, such as `if (x) { x = 5 }`. This is still allowed:

```js
if (x)
{
  x = 5
}
```

(I'm not *totally* sure it should be, but it does seem good for JS compatibility. If you want to return the object literal right after an empty `if`, I guess you add a `return`, or an explicit empty block.)

This is now ready for review.

STRd6:
I agree that we should allow the braced block on the following line for JS compatability. It's a less common style but it's not  unheard of.

STRd6 (approved):

PR #1468, opened 2024-10-14
Forbid next-line braced blocks after `return if` and `yield if`

edemaine:
During #1467, I noticed that our implicit function application magic did the following:

```js
f(x) if x
{
  x
}
↓↓↓
if (x) { f(x) }
({
  x
})
//not:
f(x)(x ? {x} : undefined)
```

I think this is reasonable, because we only need to implement braced-blocks backward compatibility with natural JS constructs, and `if` expressions are not such a construct.

This PR extends this functionality for other common patterns with `yield` and `return`:

```js
return if x
{
  x
}
↓↓↓
if (x) { return }
({
  x
})
//as opposed to the current behavior:
return x ? {x} : undefined
```

This brings argumentless `return`/`yield` into alignment with having an argument:

```js
return x if x
{
  x
}
↓↓↓ (before and after this PR)
if (x) { return x }
({
  x
})
```

### DIFF
--- a/source/parser.hera
+++ b/source/parser.hera
@@ -2650,7 +2650,7 @@ EmptyBareBlock
 
 NoBlock
   # Check that there isn't a block here. Used for empty loop bodies.
-  &( EOS ) !IndentedFurther !( Nested Then )
+  &EOS !IndentedFurther !( Nested Then )
 
 # A nonempty block that must include braces
 # This version allows same-line postfixes like `while cond`.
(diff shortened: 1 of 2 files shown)


source/parser.hera
```diff
@@ -2650,7 +2650,7 @@ EmptyBareBlock
 
 NoBlock
   # Check that there isn't a block here. Used for empty loop bodies.
-  &EOS !IndentedFurther
+  &( EOS ) !IndentedFurther !( Nested Then )
```
STRd6, line 2653: These first parens seem unnecessary.

STRd6 (approved):
Looks good, thanks!

PR #1476, opened 2024-10-19
Ranges and slices with inequalities such as `[a<..<b]`

edemaine:
Fixes #999 by allowing inequalities in `..` ranges/slices, inspired by [Swift's `a..<b` half-open range operator](https://developer.apple.com/documentation/swift/range):
* Each endpoint can be marked `<` or `<=` (or the Unicode equivalent, `≤`) for increasing loops, or `>` or `>=`/`≥` for decreasing loops.
* Inclusive (as in `<=` or `>=`) is the default behavior, but it forces a loop order, which can be useful.
* Slices forbid `>`, as we don't have a nice way to reverse slice (when dealing with strings).

Because we can now specify direction, the ranges and loops can be a bit more efficient (as well as more consistent/reliable). Compare the old `...`:

```js
[0...array#]
for i of [0...array#]
↓↓↓
((s, e) => {
  let step = e > s ? 1 : -1;
  return Array.from(
    { length: Math.abs(e - s) },
    (_, i) => s + i * step,
  );
})(0, array.length);
for (
  let end = array.length, i1 = 0, asc = 0 <= end;
  asc ? i1 < end : i1 > end;
  asc ? ++i1 : --i1
) {
  const i = i1;
}
```

versus the new `..<`:

```js
[0..<array#]
for i of [0..<array#]
↓↓↓
((s) =>
  Array.from(
    { length: array.length - s },
    (_, i) => s + i,
  ))(0);
for (
  let end = array.length, i1 = 0;
  i1 < end;
  ++i1
) {
  const i = i1;
}
```

### DIFF
--- a/civet.dev/reference.md
+++ b/civet.dev/reference.md
@@ -417,7 +417,7 @@ indices := [0...array.length]
 
 An infinite range `[x..]` is supported when [looping](#range-loop).
 
-Alternatively, you can explicitly exclude endpoints with `<` or `>`,
+Alternatively, `..` can explicitly exclude endpoints with `<` or `>`,
 or include endpoints with `<=` or `>=`. These also control loop direction,
 whereas `[a..b]` determines direction based on whether `a <= b`.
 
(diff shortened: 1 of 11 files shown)


STRd6 (approved):

STRd6:
Thanks, looks like a nice addition!

PR #1477, opened 2024-10-20
Pattern matching `catch`

### LINKED ISSUE #1473: Pattern Catching
Been discussed on the Discord before but I didn't see an issue

Prior art includes Swift [`catch`](https://docs.swift.org/swift-book/documentation/the-swift-programming-language/guidedtour#Error-Handling) and C# [`catch`-`when`](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/when#when-in-a-catch-clause)

edemaine:
Fixes #1473 by following Plan 2 from https://github.com/DanielXMoore/Civet/issues/1473#issuecomment-2420337249, which seemed the most versatile.

This PR early-rewrites multiple `catch` blocks into a `catch(e) { switch(e)` with corresponding pattern matching, so any extensions to pattern matching will automatically apply.

### DIFF
--- a/build/build.sh
+++ b/build/build.sh
@@ -17,7 +17,7 @@ civet --no-config build/esbuild.civet "$@"
 
 # built types
 for name in astro esbuild rollup unplugin vite webpack; do
-  sed 's/\.civet"/\.js"/' dist/unplugin/source/unplugin/$name.civet.d.ts >dist/unplugin/$name.d.ts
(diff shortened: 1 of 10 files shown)


STRd6 (approved):
Nice! I meant to comment on that issue earlier but this is essentially the approach that I was thinking about.

:clap:

PR #1478, opened 2024-10-20
Reverse slices and inequality slices with implicit parts

edemaine:
This was mostly for fun, and for code golfers to be able to reverse a string with `x[..>=]`. It ended up a little bigger than I expected, to get the many cases working...

* Slices remain increasing by default (even without `<`/`<=` modifier)
* With explicit `>` or `>=`, slices are reversed, similar to a range, so normally left ≥ right. For example, `x[..>=]` is equivalent to `x[-1 ..>= 0]` which is equivalent to `x[x.length - 1 ..>= 0]`
* Also fix increasing slices with implicit parts, like `x[..<]`

### DIFF
--- a/civet.dev/reference.md
+++ b/civet.dev/reference.md
@@ -451,6 +451,12 @@ Alternatively, you can exclude or include endpoints using `..` with `<` or `<=`:
 strict := numbers[first<..<last]
 </Playground>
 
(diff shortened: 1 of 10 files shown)


STRd6 (approved):
Golfers are going to go nuts over `x[..<=]`

PR #1479, opened 2024-10-20
Indented types after binary type operators

edemaine:
Allow indented type after a binary operator, avoiding `__` in this case. This is important because otherwise implicit type arguments can apply.

Context: I noticed that `source/types.civet` wasn't parsing correctly. The relevant snippet is now a test case.


STRd6 (approved):
Good to see the types section catching up to the better whitespace handling we did in the rest of the parser.

PR #1484, opened 2024-10-21
`from ... import/export` (reversed `import`/`export`)

### LINKED ISSUE #800: Python-style imports?
I'm not sure if that's possible, just been reading the reference, and realized that it would be a great development experience, when one could write
```
from "module" import { a, b}
```

thus IDE would be able to tell us what is there to import

edemaine:
Fixes #320
Fixes #800

Also cleanup some old `import`/`export` rules (merging `type`/no `type` for regular `import`, and merging explicit/implicit `import` in `operator` case)

### DIFF
--- a/civet.dev/reference.md
+++ b/civet.dev/reference.md
@@ -2219,6 +2219,8 @@ class Civet <: Animal, Named
 
 ### Decorators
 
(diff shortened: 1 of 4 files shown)


STRd6 (approved):
Looks good!

Maybe add a test case where we mix and match the `from` first / `from` last styles to see what happens.

PR #1494, opened 2024-10-22
preventExtensions on comptime functions

### LINKED ISSUE #1493: comptime serializer doesn't freeze functions
```
x := comptime
  f := function()
    5
  f.a = 6
  return Object.freeze f
```
```js
const x = Object.defineProperties(
  function () {
    return 5;
  },
  {
    a: {
      value: 6,
      writable: false,
      enumerable: true,
      configurable: false,
    },
  },
);
```

bbrk24:
Fixes #1493

### DIFF
--- a/source/parser/comptime.civet
+++ b/source/parser/comptime.civet
@@ -161,6 +161,8 @@ function serialize(value: ???, context?: NodeJS.Dict<any>): string
           // builtin, or returned from Function::bind
           throw new TypeError "cannot serialize native function"
         if /^class[\s{]/u.test string
+          if not Object.isExtensible val
+            string |>= `Object.preventExtensions(${&})`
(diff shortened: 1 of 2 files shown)


source/parser/comptime.civet
```diff
@@ -161,6 +161,8 @@ function serialize(value: ???, context?: NodeJS.Dict<any>): string
           // builtin, or returned from Function::bind
           throw new TypeError "cannot serialize native function"
         if /^class[\s{]/u.test string
+          if not Object.isExtensible val
```
edemaine, line 164:
```suggestion
          unless Object.isExtensible val
```

(also on line 203 and 265)

edemaine (approved):
Annoying to have to repeat this code, but looks good!

PR #1498, opened 2024-10-23
`:symbol` shorthand for `Symbol.symbol` or `Symbol.for("symbol")`

edemaine:


source/parser.hera
```diff
@@ -1580,7 +1597,8 @@ PropertyAccessModifier
 PropertyAccess
   # NOTE: Added shorthand x."string" -> x["string"]
   # NOTE: Added shorthand x.3 -> x[3]
-  AccessStart:dot ( TemplateLiteral / StringLiteral / IntegerLiteral ):literal ->
+  # NOTE: Added shorthand x.:symbol -> x[Symbol.for("symbol")]
```
STRd6, line 1600: I hand't thought about this but it's a nice and convenient shorthand!

STRd6 (approved):
Looks great! Glad it was so simple.

P.S. did you `git add` your tests? :P

source/parser.hera
edemaine: Thanks! I looked at all the places we had `LengthShorthand` already, and extended those in some natural ways. `x.:symbol` looks a little funny but `x:symbol` looks quite nice (we support both). Overall, I'm excited about how this makes well-known symbols so much easier to use, especially to make custom iterators and such.

edemaine: Whoops, `x:symbol` conflicts with object literals, so we can't use that. But `x.:symbol` remains.

STRd6: I'd be ok with forcing inline object literals to have whitespace but we probably want to do a brief survey of existing code to make sure it is very rare or non-existent to write them that way.

STRd6: Sounds good, we can save it for Civet 2: The Wrath of Khan

PR #1500, opened 2024-10-23
LSP completions and hover info up-to-date instead of lagging

NWYLZW:
Enhance document handling in LSP server by ensuring service availability and updating documents

### DIFF
--- a/lsp/source/lib/util.mts
+++ b/lsp/source/lib/util.mts
@@ -401,3 +401,21 @@ function diagnosticCategoryToSeverity(
       return DiagnosticSeverity.Information;
   }
 }
+
+/**
+ * https://github.com/tc39/proposal-promise-with-resolvers
+ */
+export function withResolvers<T>() {
+  let resolve: (value: T) => void;
+  let reject: (reason?: any) => void;
+  const promise = new Promise<T>((res, rej) => {
+    resolve = res;
+    reject = rej;
+  });
+  return { promise, resolve: resolve!, reject: reject! };
+}
+export interface WithResolvers<T> {
+  promise: Promise<T>;
+  resolve: (value: T) => void;
+  reject: (reason?: any) => void;
+}
--- a/lsp/source/server.mts
+++ b/lsp/source/server.mts
@@ -525,6 +525,13 @@ documents.onDidChangeContent(async ({ document }) => {
   console.log("onDidChangeContent", document.uri)
   changeQueue.add(document)
   scheduleExecuteQueue()
+
+  const sourcePath = documentToSourcePath(document)
+  assert(sourcePath)
+  const service = await ensureServiceForSourcePath(sourcePath)
+  if (!service) return
+
+  service.host.addOrUpdateDocument(document)
 });
 
 async function updateDiagnosticsForDoc(document: TextDocument) {
--- a/lsp/source/server.mts
+++ b/lsp/source/server.mts
@@ -508,11 +508,10 @@ async function executeQueue() {
   // Run all jobs in queue (preventing livelock).
   for (const document of changed) {
     await updateDiagnosticsForDoc(document)
-    documentUpdateStatus.get(document.uri)?.resolve(true)
-    Promise
-      .resolve()
+    documentUpdateStatus.get(document.uri)
+      ?.resolve(true)
       // Wait for the document to be updated before removing it from the status map
-      .then(() => documentUpdateStatus.delete(document.uri))
+      ?.then(() => documentUpdateStatus.delete(document.uri))
   }
   // Allow executeQueue() again, and run again if there are new jobs now.
   // Otherwise, schedule update of all other documents.
--- a/lsp/source/server.mts
+++ b/lsp/source/server.mts
@@ -180,13 +180,7 @@ connection.onInitialized(() => {
   }
 });
 
-const updating = async (document: { uri: string }) => {
-  const resolver = documentUpdateStatus.get(document.uri)
-  if (resolver) {
-    await resolver.promise
-  }
-}
-
+const updating = (document: { uri: string }) => documentUpdateStatus.get(document.uri)?.promise
 const tsSuffix = /\.[cm]?[jt]s$|\.json|\.[jt]sx/
 
 connection.onHover(async ({ textDocument, position }) => {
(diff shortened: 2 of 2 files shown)


lsp/source/server.mts
```diff
@@ -525,6 +525,13 @@ documents.onDidChangeContent(async ({ document }) => {
   console.log("onDidChangeContent", document.uri)
   changeQueue.add(document)
   scheduleExecuteQueue()
+
+  const sourcePath = documentToSourcePath(document)
+  assert(sourcePath)
+  const service = await ensureServiceForSourcePath(sourcePath)
+  if (!service) return
+
+  service.host.addOrUpdateDocument(document)
```
edemaine, line 534: It looks like this makes compilation synchronous again. We added the execution queue for a performance reason, and I'm worried that this will slow things down again. Maybe we can make completion await the completion of the queue instead?

NWYLZW: In fact, at the very beginning, I wrote it just as you said. I added a Promise to judge whether the queue has been updated. However, I encountered some problems.

I can't obtain the service at the queue. So, for convenience, I just wrote it here directly. If it is a performance issue, I will try to adjust it and see.

edemaine: If you could try again, that'd be great!

The code you added is already being done inside `updateDiagnosticsForDoc`, called from `updatePendingDiagnostics`, which is what's scheduled by the queue.

What I'm imagining is that some queue-related promise (which gets resolved when the queue runs) gets awaited *inside* `onCompletion` etc. Then it would hopefully be easy to still do everything else the same as we currently do...

NWYLZW: I will deal with it later. The time zone here is already four o'clock in the morning😂.

lsp/source/server.mts
```diff
@@ -179,6 +180,13 @@ connection.onInitialized(() => {
   }
 });
 
+const updating = async (document: { uri: string }) => {
+  const resolver = documentUpdateStatus.get(document.uri)
+  if (resolver) {
+    await resolver.promise
```
edemaine, line 186: Not sure if it's important, but you could just `return` the promise here, and avoid `async` in this function.

Then I guess it would just be `return documentUpdateStatus.get(document.uri)?.promise`.

lsp/source/server.mts
```diff
@@ -501,6 +514,11 @@ async function executeQueue() {
   // Run all jobs in queue (preventing livelock).
   for (const document of changed) {
     await updateDiagnosticsForDoc(document)
+    documentUpdateStatus.get(document.uri)?.resolve(true)
+    Promise
+      .resolve()
```
edemaine, line 519: Are these two lines necessary? Or could we apply `.then` to the `.resolve(true)`?

edemaine (commented):
Awesome, thanks for working on this more! The code looks good on inspection. Two comments, and meanwhile I'm going to test it out.

edemaine:
After some testing, I do think that autocompletion behavior is improved. I still don't see completions after typing `.` (is this supposed to work with `triggerCharacters` now set, or does it require robust compilation?), but I do see them after typing the first character. Without the PR, I'm having trouble getting completions to work in general.

lsp/source/lib/util.mts
```diff
@@ -401,3 +401,21 @@ function diagnosticCategoryToSeverity(
...
+ * https://github.com/tc39/proposal-promise-with-resolvers
+ */
+export function withResolvers<T>() {
+  let resolve: (value: T) => void;
+  let reject: (reason?: any) => void;
+  const promise = new Promise<T>((res, rej) => {
+    resolve = res;
+    reject = rej;
+  });
+  return { promise, resolve: resolve!, reject: reject! };
```
edemaine, line 415:
```suggestion
  return { promise, resolve!, reject! };
```

edemaine: Oops, never mind, this is TypeScript, not Civet. 😢

NWYLZW: > Oops, never mind, this is TypeScript, not Civet. 😢

In fact, at the very beginning, I also wrote it this way. 😢

lsp/source/server.mts
NWYLZW: Indeed, this will be a bit easier.

lsp/source/server.mts
```diff
@@ -179,6 +180,13 @@ connection.onInitialized(() => {
   }
 });
 
+const updating = async (document: { uri: string }) => {
+  const resolver = documentUpdateStatus.get(document.uri)
+  if (resolver) {
+    await resolver.promise
+  }
+}
+
```
NWYLZW, line 189:
```suggestion
const updating = (document: { uri: string }) => documentUpdateStatus.get(document.uri)?.promise
```

lsp/source/server.mts
```diff
@@ -501,6 +508,11 @@ async function executeQueue() {
   // Run all jobs in queue (preventing livelock).
   for (const document of changed) {
     await updateDiagnosticsForDoc(document)
+    documentUpdateStatus.get(document.uri)?.resolve(true)
+    Promise
+      .resolve()
+      // Wait for the document to be updated before removing it from the status map
+      .then(() => documentUpdateStatus.delete(document.uri))
```
NWYLZW, line 515:
```suggestion
    documentUpdateStatus.get(document.uri)
      ?.resolve(true)
      // Wait for the document to be updated before removing it from the status map
      ?.then(() => documentUpdateStatus.delete(document.uri))
```

NWYLZW:
> After some testing, I do think that autocompletion behavior is improved. I still don't see completions after typing `.` (is this supposed to work with `triggerCharacters` now set, or does it require robust compilation?), but I do see them after typing the first character. Without the PR, I'm having trouble getting completions to work in general.

Yes, if we want `triggerCharacters` to work, we need to perform some fault-tolerant compilation as mentioned in Discord. Otherwise, with the current work content directly, it will lead to many other problems.

We can support better prompts with a new pull request. However, before this, I think we need some unit tests about LSP like those in Volar to provide some functional guarantees.
https://github.com/vuejs/language-tools/blob/baa1319e37fd881967791e7ea84be73e02f8a0b3/packages/language-server/tests/completions.spec.ts#L10

edemaine (approved):
This looks great; thanks for the revisions.

PR #1501, opened 2024-10-23
Fix unplugin's webpack mode: `resolve.alias`, virtual modules

edemaine:
[tough on HN](https://news.ycombinator.com/item?id=41919111) reported that the NextJS example wasn't working. I'd never tried it, but they were right! I needed to fix two things:

* NextJS makes extensive use of `resolve.alias`es. Sadly, `resolveId` doesn't seem to get the id after such aliases are resolved, so I reimplemented the alias resolver so that they function again.
* unplugin transforms our fake filenames into virtualized filenames, but the `importer` argument didn't get correctly unvirtualized. So I added this. (Will hopefully fix upstream as well.)

The source code links to the specifics that I referred to.

With all this, I can go to `integration/unplugin-examples/nextjs`, `npm install`, (`npm link @danielx/civet` to use this new unplugin), `npm run build`, and `npm run start` to get this beautiful webpage rendered:

image

### DIFF
--- a/integration/unplugin-examples/astro/package.json
+++ b/integration/unplugin-examples/astro/package.json
@@ -15,6 +15,6 @@
     "solid-js": "^1.7.11"
   },
   "devDependencies": {
-    "@danielx/civet": "*"
+    "@danielx/civet": "latest"
   }
 }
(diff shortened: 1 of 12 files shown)


STRd6 (approved):
:clapper:

PR #1502, opened 2024-10-23
`for` loops can filter via `when` conditions (not just in CoffeeScript mode)

edemaine:
(as discussed in Discord)


STRd6 (approved):
Thanks!

PR #1505, opened 2024-10-24
Range loops can provide custom skip via `by` (not just in CoffeeScript)

edemaine:
I apparently already tried to add this, because the parse rules were all there. They just didn't activate ever because `by` wasn't in `ReservedBinary`. Now fixed, and fixed the actual code that handled `by` (previously, `by` was in the output).

I also cleaned up some literal handling code, in particular to deal with these new step values, and let us figure out whether we're stepping upward or downard when the step is a literal.

### DIFF
--- a/civet.dev/reference.md
+++ b/civet.dev/reference.md
@@ -1836,6 +1836,11 @@ for i of [0...array.length]
 </Playground>
 
 <Playground>
+for i of [0...n] by 2
+  console.log i, "is even"
+</Playground>
+
+<Playground>
 for [1..5]
   attempt()
 </Playground>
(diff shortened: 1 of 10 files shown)


STRd6 (approved):
Looks good!

Do we have any tests combining `by` and `when`? In CoffeeScript they both work in either order: https://coffeescript.org/#try:x%20for%20x%20in%20%5B0..10%5D%20when%20x%20%3C%205%20by%202%0A%0Ax%20for%20x%20in%20%5B-1..-10%5D%20by%203%20when%20x%20%3C%20-4

edemaine:
> Do we have any tests combining `by` and `when`?

Done!

> In CoffeeScript they both work in either order:

Hmm, we don't support that (in or outside of Coffee mode). Is it worth adding? I think it's probably OK to require `by` to be next to the range literal...

PR #1474, opened 2024-10-17
Fix bracing in implicitly returned one-line for loops

edemaine:
Fixes #1471

Also a couple little cleanups I noticed:
* Remove unnecessary bracing in one-line loop generators (`for*` etc)
* Remove no-longer-used "outer" mechanism in `insertReturn` and `wrapIteration...`. `insertReturn` hasn't used `outerNode` in a while, and this PR removed the use in `wrapIteration...`
* New `asserts.notNull` with a TypeScript assertion to make guarding automatic

PR #1509, opened 2024-10-25
`for some/every/count/sum/product/min/max` reduction loops, empty for loop default behavior, fix unwrapping multiple loops ina  row

edemaine:
* The automatic unwrapping of loops at the top level of a block (to avoid IIFEs) wasn't triggering when there were multiple such loops in a row (since #1474). This is now fixed.
* Expressionized `for x of y` loops with an empty body now have an implicit body of `x`, giving a simple way to arrayify (or generatorify) a given iterable `y`. The previous behavior was to return an empty array, which didn't seem helpful
* `for some/every/count/sum/product/min/max` reduction loops!

~~TODO: not working with `when` yet~~ Now works with `when` and `own`!

### DIFF
--- a/civet.dev/reference.md
+++ b/civet.dev/reference.md
@@ -1841,7 +1841,8 @@ the body values according to one of the following reductions.
 `for some` returns whether any body value is truthy,
 shortcutting once one is found (like
 [`Array.prototype.some`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some)).
-If there is no body, any iteration counts as truthy.
+If there is no body, any iteration counts as truthy,
+so it measures whether the iteration is nonempty.
 
 <Playground>
 anyEven := for some item of array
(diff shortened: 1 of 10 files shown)


bbrk24:
Do these work with `for each` and `for own`? I guess `for each every` looks a little strange but it has value

STRd6:
What if the reducers implied each, and each was just another reducer for [] and .push?

I think if we're focusing on performance for these there is value in avoiding the iterator as well.

edemaine:
I haven't tested yet, but they should work with `each`; it would be spelled `for sum each x of y` (reducer goes next to `for`, `each` goes next to thing being iterated). I agree that `for every each x of y` reads a little silly, and meant to comment on that. The others read reasonably I think.

I think of `every` and `each` as orthogonal decisions. The reducer (`every`) is about what to *emit*, while `each` is about how to *consume*. I guess really `each` is modifying `y`, not `x`, so we could rethink that syntax. But I don't think I want to merge and force `each`... It really depends what you're iterating over. I'm definitely open to better notation though... One thing I considered is `for.sum`, but I think separate words is clearer and more consistent. 

`for sum own x in y` probably doesn't work right now for the same reason that `when` doesn't (currently it's clobbering the innards of the loop). But I'll make sure it does.

bbrk24:
> it would be spelled `for sum each x of y` (reducer goes next to `for`, `each` goes next to thing being iterated).

My intuition of `for each sum` comes from C#, where `foreach` is a single keyword rather than `for` + `each`. The way you describe it makes sense.

STRd6 (approved):
Thanks!

The orthogonality of `each` and the reducers makes sense. We may consider `for x of each y` as a syntax for adjusting the consumption of `y`.

After this goes in we can explore the syntax more and see if anything comes to mind.

PR #1514, opened 2024-10-26
`resolve` of `WithResolver` is not return A `Promise`

NWYLZW:
image

image


edemaine:
Thanks! And sorry for suggesting this bug in the first place.

PR #1517, opened 2024-10-26
Allow `|> throw`

### LINKED ISSUE #1513: Miscompile when attempting to `|> throw`
What the heck is going on here
```
try
    ;
catch err
    if err <? RangeError
        new StackTracingError
            `${err.message}`
            cause: err
        ||> .addCall raw
        |> throw
    throw err
```
```js
(() => {
  throw (() => {
    throw err;
  })();
})()(
  (() => {
    try {
    } catch (err) {
      let ref;
      return err instanceof RangeError
        ? ((ref = new StackTracingError(`${err.message}`, {
            cause: err,
          })).addCall(raw),
          ref)
        : void 0;
    }
  })(),
);
```

bbrk24:
Fixes #1513 

Currently, always wraps it in an IIFE, since there's no easy way to tell whether it's in statement vs expression position.

### DIFF
--- a/source/parser.hera
+++ b/source/parser.hera
@@ -885,7 +885,7 @@ PipelineHeadItem
   ParenthesizedExpression
 
 PipelineTailItem
-  ( AwaitOp / Yield / Return ) !AccessStart -> $1
+  ( AwaitOp / Yield / Return / Throw ) !AccessStart -> $1
   "import" !AccessStart ->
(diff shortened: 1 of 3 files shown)


STRd6 (approved):
Looks good, thanks!

In the future we can refine when we add IIFEs but this is much better than it not working at all.

PR #1518, opened 2024-10-26
Multiple generator loops at top level, unwrap top-level iterations, fix implicit loop bodies

edemaine:
This started as an optimization to not wrap top-level iterations in IIFEs, which also enables use of `break`/`continue`/`return` in more cases.

As a bonus, it changes the behavior of generator iterations (e.g. `for*`) when at the top level: they now directly `yield` to the parent function, instead of only being useful as immediately returned generators. This makes it a lot easier to make generator functions without having to type `yield`!

This is also a breaking change, but only in the case of unreturned generator expressions, which should have been a bug in your code anyway (as previous discussed with @bbrk24 in Discord).

Also fix implicit `for` loop bodies in two cases (separate commits):
* Range iteration
* Generator `for*`

### DIFF
--- a/source/parser/block.civet
+++ b/source/parser/block.civet
@@ -23,6 +23,10 @@ import {
   processReturn
 } from ./function.civet
 
+import {
+  assert
+} from ./util.civet
+
 /**
  * Duplicate a block and attach statements prefixing the block.
  * Adds braces if the block is bare.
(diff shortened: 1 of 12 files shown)


source/parser/block.civet
```diff
@@ -221,11 +225,29 @@ function needsPrecedingSemicolon(exp: ASTNode)
       if exp.children
         needsPrecedingSemicolon exp.children
 
+/** Is this node a top-level statement in its block? i.e. not an expression */
+function isStatementInBlock(exp: ASTNodeObject)
```
bbrk24, line 229: imo this function name is a little misleading because it doesn't return a boolean indicating whether it is -- instead, it returns `undefined` if not, and an object containing the block and a couple other things if so.

STRd6 (approved):
Thanks!

PR #1519, opened 2024-10-27
LSP unit testing of completions

NWYLZW:


lsp/source/lib/typescript-service.mts
```diff
@@ -445,7 +445,11 @@ function TSHost(compilationSettings: CompilerOptions, initialFileNames: string[]
 }
 
 function TSService(projectURL = "./") {
-  const logger = console
+  const logger = {
```
NWYLZW, line 448: If we need to make `vscode-jsonrpc` work properly, we cannot print redundant logs in the output.

lsp/source/lib/util.mts
```diff
@@ -358,7 +358,7 @@ export function logTiming<R, A extends unknown[]>(name: string, fn: (...args: A)
     const start = performance.now(),
       result = fn(...args),
       end = performance.now();
-    console.log(`${name.padStart(32)}${(end - start).toFixed(2).padStart(8)}ms`)
+    // console.log(`${name.padStart(32)}${(end - start).toFixed(2).padStart(8)}ms`)
```
NWYLZW, line 361: Perhaps we can add a parameter or a global `logger` variable.

lsp/source/server.mts
```diff
@@ -113,6 +115,7 @@ const diagnosticsDelay = 16;  // ms delay for primary updated file
 const diagnosticsPropagationDelay = 100;  // ms delay for other files
 
 connection.onInitialize(async (params: InitializeParams) => {
+  connection.console.log(`Initializing civet server at ${new Date().toLocaleTimeString()}`);
```
NWYLZW, line 118: All logs on the `server` should use the `console` on the connection as much as possible.

lsp/source/server.mts
```diff
@@ -155,15 +157,20 @@ connection.onInitialize(async (params: InitializeParams) => {
     };
   }
 
-  // TODO: currently only using the first workspace folder
-  const baseDir = params.workspaceFolders?.[0]?.uri.toString()
-  if (!baseDir)
-    throw new Error("Could not initialize without workspace folders")
+  if (!params.rootUri) {
```
NWYLZW, line 160: Although `rootUri` has been deprecated, it is not difficult for us to support it.

lsp/test/utils.civet
```diff
@@ -0,0 +1,39 @@
...
+{ createRequire } from node:module
+
+{ PublishDiagnosticsNotification } from @volar/language-server
+{ startLanguageServer, LanguageServerHandle } from @volar/test-utils
+{ TextDocument } from vscode-languageserver-textdocument
+{ URI } from vscode-uri
+
+require := createRequire import.meta.url
+
+__dirname := import.meta.url |> .slice 7 |> dirname
```
edemaine, line 11: I'd rather use `url.fileURLToPath` than `.slice 7`.

lsp/source/lib/util.mts
edemaine: Yes, I think so. A reasonable approach might be to add an LSP command-line parameter something like `--log server.log` that redirects all logs to there, and the default behavior remains going to `console` or `connection.console`.

Actually, if we can use `connection.console` always, presumably that can be intercepted? That seems like a cleaner solution.

lsp/source/server.mts
edemaine: Is this for compatibility with tools other than VSCode (e.g. tester or Vim) that aren't updated to the new interface? Otherwise, I don't see why we'd want to support a deprecated interface.

lsp/source/lib/util.mts
NWYLZW: Indeed, having a universal behavior would indeed be better for interception. The problem here is that the required `connection` cannot be obtained.

This pull request will also address this problem.

lsp/source/server.mts
NWYLZW: In the future, there may be a demand for `Monaco` applied in browsers. For example, directly integrating civet-related language services in it. However, the version of `Monaco` will be relatively outdated. Maintaining compatibility will be very valuable here.

We can build some better online `playground`. Honestly, I think civet's work on types is very constructive. However, I didn't feel this in civet's `playground`. I think this is very regrettable.

Also, regarding unit tests, it seems that `workspaceFolders` are not needed. Here, I think the cost is not high and it can be simply supported.

lsp/source/server.mts
```diff
@@ -288,7 +294,7 @@ connection.onCompletion(async ({ textDocument, position, context: _context }) =>
   // Don't map for files that don't have a sourcemap (plain .ts for example)
   if (sourcemapLines) {
     position = forwardMap(sourcemapLines, position)
-    console.log('remapped')
+    // console.log('remapped')
```
STRd6, line 297: It's fine to delete these rather than comment them out. Or log them to connection, possibly gated by a `--verbose` setting if feeling ambitious.

STRd6:
Thanks for digging into this, the LSP definitely needs all the help it can get! :clap:

lsp/source/server.mts
NWYLZW: Yes, what you said is correct. I do want to add a `--verbose` parameter. The work here is just relatively simple brute-force commenting.

In the past two days, I mainly want to make the unit tests run. In the next work, I will improve what you said and what I am going to do.

PR #1520, opened 2024-10-27
`await`/`throw` with placeholders in pipeline, trailing `|> throw` makes statement without IIFE

edemaine:
Follow-up to #1517 to
* use `StatementExpression` instead of manual `makeIIFE`
* when dealing with `StatementExpression`, use new `blockContainingStatement` helper from #1518 to determine whether to wrap with IIFE

I tried *always* doing the latter, but it led to some broken outputs such as `check := if & then x else &` turning into the incorrect `const check = $ => if ($) x; else $`. So it only happens when explicitly requested via `maybe: true` flag.

---

I also noticed that #1517 forbade regular `throw` expressionized statements in pipelines, which could be useful when using placeholders. So I also extended `.` placeholder to lift above `await`/`throw`, which seems very similar to lifting above unary operations (indeed, `await` is mostly considered a unary operator in Civet, but `processUnaryExpression` renames it), and made sure that the special cases `|> await` and `|> throw` didn't take over these expressions.

I didn't implement `|> return ...` or `|> yield ...` which could be nice, but aren't as easy because they aren't already supported as expressions.

### DIFF
--- a/test/pipe.civet
+++ b/test/pipe.civet
@@ -141,8 +141,7 @@ describe "pipe", ->
     }
   """
 
-  // TODO: Don't wrap in IIFE when it's not necessary
-  testCase.skip """
+  testCase """
     throw statement in pipeline
     ---
     new Error
@@ -159,6 +158,14 @@ describe "pipe", ->
     |> foo
     ---
     foo((()=>{throw new Error})())
+  """
+
+  testCase """
+    throw expression with thicc pipe
+    ---
+    x |> throw ||> y
+    ---
+    let ref;(y((ref = (()=>{throw x})())),ref)
   """
 
   testCase """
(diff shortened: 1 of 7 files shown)


test/pipe.civet
```diff
@@ -161,6 +178,22 @@ describe "pipe", ->
...
+    ---
+    let ref;(y((ref = (()=>{throw x})())),ref)
+  """
+
+  testCase """
+    throw with placeholder
+    ---
+    x |> throw new Error .
+    ---
+    (()=>{return ($ => { throw new Error($)})})()(x)
```
STRd6, line 194: It might be nice to simplify this since we won't need to wrap with the IIFE if it is already a placeholder expression.

It was quite mindbending to figure out what was going on here.

STRd6 (approved):
Looks good!

Small note about avoiding a wrapper if we're already a placeholder unary function since the IIFE is unnecessary in that case.

PR #1521, opened 2024-10-27
`T?` type shorthand in named tuple elements

### LINKED ISSUE #1504: Tuple types with `?`
## Input

```js
type T = [[A, B]?]
type T = [foo: [A, B]?]
```

## [Current Output](https://civet.dev/playground?code=dHlwZSBUID0gW1tBLCBCXT9dCnR5cGUgVCA9IFtmb286IFtBLCBCXT9dCg%3D%3D)

```js
type T = [[A, B]?]
type T = [foo: [A, B]?]
```

TypeScript gives an error saying that, for this form of tuple, the question mark should be before the colon.

## Expected Output

```js
type T = [[A, B]?]
type T = [foo: [A, B] | undefined]
```

When the tuple item is named, we should use our `| undefined` expansion of `?`.

edemaine:
Fixes #1504

Also some more type AST nodes to enable detection of named tuple elements

### DIFF
--- a/source/parser.hera
+++ b/source/parser.hera
@@ -7948,11 +7948,12 @@ ImportType
   "import" InsertOpenParen Trimmed_? StringLiteral InsertCloseParen
 
 TypeTuple
-  OpenBracket AllowAll ( TypeTupleContent __ CloseBracket )? RestoreAll ->
-    if (!$3) return $skip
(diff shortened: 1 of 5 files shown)


STRd6 (approved):
Thanks!

PR #1523, opened 2024-10-27
`async` functions and arrow types auto-wrap return type in `Promise`

### LINKED ISSUE #1516: `async` arrow type
Here's a proposed simple little type improvement:

```js
declare var f: async () => void
↓↓↓
declare var f: () => Promise<void>
```

edemaine:
Fixes #1516 by making the type `async => Foo` equivalent to `() => Promise<Foo>`

I realized that this was inconsistent with how we type the return values of actual `async function`s, though, and made it consistent. So now `async function f(): Foo` is equivalent to `async function f(): Promise<Foo>`. We've discussed this before, and were worried about the case when `Foo` was already a `Promise`-wrapped type. Luckily, this [doesn't affect type checking](https://www.typescriptlang.org/play/?#code/IYZwngdgxgBAZgV2gFwJYHsLwBQQFwwQIC2ARgKYBOAlAQAqXrGojkA8DTL7RZVAfPxgBvALAAoGFJjAA7sFTIYnZqwB0lciHQAbAG7ls1CdJibkCSlggSAvifgSA9E4B6AfiA) (though it does affect hover documentation, annoyingly).

In both cases, wrapping is suppressed if the type is already `Promise`-wrapped right there. In the distant future, we could inspect the type and avoid the redundant wrapper.

### DIFF
--- a/civet.dev/reference.md
+++ b/civet.dev/reference.md
@@ -2445,7 +2445,7 @@ declare function jsonParse(json: string): ???
 declare function sign(n: number): -1 | 0 | +1
 </Playground>
 
-### Arrow Types
+### Function Types
 
 Like [arrow functions](#arrow-functions),
 arrow types can use `=>` or `->` (with equivalent meanings)
@@ -2463,6 +2463,21 @@ return types:
 function f(callback: async =>)
   callback()
 </Playground>
+
+Similarly, `async` functions (except generators) get their return type
+automatically wrapped in `Promise`:
+
+<Playground>
+function f(): number
+  await fetch 'https://civet.dev'
+  .status
+</Playground>
+
+::: info
+You can still include explicit `Promise` wrappers in your return types.
+If the wrapper is hidden behind a `type` declaration, the output will include
+an extra `Promise` wrapper, but this does not affect typechecking.
+:::
 
 ### Conditional Types
 
--- a/civet.dev/reference.md
+++ b/civet.dev/reference.md
@@ -2445,6 +2445,25 @@ declare function jsonParse(json: string): ???
 declare function sign(n: number): -1 | 0 | +1
 </Playground>
 
+### Arrow Types
+
+Like [arrow functions](#arrow-functions),
+arrow types can use `=>` or `->` (with equivalent meanings)
+and can omit parameters and/or return type:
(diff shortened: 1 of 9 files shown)


STRd6 (approved):
Thanks!

bbrk24:
It does affect type checking https://www.typescriptlang.org/play/?#code/MYewdgzgLgBAZgLhgCgJQwLwD4YAUBOIAtgJYQCmAPAcWVWAK5EBG5+WOGMAhhAJ5hgKdNhgBWANwBYAFBA

STRd6:
> It does affect type checking

:crying_cat_face:

STRd6:
I suppose we may have to add `Awaited` in situations where we wrap for type aliases and maybe other cases:

```
type T = Promise<number>

const f: () => Promise<Awaited<T>> = async () => 5;
```

edemaine:
Thanks for catching this @bbrk24! So we need to avoid extra `Promise` wrappers.

To make the hover documentation look nicer, I defined an `AutoPromise<T>` helper. It's a little different than `Promise<Awaited<T>>`: it avoids messing with `T` whenever `T` is already a `Promise`. (I believe we want `Promise` here, not `PromiseLike`, because [TypeScript insists](<https://www.typescriptlang.org/play/?#code/IYZwngdgxgBAZgV2gFwJYHsIwBYAoCUAXDAAoBO6AtqiAKYAyqA1rQDwQKUBGtZAfDADeAWABQMCTGAB3YKmQwOAGyVjJMMrWQIyWAKxiAvkA>) that `async` functions return `Promise`.)

[Example of this working](https://www.typescriptlang.org/play/?#code/C4TwDgpgBAggrsA9gBQE6ILYEsDOEA8AKgHxQC8UhUEAHsBAHYAmOUamuBcDA1g4gHcGpAPyUoALjbpseIsQCwAKADGiBjmBQAZlIAUASnKl4SdrILnO+BnAwAjCKmKkKAQxwgGKqIeNQ3ATcsLQBWAG5lNQ0tAHN9IzITBBQZa1sHJ1cAz29fRNJA4LDIpSA).

STRd6 (approved):

PR #1547, opened 2024-10-30
Inequality slicing without `..`: `x[<i]` etc.

edemaine:
As suggested by @STRd6 on Discord

### DIFF
--- a/civet.dev/reference.md
+++ b/civet.dev/reference.md
@@ -461,8 +461,8 @@ If you just want to specify one endpoint of an increasing slice,
 you can avoid `..` altogether:
 
 <Playground>
-x is x[<=i] + x[>i]
-x is x[<i] + x[>=i]
+[left, right] = [x[<=i], x[>i]]
+[left, right] = [x[<i], x[>=i]]
 </Playground>
 
 ## Strings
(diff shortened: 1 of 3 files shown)


civet.dev/reference.md
```diff
@@ -457,6 +457,14 @@ Slices are increasing by default, but you can reverse them with `>` or `>=`:
 reversed := x[..>=]
 </Playground>
 
+If you just want to specify one endpoint of an increasing slice,
+you can avoid `..` altogether:
+
+<Playground>
+x is x[<=i] + x[>i]
```
STRd6, line 464: We should remove the `x is` and `+` from these examples because `===` and `+` on slices aren't how one may expect.

STRd6 (approved):

JayXon:
This is a little confusing because I thought `x[>=i]` is a reverse slice where `..` is omitted

edemaine:
We discussed this some in Discord (feel free to join us!).

This PR's notation is more like the binary operator sections `(< i)` and `(>= i)`, and clearly specify the desired index range.

We could of course also spell `x[>=i]` as `x[i<=]` (which wouldn't have the same confusion), and I'd be open to adding that as an alternative spelling. But I worry about making `x[>=i]` a reverse slice: my feeling is that reverse slices are a niche feature, so don't need concise notation, and I don't think it's particularly intuitive that `x[>=i]` is reversed, so it's not worth making it so for consistency. But happy to continue discussing!

PR #1548, opened 2024-10-30
Allow multiple indented blocks of arguments in function call

edemaine:
Argument parsing redux, part 721.

@bbrk24 reported that we couldn't mix bulleted arrays with other indented arguments. This was a holdover from CoffeeScript-style argument where we required commas between implicit objects and other items, because that's all that could work.

I ended up being able to make commas optional before indented blocks of arguments (which can be regular arguments, bulleted list, or an implicit object). This simplified the grammar somewhat, even. And with a bit of work, it lets you flexibly switch between the different styles of indented argument blocks.


STRd6 (approved):
Nice!

PR #1556, opened 2024-10-31
Avoid duplicate calls in relation chains and `@` bind shorthand via refs 

### LINKED ISSUE #50: Use refs for non-simple chained comparisons
Use a ref in non-simple chained comparisons so not to repeat evaluations of complex expressions.

```civet
f1(x) < f2(x) < f3(x) < f4(x)
```

```
let ref
f1(x) < (ref = f2(x)) && ref < (ref = f3(x)) && ref < f4(x)
```

edemaine:
Fixes #50, but I still need to think about the TODO (out of time for now).

### DIFF
--- a/source/parser.hera
+++ b/source/parser.hera
@@ -1272,7 +1272,11 @@ FieldDefinition
     }
 
 ThisLiteral
-  This
+  This -> {
+    type: "Identifier",
+    name: "this",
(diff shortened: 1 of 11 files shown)


STRd6 (approved):
Thanks!

PR #1560, opened 2024-11-01
Robust parsing of `foo.` with error node

edemaine:
@NWYLZW Can you make use of this for completions after typing `.`? I still couldn't get completions to work in this case, but at least I get other TypeScript errors after typing a `.`...

### DIFF
--- a/source/parser.hera
+++ b/source/parser.hera
@@ -1681,6 +1681,18 @@ PropertyAccess
       children: [ dot, ...comments, ...id.children ],
     }
 
+  # NOTE: For LSP completion / robustness, parse `foo.` but with an error node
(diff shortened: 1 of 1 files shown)


NWYLZW:
I merge it into my branch and try it.

seanstrom:
I tested this by running the Civet CLI locally with some sample code of `foo. `, and I definitely get the incomplete property error. Tests run well for met too

Although, when I attempt to use the LSP extension (built and packaged from this branch's source), I don't seem to get the missing property error, and instead get a "Failed to Parse" error.

Should I be able to see the "Incomplete Property" error when using the LSP extension?

PR #1561, opened 2024-11-01
LSP completions show details and documentation

edemaine:
The LSP protocol has a way to ask for more details about completions. I struggled through recreating what the TypeScript LSP does, forwarding the request on to TypeScript and then redoing the format completely (ugh). This roughly duplicates a bunch of code from https://github.com/microsoft/vscode/blob/main/extensions/typescript-language-features/src/languageFeatures/util/textRendering.ts in particular, but I didn't see a way around it (and it's MIT licensed at least, but maybe needs a better credit?).

image

It crashes for some completions where the completion provides a "source" (like the `mocha` example above). I don't really know what the issue is. (It says `Cannot read properties of undefined (reading 'options')` but the error occurs within TypeScript.) Maybe we should catch the error and display it in logs but not crash? But ideally we'd fix whatever's causing the problem. Still, perhaps better to have this PR, mostly working, than not have details/documentation at all.

### DIFF
--- a/lsp/source/lib/textRendering.mts
+++ b/lsp/source/lib/textRendering.mts
@@ -1,0 +1,162 @@
+import type ts from 'typescript'
+type SymbolDisplayPart = ts.server.protocol.SymbolDisplayPart
+type FileSpan = ts.server.protocol.FileSpan
+type JSDocTagInfo = ts.server.protocol.JSDocTagInfo
+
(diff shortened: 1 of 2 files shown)


STRd6 (approved):

STRd6:
Thanks! It seems like copying the TS stuff is largely unavoidable (at least I haven't found a clean way to get the features out of the existing TS language server or extend it)

Issue #439, opened 2023-03-08
Object and advanced array comprehensions


edemaine:
## DONE: Object comprehensions

(based on Python's dict comprehensions)

```coffee
obj := {for item of list
   [item.key]: item
   [item.key.toLowerCase()]: item
}
---
const obj = (()=>{const results = {}
  for (const item of list) {
    Object.assign(results, {
      [item.key]: item,
      [item.key.toLowerCase()]: item,
    })
  }
})()
```

## Array flattening via spreads

```coffee
array := for item of list
  if Array.isArray item
    ...item
  else
    item.toString()
---
const array = (()=>{const results = []
  for (const item of list) {
    if (Array.isArray(item)) {
      results.push(...item)
    } else {
      results.push(item.toString())
    }
  }
})()
```

This could also be achieved via `for.value.push(...item); continue` given #362, but this seems like more natural notation.

STRd6:
Spreads could also apply to object results as well.

edemaine:
Another application would be the ability to flatten the results of nested `for` loops. Ideally we could simplify the output to not construct an intermediate array, as follows:

```js
array := for sublist of list
  ...for item of sublist
    item ** 2
---
const array = (function(){const results = []
  for (const sublist of list) {
    for (const item of sublist) {
      results.push(item ** 2)
    }
  }
})()
```

This gives us a very flexible equivalent of [Janet's `loop` primitive](https://janet.guide/control-flow/).

gwhitney:
+1 on the object comprehensions: looks/feels totally natural, and with significant Python exposure prior to any Java/TypeScript, something I've been feeling the lack of basically since day 1.

edemaine:
For anyone following this issue, #1563 gave us object comprehensions! See https://civet.dev/reference#object-comprehensions

PR #1563, opened 2024-11-02
Object comprehensions via loops in braces

peey:
Work-in-progress to address the "object comprehension" part of issue #439.

Implementation rationale is that whenever  the following is seen:
```
{ 
  for identifier of expression 
    [a] : b
    [c]:  d
    //...
}
```

Then it should be recognized as object comprehension (instead of being recognized as a block containing a ForOf statement).

This works because `for identifier of expression` being followed by a computed property key-value pair is invalid syntax in js.

I need some guidance:

### 1

My rules recognize the syntax but it isn't prioritized over "block containing a for-of" rule when the [a] : b are indented. This results in the following input-output pair:

```
{ for x of [1, 2, 3]
  [x]:  2 * x
  [x * 2]: 4 * x
}
// Becomes:
 for (const x of [1, 2, 3]) {
  ({[x]:  2 * x,
  [x * 2]: 4 * x})
}
}

```

This is also a test (last one) which fails. If I remove the leading whitespace, then it works fine

```
{ for x of [1, 2, 3]
  [x]:  2 * x
  [x * 2]: 4 * x
}
// Becomes:
(()=>{const _results = {};
  for (const x of [1, 2, 3]) {
    Object.assign(_results, {
      [x]:  2 * x,
      [x * 2]: 4 * x
  })}; 
  return _results
})()
```

Which provides the output `{ '1': 2, '2': 4, '3': 6, '4': 8, '6': 12 }` if you run it. Note that the indentation in this output has been done by me manually for readability.

### 2

How should I deal with the indentation in output? I'm introduce an IIFE, a for loop, and an `Object.assign`. That's 3 extra levels of nesting. Should I just add two spaces for each level?

### 3

The second test case transpiles but it's incorrect because the inner `_results` shadows the outer `_results` and I need to use the variable I introduce in the inner scope. How do I create a variable that's guaranteed to not conflict with the inner scope? 

### 4

I've noticed that my code differs from the existing code in following ways:

1. Most rules return an object with `type`
2. There's some way to deal with source locations so that it's useful for source mapping

I'll address these in future commits. Tell me if there are other conventions that I should learn about!

### DIFF
--- a/civet.dev/reference.md
+++ b/civet.dev/reference.md
@@ -1937,6 +1937,31 @@ min := for min item of array
 max := for max item of array
 </Playground>
 
+### Object Comprehensions
+
+Loops can also accumulate their body values into an object.
+When any loop is found is found within an object expression,
+its body value is spread into the containing object.
+
+<Playground>
+object  := {a: 1, b: 2, c: 3}
+doubled := {
+  for key in object
+    [key]: 2 * object[key]
+}
+</Playground>
+
+The loop can exist anywhere a property is expected.
+It can be freely mixed with other object properties.
+
+<Playground>
+rateLimits := {
+  admin: Infinity,
+  for user of users:
+    [user.name]: getRemainingLimit(user)
+}
+</Playground>
+
 ### Infinite Loop
 
 <Playground>
--- a/civet.dev/reference.md
+++ b/civet.dev/reference.md
@@ -1940,7 +1940,7 @@ max := for max item of array
 ### Object Comprehensions
 
 Loops can also accumulate their body values into an object.
-When any loop is found is found within an object expression,
+When any loop is found is found within a braced object expression,
 its body value is spread into the containing object.
 
 <Playground>
@@ -1951,13 +1951,22 @@ doubled := {
 }
 </Playground>
 
+<Playground>
+i .= 1
+squares := {
+  do
+    [i]: i * i
+  while i++ < 10
+}
+</Playground>
+
 The loop can exist anywhere a property is expected.
 It can be freely mixed with other object properties.
 
 <Playground>
 rateLimits := {
   admin: Infinity,
-  for user of users:
+  for user of users
     [user.name]: getRemainingLimit(user)
 }
 </Playground>
--- a/source/parser.hera
+++ b/source/parser.hera
@@ -3284,20 +3284,6 @@ ObjectLiteral
   BracedObjectLiteral
   NestedImplicitObjectLiteral
   InlineObjectLiteral
-  ObjectComprehension
-
-ObjectComprehension
-  OpenBrace __ "for" _ Identifier:id _ "of" _ Expression:exp __ ObjectComprehensionProps:props __ CloseBrace ->
-    return ["(()=>{const _results = {};\n",
-             "for (const ", id, " of ", exp, ") {\n",
-             "Object.assign(_results, {\n",
-             props, "\n",
-             "})}; return _results})()"
-           ]
-
-ObjectComprehensionProps
-  (ComputedPropertyName __ ":" __ Expression __)+ ->
-    return $0.map(entry => entry.concat([","]))
 
 BracedObjectLiteral
   OpenBrace:open AllowAll ( BracedObjectLiteralContent __ CloseBrace )? RestoreAll ->
@@ -3314,7 +3300,7 @@ BracedObjectLiteral
 BracedObjectLiteralContent
   # Allow unnested comma-separated properties on first line, optionally
   # followed by nested comma-separated properties on further lines
-  ( PropertyDefinition ObjectPropertyDelimiter )*:line NestedPropertyDefinitions?:nested ->
+  ( ObjectPropLine )*:line NestedPropertyDefinitions?:nested ->
     line = line.flatMap(([prop, delim]) => {
       // Allow each prop to be a single Property object or an array of such
       prop = Array.isArray(prop) ? prop : [prop]
@@ -3329,7 +3315,7 @@ BracedObjectLiteralContent
     })
     return line.concat(nested || [])
   # As a backup, allow for arbitrary untracked indentation
-  ( __ PropertyDefinition ObjectPropertyDelimiter )+ ->
+  ( __ ObjectPropLine )+ ->
     return $0.flatMap(([ws, prop, delim]) => {
       // Allow each prop to be a single Property object or an array of such
       prop = Array.isArray(prop) ? prop : [prop]
@@ -3343,6 +3329,14 @@ BracedObjectLiteralContent
       }
       return [...prop.slice(0, prop.length-1), last]
     })
+
+
+ObjectPropLine
+  PropertyDefinition ObjectPropertyDelimiter
+
+InsertSpread
+  "" ->
+    return { $loc, token: "..." }
 
 # NOTE: Nested implicit object literal starts with an indentation.
 # All properties can be spaced out and must have a colon.
@@ -3377,7 +3371,7 @@ NestedPropertyDefinitions
 
 NestedPropertyDefinition
   # TODO: This may be a little weird/ambiguous with single identifier shorthand
-  Nested:ws ( PropertyDefinition ObjectPropertyDelimiter )+:inlineProps ->
+  Nested:ws ( ObjectPropLine )+:inlineProps ->
     return inlineProps.flatMap( ([prop, delim], i) => {
       // Allow each prop to be a single Property object or an array of such
       if (!Array.isArray(prop)) prop = [prop]
@@ -3553,6 +3547,15 @@ PropertyDefinition
   # NOTE: basic identifiers are now part of the rule above
   #_?:ws IdentifierReference:id ->
   #  return prepend(ws, id)
+  InsertSpread:dots IterationExpression:exp ->
+    exp.statement.object = true
+    return {
+      type: "SpreadProperty",
+      children: [dots, exp],
+      names: exp.names,
+      dots,
+      value: exp,
+    }
 
 NamedProperty
   # NOTE: CoverInitializedName early error doesn't seem necessary with this parser
--- a/source/parser.hera
+++ b/source/parser.hera
@@ -3424,7 +3424,6 @@ ObjectPropertyDelimiter
 PropertyDefinition
   # Safe to prioritize this over others as it's unlikely that a loop is matched unless the user intended to put the loop there. This is also required for giving braceless-do loop priority over interpreting do as a PropertyName
   _?:ws InsertDotDotDot:dots IterationExpression:exp ->
-    debugger
     let { statement } = exp
 
     // immutably set exp.statement.object = true
--- a/source/parser.hera
+++ b/source/parser.hera
@@ -3284,6 +3284,20 @@ ObjectLiteral
   BracedObjectLiteral
   NestedImplicitObjectLiteral
   InlineObjectLiteral
+  ObjectComprehension
+
+ObjectComprehension
+  OpenBrace __ "for" _ Identifier:id _ "of" _ Expression:exp __ ObjectComprehensionProps:props __ CloseBrace ->
+    return ["(()=>{const _results = {};\n",
+             "for (const ", id, " of ", exp, ") {\n",
+             "Object.assign(_results, {\n",
+             props, "\n",
+             "})}; return _results})()"
+           ]
+
+ObjectComprehensionProps
+  (ComputedPropertyName __ ":" __ Expression __)+ ->
+    return $0.map(entry => entry.concat([","]))
 
 BracedObjectLiteral
   OpenBrace:open AllowAll ( BracedObjectLiteralContent __ CloseBrace )? RestoreAll ->
--- a/source/parser.hera
+++ b/source/parser.hera
@@ -3300,7 +3300,7 @@ BracedObjectLiteral
 BracedObjectLiteralContent
   # Allow unnested comma-separated properties on first line, optionally
   # followed by nested comma-separated properties on further lines
-  ( ObjectPropLine )*:line NestedPropertyDefinitions?:nested ->
+  ( PropertyDefinition ObjectPropertyDelimiter )*:line NestedPropertyDefinitions?:nested ->
     line = line.flatMap(([prop, delim]) => {
       // Allow each prop to be a single Property object or an array of such
       prop = Array.isArray(prop) ? prop : [prop]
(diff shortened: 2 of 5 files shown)


edemaine:
Good start!

Instead of making a separate rule `ObjectComprehension`, I'd suggest adding `for` loops to `ObjectLiteralContent`. This would allow multiple `for` loops, mixing non-`for` content, etc. For example:

```js
{
  name: "Upgrade"
  for key, value in object
    [key.toLowerCase()]: value + 1
}
```

Admittedly, this will be harder to compile. I'm imagining this:

```js
{
  name: "Upgrade",
  ...(() => {const results=[]; for (key in object) {const value = object[key];
    Object.assign(results, {[key.toLowerCase()]: value + 1})
  })()
}
```

The code above might look complicated, but it's [mostly what we already get from the `for` loop currently](https://civet.dev/playground?code=eCA9IChmb3Iga2V5LCB2YWx1ZSBpbiBvYmplY3QKICBba2V5LnRvTG93ZXJDYXNlKCldOiB2YWx1ZSArIDEKKQ%3D%3D): the only difference is `results.push(` vs. `Object.assign(results,`.

You might check out the recent PR #1509 that provided alternatives to `results.push`, such as `results +=` when using `for sum`. In the parser, I think you could just allow arbitrary `for` loops, and attach an extra property like `object: true` to denote that they're an object comprehension. Then, in `iterationDeclaration` where this code gets generated, you could use `Object.assign` when the `object` property is true.

I think this approach might also fix your first issue about priority of blocks vs. object literals, because object comprehensions will now be considered an object literal. (I'm not totally sure, but generally we favor object literals.)

> How should I deal with the indentation in output? I'm introduce an IIFE, a for loop, and an `Object.assign`. That's 3 extra levels of nesting. Should I just add two spaces for each level?

No, try to preserve the indentation in the original input. Also try not to add any extra lines (except what `for` already does — we aim to remove those eventually, but obviously not this PR). So just one-line whatever you need to add.

> The second test case transpiles but it's incorrect because the inner `_results` shadows the outer `_results` and I need to use the variable I introduce in the inner scope. How do I create a variable that's guaranteed to not conflict with the inner scope?

Use `makeRef("results")` to create a variable named `results` or `results1` or `results2` or ..., whatever avoids conflict with existing variable names.

> 2\. There's some way to deal with source locations so that it's useful for source mapping

To help with this, use tokens as much as possible. For example, instead of using `"for"`, use `For`, which attaches `$loc` data for source mapping. That said, I don't think you need `for` at all; ideally, you'd use `ForStatement` directly (though you may need to manually wrap this in an `IterationExpression` node so that it gets expressionized) and support the many many `for` loops that Civet already supports.

Hope this helps!

peey:
Thanks, very helpful! I did not even realize that Civet already has these for loops. The fact that they are expression-like (they evaluate to their results) is quite interesting and changes the way I was thinking about incorporating loops into Object Comprehensions.

I'll study these a bit and then proceed.

peey:
Hey @edemaine, look at this! I think it's exactly how you wanted to implement it?

When a loop (`IterationExpression`) is found where you'd normally expect a property definition, we now expressionize it and spread it into the containing object.

A few notes:

1. We can limit it to `ForStatement`, but since the loop is expressionized and loop's value is what's `Object.assign`ed to the result, so in principle it doesn't matter. I'm yet to investigate utility of other loops and add tests for them.
2. Initially I wanted to restrict the loop body to be only computed-key computed-value pairs, but I now believe that's both unnecessary (see the last test case) and less clean to implement
3. Conceptually, this is actually a "reduction" much like the other reductions: we have an accumulator `{}`, an update operation, and we return the accumulated value. But in code, `statement.reduction` corresponds to a particular parsing and transpiling logic belonging to the for reduction loops (#1509). 

Technically this is a "support loops inside object expressions" implementation which is rather clean, and as a special and most-commonly used case, it provides for syntax such as in #439.

As a consequence, here's a nice self-describing program 

```coffeescript
console.log({ 
  name: 'Square table generator'
  description: 'This table contains numeric properties mapped to their squares'
  for x of [1, 2, 3]
    [x]:  x * x
    [x * 2]: 4 * x * x
})
```

That would produce:

```js
{
  name: 'Square table generator',
  description: 'This table contains numeric properties mapped to their squares',
  '1': 1,
  '2': 4,
  '3': 9,
  '4': 16,
  '6': 36
}
```

What do you think?

source/parser.hera
```diff
@@ -3539,6 +3539,15 @@ PropertyDefinition
   # NOTE: basic identifiers are now part of the rule above
   #_?:ws IdentifierReference:id ->
   #  return prepend(ws, id)
+  InsertDotDotDot:dots IterationExpression:exp ->
+    exp.statement.object = true
```
edemaine, line 3543: We shouldn't modify objects like this in `parser.hera` (only in postprocessing steps), because it can mess up caching. Please change to be an immutable update, like
```js
let { statement } = exp
statement = { ...statement, object: true }
exp = {
  ...exp,
  statement,
  children: exp.children.map(($) => $ === exp.statement ? statement : $),
}
```

source/parser/function.civet
```diff
@@ -661,6 +661,7 @@ function iterationDeclaration(statement: IterationStatement | ForStatement)
     bindings: []
   }
 
+  object := statement.type is "ForStatement" and statement.object
```
edemaine, line 664: I like your idea of allowing any iteration! So I think you can remove `statement.type is "ForStatement"`. Also, this probably makes sense to move up to where `reduction` is defined.

source/parser.hera
edemaine: This will also generate weird code when the iteration is a reduction, given that it takes priority over `object` behavior. Perhaps we should raise an error in this case? Something like
```js
const children = [ dots, exp ]
if (exp.statement.reduction) {
  children.unshift({
    type: "Error",
    message: "Reduction loops are forbidden in object literals",
  })
}
```

Alternatively, the error detection could be done in `iterationDeclaration` below.

test/object-comprehensions.civet
```diff
@@ -0,0 +1,118 @@
...
+        [x * 2]: 4 * x
+    }
+    ---
+    ({
+      ...(()=>{const results={};for (const x of [1, 2, 3]) {
+        Object.assign(results,({[x]:  2 * x,
+        [x * 2]: 4 * x}))
+      }return results})()
+    })
+  """
```
edemaine, line 19: An interesting issue is that [TypeScript completely gets the types wrong here](https://www.typescriptlang.org/play/?#code/MYewdgzgLgBCMF4bIBQG8CwAoZuYDpCUUBKBAPjVElgCcBTCAVwBsoIE0BfAbgDMQtGCmrQYADzh8YAbQCMAGhgAmJQGYAuiRiYceXAHkARgCt6wKPgCGECAEsA5mBQNmbCAvQzxGgFzJlGAAqCQVsfVxvYJU-GAAWaPEuEhJw-S4GKCZaMBhXVnZk0jTkZOw0kGwAeiqAPQB+IA); I guess it just ignores `Object.assign`. At least it doesn't error, but it's sad that it doesn't figure out that it's `Record<string, number>`.

I don't know a way to fix this. (I tried `let results;results={}` but it behaved the same.) I guess we can always explicit typecast, e.g. like this:
```js
    {
      for x of [1, 2, 3]
        [x]:  2 * x
        [x * 2]: 4 * x
    } as Record<string, number>
```

source/parser/function.civet
```diff
@@ -670,6 +671,8 @@ function iterationDeclaration(statement: IterationStatement | ForStatement)
         when "max" then "-Infinity"
         when "product" then "1"
         else "0"
+  else if object
+    declaration.children.push "={}"
```
edemaine, line 675: This `if`/`else if` gives priority to `reduction` over `object` here, whereas below you give priority to `object`. I think these should be consistent. Maybe `object` should be given priority, given the `...` that precedes it.

(You could also do error detection (`reduction and object`) here, instead of in `parser.hera`.)

edemaine (commented):
This looks great!! Awesome how simple the code is now.

I like your idea of allowing arbitrary iteration expressions. I can imagine `while` making sense, and especially `do` being helpful (to allow for intermediate computations). Adding some tests for these would be good.

Actually, it'd be cool to allow `if` expressions too. (But feel free to leave that for another PR.)

Does this PR work with postfixed iteration, such as `{[i]: i*2} for i of [1,2,3]`? I'm guessing not. I'm not really sure whether it's a good idea... (`i: i*2 for i of [1,2,3]` would be neat but that already has a different meaning: `i: (i*2 for i of [1,2,3])`.)

test/object-comprehensions.civet
```diff
@@ -0,0 +1,266 @@
...
+  testCase """
+    non-comprehension while loop
+    ---
+    o := {...while(predicate(x)) [x]: f(x)}
+    ---
+    const o = {...(()=>{const results=[];while(predicate(x)) results.push(({[x]: f(x)}));return results})()}
+  """
+
+  testCase """
+    do while object comprehension
```
peey, line 222: @edemaine I discovered one fundamental design issue here.

While this one-liner do while loop works fine, I cannot support a multi-line braceless do-while loop.

So for instance, this works:
```
o := {
  do {
    [x]: f(x)
  } while(predicate(x)) 
}
```

But this does not work:

```
o := {
  do
    [x]: f(x)
  while(predicate(x)) 
}
```

When there is nothing after `do`  except a newline, it's interpreted as a `NamedProperty`. This does not affect for and while because they have some tokens before a newline.

I can prioritize my rule by putting it above all the other rules in `PropertyDefinition`, and surprisingly that breaks only one test!

But that test is https://github.com/DanielXMoore/Civet/blob/b084f871269562be1e3ef12f8fe8f114577f5a44/test/object.civet#L1409 , which is about Civet's feature that there are no reserved words. 

If I prioritize loops, then "do" becomes a reserved word. If I don't, I can't support multi-line do.

peey (commented):
Thanks for your review!

I currently do not support postfixed iteration, I have not thought about it much yet.

**EDIT:** In an earlier version of comment I had misunderstood what you meant by postfixed iteration. 

I'll also leave the `if` expressions out of the scope of this PR, but I agree that it could be interesting to have values conditionally slurped into the object, e.g.

```
{ 
  a: 1
  if x
    b: 2
}
```

Currently, the way to do that is not very friendly:

```
{
  a: 1,
  ...(x? {b : 2} : {})
}
```

I also added some docs!

test/object-comprehensions.civet
```diff
@@ -0,0 +1,266 @@
...
+    ({
+      ...(()=>{const results={};for (const x of [1]) {
+        foo(bar)
+        Object.assign(results,({[x]:  2 * x}))
+      }return results})()
+    })
+  """
+
+  testCase """
+    loop body with additional statements: multi prop style 1
```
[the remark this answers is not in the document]
edemaine: This is deliberate; it's how implicit object literals work.

test/object-comprehensions.civet
edemaine: Good catch! Yes, I think you should move your rule up for higher priority, and remove the looping reserved words from the test.

civet.dev/reference.md
```diff
@@ -1937,6 +1937,31 @@ min := for min item of array
 max := for max item of array
 </Playground>
 
+### Object Comprehensions
+
+Loops can also accumulate their body values into an object.
+When any loop is found is found within an object expression,
```
edemaine, line 1943: You should mention *braced* object expressions (or literals) here. This won't work with implicit (unbraced) object literals.

civet.dev/reference.md
```diff
@@ -1937,6 +1937,31 @@ min := for min item of array
...
+}
+</Playground>
+
+The loop can exist anywhere a property is expected.
+It can be freely mixed with other object properties.
+
+<Playground>
+rateLimits := {
+  admin: Infinity,
+  for user of users:
```
edemaine, line 1960: What does this colon do?

peey: Oversight, thanks for catching!

test/object.civet
```diff
@@ -1408,9 +1408,9 @@ describe "object", ->
     testCase """
       with reserved word keys
       ---
-      state.{and,await,break,case,catch,class,const,continue,debugger,default,delete,do,else,enum,export,extends}
+      state.{and,await,break,case,catch,class,const,continue,debugger,default,delete,else,enum,export,extends}
       ---
-      ({and:state.and,await:state.await,break:state.break,case:state.case,catch:state.catch,class:state.class,const:state.const,continue:state.continue,debugger:state.debugger,default:state.default,delete:state.delete,do:state.do,else:state.else,enum:state.enum,export:state.export,extends:state.extends})
+      ({and:state.and,await:state.await,break:state.break,case:state.case,catch:state.catch,class:state.class,const:state.const,continue:state.continue,debugger:state.debugger,default:state.default,delete:state.delete,else:state.else,enum:state.enum,export:state.export,extends:state.extends})
```
edemaine, line 1413: Oh, I didn't realize that it's property globs that are failing. This is annoying. But fixing it will require some work on `PropertyGlob` to not use `BracedObjectLiteral`, but rather just use a subset of those rules. I can work on this in another PR.

peey: Thanks for clarifying. I do not know much about property globs so I'll safely leave it with you.

source/parser.hera
```diff
@@ -3422,6 +3422,34 @@ ObjectPropertyDelimiter
 # https://262.ecma-international.org/#prod-PropertyDefinition
 # NOTE: Must start on same line
 PropertyDefinition
+  # Safe to prioritize this over others as it's unlikely that a loop is matched unless the user intended to put the loop there. This is also required for giving braceless-do loop priority over interpreting do as a PropertyName
+  _?:ws InsertDotDotDot:dots IterationExpression:exp ->
+    debugger
```
STRd6, line 3427: Remove this `debugger` statement.

peey:
> Possible TODO: if there's just a single loop and nothing else in the object literal, then we don't need a { ...(comprehension) } wrapper so should avoid emitting one. Also postfix loops remains an interesting TODO.

I also wish to leave this optimization for later. I could definitely investigate both of these myself, but only later and not now.

STRd6 (approved):
Thanks @peey for all the work on this and @edemaine for the feedback and review! This is a great extension to the language.

I think we may want to use direct assignment of the result rather than `Object.assign` but that can be done later as it is pretty minor.

Also added a note about a debugger statement that snuck in.

source/parser.hera
peey: I'll do that right away!

test/object.civet
edemaine: I was able to fix this without messing with globs; just forbidding implicit blocks with `do` and `loop`.

PR #1572, opened 2024-11-04
Playground buttons to copy input or output

edemaine:
Follow up to #1569 to
* Buttons for copying both input or output to clipboard
* Show buttons in all cases, not just when it's long
* Improve looks somewhat: `copy` text wasn't obviously a button; now it's a clipboard icon, with feedback via a green checkmark or a red X

image


STRd6 (approved):
Thanks, a good improvement!

PR #1575, opened 2024-11-05
Object and array literals via `{}` and `[]` followed by items

edemaine:
Braced object literals have many more features than unbraced object literals (from CoffeeScript). This PR transforms an unbraced object literal into a braced one by adding `{}` in front of it, both in the inline case (`{} foo, bar`) and in the nested case with a layer of indentation:

```js
{}
  foo
  bar
```

Overall, the syntax aligns with implicit function calls, and `{}` is never a function that can be called, so there's effectively no ambiguity here.

Similarly, `[]` constructs array literals. This isn't as important, especially given that we have bulleted arrays (which are still useful because they allow clear delineation between unbraced object literal elements), but it does allow for some fun things:

```js
[] [] [] [] x
↓↓↓
[ [ [ [ x]]]]
```

```js
consList := [] x, [] y, [] z
↓↓↓
const consList = [ x, [ y, [ z]]]
```


STRd6 (approved):
I'm scared...

... scared I might like it :fearful: :relieved:

edemaine:
Let's see how it feels! I was a bit scared about this one for a while, until I realized we're only replacing function calls that didn't make sense.

PR #1578, opened 2024-11-05
Syntax highlighting: fix leading `_` being treated as a number

zoodogood:
- [x] Tested via civet/lsp debug mode
Correctly display digits like 1_000


edemaine (approved):
LGTM. I changed the title to `Syntax highlighting: fix leading `_` being treated as a number` because `_` in number literals was already "allowed", just overzealous; it seems like you're forbidding e.g. `_5` from being treated like a number.

PR #1580, opened 2024-11-05
Syntax highlighting(import)

zoodogood:
The import path now has the same highlighting as the string literal (string.unquoted.module-reference.civet). Previously it could be colored with anything

image

I have been unable to determine the correct method for intercepting the block completion
```civet
with
  type: "json"
```
so this part is not part of the import expression, which is semantically incorrect. Any criticism is welcome, I'm inexperienced


edemaine:
Could you rebase against our main branch? There seems to be a lot of other commits here.

zoodogood:
Oops. I'm redoing the pull request now. Or does it merge into pure history?

zoodogood:
Yeah, I noticed that, too. It's gonna take me a little while

edemaine:
If you setup a remote of `upstream` to refer to our repo/fork, then `git rebase -i upstream/main`, then `git push -f` (after checking everything is still OK), the PR will update.

zoodogood:
> If you setup a remote of `upstream` to refer to our repo/fork (`git remote add upstream git@github.com:DanielXMoore/Civet.git`), then `git rebase -i upstream/main`, then `git push -f` (after checking everything is still OK), the PR will update.  

Thank you, that helped.

lsp/syntaxes/civet.json
```diff
@@ -167,7 +167,22 @@
...
+          "patterns": [
+            {"include": "source.civet"}
+          ]
+        },
+        "4": { "name": "keyword.control.from.civet" },
+        "5": { "name": "string.unquoted.module-reference.civet" }
+      }
+    },
+    {
+      "match": "\\b(?<![\\.\\$])(break|by|case|catch|continue|else|finally|for|in|of|if|return|switch|then|throw|try|unless|when|while|until|loop|do|export|import|default|from|as|yield|async|await|with|(?<=for)\\s+own)(?!\\s*:)\\b",
```
edemaine, line 185: Not related to this PR, but while I look at this, we have a bunch more `for` keywords, such as `each` and the reducers `sum` etc.

zoodogood:
@edemaine I think it's ready. Can you check?

Thank you for taking the time to check the code and suggesting ways to improve it

lsp/syntaxes/civet.json
```diff
@@ -166,6 +166,24 @@
         }
       ]
     },
+
+    {
+      "name": "string.unquoted.module-reference.civet",
+      "match": "(?<=(^\\s*(import|export)|^)\\s*(type\\s+)?((\\w+\\s*,\\s*)?\\{(.|\\n)*\\}|\\w+|(\\*\\s+as\\s+\\w+))\\s+(from)\\s+)([^\\(\\{\\[\"';\\s=>][^\"';\\s=>]+|\".+?\"|'.+?')(?=[^()]|$)"
```
edemaine, line 172: `\\{(.|\\n)*\\}` seems rather inclusive. Perhaps `(.|\\n)` should be replaced by `[^{}]`? Or at least `*` should be `*?`, like the string regexes.

I also realize that, if you're just trying to match unquoted module references here, you don't actually need `|\".+?\"|'.+?'`. (If you keep them, shouldn't `+` be `*`?)

PR #1589, opened 2024-11-09
Fix typo in Coffeescript comparison

bbrk24:
Not only is `CoffeeOf` the wrong capitalization, `coffeeOf` is the wrong directive.


STRd6:
Thanks!

PR #1592, opened 2024-11-10
Fix range `for` loop with complex left-hand side

### LINKED ISSUE #1587: Range for loop doesn't support complex left-hand sides
Input:

```js
for chosen[i] of [0...10]
```

Error:

```
Assertion failed [Internal error: Coffee-style for loop must be an assignment expression]: MemberExpression !== AssignmentExpression
```

edemaine:
Fixes #1587

### DIFF
--- a/source/parser/for.civet
+++ b/source/parser/for.civet
@@ -1,4 +1,5 @@
 import type {
+  AssignmentExpression
   ASTNode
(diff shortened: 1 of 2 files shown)


STRd6 (approved):
Thanks!

PR #1598, opened 2024-11-11
Fix reduction inside conditions, and other subtle aliasing issues

### LINKED ISSUE #1586: Reduction inside `unless` seems broken
```js
return false unless for every f of facesHit
```

Error:

```
Could not find iteration statement in iteration expression
```

edemaine:
Fixes #1586 by preserving aliasing in our objects during `insertTrimmingSpace`. This was tricky to debug! Hopefully this fixes everything.


STRd6 (approved):
Thanks!

PR #1609, opened 2024-11-18
Fix trailing member/call behavior

edemaine:
Fixes @bbrk24's reported regression (which is now a test).
* Call arguments now check for trailing call expressions instead of trailing member expressions.
* I never understood why we had both trailing member and trailing call rules. We no longer use the member ones, so the grammar got a bunch simpler.
* Missing `.flat()` in handling trailing call expressions (else a JSX test would fail).


STRd6 (approved):
Thanks, glad to see things getting simpler!

PR #1612, opened 2024-11-20
Allow postfix in indented `if` conditions

edemaine:
This example came up in Discord: (where `1` and `2` are placeholders for more complex expressions)

```js
if
  1 for every x in y
then
  2
```

Without this PR:

```js
if(
  1) (()=>{let results=true;for (const x in y)if (!(void 0)) {results = false; break};;return results})()(
  2)
```

With this PR:

```js
if((()=>{let results=true;for (const x in y) { if (!(
  1)) {results = false; break} }return results})()) {
  2
}
```

### DIFF
--- a/source/parser.hera
+++ b/source/parser.hera
@@ -5218,7 +5218,7 @@ Condition
     }
 
   # NOTE: Indented condition allows for further nested function application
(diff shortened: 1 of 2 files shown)


STRd6 (approved):
Definitely seems better this way

PR #1613, opened 2024-11-20
`for` reduction implicit body can destructure, fix implicitly returned patterns in some cases

edemaine:
My initial goal was to make `for sum [x] of y` use an implicit body of `x` instead of `[x]`.

Along the way, I ran into some bugs in `patternAsValue`, which led to cleaning up of some pattern matching code to correctly handle the previous addition of `BindingElement`.


STRd6 (approved):
Thanks! Glad to see the patterns and bindings getting some cleanup.

PR #1615, opened 2024-11-21
Fix fallthrough in implicitly returned switch with semicolon, improve `hasExit` heuristic

### LINKED ISSUE #1602: Semicolon causes implicit fallthrough in switch
When the `switch` is the last statement in the function, each branch gets an implicit `return`. If you use `;` to suppress the `return`, there is no `break` to replace it, so the branch falls through:
```civet
->
  switch foo
    when 0
      // implicit `break`
    else
      console.log 'Not zero'
  switch bar
    when 0
      /* implicit `return` */ 0
    when 1
      ; // no `break` or `return` -- falls through!
    else
      throw new RangeError 'bar must be 0 or 1'
```

edemaine:
Fixes #1602

This was trickier than expected. One fun issue is that SemicolonDelimiter can end with a trailing Comment. This hasn't been an issue in the past because we avoid adding `return` in such scenarios. But now we need to, so I added a `hasTrailingComment` helper for this.

### DIFF
--- a/source/parser.hera
+++ b/source/parser.hera
@@ -2807,7 +2807,7 @@ SingleLineStatements
       maybeComment?.type === "Comment" && maybeComment.token.startsWith("//")
 
     const children = [expressions]
-    if (hasTrailingComment) children.push("\n")
+    if (hasTrailingComment) children.push(["\n"])
 
     return {
       type: "BlockStatement",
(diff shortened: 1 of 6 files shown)


STRd6 (approved):
Nice!

PR #1626, opened 2024-11-29
make tab key insert tabs and set tab size to 4 in playground

samualtnorman:
these changes make it so that you can easily use tabs in the playground


edemaine:
Is this accessible? How would a keyboard user navigate away from the text box? (We've discussed the escape key working in this case. Is that automatic somehow or do you need to add code?)

STRd6:
I think using escape to unfocus the code editor and allow navigating to the next interactive element with tab is a good idea.

It would also be nice to indent every line within the selection, if present, as well.

samualtnorman:
escape to unfocus sounds like a good idea, indenting every selected line sounds a little tricky to implement so probably out of scope for this PR

STRd6 (approved):

PR #1627, opened 2024-12-02
Fix implicit `async` and `*` in methods

### LINKED ISSUE #1621: `do*` works differently in top-level functions vs in methods
At the top level:
```js
->
  do*
   yield foo
   yield bar
// vvv
function () {
  return (function* () {
    {
      yield foo;
      yield bar;
    }
  })();
}
```
In a method:
```js
{
  f()
    do*
     yield foo
     yield bar
}
// vvv
{
  *f() {
    (function* () {
      {
        yield foo;
        yield bar;
      }
    })();
  }
}
```
I think the latter is a bug, since it never returns or yields anything.

edemaine:
Fixes #1621 by unifying implicit async/generator handling in methods with functions. Previously, methods did the implicit additions too early, during the grammar which is before `async do` and `do*` gain wrappers.

### DIFF
--- a/source/parser.hera
+++ b/source/parser.hera
@@ -3727,60 +3727,22 @@ MethodDefinition
       abstract: true,
       signature,
       parameters: signature.parameters,
-      ts: true
+      async: signature.async,
+      generator: signature.generator,
+      ts: true,
     }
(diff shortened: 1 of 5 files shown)


STRd6 (approved):
:clap:

PR #1628, opened 2024-12-02
Fix object literals and bulleted lists in JSX indented attributes, fix indentation detection

### LINKED ISSUE #1622: JSX indented attributes
This works:

```js
<.glue style=
  background: "white"
>
```

This fails to parse:

```js
<.glue style=
  background: "white"
  color: "black"
>
```

I'm guessing we need to add explicit support for nested object literals, and probably also bulleted lists.

edemaine:
The `MaybeNested...Expression` rules were surprisingly quite broken. Can you spot the bug?

```js
MaybeNestedExpression
  NestedBulletedArray
  PushIndent ( Nested Expression )? PopIndent ->
    if ($3) return $3
    return $skip
  Expression
```

"Luckily" `PopIndent` was always false, so the indented rule never applied. But we weren't correctly tracking indentation in this case.

With this fixed, we no longer supported dedented trailing call expressions after a nested expression. So I added `AllowedTrailingCallExpression?` after `PopIndent` in the rule above (and its friends).

But adding that revealed a different bug.  In particular, the following was treated as a call before:

```js
x =
    a
  b
```

So I modified `AllowedTrailingCallExpression` to not check for implicit calls; they shouldn't be relevant for trailing call expressions.

With all that fixed, it was relatively easy to fix #1622 JSX indented attributes to use something like this. I couldn't use the exact mechanism because (1) it didn't want the `Maybe` and (2) I didn't want to allow for dedented trailing call expressions in JSX, because e.g. a dedented `.foo` should be treated as JSX `class` shorthand, not a trailing member access.

### DIFF
--- a/source/parser.hera
+++ b/source/parser.hera
@@ -5472,30 +5472,37 @@ Debugger
 
 MaybeNestedNonPipelineExpression
   NestedBulletedArray
-  PushIndent ( Nested NonPipelineExpression )? PopIndent ->
-    if ($3) return $3
-    return $skip
+  NestedImplicitObjectLiteral
+  PushIndent ( Nested NonPipelineExpression )? PopIndent AllowedTrailingCallExpressions? ->
+    if (!$2) return $skip
+    if (!$4) return $2
+    return [ $2, $4 ]
   NonPipelineExpression
 
 MaybeNestedPostfixedExpression
   NestedBulletedArray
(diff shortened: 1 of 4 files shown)


STRd6 (approved):
Nice, good clean up!

The indentation stuff is always a little zany.

PR #1630, opened 2024-12-02
Fix placeholders in `for each` loops

### LINKED ISSUE #1603: `for each` with `&`
## Input

```js
for each bit of &
  Number bit
```

## Output

```js
$ => { const results=[];for (let i = 0, len = &.length; i < len; i++) {$ => { const bit = $1[i];return};
  results.push(Number(bit))
};return results;}
```

Lots wrong here: `&.length`, two levels of `$ =>`

By contrast, `for` (without `each`) works fine.

edemaine:
Fixes #1603

There might be other situations where we need to `implicitLift` (and maybe it will need to be a number instead of just 1 or 0) but at least we have a mechanism now.

(also open to better names than "`implicitLift`")

### DIFF
--- a/source/parser/for.civet
+++ b/source/parser/for.civet
@@ -296,6 +296,7 @@ function processForInOf($0: [
         type: "Declaration"
         children: [declaration, " = ", trimFirstSpace(expRef), "[", counterRef, "]"]
         names: assignmentNames
+        implicitLift: true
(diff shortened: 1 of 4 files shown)


STRd6 (approved):
Looks good. I can't come up with a better name than `implicitLift` on the spot.

PR #1631, opened 2024-12-02
Fix mixing access modifiers (e.g. `readonly`) with `@` arguments in constructors

### LINKED ISSUE #1601: TypeError when compiling `readonly @prop` in c'tor
both of these work fine:
```
class Foo
  @(readonly bar)

class Foo
  @(@bar)
```

but you can't combine them.

edemaine:
Fixes #1601

### DIFF
--- a/source/parser.hera
+++ b/source/parser.hera
@@ -2001,6 +2001,7 @@ NWBindingIdentifier
       type: "AtBinding",
       children: [ref],
       ref,
+      names: [],
     }
   # Likewise for private identifiers
   Hash AtIdentifierRef:ref ->
@@ -2009,6 +2010,7 @@ NWBindingIdentifier
       type: "AtBinding",
       children: [ref],
       ref,
+      names: [],
     }
   Identifier:id
   # NOTE: Support for return := 1 and let return: number
--- a/source/parser/binding.civet
+++ b/source/parser/binding.civet
@@ -2,7 +2,6 @@ import type {
   ArrayBindingPattern
   ASTNode
   ASTNodeObject
-  AtBinding
   BindingPattern
   BindingRestElement
   Children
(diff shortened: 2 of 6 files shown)


STRd6 (commented):
We may need to handle the TS output slightly differently since it looks like TS also inserts the assignment: https://www.typescriptlang.org/play/?#code/MYGwhgzhAEBiD29oG8CwAoa1jwHYQBcAnAV2APiIAoiBTMAEzxAE9oAjMIgSmQIAsAlhAB0nItAC8HLgG4AvhnlA

edemaine:
I agree that it's redundant, but at least the double assignment is not incorrect...

I could imagine the current behavior of this PR being somewhat helpful in some scenarios:
* When type checking, you want `readonly` typing
* When compiling with `js: true`, you want the assignment

Probably better would be for Civet to support `readonly` (and `public` etc.) in `js: true` mode, turning into an assignment just like an `@` binding. I assume other TypeScript transpilers support this? Yes, [esbuild does](https://esbuild.github.io/try/#dAAwLjI0LjAAeyBsb2FkZXI6ICJ0cyIgfQBjbGFzcyBGb28gewogIGNvbnN0cnVjdG9yKHJlYWRvbmx5IGJhcil7dGhpcy5iYXIgPSBiYXI7fQp9). Anyway, then we could detect and remove the duplicate, or rather, trigger once when there's an `@` binding or has a TypeScript specifier.

STRd6 (approved):

edemaine:
> Probably better would be for Civet to support `readonly` (and `public` etc.) in `js: true` mode, turning into an assignment just like an `@` binding

Actually, it looks like we already did this, some time ago.

Here's a fun situation:

```js
class Foo
  @(public @var)
```

Currently we compile this to:

```js
class Foo {
  constructor(public _var){this.var = _var;}
}
```

This declares the wrong thing `public`. But we also don't want to omit the `_`. I think we need to compile to

```js
class Foo {
  public var
  constructor(_var){this.var = _var;}
}
```

So I'm tempted to lift `public`/`readonly`/etc. to the type declaration, using the existing type declaration lifting. But unlike type lifting, we *won't* duplicate the modifiers; we'll remove them from `constructor`.

edemaine:
OK, there shouldn't be any duplicate `this` assignment when mixing access modifiers with `@` bindings now, using the alternate compilation above.

STRd6 (approved):
Awesome!

PR #1632, opened 2024-12-03
Range literal `[a..b]` defaults to increasing; adaptive behavior behind `"civet coffeeRange"` directive

edemaine:
Now that we have the ability to change the range literal direction via `<`/`>`/`<=`/`>=` modifiers, change the default behavior to `<=`. Old behavior available via `"civet coffeeCompat"` or `"civet coffeeRange"`.

Arguments for this change in default behavior from Discord:
* Brings range literal `[a..b]` into alignment with slices `x[a..b]` (which were already increasing by default, unless you explicitly write `>` or `>=`)
* Avoids bugs where you write e.g. `[1...n]` and fail to consider the case `n = 0`, so you end up with `[1, 0]` whereas you meant `[]`.

BREAKING CHANGE: `[a..b]` and `[a...b]` no longer creates a decreasing range without an explicit `>` or `>=`; add one where needed.

~~TODO: fix tests~~
~~TODO: some character loops are breaking~~


edemaine:
Turns out `for x of ['a'<..'z']` was previously broken in two ways. Now fixed and ready for review!

STRd6 (approved):
Thanks!

PR #1635, opened 2024-12-05
`"""` interpolations require `coffeeInterpolation`, `///` respects `coffeeInterpolation`, `coffeeComment`, `coffeeDiv`

edemaine:
* Bring `"""` strings into alignment with `"` strings: `#{expr}` interpolation is supported only when `coffeeInterpolation` is set. Use ` ``` ` and `${expr}` to interpolate
* Bring `///` regexes into alignment with this: `#{expr}` interpolation is supported only when `coffeeInterpolation` is set. Use `${expr}` interpolation instead.
* Fix a bug where `///foo${expr}///` did not interpolate
* Change comment behavior in `///`:
  * `#` is a comment only when `coffeeComment` is enabled
  * `//` is a comment only when `coffeeDiv` is disabled (the idea is that `coffeeDiv` sacrifices `//` working as a comment)

BREAKING CHANGE: `"""` and `///` no longer support `#{expr}` interpolation unless you set `coffeeInterpolation`; `///` no longer supports `#` comments without `coffeeComment`; `///` no longer supports `//` comments with `coffeeDiv`

### DIFF
--- a/civet.dev/config.md
+++ b/civet.dev/config.md
@@ -69,11 +69,11 @@ For now, we have the following related options:
 | [`coffeeBooleans`](reference#coffeescript-booleans) | `yes`, `no`, `on`, `off` |
 | [`coffeeClasses`](reference#coffeescript-classes) | CoffeeScript-style `class` methods via `->` functions |
 | [`coffeeComment`](reference#coffeescript-comments) | `# single line comments` |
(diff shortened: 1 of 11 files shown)


STRd6 (approved):
Thanks! Definitely better to get this sorted out now rather than years from now.

PR #1636, opened 2024-12-05
Fat pipe `||>` uses semicolons at statement level

### LINKED ISSUE #1339: Fat pipe (`||>`) does not play nice with typescript-eslint v8
Evidence: bbrk24/Trilangle#142

> ```error  Expected an assignment or function call and instead saw an expression  @typescript-eslint/no-unused-expressions```

edemaine:
Fixes #1339

Functionally equivalent, but good for linting. Also generates somewhat nicer looking code, because implicit `return`s get moved to the end of a chain.

It was a little trickier than the idea in https://github.com/DanielXMoore/Civet/issues/1339#issuecomment-2495547043 because of implicit returns.

### DIFF
--- a/source/parser/function.civet
+++ b/source/parser/function.civet
@@ -451,6 +451,13 @@ function assignResults(node: StatementTuple[] | ASTNode, collect: (node: ASTNode
       for each block of exp.blocks
         assignResults block, collect
       return
+    when "PipelineExpression"
+      // At statement level, pipeline generates semicolons between statements
+      // Return the last one
+      semi := exp.children.indexOf ";"
+      if 0 <= semi < exp.children# - 1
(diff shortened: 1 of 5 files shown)


source/parser/function.civet
```diff
@@ -573,6 +580,13 @@ function insertReturn(node: ASTNode): void
         insertReturn block
       // NOTE: do not insert a return in the finally block
       return
+    when "PipelineExpression"
+      // At statement level, pipeline generates semicolons between statements
+      // Return the last one
+      semi := exp.children.indexOf ";"
```
STRd6, line 586: Should this be `lastIndexOf` or is it always only going to have one?

edemaine: Indeed! Thanks for catching!

PR #1640, opened 2024-12-07
Fix #1637: prevent parser crash on blocks inside switch

### LINKED ISSUE #1637: TypeError with object literal in switch?
```civet
nextLocation := switch direction
    when Direction.Up
        { location.y - 1 }
```
errors so badly the LSP refuses to elaborate. Best I can tell it's calling `undefined.some()` which is erroring

STRd6:
`node` is sometimes the Statement tuple so using `exp` which is always the actual block. TS actually caught the error.

Note that this is being interpreted as a block statement not an object literal at present.


edemaine (approved):
Whoops! Good catch. This PR should also fix the labeled statement case.

PR #1642, opened 2024-12-12
`break/continue loop/while/until/for/do` refer to anonymous containing iteration

edemaine:
As mentioned in Discord, this seems in line with `continue switch` and fairly intuitive.

It is technically a breaking change, but these seem like unlikely constructions:
* old `break loop` (infinitely many `break`s) is equivalent to `break`
* old `break while cond()` is equivalent to `break if cond()` (and is now a parse error)
* old `break until cond()` is equivalent to `break unless cond()` (and is now a parse error)

### DIFF
--- a/civet.dev/reference.md
+++ b/civet.dev/reference.md
@@ -2107,6 +2107,21 @@ except for Civet reserved words (e.g. `and`) where a colon is required.
 JavaScript reserved words are invalid as labels.
 :::
 
+Iterations also get implicit labels if you refer to them by type,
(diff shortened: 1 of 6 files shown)


STRd6 (approved):
Nice, looks good!

For the compatibility of previous `break while cond()` is closer to equivalent to `cond()` since it checks the cond and breaks from inside the wile, it wouldn't affect outside. 

I don't think we need to worry about it since probably no one used it unless by accident.

PR #1648, opened 2024-12-18
Non-end rest parameters support types

### LINKED ISSUE #929: Rest elements at middle/end position and types
When middle/end positioned rest params/items are used with typescript types the type is wrongly moved into the destructuring which is also invalid ts.
```civet
function foobar(...foo: number[], bar: string)
  ;
```
⬇️⬇️⬇️
```ts
function foobar(...foo: number[]) {
  let [bar: string] = foo.splice(-1);
  ;
}
```
[playground](https://civet.dev/playground?code=ZnVuY3Rpb24gZm9vYmFyKC4uLmZvbzogbnVtYmVyW10sIGJhcjogc3RyaW5nKQogIDs%3D)

A possible solution would be to both use a spread in the function signature and an assertion with the splice, because `Array.prototype.splice` otherwise returns the union from the tuple's items.

```ts
function foobar(...foo: [...number[], string]) {
  let [bar] = foo.splice(-1) as [string];
  ;
}
```

Even better version gets the type of `foo` correct ([playground](https://tsplay.dev/N92K7m)):

```ts
function foobar(..._args: [...number[], string]) {
  let [bar] = _args.splice(-1) as [string];
  let foo = _args as number[]
  return bar
}
```

edemaine:
Fixes #929 by
* Typing new end rest parameter as tuple type
* Add type casting of `splice` return value
* Add ref to type cast remaining array, but only for TypeScript output

Lots of other internal cleanup along the way:

* Fix traversal functions to allow non-`Predicate` arguments instead of erroring. In this case, they just don't filter.
* Move `NonEmptyParameters` handler into `processParams`
* Uniform use of `PostRestBindingElements` in `blockPrefix`
* `Parameters` node now has `parameters` property with array of actual parameters
* Internal type fixes

### DIFF
--- a/source/parser.hera
+++ b/source/parser.hera
@@ -792,6 +792,7 @@ ArrowFunction
         modifier: {
           async: !!async.length,
         },
+        parameters,
         returnType,
       },
       parameters,
--- a/source/parser.hera
+++ b/source/parser.hera
@@ -48,6 +48,7 @@ import {
   modifyString,
   negateCondition,
   prepend,
+  preprocessParams,
   processAssignmentDeclaration,
   processBinaryOpExpression,
   processCallMemberExpression,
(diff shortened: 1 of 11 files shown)


STRd6 (approved):
Thanks, great improvement! I’m glad the TS types ended up being so simple.

PR #1651, opened 2024-12-19
Pin parameter `^p` assigns function parameter to outer variable

edemaine:
As discussed on Discord, this seems like a natural extension of our pinned patterns. Originally motivated by simulating `Promise.withResolvers` in this natural way:

image


STRd6 (approved):
👏

PR #1653, opened 2024-12-20
List of ASCII symbols in cheatsheet

edemaine:
This started as an exercise to see which symbols we use the least (e.g. `~`). But I think it could be helpful also as:
* a list of operator-based features
* reading code (what does `#` mean here?)

I'm open to better ideas of where to put it, though. Also let me know if I missed some!


civet.dev/cheatsheet.md
```diff
@@ -315,3 +315,39 @@ floor # / 2
...
+
+## ASCII Symbols
+
+Here is a list of ASCII symbols and their various contextual meanings in Civet.
+(There are also more [Unicode symbols](/reference#unicode-operators).)
+
+| Symbol | Meanings |
+|--------|----------|
+| space | Function call |
+| `?` | Non-null test `x?`; non-null access `x?.y`, `x?y`, `x?(...)`, `x?[...]`; null coalescing `x ?? y` (`x ? y` with `coffeeBinaryExistential`); TypeScript optionals `(x?: number) => x`; if/then/else ternary `x ? y : z` |
```
bbrk24, line 327: Optional types `number?`, `number??`, optional declarations `let x?: number`, optional check `if x? := foo()`

As an aside, `let x?` with no type annotation is currently left alone during compilation.

civet.dev/cheatsheet.md
```diff
@@ -315,3 +315,39 @@ floor # / 2
...
+| `!` | Negation `!x`; object flag `{!x}`; TypeScript non-null assertion `x!`; non-null assertion access `x!.y`, `x!y` |
```
bbrk24, line 328: Non-nullable types `let x: T!`, force-cast `as!`

Should we also mention that it can be interleaved with certain operators like `!in`?

civet.dev/cheatsheet.md
```diff
@@ -315,3 +315,39 @@ floor # / 2
...
+| `#` | private properties `x.#y`, `#y`; length shorthand `x#`, `#`; block comments `### ... ###`; one-line comments `# ...` with `coffeeComment`; CoffeeScript string interpolations `"x#{y}"` with `coffeeInterpolation` |
+| `$` | JavaScript identifier character; template interpolations `` `x${y}` `` |
+| `_` | JavaScript identifier character |
+| `%` | Modulo operators `x % y`, `x %% y`; modulo indexing `x[i %]`; integer division `%/` |
+| `^` | Binary XOR `x ^ y`; logical XOR `x ^^ y`; logical XNOR `x !^ y`, `x !^^ y`; pin patterns `(^x) =>` |
+| `&` | Binary AND `x & y`; logical AND `x && y`; TypeScript AND `number & string`; placeholder for function shorthand `&[x]+1` |
+| `\|` | Binary OR `x \| y`; logical OR `x \|\| y`; TypeScript OR `number \| string`; pipeline `\|>`, `\|\|>` |
+| `+` | Addition `x + y`; positive `+x`; increment `x++`, `++x`; object flag `{+x}`; concatenation `x ++ y` |
+| `-` | Subtraction `x - y`; negation `-x`; decrement `x--`, `--x`; object flag `{-x}`; function arrows `->`; property name symbol `x-y: z` |
+| `*` | Multiplication `x * y`; exponentiation `x ** y`; generators `function*`, `*method()`, `for*`; comments `/* x */` |
```
bbrk24, line 339: We might want to explicitly mention `yield*` as a special thing relating to generators

civet.dev/cheatsheet.md
```diff
@@ -315,3 +315,39 @@ floor # / 2
...
+| `/` | Division `x / y`; integer division `x %/ y`; comments `// x`, `/* x */`; regular expressions `/x/`, `///x///`; TypeScript triple-slash directive `/// <reference/>` |
+| `\` | String escape `"\""`; regular expression escape `/\//` |
+| `=` | Assignment `x = y`, `x op= y`; declaration shorthand `.=`, `:=`, `::=`; function arrows `=>` |
```
bbrk24, line 342: `==`, `!=`, `===`, `!==` (also for `!`)

civet.dev/cheatsheet.md
```diff
@@ -315,3 +315,39 @@ floor # / 2
...
+| `~` | Binary NOT `~x` |
+| `.` | Property access `x.y`; property access function `.x`; argument placeholder `f(x, .)`; bulleted list `. x` (after indentation); let shorthand `x .= y`; range literals `[x..y]`; spreads `...x` |
+| `:` | Labels `:x`; symbols `:x`; object properties `x: y`, `{x: y}`; TypeScript types `let x: number`, `(x: number) =>`; property typing `{x:: number}`, `[x:: number]`; if/then/else ternary `x ? y : z`; const shorthand `x := y`; type shorthand `T ::= x`; prototype shorthand `x::y` with `coffeePrototype` |
+| `,` | Argument separator `f x, y`; comma operator `x, y` |
+| `'` | Strings `'hello'`, `'''hello'''`; template strings `f'hello'`; property access `x's y` |
+| `"` | Strings `"hello"`, `"""hello"""`; template strings `f"hello"` |
+| `` ` `` | Template strings `` `x${y}` ``, ` ```x${y}``` `, ``f`x${y}` `` |
```
bbrk24, line 349: For a moment I misread `f"hello"` like a Python-style f-string. Maybe "tag template" or similar to clarify it's the function call kind?

civet.dev/cheatsheet.md
```diff
@@ -315,3 +315,39 @@ floor # / 2
...
+| `<`/`>` | Comparison operators `x < y > z`; JSX `<div>`; TypeScript parameter `Set<number>`; range literals `[x<..<y]`; function arrows `->`, `=>`; pipelines `|>`, `||>` |
```
bbrk24, line 350: `<` extends; `<:` implements (also under `:`), `<?` typeof/instanceof (also under `?`)

bbrk24: Also `>` can be used for backwards extends in type conditionals: `type A<B> = B > C ? D : E` compiles to `type A<B> = C extends B ? E : D`.

civet.dev/cheatsheet.md
```diff
@@ -315,3 +315,39 @@ floor # / 2
...
 </div>
+
+## ASCII Symbols
+
+Here is a list of ASCII symbols and their various contextual meanings in Civet.
+(There are also more [Unicode symbols](/reference#unicode-operators).)
+
+| Symbol | Meanings |
+|--------|----------|
+| space | Function call |
```
bbrk24, line 326: Actually, if we're listing space here, it can also be a TS type argument (`Set number` = `Set<number>`)

STRd6 (approved):
Awesome, I’ve always wanted something like this! Eventually we can also add a search/filter box so we can type in a symbol and see what it could be.

civet.dev/cheatsheet.md
STRd6: And JSX code children 🤫

civet.dev/cheatsheet.md
edemaine: I wonder what `let x?` should mean. I think most intuitively it would say "allow `x` to be unassigned on first access" but I'm not sure this is possible in TypeScript. Perhaps a helpful error message for now.

PR #1654, opened 2024-12-21
`let x?` allows initial `undefined` value + type inference; `let x? = y` for simple `y`

edemaine:
Previously, `let x?` compiled to the invalid TypeScript `let x?`. Now it compiles to `let x = undefined`. As [this playground shows](https://www.typescriptlang.org/play/?#code/CYUwxgNghgTiAEA3W8wHsB2wBc8BGaaEIUGAsAFDEAu8AHvALzwCuWIAZgJYYjCXoMAZyIgAdBDQBzABR0AlJXjKVqgPRqAegH5KXDvBmDg8+AG8l9JvACslhgGpmARgAMlAL4DMI4hOlyihSqIfAaOkA), this seems to allow for the initial `undefined` value (no error about access before initialized) as well as type inference.

This PR also makes `let x? = 5` compile to `let x = 5` (instead of leaving the invalid `?` in there). I wanted to compile it to `let x: undefined | typeof 5 = 5`, but sadly this [isn't valid TypeScript](https://www.typescriptlang.org/docs/handbook/2/typeof-types.html#limitations):

> TypeScript intentionally limits the sorts of expressions you can use `typeof` on. Specifically, it’s only legal to use `typeof` on identifiers (i.e. variable names) or their properties.

### DIFF
--- a/civet.dev/reference.md
+++ b/civet.dev/reference.md
@@ -2548,7 +2548,7 @@ function f(x?: string)?: string
   x
 </Playground>
 
-More generally, `T?` allows for `undefined` and
+More generally, type `T?` allows for `undefined` and
 `T??` additionally allows for `null`:
 
 <Playground>
@@ -2560,6 +2560,14 @@ To allow for type inference and the initial `undefined` value:
 
 <Playground>
 let x?
+</Playground>
+
+To allow for later assignment of `undefined` while specifying an initial value
+(currently limited to a literal or member expression):
+
+<Playground>
+let x? = 5
+let y? = x
 </Playground>
 
 ### Non-Null Types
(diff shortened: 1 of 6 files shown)


STRd6 (approved):
A good improvement. We could check for common literal types: string, number, bigint, and generate the corresponding type | undefined in those cases.

edemaine:
Ah, good idea. I had actually written code to do the rewriting into `typeof` before realizing that it didn't always work, so I could use that, modified to work in just the good cases (member expression and identifier). And I added a new `literalType` helper for literal types, and error in the other cases.

STRd6 (approved):

Issue #1649, opened 2024-12-18
Guaranteed `break with`


edemaine:
This seems like a useful construction, analogous to `Array.prototype.find`:

```js
result := for x of y when condition(x)
  break with x
```

The generated code

```js
let results;results=[];for (const x of y) {if (!condition(x)) continue;
  results = x;break
};const result =results
```

sets `results = []`. It would type better without that. We can detect this by seeing that the loop always exits.

EDIT: I realize that this wouldn't really work, because the loop might not iterate at all, in which case, what should `result` be? Until we have a way to specify the initial value, I don't think this proposal makes sense.

---

Related, I wonder if there should be a reducer for this, called `for find` or `for first`. It's similar to `for some` but returning the first item instead of a boolean.

PR #1657, opened 2024-12-23
`for concat` reduction for concatenating arrays, `for first` reduction for finding elements

edemaine:
`for concat` concatenates using `++`. Like #1656, this is another approach to flattening, which I imagine would be useful in some cases. For example, `array.flat(1)` is equivalent to

```js
for concat subarray of array
```

---

`for first` returns the first body, as suggested in #1649, or else `undefined`. For example, this can simulate `Array.prototype.find`: `array.find(cond)` is equivalent to

```js
for first item of array when cond(item)
```

If `cond` isn't already a function, though, it can make nicer code (no function wrapper necessary).
`for first` also lets you choose *what* to return when found. For example, you can simulate `array.findIndex(cond)` like so:

```js
for first item, i of array when cond(item)
  i
```

I wasn't sure whether to return the first *truthy* body. This would behave more like `for some`, just changing the return value. But I think just returning the first body offers more consistent behavior (maybe you want to return a falsey value), and you can add truthy checking via `when`. But this is partly why it's called `for first`, not `for find`.

### DIFF
--- a/civet.dev/reference.md
+++ b/civet.dev/reference.md
@@ -1964,6 +1964,19 @@ numEven := for count item of array
 numKeys := for count key in object
 </Playground>
 
+`for first` returns the first body value.
+If there is no body, it uses the item being looped over.
+Combined with a `when` condition, this can act like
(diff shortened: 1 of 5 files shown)


STRd6 (approved):
Nice!

PR #1663, opened 2024-12-27
Pattern `name^: value` binds `name`, while `name: value` never does

### LINKED ISSUE #1660: Pattern match creates unused variable
```coffee
switch person
  {age: 100}
    console.log("old")
```

generates:

```typescript
if (
  typeof person === "object" &&
  person != null &&
  "age" in person &&
  person.age === 100
) {
  const { age } = person;
  console.log("old");
}
```

Which causes this TypeScript error:
`'age' is declared but its value is never read.typescript(6133)`

edemaine:
Fixes #1660 by no longer binding `name` in object pattern properties with literal values: `name: "string"`, `name: 42`, or `name: /regex/`. The general rule is now this: if a property has a value, then it doesn't get bound. This seems like a natural extension of how destructuring assignment works.

New feature `name^: value` binds `name`. This also works when checking other properties. For example, the pattern `{parent^: {type: "BlockStatement", expressions}}` binds `parent` and `expression` in two destructurings: `{parent} = x, {expressions} = parent`.

BREAKING CHANGE: Pattern matching with `{name: literal}` no longer binds `name`; use `{name^: literal}` to bind.

### DIFF
--- a/civet.dev/reference.md
+++ b/civet.dev/reference.md
@@ -1646,6 +1646,16 @@ switch x
     type
 </Playground>
 
+Object properties with value matchers are not bound by default (similar to
+[object destructuring assignment](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment)).
+Add a trailing `^` to bind them:
+
+<Playground>
+switch x
+  {type^: /list/, content^: [first, ...]}
+    console.log type, content, first
+</Playground>
+
 Use `^x` to refer to variable `x` in the parent scope,
 as opposed to a generic name that gets destructured.
 (This is called "pinning" in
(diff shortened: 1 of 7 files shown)


STRd6 (approved):
Looks good! It seems natural to extend to regular destructuring at some point as well.

edemaine:
Oh, that's a fun idea, like `{x^: [y]} := z`. (It already works within `if`.)

I wonder if there's a natural way to do it at the top level too, not just for object properties. Like I want `function Comp(props)` to bind `props` but also destructure... Something like `function Comp(props^: {name})`. This is related to the desire to have `as` in a pattern, like `[a, b] as arr`. I'm not sure we want to call it `as` because it's not a type operation. So maybe something involving `^`.

PR #1666, opened 2024-12-30
Worker threads fixes and SourceMap revamp

edemaine:
**Worker threads** (`CIVET_THREADS` feature) was broken in a few ways:
* Errors weren't serialized, so we got the empty `{}` exception if there was an error
* `errors` option didn't properly modify the array
* SourceMap outputs were crashing when serializing into a worker response message (because of function properties), which caused an error (see below). This caused `yarn test` to fail (before #1665), and also caused `civet filename.civet` to fail. 🙀 

**Internal API change:** `SourceMap` is now a `class`. This seems cleaner overall, and helps with serialization for worker messages. But it changes the API:
* Instead of `SourceMap(source)` we need to use `new SourceMap(source)`.
* `lines` etc. are properties instead of part of the `data` field. In code that deals with multiple versions of Civet (e.g. LSP), I included a fallback for the old format.
* `source` is now a property, not a function. I don't think we use it anywhere though.
* `updateSourceMap` needs to be `bind`ed

Also the ESM register plugin tells you which file is erroring (which has been an issue in the past).

Also a small fix so that `CIVET_THREADS=0 yarn test` disables Mocha's parallel mode (to bypass `export CIVET_THREADS=4`). Previously `0` was treated as truthy.

### DIFF
--- a/build/test.sh
+++ b/build/test.sh
@@ -5,7 +5,7 @@ set -euo pipefail
 # Translate CIVET_THREADS into Mocha's --parallel
 # In particular, CIVET_THREADS Workers don't work within Mocha
 args=""
-if [ ! -z "${CIVET_THREADS:-}" -a "$CIVET_THREADS" != 0 ]; then
+if [ "${CIVET_THREADS:-0}" != 0 ]; then
   args="--parallel -j $CIVET_THREADS"
   export CIVET_THREADS=
 fi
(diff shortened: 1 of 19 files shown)


STRd6 (approved):
Thanks! We're finally migrating away from the CoffeeScript circa 2011 coding style :P

PR #1668, opened 2025-01-03
Named binding patterns `name^pattern` in pattern matching, function parameters, declarations, `for` loops; fix complex bindings in `for` loops

edemaine:
New features:
* `name^pattern` or `name^ pattern` is equivalent to `pattern`, but also binds the whole value to `name`.
  * For example, `function Comp(props^{a, b})` binds `props`, `a`, and `b`
  * This is like `name@pattern` in Haskell, and `pattern as name` in Ocaml and Python. It's motivated by our existing `{name^: pattern}` syntax.
  * It works in pattern matching, function parameters, and declarations (`:=`, `.=`, `const`, `let`, `var`).
  * It works in `=` assignment, ~~but not yet at the top level~~. Note a small **ambiguity**: `x^{y} = z` previously meant `x ^ ({y} = z)`. I think this is worth overriding, as it's not useful to binary XOR with an object or array. (It's also not particularly critical at the top level, as all it's doing is making a copy of `z`. But I think it's cleaner and more symmetric to allow it.)
  * It works in `for ... of/in` loops. These were actually broken with existing `@prop` and splices. The following now work and did not before:
    * `for @x of y`
    * `for [...x, y] of z`
* `{name^: pattern}` is now shorthand for `{name: name^pattern}`, and it is now implemented this way.
  * As a consequence, this now works in function parameters and declarations and assignments and `for` loops, not just pattern matching, as suggested in https://github.com/DanielXMoore/Civet/pull/1663#issuecomment-2563886636
* In pattern matching, `name < 5` now acts like the pattern `< 5` but also binds the matching value to `name`.
  * This is especially useful for matching exceptions. We can now use `try ... catch e <? RangeError` to both match and get the exception `e`.

There's a new `gatherSubbindings` helper that maybe should be part of `gatherBindingCode`. They're now almost always used together. However, in pattern matching, they're used with somewhat different arguments, so I haven't been able to unify yet. (There's also lots of `splices` and `thisAssignments` code that's mostly duplicated in various places, so I guess this is par for the course.)

TODO: Documentation (once the feature/syntax is approved)

### DIFF
--- a/civet.dev/reference.md
+++ b/civet.dev/reference.md
@@ -1631,9 +1631,7 @@ switch x
     console.log "leading type:", type
 </Playground>
 
-::: info
-You can also use condition fragments as patterns.
-:::
+You can also use condition fragments as patterns:
 
 <Playground>
 switch x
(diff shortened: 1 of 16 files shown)


STRd6 (approved):
Thanks, very nice improvements!

Seeing the examples helps me appreciate the syntax. It seems powerful and natural with room to expand in the future as well.

PR #1669, opened 2025-01-04
Fix property `::` typing with private fields

### LINKED ISSUE #1667: Destructuring type error in private field
https://civet.dev/playground?code=CmNsYXNzIENodW5rIHsKICBAKHsKICAgI2Zvbzo6IHN0cmluZwogICAjYmFyOjogbnVtYmVyCiAgfSkKfQ%3D%3D

```civet
class Chunk {
  @({
   #foo:: string
   #bar:: number
  })
}
```

```ts
class Chunk {
  #foo: string;
  #bar: number;
  constructor({
    foo,
    bar,
  }: {
    #foo: string;
    ^^^^^^^^^^^^^
    #bar: number;
    ^^^^^^^^^^^^^
  }) {
    this.#foo = foo;
    this.#bar = bar;
  }
}

```

edemaine:
Fixes #1667 

Also move `populateRefs` to ref.civet where it seems to belong.  (No functional changes to this function, just Civetification.)

### DIFF
--- a/source/parser/binding.civet
+++ b/source/parser/binding.civet
@@ -3,7 +3,6 @@ import type {
   ASTNode
   ASTNodeObject
   BindingPattern
-  BindingProperty
   BindingRestElement
   Children
   HasSubbinding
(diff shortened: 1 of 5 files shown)


STRd6 (approved):
Thanks! I appreciate the comment about `base` and `id` as well!

edemaine:
Yeah, I kept getting confused between them. 😅

PR #1670, opened 2025-01-04
`yield*` in pipeline

### LINKED ISSUE #1671: `yield*` not supported in pipelines (unless surrounded with parens)
🚫 These do not work as expected
```js
foo |> yield*
// expected:
//   yield * foo;
// got:
//   foo |> yield*
//               ^ Failed to parse

foo |> yield* bar
// expected:
//   (yield * bar)(foo);
// foo |> yield* bar
//             ^ Failed to parse
```

✅ These all **do** work:

```js
foo |> (yield*)  //=> yield * foo;
foo |> (yield *) //=> yield * foo;
foo |> yield bar // => (yield bar)(foo);
foo |> (yield* bar) // => (yield * bar)(foo);
```

edemaine:
Fixes #1671

### DIFF
--- a/source/parser.hera
+++ b/source/parser.hera
@@ -967,9 +967,8 @@ PipelineHeadItem
 
 PipelineTailItem
   ( AwaitOp / Return / Throw ) !AccessStart !MaybeParenNestedExpression -> $1
-  Yield $( _? Star )? !AccessStart !MaybeParenNestedExpression ->
-    if (!$2) return $1
-    return { ...$1, token: $1.token + $2 }
+  $( Yield ( _? Star )? ) !AccessStart !MaybeParenNestedExpression ->
+    return { $loc, token: $1, type: "Yield" }
   "import" !AccessStart ->
     return {
       type: "Identifier",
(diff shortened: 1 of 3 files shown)


source/parser.hera
```diff
@@ -966,7 +966,10 @@ PipelineHeadItem
   OptimizedParenthesizedExpression
 
 PipelineTailItem
-  ( AwaitOp / Yield / Return / Throw ) !AccessStart !MaybeParenNestedExpression -> $1
+  ( AwaitOp / Return / Throw ) !AccessStart !MaybeParenNestedExpression -> $1
+  Yield $( _? Star )? !AccessStart !MaybeParenNestedExpression ->
```
STRd6, line 970: Could probably simplify to:

```
  $( Yield ( _? Star )? ) !AccessStart !MaybeParenNestedExpression ->
    return { $loc, token: $1, type: "Yield" }
```

This way will have slightly better source mapping.

STRd6 (approved):
:+1:

source/parser.hera
edemaine: I thought `$loc` just had the start location, which is already in the return value of `Yield`? Or does it also have a length?

edemaine: Just checked and indeed `$loc` has a length. Fixed!

PR #1676, opened 2025-01-05
Fix implicit return of pipeline with `return`/`throw`

### LINKED ISSUE #1675: `return` at end of pipeline that causes `return return` if the pipeline was at the end of a function
# Input:
```
function f()
  1 |> return
```

# Actual Output:
```js
function f() {
  return return 1
}
```

When you run it you'll get `SyntaxError: expected expression, got keyword 'return'`

# Expected Output:
```js
function f() {
  return 1
}
```

Seem like either:
1. the extra return should be removed
2. it should be a syntax error at compile time

edemaine:
Fixes #1675

### DIFF
--- a/source/parser/pipe.civet
+++ b/source/parser/pipe.civet
@@ -127,146 +127,142 @@ function constructPipeStep(fn: ExprWithComments, arg: ASTNode!, returning: ASTNo
 // body: [ws, pipe, ws, expr][]
 
 function processPipelineExpressions(statements): void
-  gatherRecursiveAll(statements, (n) => n.type is "PipelineExpression")
-    .forEach (s) =>
-      [ws, , body] := s.children
-      [, arg] .= s.children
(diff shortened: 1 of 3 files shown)


source/parser/pipe.civet
```diff
@@ -127,146 +127,142 @@ function constructPipeStep(fn: ExprWithComments, arg: ASTNode!, returning: ASTNo
...
-          usingRef = usingRef or ref
-          arg = {
-            type: "ParenthesizedExpression",
-            children: ["(", {
-              type: "AssignmentExpression"
-              children: [usingRef, " = ", arg]
-            }, ")"],
-          }
-          returning = usingRef
+  for each s of gatherRecursiveAll statements, .type is "PipelineExpression"
```
edemaine, line 130: Apologies for the big diff; this is mostly just a dedentation.

STRd6 (approved):

PR #1677, opened 2025-01-05
Fix immediately called `(op)` in pipeline context 

### LINKED ISSUE #1672: Compiler barfs on `foo |> (*) bar`
Error in Playground dev console is: `TypeError: i is undefined`.

🚫 Each of these lines cause a TypeError instead of a parse error.
```js
foo |> (*) bar
foo |> (*) 1
foo |> (+) bar
foo |> (-) bar
foo |> (and) bar
```

✅ Each of these lines correctly compile or correctly fail to parse
```js
foo |> (*) * // Failed to parse
foo |> (baz) bar // baz(bar)(foo);
```

Also, the playground doesn't handle the error very well either:

CleanShot 2025-01-04 at 18 15 27@2x

edemaine:
Fixes #1672

### DIFF
--- a/test/function-block-shorthand.civet
+++ b/test/function-block-shorthand.civet
@@ -921,6 +921,14 @@ describe "(op) shorthand", ->
   """
 
   testCase """
+    apply binary op with one argument in pipeline
+    ---
+    foo |> (+) x
+    ---
+    (x)(foo)
+  """
+
+  testCase """
     apply binary op with no arguments
     ---
     (+)()
--- a/source/parser/lib.civet
+++ b/source/parser/lib.civet
@@ -562,12 +562,18 @@ function processCallMemberExpression(node: CallExpression | MemberExpression): A
         (isComma(arg) as ASTLeaf).token = `)${op.token}(`
         commaCount++
     // Don't mess with (+)()
-    if args.length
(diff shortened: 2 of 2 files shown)


STRd6 (approved):

test/function-block-shorthand.civet
```diff
@@ -920,6 +920,14 @@ describe "(op) shorthand", ->
     (1)
   """
 
+  testCase """
+    apply binary op with one argument in pipeline
+    ---
+    foo |> (+) x
+    ---
+    (x)(foo)
```
vendethiel, line 928: Why is that the desired result?

vendethiel: Oh, `(+) x` instantly unwraps, it’s not a section

edemaine: I'm not sure it's perfect, but the current behavior is that `(+)` can be followed by *any number* of arguments, and they get added together. In the special case of 1, it's not very useful, but maybe helpful in a setting where you're adding/removing arguments like so:

```js
(+)
  part1
  part2
//edit from/to:
(+)
  part1
  //part2
```

(similar to how we support empty blocks in `if` etc.)

PR #1688, opened 2025-01-22
Allow binary operators after nested object literals and bulleted arrays, fix `++` with bulleted arrays

### LINKED ISSUE #1687: Regression with dedented postfix after implicit object literal
This used to work, but is now a parse error:

```js
defaults :=
  text: ''
  quality: 'M'
  margin: 1
  stroke: 0.2
as const
```

Workaround:

```js
defaults := {}
  text: ''
  quality: 'M'
  margin: 1
  stroke: 0.2
as const
```

edemaine:
Fixes #1687. This bug was introduced by #1628 which innocently added checking for `NestedImplicitObjectLiteral`s next to `NestedBulletedArray`s in `MaybeNestedExpression`.

It turns out the "right" answer is to check for neither! Otherwise, we don't allow for binary operators after dedent after the nested object literal / bulleted array. Conveniently, we already check for `NestedImplicitObjectLiteral` and `NestedBulletedArray` within `PrimaryExpression` (as part of `ObjectLiteral` and `ArrayLiteral` respectively), so we don't need to do anything special here, except for *stopping* the `PushIndent Nested PopIndent` check in `MaybeNestedExpression`.

Along the way, I fixed another issue with the concat operator `++` followed by a bulleted array. Previously, this broke for two reasons: (1) `UpdateExpression` thought that this was an increment, despite the indentation after `++` (needed to explicitly `!Whitespace` to match the comment). (2) `WRHS` *did* need to check explicitly for `NestedImplicitObjectLiteral` and `NestedBulletedArray`, like `MaybeNestedExpression` used to. (Here it's the right thing to do, because any further binary operators will be handled by the parent `BinaryExpression`.)

Also fixed some unrelated CI issues that seem to have come up just from 2025 deprecations.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -72,8 +72,10 @@ jobs:
         cache: yarn
         node-version: 20
 
-    - name: Install and Test
+    - name: Install and Build
       # `yarn docs:build` includes `yarn build`
       run: |
         yarn
         yarn docs:build
+      env:
+        API_TOKEN: ${{ secrets.GITHUB_TOKEN }}
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -29,7 +29,7 @@ jobs:
         yarn build
         yarn test
 
-    - uses: actions/upload-artifact@v3
+    - uses: actions/upload-artifact@v4
       with:
         name: browser.js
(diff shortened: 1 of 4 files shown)


STRd6 (approved):
Thanks!

PR #1694, opened 2025-02-05
Bun --watch flag issue

jtenner:
Hey! Thank you for this amazing software.

The only issue I had was an issue with `bun` when using the `--watch` flag.

When the compilation of a given module fails with a syntax error, it prevents the plugin from generating any more modules. This breaks the `--watch` flag features.

Instead of failing, it makes sense to catch the error, log it to stderr, and then use a previously cached version of that module.

Features added:

- Bugfix: Prevent plugin from crashing
- Bugfix: Report errors to console for end users

A few questions:

1. Is there a specific coding style I should conform to? Or is this solution fine?
2. How can we "test" this, or should we just accept that it can't be tested well?

To try it out locally:

1. [Install Bun](https://bun.sh/)
2. Initialize a bun project `bun init -y`
3. Add `bunfig.toml` with `preload = ["./dist/bun-civet.mjs"]`
4. Create `index.civet` with `Bun.write "./test.txt", "Hello world!"`
5. `bun --watch index.civet`
6. Add a syntax error (like a bunch of `=`'s) and change the `"Hello world!"` string to something else
7. Notice that `./text.txt` remains unmodified
8. Remove the syntax errors from `index.civet`
9. Check `./text.txt` for the corresponding changes

I'm very happy to help or make modifications.

### DIFF
--- a/source/bun-civet.civet
+++ b/source/bun-civet.civet
@@ -17,19 +17,23 @@ import { plugin, file } from 'bun'
 await plugin {
   name: 'Civet loader'
   setup(builder)
-    { compile } := await import('./main.mjs')
+    { compile } := await import './main.mjs'
     //transpiler := new Transpiler loader: 'tsx'
-
-    builder.onLoad filter: /\.civet$/, ({path}) =>
-      source := file path |> .text() |> await
-      // Compiling and running at the same time, so enable comptime
-      // to ensure any async promises get resolved
-      contents .= await compile source, comptime: true
-      // Transpile directly as workaround to loader spec below not working
-      //contents = await transpiler.transform contents
-
-      return {
-        contents
-        loader: 'tsx'
-      }
+    const cache = new Map<string, string>
+    builder.onLoad filter: /\.civet$/, ({ path }) =>
+      contents:
+        try
+          source := file path |> .text() |> await
+          // Compiling and running at the same time, so enable comptime
+          // to ensure any async promises get resolved
+          contents .= await compile source, comptime: true
+          // Transpile directly as workaround to loader spec below not working
+          // contents = await transpiler.transform contents
+          cache.set path, contents
+          contents
+        catch ex
+          // log the error, use the cache or an empty module
+          console.error ex
+          (cache.get path) ?? 'void 0'
+      loader: 'tsx'
 }


source/bun-civet.civet
```diff
@@ -17,19 +17,23 @@ import { plugin, file } from 'bun'
...
-      // to ensure any async promises get resolved
-      contents .= await compile source, comptime: true
-      // Transpile directly as workaround to loader spec below not working
-      //contents = await transpiler.transform contents
-
-      return {
-        contents
-        loader: 'tsx'
-      }
+    const cache = new Map<string, string>
```
edemaine, line 22:
```suggestion
    cache := new Map<string, string>
```

edemaine (commented):
I'm worried about this change, because it would cause `bun foo.civet` to merely display an error, but not give an exit code, when `foo.civet` has a parse error. I'm trying to figure out whether there's a better approach, though Bun's `onLoad` doesn't seem to have a way to return an error other than throwing one... Possibly this is just a bug in `bun --watch`.

edemaine:
I've reported as https://github.com/oven-sh/bun/issues/17270 — we'll see what the Bun folks think about it.

uinz:
> https://github.com/oven-sh/bun/issues/4689#issuecomment-2888634558

try this

```toml
# bunfig.toml
preload = ["./plugin.ts"]
```

```ts
// plugin.ts
import { compile } from "@danielx/civet";
import { plugin } from "bun";

await plugin({
  name: "Civet loader",
  async setup(builder) {
    builder.onLoad({ filter: /\.civet$/ }, async ({ path }) => {
      const mod = await import(`${path}?`, { with: { type: "text" } });
      //          ^^^^^^^^^^^^

      const contents = await compile(mod.default);

      return {
        contents,
        loader: "tsx",
      };
    });
  },
});
```

STRd6:
Retested on bun 1.3.14:

- The original repro (syntax error in the **entry-point** `.civet` under `--watch`) no longer reproduces — the error is reported and fixing the file recovers, with the unmodified plugin. Bun covers the entry point via a fallback.
- However, **imported** `.civet` files are still never registered with bun's watcher at all (oven-sh/bun#17270; fix oven-sh/bun#30608 not yet merged) — even valid edits to an imported file trigger no reload. The caching approach in this PR can't help that case, since no reload event ever fires.
- @uinz's text-import suggestion fixes the import case: routing the source read through the module loader registers the file with the watcher. Verified the full edit → error → fix cycle recovers, and plain `bun run` / `bun build` are unaffected.

I've applied that workaround in #2156. Thanks @jtenner for the report and repro steps, and @uinz for the workaround — closing this in favor of #2156; once oven-sh/bun#30608 ships the workaround can be reverted entirely.

PR #1703, opened 2025-02-24
Document `class with A, B` mixins

### LINKED ISSUE #1697: Document `class with ...` mixin examples
I wanted to look this up today but didn't find any examples in the docs. This is a reminder that we should add docs at some point.

edemaine:
Fixes #1697 

When writing this, I wondered one thing: should we instead use the following existing syntax? (though it doesn't currently compile)

```js
class Civet extends Animal |> Mixin1 |> Mixin2
```

(Or maybe we discussed this already?)

It's more annoying if you want to write `class Civet with Mixin1, Mixin2`:

```js
class Civet extends Object |> Mixin1 |> Mixin2
```

### DIFF
--- a/civet.dev/reference.md
+++ b/civet.dev/reference.md
@@ -2634,6 +2634,23 @@ class Civet < Animal <: Named
 class Civet <: Animal, Named
 </Playground>
 
+### Mixins
+
+Civet experimentally defines a "mixin" to be a function mapping a class to
+a class (e.g., returning a subclass with added functionality),
+and provides syntax for applying one or more mixins to the base class
+you're extending:
+
+<Playground>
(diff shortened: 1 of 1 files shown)


STRd6:
It might get weird with a nested class body where it would be implied to be inside of `Mixin2`. It may make sense as an alternative syntax but it isn't as intuitive to me.

STRd6 (approved):
Looks good, but might be nice to add a brief class body as well.

edemaine:
I opted to not include types for the example, given that the current approach to typing is slightly messy and we aim to clean that up in the future.

STRd6 (approved):

PR #1704, opened 2025-02-24
Document how to use VSCode plugin

### LINKED ISSUE #1086: LSP -- exclude JS files from analysis
If I have two files with the same base name, but one is `.civet` and the other is `.js`, the LSP ignores the .js file for analysis. However, in my project, I have a couple files where the compiled output is a different name entirely to the source (e.g. `worker.civet` compiles to `pre.js`), and the LSP complains about this, even if I add it to the `exclude` in my tsconfig:

image

### LINKED ISSUE #1699: LSP Incorrectly Flags Errors In `coffeeCompat` Mode
Given the following Civet code:

```civet
"civet coffeeCompat"

class A
  a: ->
    for await x from y
      continue

  b()
    for await x from y
      continue
```

…the VSCode extension flags `await` in the first method definition, but not the second.

Image

The expected behavior would be for neither case to be flagged.

edemaine:
Fixes #1699 I think.

I think we want a VSCode plugin option to disable type checking altogether, but I don't actually know how to disable type checking... 😅

### DIFF
--- a/civet.dev/getting-started.md
+++ b/civet.dev/getting-started.md
@@ -44,7 +44,13 @@ npm i -g @danielx/civet
 The instructions below assume such a global install;
 if you do not want to, use `npx @danielx/civet` in place of `civet`.
 
-To use TypeScript for type checking, create a `tsconfig.json` file. For example:
+If you use VSCode, install the
+[Civet VSCode extension](https://marketplace.visualstudio.com/items?itemName=DanielX.civet).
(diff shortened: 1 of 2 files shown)


STRd6 (approved):

STRd6:
It might be a good default to disable type checking if there is no tsconfig. We may need to manually filter out the type checking messages in the LSP unless TS has a built in option for that... or maybe we can abuse `exclude` or something.

Issue #1712, opened 2025-03-14
Awaited dynamic import fails to parse at top level or within functions


STRd6:
```civet
await import pathToFileURL(tmpPath)
```

It seems to work outside of `await` though:

```civet
->
  x := import pathToFileURL(tmpPath) // this is ok
  await x
```

edemaine:
You've uncovered an ambiguity in our `from...import` shorthand:

```js
await import pathToFileURL
// short for
from await import pathToFileURL
// equivalent to
import pathToFileURL from 'await'
```

We can certainly forbid `await` from being an import name to fix this particular case. But it makes me worry about other useful cases. Consider:

```js
callback import foo
// old meaning:
callback(import(foo))
// current meaning:
import foo from 'callback'
```

I'm tempted to roll back something here. I see at least two options:

1. Require leading `from` in inverted `import`s, as in `from x import y`. Remove `x import y` shorthand.
2. Allow `x import y` but require `x` to have quotes in this case, as in `'x' import y` (unambiguous given that we wouldn't want to call a string).

Am I missing an option? I guess option 1 is simpler, but option 2 is most flexible?

Admittedly, `x from y` shorthand has a similar ambiguity. I think it's just less common for `from` to mean a function call (though it's not a reserved word), unlike `import`.

STRd6:
I think removing the `x import y` shorthand is fine. If there is pressure to keep it then restricting it to `'x' import y` or `./pathlike import y` may be an acceptable compromise. Could still be ambiguous with something like `@danielx/civet import parse` (this, snug division, ...)

edemaine:
It looks like [I've used this feature a fair amount](https://github.com/search?q=owner%3Aedemaine+%2F%5B%5E%5Cs%5D+import+%5C%7B%2F+path%3A*.civet&type=code), but [all other public uses on GitHub](https://github.com/search?q=%2F%5B%5E%5Cs%5C%2F%3D%7D%5D+import+%5C%7B%2F+path%3A*.civet&type=code) have a leading `from`. My uses happen to be unambiguous because of the `-` or `./` in the path name, but it feels too flaky to rely on, so I'll plan to just remove it. (We might want to version bump given the breaking change though...)

PR #1715, opened 2025-03-16
Require `from` in backwards `import`/`export` to fix ambiguity with dynamic `import`

edemaine:
**New behavior:** `a import b` means `a(import(b))`
**Old behavior:** `a import b` means `import b from a`

In particular, this is definitely what we want when `a` is `await`, fixing #1712.

BREAKING CHANGE: `module import spec` needs to written as `from module import spec` (`from` can no longer be omitted)


STRd6 (approved):
Thanks!

PR #1717, opened 2025-03-16
Fix `export`ing statement expressions

### LINKED ISSUE #1708: exported const value with conditional exports the wrong variable
## Input

```
export anExportedProperty := if x then y else z
```

## ❌ Output

```ts
export let ref;
if (x) ref = y;
else ref = z;
const anExportedProperty = ref;
```

It exports `ref` instead of `anExportedProperty`

## Expected output

```ts
let ref;
if (x) ref = y;
else ref = z;
export const anExportedProperty = ref;
```

## Workaround

Assign to a variable before exporting. e.g.:

```
anExportedProperty := if x then y else z
export anExportedProperty
```

edemaine:
Fixes #1708 

The attempt to lift statement expressions out of the initializers of declarations was missing a check to see whether the declaration is top level. I think we didn't think about this case because declarations are mostly forced to be top-level because of the grammar, but `ExportDeclaration` is at least one thing that could surround them. By contrast, assignments (see `processAssignments`) already had such a check (`blockContainingStatement`), so I basically just copied that.

We could of course consider a smarter compilation specific to `export`:

```js
export x := if a then b else c
---
let ref
if (a) { x = b }
else { x = c }
export const x = ref
```

This could still be done in the future. In any case I think we should have this check.

### DIFF
--- a/source/parser/declaration.civet
+++ b/source/parser/declaration.civet
@@ -19,6 +19,7 @@ import type {
 } from ./types.civet
 
 import {
+  blockContainingStatement
   blockWithPrefix
   braceBlock
   makeEmptyBlock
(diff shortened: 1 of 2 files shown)


STRd6 (approved):
Agreed that we could optimize further later but this is a nice improvement!

PR #1718, opened 2025-03-17
Fix complex one-line `if` then blocks via bracing

### LINKED ISSUE #1706: Extra semicolon in oddly-formatted `if ... if ... else ... else`
```
function foo(a: boolean, b: boolean)
    if (a) if (b) {
        return "1";
    } else {
        return "2";
    } else {
        return "3";
    }
```
becomes:
```js
function foo(a: boolean, b: boolean) {
    if (a) if (b) {
        return "1";
    } else {
        return "2";
    }; else {
        return "3";
    }
}
```
which is invalid JS.

edemaine:
Fixes #1706 as discussed in https://github.com/DanielXMoore/Civet/issues/1706#issuecomment-2730164047

> My only minor suggestion is replacing the first space, if it exists, with the open brace and a trailing space, if it exists, with the closing brace to keep the source alignment as close as possible with the transpiled code.

I'm not sure this is that easy to do, given that we'd like to do this local to the block. It does seem that the leading space is often part of the block, but the trailing space isn't. Notably, here's the relevant code in `braceBlock`:

```js
    block.children.unshift " {"
    block.children.push "}"
```

I've always found this asymmetric. I could see
1. removing the space before the `{`, or
2. adding a space before the `}`.

Having half of the braces be spaced feels a bit weird... but also fine. I'd like to settle on what changes to make (if any) before changing anything, because this affects 135 tests. Mostly in pattern matching, e.g.:

```js
switch x
  < 0 then console.log "it's negative"
  > 0 then console.log "it's positive"
  is 0 then console.log "it's zero"
  else console.log "it's something else"
---
// current
if(x < 0) { console.log("it's negative")}
else if(x > 0) { console.log("it's positive")}
else if(x === 0) { console.log("it's zero")}
// option 1
if(x < 0){ console.log("it's negative")}
else if(x > 0){ console.log("it's positive")}
else if(x === 0){ console.log("it's zero")}
// option 2
if(x < 0) { console.log("it's negative") }
else if(x > 0) { console.log("it's positive") }
else if(x === 0) { console.log("it's zero") }
```

Here's an example where we're inconsistent (we favor spaces in a different bracing application):

```js
=> for x of y
=> for [x] of y
---
// current
() => { const results=[];for (const x of y) {results.push(x)};return results; };
() => { const results1=[];for (const [x] of y) {results1.push([x])};return results1; }
// option 1
() => { const results=[];for (const x of y){results.push(x)};return results; };
() => { const results1=[];for (const [x] of y){results1.push([x])};return results1; }
// option 2
() => { const results=[];for (const x of y) { results.push(x) };return results; };
() => { const results1=[];for (const [x] of y) { results1.push([x]) };return results1; }
```

### DIFF
--- a/source/parser.hera
+++ b/source/parser.hera
@@ -13,6 +13,7 @@ import {
   append,
   blockWithPrefix,
   braceBlock,
+  bracedBlock,
   convertNamedImportsToObject,
   convertObjectToJSXAttributes,
   convertWithClause,
@@ -4624,12 +4625,7 @@ IfStatement
     }
     // Ensure semicolon separating single line from `else`
     if (block.bare && e) {
-      const semicolon = ";"
-      block = {
-        ...block,
-        semicolon,
-        children: [...block.children, semicolon],
-      }
+      block = bracedBlock(block)
     }
     return {
       type: "IfStatement",
@@ -4643,11 +4639,7 @@ IfStatement
   IfClause:clause BlockOrEmpty:block ElseClause?:e ->
     // Ensure semicolon separating single line from `else`
     if (block.bare && e) {
-      block = {
-        ...block,
-        semicolon: ";",
-        children: [...block.children, ";"],
-      }
+      block = bracedBlock(block)
     }
     return {
       type: "IfStatement",
--- a/source/parser/block.civet
+++ b/source/parser/block.civet
@@ -51,12 +51,12 @@ function blockWithPrefix(prefixStatements: StatementTuple[] | undefined, block: 
 
   return block
 
-// Add braces if block lacks them
+/** Add braces if block lacks them */
 function braceBlock(block: BlockStatement): void
   if block.bare and not block.root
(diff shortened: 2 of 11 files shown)


STRd6 (approved):
Looks good!

Agreed that the spacing isn't worth changing all those tests for.

My current thoughts about inserting characters are to insert minimal spaces/newlines so as to keep the transpiled code as close to the source as possible. If people want specific spacing of the transpiled output we should delegate to other tools like prettier. Maybe later on we can do a spacing cleanup but it's not a priority.

edemaine:
Sounds good. I think added newlines are most important to avoid, as line numbers are most important for traceback. I don't think spaces are as important, but it'd be nice to have a consistency pass at some point. Minimizing extra spaces sounds like a fine objective, and will help break ties, but we definitely don't do that currently. 😅

PR #1719, opened 2025-03-17
Indented forms of `implements` and `with`

### LINKED ISSUE #1707: Extend `with` To Support Indented Blocks
Allow for mixin combinators (functions that return mixins) to be used more easily by introducing a block syntax for `with`. This would allow us to write:

```js
class MyComponent extends Component
  with
    click -> console.log "clicked!"
    change -> console.log "changed!"
// equivalent to the following, which already works
class MyComponent extends Component
  with
    (click -> console.log "clicked!"),
    (change -> console.log "changed!")
```

While we can already write this by wrapping each mixin expression in parenthesis and separating them by commas, the block syntax would introduce implicit commas between lines, like `[]` array literals.

edemaine:
Fixes #1707 by adding a nested list form of both `implements` and `with`.
Also convert some `__` to `NotDedented`, allow spaces before commas, etc.

### DIFF
--- a/civet.dev/reference.md
+++ b/civet.dev/reference.md
@@ -2634,6 +2634,16 @@ class Civet < Animal <: Named
 class Civet <: Animal, Named
 </Playground>
 
+Long `implements` lists can be indented:
+
+<Playground>
+class Civet
+  extends Animal
+  implements
+    Named
+    Character
+</Playground>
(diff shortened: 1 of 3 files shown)


STRd6 (approved):
:sunglasses:

PR #1722, opened 2025-03-19
Fix hot reload via unplugin in NextJS, fix sourcemaps with `ts: "preserve"`

### LINKED ISSUE #1693: Hot Reload broken in Next.JS starter template
Unplugin works fine for building, but when you run next on dev mode it will not hot reload modules as expected.

Got it working by disabling plugin cache, passing {cache: false} to civetWebpackPlugin.

The cache implementation seems fine https://github.com/DanielXMoore/Civet/blob/997f5f0190f8d17e94d9881e26afb9f3346c4560/source/unplugin/unplugin.civet#L419

edemaine:
Fixes #1693

* Update NextJS example to what I got from a fresh `npx create-next-app@latest` which uses Next's new app router instead of pages router. I think this is maybe why the old code was working and new code wasn't.
* Add `addWatchFile` even if we have a cache mix. This is the crucial fix.
* Also support `stat` failing (and skipping cache), just in case. This probably isn't important, given that we need to load the file, but maybe there's a weird filesystem out there...
* Fix sourcemap when specifying `ts: "preserve"`; we were accidentally dropping them.


STRd6 (approved):
Awesome, thanks!

PR #1723, opened 2025-03-19
Farm and Rolldown bundler support

edemaine:
unplugin added support for Farm. I tried it out, along with rolldown, and they work well!

Unfortunately, rspack and rsbuild don't work with our current setup, at least on Windows. (Looks like a path issue.) That can be dealt with later, but for now I've removed our claims to work with rspack.


STRd6 (approved):
:clap:

PR #1726, opened 2025-04-14
Fix unplugin source map directories

STRd6:
This adds the directory paths to the sourcemaps so they don't all end up at the root.


edemaine (approved):
This definitely makes sense to me. Is there an easy scenario to test this out?

STRd6:
Looks like we already had an integration test that demonstrates the change. I tested it locally on a project I had by using `yarn link` and esbuild `ctx.serve`.

PR #1727, opened 2025-04-14
Explicit changelog build script shell

STRd6:
Changelog script wasn't working for me unless I was explicit about the shell.

### DIFF
--- a/build/changelog.civet
+++ b/build/changelog.civet
@@ -13,6 +13,7 @@ process.chdir join (dirname fileURLToPath import.meta.url), '..'
 
 function run(command: string, args: string[], options?: object): string
   sub := spawnSync command, args, {
+    shell: 'bash'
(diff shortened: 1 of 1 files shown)


edemaine:
Unfortunately, this breaks the script for me on Windows, I guess because Node isn't Cygwin-enabled so it's not finding `/bin/bash` (it's in `/cygwin64/bin/bash`).

Does `shell: 'bash'` work for you? That searches my path and works for me.

Incidentally, what was the error you were getting without calling `shell`? I'm a little surprised it's necessary, but I guess cross-platform is hard.

STRd6:
`shell: 'bash'` works for me. The error I was getting was that the list of revisions would end up being empty without the explicit shell and quotes.

edemaine (approved):

PR #1733, opened 2025-04-27
Fix `outputExtension` option in unplugin, enabling `.svelte.ts`/`.svelte.js` for Svelte Runes reactive compatibility

adam2am:
## Summary

- Expand handling of `.civet` files to support output extensions like `.svelte.ts` or `.svelte.js`.
- Update `isCivetTranspiled` regex to match `.civet.svelte.ts` and `.civet.svelte.js`.
- Update `cleanCivetId()` to correctly strip new suffixes (`.svelte.ts`, `.svelte.js`).
- This enables `.civet` files to output TypeScript-compatible `.svelte` files, which are needed for Svelte's new Runes API.
- No breaking changes - `.civet.jsx` / `.civet.tsx` are still supported as before.

## Motivation

- Svelte's new Runes system.
- Civet users can now fully integrate with workflows by outputting `.svelte.ts` from `.civet` files.
(even tho for now IDE will not recognize syntax of it out of the box, have to manually do .d.ts)

Thanks for reading! 🚀 Happy to adjust anything if needed

### DIFF
--- a/source/unplugin/unplugin.civet
+++ b/source/unplugin/unplugin.civet
@@ -43,8 +43,8 @@ export type PluginOptions
   threads?: number
   /** Cache compilation results based on file mtime (useful for serve or watch mode) */
   cache?: boolean
-  /** config filename, or null to not look for default config file */
-  config?: string? | null
+  /** config filename, or false/null to not look for default config file */
+  config?: string | false | null
   parseOptions?: ParseOptions
 
 type CacheEntry
@@ -150,9 +150,9 @@ export const rawPlugin: Parameters<typeof createUnplugin<PluginOptions>>[0] =
     enforce: 'pre'
 
     async buildStart(): Promise<void>
-      const civetConfigPath = 'config' in options
-        ? options.config
-        : await findInDir(process.cwd())
+      civetConfigPath .= options.config
+      if civetConfigPath is undefined
+        civetConfigPath = await findInDir process.cwd()
       if civetConfigPath
         compileOptions = await loadConfig civetConfigPath
       // Merge parseOptions, with plugin options taking priority
--- a/source/unplugin/unplugin.civet
+++ b/source/unplugin/unplugin.civet
@@ -55,20 +55,9 @@ type CacheEntry
 civetExtension := /\.civet$/
 // Normally .jsx/.tsx extension should be present, but sometimes
 // (e.g. esbuild's alias feature) loads directly without resolve
-isCivetTranspiled := /(\.civet)(\.(?:[jt]sx|svelte\.(?:js|ts)))?([?#].*)?$/
 postfixRE := /[?#].*$/s
 isWindows := os.platform() is 'win32'
 windowsSlashRE := /\\/g
-
-// Extracts query string and hash, and removes tsx/jsx extension
-function cleanCivetId(id: string): {id: string, postfix: string}
-  let postfix = ''
-  id = id
-  .replace postfixRE, (match) =>
-    postfix = match
-    ''
-  .replace /\.(?:[jt]sx|svelte\.(?:js|ts))$/, ''
-  {id, postfix}
 
 function tryStatSync(file: string): fs.Stats?
   try
@@ -116,9 +105,23 @@ export const rawPlugin: Parameters<typeof createUnplugin<PluginOptions>>[0] =
     ts = "civet"
 
   transformTS := options.emitDeclaration or options.typecheck
+  implicitExtension := options.implicitExtension ?? true
+  
   outExt :=
     options.outputExtension ?? (ts is "preserve" ? ".tsx" : ".jsx")
-  implicitExtension := options.implicitExtension ?? true
+  cleanOutExt := outExt.replace(/\./g, '\\.')
+  cleanOutputExtRE := new RegExp(`${cleanOutExt}$`)
+  isCivetTranspiled := new RegExp(`(\\.civet)(?:${cleanOutExt})?([?#].*)?$`)
+
+  function cleanCivetId(id: string): {id: string, postfix: string}
+    let postfix = ''
+    id = id
+    .replace postfixRE, (match) =>
+      postfix = match
+      ''
+    .replace cleanOutputExtRE, ''
+    {id, postfix}
+
   let aliasResolver: (id: string) => string
 
   fsMap: Map<string, string> .= new Map
--- a/source/unplugin/unplugin.civet
+++ b/source/unplugin/unplugin.civet
@@ -52,22 +52,20 @@ type CacheEntry
   result?: TransformResult
   promise?: Promise<void>
 
-civetExtension := /\\.civet$/
-// Normally .jsx/.tsx extension should be present, but sometimes
-// (e.g. esbuild's alias feature) loads directly without resolve
 postfixRE := /[?#].*$/s
 isWindows := os.platform() is 'win32'
 windowsSlashRE := /\\/g
 civetSuffix := '.civet'
 
 // Validate and extract the base Civet filename dynamically using outExt
-function validateAndExtractCivetBase(id: string, outputExtension: string): string | null
-  unless id? and outputExtension? return null
-  cleanedId := id.replace postfixRE, ''
-  unless cleanedId.endsWith outputExtension return null 
-  base := cleanedId.slice 0, -outputExtension.length
-  unless base.endsWith civetSuffix return null 
-  base
+function extractCivetFilename(id: string, outputExtension: string): string?
+  cleanedId := normalizePath(id.replace postfixRE, '')
+  lCleanedId := cleanedId.toLowerCase()
+  lOutputExtension := outputExtension.toLowerCase()
+  if lCleanedId.endsWith lOutputExtension
+    base := cleanedId[<-outputExtension.length]
+    lBase := base.toLowerCase()
+    return base if lBase.endsWith civetSuffix
 
 // Separate the ID from its query/hash postfix
 function splitIdPostfix(id: string): {idWithoutPostfix: string, postfix: string}
@@ -366,8 +364,8 @@ export const rawPlugin: Parameters<typeof createUnplugin<PluginOptions>>[0] =
                     filePath = filePath[0...-5]
                   else
                     console.log `WARNING: No .d.ts extension in ${filePath}`
-                  if match := filePath.match civetExtension
-                    filePath = filePath[0...-match[0]#]
+                  if filePath.endsWith civetSuffix
+                    filePath = filePath[0...-civetSuffix#]
                   else
                     console.log `WARNING: No .civet extension in ${filePath}`
                   filePath += options.declarationExtension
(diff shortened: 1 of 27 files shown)


edemaine (commented):
Thanks for this PR, and sorry for the delay. Do you think it would be possible to use `outputExtension` instead of hard-coding Svelte support? That would be better, I think, but it may require a bit more restructuring.

adam2am:
> Thanks for this PR, and sorry for the delay. Do you think it would be possible to use `outputExtension` instead of hard-coding Svelte support? That would be better, I think, but it may require a bit more restructuring.

You mean something like this? 
```ts
  outExt :=
    options.outputExtension ?? (ts is "preserve" ? ".tsx" : ".jsx")
  cleanOutExt := outExt.replace(/\./g, '\\.')
  cleanOutputExtRE := new RegExp(`${cleanOutExt}$`)
  isCivetTranspiled := new RegExp(`(\\.civet)(?:${cleanOutExt})?([?#].*)?$`)

  function cleanCivetId(id: string): {id: string, postfix: string}
    let postfix = ''
    id = id
    .replace postfixRE, (match) =>
      postfix = match
      ''
    .replace cleanOutputExtRE, ''
    {id, postfix}
 ```

      
Also while experimenting I noticed it's adding it to the end, so it's file.civet.ts / file.civet.svelte.ts without stripping
 Do you think would it be a cool idea to optionally add something like 
 
 ```ts
    stripCivet  := options.stripCivetExtension ?? false
 ```
 
and some extra logic to make it possible > file.ts / file.svelte.ts / file.tsx or there is no point for doing that?

**Comment by edemaine** [comment]
If we stick to regular expressions, you'll need to escape all active regexp characters, not just `.`. (Can't wait for [RegExp.escape](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/escape)...) Something like

```js
cleanOutExt := outExt.replace(/[.*+?^${}()|[\]\\]/g, '\\$&')
```

But I wonder if we could instead just remove do the `postfixRE` removal, then detect whether the filename `endsWIth(outExt)`, strip that, and then check for `/\.civet$/`, instead of building a big regular expression.

> Do you think would it be a cool idea to optionally add something like
> 
> ```ts
>    stripCivet  := options.stripCivetExtension ?? false
> ```

I don't think this can work in the unplugin, because we wouldn't be able to detect transformed files, because they no longer have `.civet` in them. But I'm also not quite sure why it really matters: I don't think the unplugin ever emits a file with the output extension in it, does it? (The CLI already supports writing `.ts` files instead of `.civet.ts` files.) I think the unplugin's filenames are all for internal processing purposes. The point of this PR is to allow postprocessing triggered by extensions other than `.tsx`, including `.svelte.ts`.

source/unplugin/unplugin.civet
```diff
@@ -52,23 +52,30 @@ type CacheEntry
   result?: TransformResult
   promise?: Promise<void>
 
-civetExtension := /\.civet$/
+civetExtension := /\\.civet$/
```
edemaine, line 55: This looks wrong (too many `\`s) but also no longer needed; remove.

source/unplugin/unplugin.civet
```diff
@@ -403,15 +411,15 @@ export const rawPlugin: Parameters<typeof createUnplugin<PluginOptions>>[0] =
       if (options as! {scan?: boolean}).scan and meta.framework is 'vite'
         resolvedId = `\0${resolvedId}`
 
+      // Add back the original postfix at the end
       return resolvedId + outExt + postfix
 
     loadInclude(id)
-      isCivetTranspiled.test id
+      validateAndExtractCivetBase(id, outExt) is not null
```
edemaine, line 418:
```suggestion
      validateAndExtractCivetBase(id, outExt)?
```

source/unplugin/unplugin.civet
```diff
@@ -52,23 +52,30 @@ type CacheEntry
...
-
-// Extracts query string and hash, and removes tsx/jsx extension
-function cleanCivetId(id: string): {id: string, postfix: string}
-  let postfix = ''
-  id = id
-  .replace postfixRE, (match) =>
+civetSuffix := '.civet'
+
+// Validate and extract the base Civet filename dynamically using outExt
+function validateAndExtractCivetBase(id: string, outputExtension: string): string | null
```
edemaine, line 64:
```suggestion
function extractCivetFilename(id: string, outputExtension: string): string?
```

I think it's easier to just use `undefined` for the failure value. Also suggest a simpler name for the function.

source/unplugin/unplugin.civet
```diff
@@ -52,23 +52,30 @@ type CacheEntry
   result?: TransformResult
   promise?: Promise<void>
 
-civetExtension := /\.civet$/
+civetExtension := /\\.civet$/
 // Normally .jsx/.tsx extension should be present, but sometimes
 // (e.g. esbuild's alias feature) loads directly without resolve
```
edemaine, line 57: This comment should be removed or moved, given that you removed the regular expression.

But I'm worried that a missing `.jsx`/`.tsx` (what this comment is about) won't be handled correctly anymore. Before the `(\.[jt]sx)?` did the work. I've added a comment below where I think it needs to be fixed.

source/unplugin/unplugin.civet
```diff
@@ -52,23 +52,30 @@ type CacheEntry
...
-// Extracts query string and hash, and removes tsx/jsx extension
-function cleanCivetId(id: string): {id: string, postfix: string}
-  let postfix = ''
-  id = id
-  .replace postfixRE, (match) =>
+civetSuffix := '.civet'
+
+// Validate and extract the base Civet filename dynamically using outExt
+function validateAndExtractCivetBase(id: string, outputExtension: string): string | null
+  unless id? and outputExtension? return null
```
edemaine, line 65:
```suggestion
  return unless id? and outputExtension?
```

Actually, I think this line can safely be removed: typing should guarantee we never hit this condition. We don't need to guard against misuse errors given that we're the only callers.

source/unplugin/unplugin.civet
```diff
@@ -52,23 +52,30 @@ type CacheEntry
...
+  cleanedId := id.replace postfixRE, ''
+  unless cleanedId.endsWith outputExtension return null
```
edemaine, line 67:
```suggestion
  return unless cleanedId.endsWith outputExtension
```

Actually, to handle missing added `outputExtension`, I think this should be more like

```suggestion
  if cleanedId.endsWith outputExtension
```

(and inside the `if` block, maybe update `cleanedId` like the next line)

We'll still return `undefined` if there isn't then a `.civet` extension.

source/unplugin/unplugin.civet
```diff
@@ -52,23 +52,30 @@ type CacheEntry
...
-  id = id
-  .replace postfixRE, (match) =>
+civetSuffix := '.civet'
+
+// Validate and extract the base Civet filename dynamically using outExt
+function validateAndExtractCivetBase(id: string, outputExtension: string): string | null
+  unless id? and outputExtension? return null
+  cleanedId := id.replace postfixRE, ''
+  unless cleanedId.endsWith outputExtension return null 
+  base := cleanedId.slice 0, -outputExtension.length
```
edemaine, line 68:
```suggestion
  base := cleanedId[< -outputExtension.length]
```

source/unplugin/unplugin.civet
```diff
@@ -52,23 +52,30 @@ type CacheEntry
...
+  unless base.endsWith civetSuffix return null
```
edemaine, line 69:
```suggestion
  return unless base.endsWith civetSuffix
```

source/unplugin/unplugin.civet
```diff
@@ -403,15 +411,15 @@ export const rawPlugin: Parameters<typeof createUnplugin<PluginOptions>>[0] =
...
     loadInclude(id)
-      isCivetTranspiled.test id
+      validateAndExtractCivetBase(id, outExt) is not null
 
     async load(id)
-      match := isCivetTranspiled.exec id
-      return null unless match
-      basename := id[...match.index + match[1]#]
+      basename := validateAndExtractCivetBase(id, outExt)
+      unless basename return null
```
edemaine, line 422:
```suggestion
      return null unless basename
```

edemaine (changes requested):
I really like the new approach!  This is exactly what I was thinking.

I posted a few comments, the most important of which is about handling missing `outputExtension`. The rest are mostly about Civet code style.

Once implemented, I'll do some thorough testing, and then we can finally merge. Sorry for the long delay!

source/unplugin/unplugin.civet
adam2am: I just deleted this comment. 
.jsx/.tsx extensions still present if you mention it. When I change .svelte.ts to .tsx 		
```ts
civetVitePlugin({
	outputExtension: '.tsx',

```
it would throw the error:

1:18:48 PM [vite] (client) ✨ optimized dependencies changed. reloading
Svelte error: rune_outside_svelte
The `$state` rune is only available inside `.svelte` and `.svelte.js/ts` files
https://svelte.dev/e/rune_outside_svelte
    at <instance_members_initializer> (/home/user/Documents/repos/sv-civet-proof-of-concept/src/routes/class.civet.tsx:20:10)

source/unplugin/unplugin.civet
adam2am: I went with this

```ts
function extractCivetFilename(id: string, outputExtension: string): string?
  cleanedId := normalizePath(id.replace postfixRE, '')
  lCleanedId := cleanedId.toLowerCase()
  lOutputExtension := outputExtension.toLowerCase()
  if lCleanedId.endsWith lOutputExtension
    base := cleanedId[<-outputExtension.length]
    lBase := base.toLowerCase()
    return base if lBase.endsWith civetSuffix
```
as extra covering edge-cases with paths and uppercase

edemaine (changes requested):
Thanks for the edits!  I did some more thorough testing, and found and fixed some issues:

1. `extractCivetFilename` still isn't correct: if the cleaned id doesn't end with outputExtension, we still want to check for `.civet` extension, as explained in the comment you deleted.
2. We also need to remove the outputExtension in `resolveId` (at least in Vite, because of HTML `<script>` tag rewriting), but your changes removed this stripping. So I think `extractCivetFilename` and `splitIdPostfix` should go back to being unified.
3. I don't see why we should lower-case when testing for `outputExtension`, because it's always something we're adding manually, so case should not have changed. Arguably we *should* allow for `.CiVeT`, but we don't currently, so let's leave that for another time.

I fixed these and did general cleanup/rewrite in https://github.com/DanielXMoore/Civet/tree/unplugin-extension branch. Unfortunately, because you PR'd off your `main` branch, I can't push to it. (Please PR off a non-`main` branch in the future.) Can you pull from https://github.com/DanielXMoore/Civet/tree/unplugin-extension ?

With my commit, the `integration/unplugin-examples` seem to work, as does a real-world Vite example.

adam2am:
> 3\. lower-case when testing for `outputExtension`
Was just thinking about a few weird edge cases, same thing with normalizePath. But yeah, overall it's looking solid now, your code’s way cleaner to read too, holy amazing. Totally forgot to test it with integrations before

edemaine:
Thanks! Are you able to pull my changes into your branch? If not, we can make a new PR on my branch.

adam2am:
> Thanks! Are you able to pull my changes into your branch? If not, we can make a new PR on my branch.

My bad, got distracted before, merged your code into the branch

edemaine (approved):
Thanks!

PR #1734, opened 2025-04-27
ESM loader supports civetconfig by default, and overriding configuration

### LINKED ISSUE #1315: Problems with `node --test`s written in Civet (or possibly wider issue with ESM reading config)
1. `node --import @danielx/civet/register --test tests/*.civet` does not read or take into account any civet config.

2. `node --import @danielx/civet/register --test tests/*.civet --experimental-test-coverage` throws a `Warning: Could not report code coverage. TypeError: Cannot read properties of undefined (reading 'startOffset')` error, so no test coverage :(

> [!NOTE]
> This could just be symptomatic of a wider issue where executing ESM `civet *.civet` files ignores config entirely.

edemaine:
* ESM loader by default now looks for civetconfig files in the directory of the source file and its ancestors, like the CLI already did for CJS files. This should also fix the CLI in ESM mode, which relied on the ESM loader. Fixes #1315 and part of #1239.
* `@danielx/civet/register-noconfig` disables this new behavior, but otherwise acts like `@danielx/civet/register`.
* ESM loader more generally supports configuration via `initialize` hook, specifiable via `data` option in [`module.register`](https://nodejs.org/api/module.html#moduleregisterspecifier-parenturl-options). This lets you build custom register modules with custom parse options, e.g. for testing.

Also fix a small issue with tests not working in Windows.

BREAKING CHANGE: If you don't use civetconfig files and want to be robust against their existence in ancestor directories, or want to maximize performance, use `@danielx/civet/register-noconfig` instead of `@danielx/civet/register`

BREAKING CHANGE: In unplugin settings, `config: undefined` no longer disables searching for config files. Use `config: false` or `config: null` instead.

### DIFF
--- a/source/esm.civet
+++ b/source/esm.civet
@@ -46,7 +46,8 @@ register('@danielx/civet/esm', import.meta.url, {data: {
 type { ParseOptions } from @danielx/civet
 
 export type RegisterOptions
-  config?: string | null
+  /** config filename, or false/null to not look for default config file */
+  config?: string | false | null
   parseOptions?: ParseOptions
 
 baseURL := pathToFileURL(process.cwd() + '/').href
@@ -78,7 +79,7 @@ export async function load(url: string, context: any, next: any)
     { config } .= globalOptions
     if config is undefined
       config = await findConfig dirname path
-    if config <? "string"
+    if config
       loadedConfig = await loadConfig config
 
     options: Parameters<typeof compile>[1] & { sourceMap: true } :=
--- a/source/esm.civet
+++ b/source/esm.civet
@@ -21,16 +21,38 @@ Previously depended on ts-node esm loader being downstream:
 ```bash
 node --loader ts-node/esm --loader @danielx/civet/esm source.civet
 ```
+
+If you want to configure this ESM loader, you can make your own
+`register.mjs` along these lines:
+
+@example
+```javascript
+import { register } from 'node:module';
+register('@danielx/civet/esm', import.meta.url, {data: {
+  parseOptions: {
+    // Add your parse options here
+  },
+}});
+```
+
 */
 
 { readFileSync } from fs
 { pathToFileURL, fileURLToPath } from url
 
-Civet from ./main.civet
-{ compile, SourceMap } := Civet
+{ compile, SourceMap } from ./main.civet
+type { ParseOptions } from @danielx/civet
+
+export type RegisterOptions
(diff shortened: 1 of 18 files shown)


edemaine:
By the way, you can test the `register` scripts as follows:

```sh
cd test/infra/config
node --import ../../../register.js needs-coffee.civet
node --import ../../../register-noconfig.js needs-coffee.civet
```

(The latter is supposed to crash.)

source/esm.civet
```diff
@@ -21,17 +21,42 @@ Previously depended on ts-node esm loader being downstream:
...
 { pathToFileURL, fileURLToPath } from url
 
-Civet from ./main.civet
-{ compile, SourceMap } := Civet
+{ compile, SourceMap } from ./main.civet
+{ findConfig, loadConfig } from ./config.civet
+type { ParseOptions } from @danielx/civet
+
+export type RegisterOptions
+  config?: string | null
```
STRd6, line 49: I'd prefer keeping this api as `string | false`. I think `config: false` reads slightly better than `config: null` in terms of intent.

STRd6 (approved):
Looks good, just added one comment about preferring `config: false` for the API to disable config.

Clever trick with the register-noconfig sed comments!

Thanks!

source/esm.civet
edemaine: I'm fine with changing to/allowing `false`, though we should probably continue to support `null` because that's the current API for the unplugin (the only place where `config` can currently be specified):

https://github.com/DanielXMoore/Civet/blob/0039cdd9c0351815adf810eb2f3c39276472b450/source/unplugin/unplugin.civet#L46-L47

Maybe `| false | null` for all of them?

STRd6: :+1: Allowing both is fine.

edemaine: Reviewing the code closer, the unplugin actually supported `config: undefined` to disable config files. I think this is a bit too much: it's easy to do something like `{config: someOtherObject?.config}` and accidentally set `undefined` for an absent property. So I refactored a bit to treat `undefined` just like an unspecified property, i.e., to search for a config file, matching the documentation.

Anyway, now we support `config: false` and `config: null` (and probably any falsey value other than `undefined`), and it should be the same in the unplugin and the ESM loader (and the CLI too, though there it's just an internal notion of `config: false`, now restored).

PR #1737, opened 2025-04-28
Fix implicit returns in IIFEs with "civet -implicitReturns"

### LINKED ISSUE #1735: assignment fails without implicit returns
civet version 0.10.2 on Windows 11

The following code:

```
srcLib := 'src/lib/datatypes.lib.ts'
importCode := (
	if (srcLib == undefined)
		''
	else if srcLib.match(/^[\@\.\/]/)
		"import voidFunc from '#{srcLib}';"
	else
		"import voidFunc from './#{srcLib}';"
	)

console.log importCode
```
is compiled  to:

```
"use strict";
const srcLib = 'src/lib/datatypes.lib.ts'
const importCode = (
	(()=>{if (srcLib === undefined) {
		''
	}
	else if (srcLib.match(/^[\@\.\/]/)) {
		`import voidFunc from '${srcLib}';`
	}
	else {
		`import voidFunc from './${srcLib}';`
	}})()
	)

console.log(importCode)
```
The code inside the anonymous function will never return anything, so the variable importCode remains undefined. That's certainly not the intention of the civet code. Furthermore, I'm sure that this is a breaking change in the civet compiler (I recently had to update my civet version to 0.10.2). 

I've found that if, in my `civetconfig.civet` file, I set implicitReturns to true, the code works, i.e. the variable importCode is correctly set. (it also works to add the 'return' keyword in the civet code). In my opinion, implicit returns are horrible. If you intend to return something from a function, you should explicitly return it. However, in this case (i.e. in the original civet code above), I am simply assigning a value to a constant. There is no function there at all (though, I admit, the generated TypeScript code implements my intentions using a function). More importantly, I believe that this is a breaking change from previous versions.

edemaine:
Fixes #1735

### DIFF
--- a/source/parser/function.civet
+++ b/source/parser/function.civet
@@ -195,7 +195,8 @@ function processReturn(f: FunctionNode, implicitReturns: boolean): void
   if returnType and returnType.optional
     convertOptionalType returnType
 
-  if not processReturnValue(f) and implicitReturns
+  if not processReturnValue(f) and
+     (implicitReturns or f.signature.implicitReturn)
     { signature, block } := f
     { modifier, name, returnType } := signature
     { async, generator, set } := modifier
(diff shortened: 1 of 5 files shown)


STRd6 (approved):
Thanks!

PR #1747, opened 2025-06-16
lsp: async Civet config load, shadow-.civet resolution & opt-out on .ts/.js

adam2am:
### Motivation  
1.  **Random parseOptions not loaded errors** - when the TS server restarted it could compile before the Civet config promise resolved => linter errors in files
2.  **Duplicate diagnostics** - sometimes 2 linter errors in typescript files for the same thign (Civet LSP + TS).  
3.  **IDE navigation to a .civet and not generated .ts file** - when `foo.ts` and `foo.civet` coexist, we want the language server to prefer the Civet source first. Smooth exp with civetman (if there is a .ts is generated from .civet)

### What’s in this PR  
1. **Make `TSService` asynchronous**  
   * Await `CivetConfig.findConfig / loadConfig` so the service starts with the correct `CompileOptions`.  
2. **Opt-out flag for .ts / .js handling**  
   * New compiler-host switch (`allowTsJs?: boolean`, default `true`).  
   * When disabled the Civet LSP ignores plain `.ts/.js` files, avoiding duplicate diagnostics.  
3. **Shadow-file module resolution**  
   * In `resolveModuleNames` prefer a sibling `.civet` file over `.ts/.tsx/.mts/.cts`.  
   * Generates a virtual `foo.civet.tsx` file so TypeScript tooling continues to work.  

Let me know if you have any thoughts

### DIFF
--- a/lsp/package.json
+++ b/lsp/package.json
@@ -65,7 +65,18 @@
         "title": "Restart Civet Language Server",
         "category": "Civet"
       }
-    ]
+    ],
+    "configuration": {
+      "type": "object",
+      "title": "Civet",
+      "properties": {
+        "civet.langServer.includeTypescript": {
+          "type": "boolean",
+          "default": true,
+          "description": "If false, the Civet language server will not attach to JavaScript/TypeScript files, preventing duplicate definitions when another TS server is active."
+        }
+      }
+    }
   },
   "mocha": {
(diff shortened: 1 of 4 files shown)


adam2am:
Gonna craft modular PRs with separate, more clear responsibilities later

PR #1750, opened 2025-06-18
Fix CoffeeScript comments getting treated as `length` shorthand

### LINKED ISSUE #1749: bug in handling CoffeeScript style comments
By "CoffeeScript style comments", I mean comments starting with the `#` character. Illustrating this bug is pretty straightforward:

```
# temp.civet

class Parser
	text: string      # a text string
	lRules: string[]

	constructor(text: string, lRules: string[])
		@text = text     # a text string
		@lRules = lRules

# --- Create a parser
parser := new Parser(
		"3+4",    # --- string to parse
		['3', '4', '+'],
		)
console.dir parser, {depth: null}    # log it

```
Trying to compile with civet, i.e. `civet -o .ts temp.civet` gives error:

```
ParseError2: temp.civet:15:18 Failed to parse
Expected:
       ... bunch of text
Found: ","
    at validate 
   ...bunch of text
  header: 'Failed to parse',
  body: 'Expected:\n' +
   ...bunch of text
    'Found: ","\n',
  filename: 'temp.civet',
  line: 15,
  column: 18,
  offset: 277
}

```
However, if you remove the comment `# --- string to parse`, the file compiles and runs without error. I've included a bunch of other CoffeeScript style comments just to illustrate that they're usually handled without problems. Also, I can clearly remove that comment, so I do have a workaround.

Here is my `civetconfig.civet` file (I have coffeeComment set to true):

```
export default
	parseOptions:
		tab: 3
		strict: true
		implicitReturns: false
		rewriteTsImports: false

		coffeeClasses: false
		coffeeComment: true
		coffeeEq: true            // == means ===, != means !==
		coffeeInterpolation: true // interpolate via "#{}"
		objectIs: true

		autoConst: false
		autoLet: false
		autoVar: false
		coffeeBooleans: false
		coffeeDo: false
		coffeeOf: false
		coffeeDiv: false
		coffeeIsnt: false
		coffeeLineContinuation: false
		coffeeForLoops: false         // for in and for of are not swapped

```

STRd6:
Fixes #1749

### DIFF
--- a/source/parser.hera
+++ b/source/parser.hera
@@ -1479,7 +1479,7 @@ HashThis
     }
 
 LengthShorthand
-  Hash NonIdContinue ->
+  !CoffeeCommentEnabled Hash NonIdContinue ->
     const id = "length"
     return {
       type: "Identifier",
--- a/test/comment.civet
+++ b/test/comment.civet
@@ -188,13 +188,13 @@ describe "comment", ->
   testCase """
     multiline comment among directives
     ---
-    "civet -implicitReturns" /* 
+    "civet -implicitReturns" /*
       multiline comment
     */
     "civet autoLet" // hello there!
     a = -> console.log 'hi'
     ---
-    /* 
+    /*
       multiline comment
     */
     // hello there!
@@ -206,7 +206,7 @@ describe "comment", ->
     ---
     "civet autoLet" /* hi! */
     "civet -implicitReturns"
-    "civet coffeeComment"       
+    "civet coffeeComment"
     "civet coffeeBooleans" // hello!
     ---
     /* hi! */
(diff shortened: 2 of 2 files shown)


edemaine (approved):
I was hoping the fix was this easy!  Nice.

For a while I wondered whether we should support *some* length notation like `foo#`, by moving the `!CoffeeComments` check to `HashThis` specifically. But I realize that's not CoffeeScript compatible, so I'm not sure which is better.

Related, I realize that `foo#bar` and `foo.#bar` also do not treat the `#` as a comment. But here adding a space (`foo #bar`) fixes the issue. Should we aim to be consistent with that, or more exactly Coffee-compatible?

STRd6:
I think we can wait until someone raises the issue before making it more compatible. It's probably rare for Coffee comments to match that pattern but if someone says it's affecting them then we should lean towards compatibility.

PR #1754, opened 2025-06-25
early exit if no transpiled doc

STRd6:


edemaine:
Looks like this was already added: https://github.com/DanielXMoore/Civet/blob/main/lsp/source/server.mts#L1077-L1080
from #1772

Issue #40, opened 2022-12-13
LSP Semantic Token Provider


STRd6:
Trying to merge the TypeScript TextMate grammar and an updated CoffeeScript grammar seems like it would be an unspeakable nightmare. So it would be nice if the Civet parser / LSP could provide semantic tokens.

Essentially tokens that have `{$loc}` would be source tokens styled from the nearest ancestor that has some sort of `{sematicToken}` property. It might require passing a flag into the parser and skipping normal AST processing instead constructing a stream of semantic tokens.

https://code.visualstudio.com/api/references/vscode-api#DocumentSemanticTokensProvider (looks like it is very similar to how source maps work)

edemaine:
As discussed in Discord: We should aim to solve syntax highlighting at the same time. One approach is for each leaf token (with `$loc`) to have a `group` that defines what type of token it is (e.g. `keyword`), where `group` could be from [this TextMate list](https://macromates.com/manual/en/language_grammars#naming_conventions). Input tokens that don't appear in the output could still received a group via `{$loc, group: <something>, token: ''}`.  A highlighting pass would take the ASP, visit all tokens with `group`, highlight those accordingly, maybe merge together adjacent equal `group`s, and assume everything else isn't highlighted in any way.

Semantic Tokens might need more of a hierarchy (e.g. for foldable sections) so the `semanticToken` parent structure might be more useful there. Or some combination thereof.

PR #1757, opened 2025-07-10
lsp: auto-regenerate textmate grammar based on a project config

adam2am:
### The Motivation: 
Previously, enabling a feature like coffeeComment in civetconfig.json would not visually update the syntax highlighting, linter error is gone, but syntax highlighting is static, it would still highlight different words with color inside of the # comment instead of it being a comment

### The Solution:
This PR implements a regeneration of a civet.json syntax file based on config (currently only for coffeeComment parse option), and a file watcher that monitors Civet config files. On any change, it automatically regenerates the TextMate grammar (if needed) while keeping previous behavior of restarting a language-server. 

The only thing user need to do -> reload the window. The toast appears for that, where you can click the  `Reload Window`

### Key Benefits
* Instant Visual Feedback: What you configure is what you see. Changes to syntax-related flags in civetconfig.json are immediately reflected in the editor after a quick reload.
* Future-Proof & Extensible: The new grammarList.civet provides a data-driven way to add new syntax features. Adding support for another config flag is now as simple as adding a new entry to that list.

How It Works:
1. A FileSystemWatcher in extension.civet listens for changes to config files.
2. On change, it calls the new regenerateTextmate command.
3. regenerateTextmate reads the project's config, iterates through features defined in grammarList.civet, and surgically injects or removes rules from syntaxes/civet.json.

### DIFF
--- a/lsp/source/grammarList.civet
+++ b/lsp/source/grammarList.civet
@@ -26,11 +26,13 @@ CoffeeComment:
                           -   /       -> ‘divide-by’ operator
                           -   ]       -> length token inside brackets
 
-Edge-cases:
-  array.#          // length shorthand
-  object.#private  // JS private field
-  @[index %% #]    // length token in brackets
-  floor # / 2      // arithmetic
-  "#{interpolate}" // string interpolation
-  ### ###         // triple-hash block comments
+Edge-case matrix covered:
+  # comment               + comment
+  code  # trailing        + comment
+  array.# length          X not comment
+  object.#private         X not comment
+  @[index %% #]           X not comment
+  floor # / 2             X not comment
+  "#{interpolate}"        X not comment
+  ### ###             handled by separate block-comment rule
 ###
--- a/lsp/source/grammarList.civet
+++ b/lsp/source/grammarList.civet
@@ -4,7 +4,7 @@ export features = [
     marker: 'hash_comment' // The key for the repository
     snippet: `{
       "name": "comment.line.number-sign.civet",
-      "begin": "(?<=^|\\\\s)(?<!\\\\.)#(?!\\\\{)(?!\\\\s*/|\\\\s*\\])",
+      "begin": "(?<=^|\\\\s)(?<!\\\\.)#(?!(\\\\{|##|\\\\s*/|\\\\s*\\\\]))",
       "end": "$",
       "beginCaptures": {
         "0": {
@@ -14,3 +14,23 @@ export features = [
     }`
   }
 ]
+
+###
+CoffeeComment:
+  (?<=^|\s)      - preceded by BOL or whitespace
+  (?<!\.)        - NOT preceded by a dot
+  #              - the hash itself
+  (?!(\{|##|\s*/|\s*\]))   - NOT followed by any of:
+                          -   {       -> interpolation '#{'
+                          -   ##      -> triple-hash block comment
+                          -   /       -> ‘divide-by’ operator
+                          -   ]       -> length token inside brackets
+
+Edge-cases:
+  array.#          // length shorthand
+  object.#private  // JS private field
+  @[index %% #]    // length token in brackets
+  floor # / 2      // arithmetic
+  "#{interpolate}" // string interpolation
+  ### ###         // triple-hash block comments
+###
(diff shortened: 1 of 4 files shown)


lsp/source/regenerateTextmate.civet
```diff
@@ -0,0 +1,174 @@
...
+      return { modified: false, parseOptions: civetParseOpts }
+    
+    grammarFile := path.join ext.extensionPath, 'syntaxes', 'civet.json'
+    cacheFile := path.join ext.extensionPath, 'syntaxes', '.civet-grammar-cache.json'
+    out.appendLine(`Found grammar file at: ${grammarFile}`)
+
+    // Check if options have changed by comparing with cache (async)
+    cachedOpts .= {}
+    try
+      await fsPromises.access cacheFile
```
STRd6, line 72: We shouldn't call `access` before `readFile`. There's no guarantee that the file remains in the same state after the call to access and readFile will error anyway if the file is unreadable.

https://nodejs.org/api/fs.html#fspromisesaccesspath-mode

STRd6 (dismissed):
Thanks, looks like a good overall approach and it's nice to see the grammar getting some enhancements.

I'll play around with it and get it merged/deployed soon.

lsp/source/grammarList.civet
```diff
@@ -0,0 +1,38 @@
...
+  (?!(\{|##|\s*/|\s*\]))   - NOT followed by any of:
+                          -   {       -> interpolation '#{'
+                          -   ##      -> triple-hash block comment
+                          -   /       -> ‘divide-by’ operator
+                          -   ]       -> length token inside brackets
+
+Edge-case matrix covered:
+  # comment               + comment
+  code  # trailing        + comment
+  array.# length          X not comment
```
STRd6, line 32: I think this now parses as a comment after https://github.com/DanielXMoore/Civet/pull/1750

STRd6:
Hmm... actually looking a little closer this has some issues with multiple projects. Since it rewrites the grammar file and requires a restart of VSCode it won't work well if there are multiple projects with different configuration.

I was hoping VSCode had some kind of way of modifying the grammars but this may be overly complex for how many additional edge cases come up.

One approach we've been thinking about was using the TypeScript semantic tokens to provide more accurate syntax highlighting. It may be a bit more work but it should work in much broader cases as well.

STRd6:
Another possibility is that we can provide an alternative grammar that can be toggled in the file type setting. This way someone can choose the grammar based on their setting and we can add more as needed.

So for now `civet` and `civet-coffeeComment`.

We may even be able to reuse large parts of the grammar if they have some kind of include/import capability.

STRd6:
I'm fine with generating up to `5!` but in practice the top 10-20 most popular settings should cover 95% of people's use cases.

adam2am:
> I'm fine with generating up to `5!` but in practice the top 10-20 most popular settings should cover 95% of people's use cases.

I'm okay with any of the approaches. Whatever it's a `file type setting`/`command>reload`/`auto-detect>reload`. The lazy-generation seems promising, but we can start with file type and iterate, if it doesn't scale, we can always pivot to a more flexible solution later when needed

My main question here is about the file type approach: can an extension easily change the language association for all relevant files in a workspace? Or would this require users to manually configure files.associations in their settings, which might be a usability hurdle?

edemaine:
> can an extension easily change the language association for all relevant files in a workspace?

I think the extension can set the language association for each file it's applied to, via `setTextDocumentLanguage`. So each instance can check the config and top `"civet"` pragma and set its language accordingly.

adam2am:
> what sort of syntax highlighting differences are there with the config changes?
> 
> >

parseOptions : coffeeComment is bringing coffeescript type comments

```
Edge-case matrix:
# comment               + comment
code  # trailing        + comment
  array.# length          X not a comment
  object.#private         X not a comment
  @[index %% #]           X not a comment
  floor # / 2             X not a comment
  "#{interpolate}"        X not a comment
  ### ###             handled by separate block-comment rule
```

some llms when generating civet sometimes would confuse it for a coffeescript and would generate this type of comments instead of javascript type

edemaine:
@adam2am Semantic provider sounds exciting! This may solve #40 as well.

STRd6:
Closing this now that we have semantic token highlighting. Let me know if there are any things that are still missing from semantic tokens, thanks!

PR #1758, opened 2025-07-13
Comment and tidy up source mapping util

STRd6:


source/sourcemap.civet
```diff
@@ -1,8 +1,31 @@
-export type SourceMapEntries = (
-  [number, number, number, number, number] |
-  [number, number, number, number] |
-  [number]
-)[][]
+/** A source map entry from the spec with deltas for all fields */
+export type SourceMapEntry =
+| [generatedColumnDelta: number, sourceFileDelta: number, sourceLineDelta: number, sourceColumnDelta: number, sourceNameDelta: number]
```
edemaine, line 3:
```suggestion
| [generatedColumnDelta: number, sourceFileDelta: number, sourceLineDelta: number, sourceColumnDelta: number, sourceNameDelta?: number]
```

FWIW, I think this will let you remove the next line and thereby avoid repetition.

source/sourcemap.civet
```diff
@@ -1,8 +1,31 @@
...
+| [generatedColumnDelta: number, sourceFileDelta: number, sourceLineDelta: number, sourceColumnDelta: number]
+| [generatedColumnDelta: number]
+
+/** A source map entry with absolute source lines and columns, the other fields are still deltas */
+export type ResolvedSourceMapEntry =
+| [generatedColumnDelta: number, sourceFileDelta: number, sourceLine: number, sourceColumn: number, sourceNameDelta: number]
```
edemaine, line 9: Ditto here.

source/sourcemap.civet
```diff
@@ -102,88 +122,91 @@ export class SourceMap
...
-
-  if sourceMapText
-    parsed := parseWithLines sourceMapText
-    composedLines := composeLines upstreamMap.lines, parsed.lines
-    upstreamMap.lines = composedLines
+  /**
+  Remap a string with compiled code and a source map to use a new source map
+  referencing upstream source files.
+  */
+  @remap := (codeWithSourceMap: string, upstreamMap: SourceMap, sourcePath: string, targetPath: string) =>
```
edemaine, line 141: Should we use static methods (`@remap(...)`) instead of arrow functions? I'm almost surprised `@method` works, but I guess it does either way?

source/sourcemap.civet
```diff
@@ -102,88 +122,91 @@ export class SourceMap
...
+    return remappedCodeWithSourceMap
 
-      [colDelta, sourceFileIndex, srcLine, srcCol] := entry
-      srcPos := remapPosition [srcLine, srcCol], upstreamMapping
+  /**
+  Compose lines from an upstream source map with lines from a downstream source map.
+  */
+  @composeLines := (upstreamMapping: SourceMapLines, lines: SourceMapLines): SourceMapLines =>
+    lines.map (line) =>
+      line.map (entry) =>
```
edemaine, line 160: Not related to this PR: These might be faster with `for each` (but can no longer use `return`). Given that this can be pretty big, might be worth a try.

test/sourcemap.civet
```diff
@@ -1,34 +1,9 @@
...
-      y++
-
-      y = x + a
-    """
-
-    {sourceMap} := await compile(src, {
-      sourceMap: true
-    })
-
-    srcMapJSON := sourceMap.json("yo.civet", "yo.ts")
```
edemaine, line 19: What happened to this test? Admittedly we never checked anything in the resulting JSON, but perhaps we should?

source/sourcemap.civet
STRd6: I used the arrow function so that `this` binds to the class and the function can be passed around easier without `.bind`.

test/sourcemap.civet
STRd6: This was essentially duplicated by the following test.

PR #1765, opened 2025-07-21
feat: add nix packages and devshells

kasimeka:


kasimeka:
to update `yarnDepsHash` after modifying `yarn.lock` files a maintainer should
1. omit the `yarnDepsHash` field from `mkCivetPackage`'s arguments for their package
2. build the target package with `nix build .#<output-name>`
3. wait for the new dependencies to be downloaded and hashed, an expected error will be emitted from this step as follows:
   ```console
   error: hash mismatch in fixed-output derivation '/nix/store/sgqsgja2ax43633r4jjdvh9inb8inxqj-offline.drv':
            specified: sha256-AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA=
               got:    sha256-pEpST+ZWDJYzIaxeb4ou8rLtNwgnKv0SdahHJ15bgn8=
   error: Cannot build '/nix/store/vwm5gkrk9gl26s26vgkmcyji7barafv1-civet-0.10.5.drv'.
          Reason: 1 dependency failed.
          Output paths:
            /nix/store/1fkmcwawv0zzkaninsy90nynci314i5q-civet-0.10.5
   ``` 
5. copy the value of "got", which in this example is `sha256-pEpST+ZWDJYzIaxeb4ou8rLtNwgnKv0SdahHJ15bgn8=` and use it as the new `yarnDepsHash` value
6. rerun the build step after updating the hash and committing the change after it builds successfully

there are tools that _attempt to automate_ this process but i don't find any of them convenient enough to use.

STRd6:
We could probably make a bash script that automates the update.

kasimeka:
@STRd6 added an update script

.gitignore
```diff
@@ -4,3 +4,6 @@ coverage/
 dist/
 node_modules/
 register-noconfig.js
+
+# nix build outputs path
```
edemaine:
```suggestion
# nix build output path
```

kasimeka:
the cli issue should be fixed now. it was based on a misunderstanding on my end of how the cli is bundled. now instead of overriding the `installPhase` added by `yarnInstallHook` i moved the entrypoint creation code to a `fixupPhase` that runs after the installation code. this results in a `civet` cli entrypoint that has access to `node_modules` during runtime.

seanstrom:
Hey @kasimeka , could you rebase this branch 🙏

kasimeka:
ive rebased the branch and updated nix derivation hashes

PR #1769, opened 2025-08-07
Fix REPL handling of blank lines on Node 24

edemaine:
On Node 24 (.4.1 and .5 at least), the REPL is broken in that it no longer correctly detects a blank line to execute or compile the code.  Apparently new Nodes provide an input that contains `\r` for newlines for all but the final `\n`.  (Tested on both Windows and Linux, so doesn't seem to be OS dependent.)


STRd6 (approved):
Thanks, Node remains a mystery :)

PR #1773, opened 2025-08-13
lsp(feat): Add option to disable language server for JS/TS files

adam2am:
This PR introduces a new configuration setting to prevent the Civet language server from providing diagnostics for JavaScript and TypeScript files.

### The Problem

When working in a project that already has a dedicated TypeScript language server active (built into VS Code), the Civet language server also attaches to `.js` and `.ts` files and 

This results in **duplicate diagnostics** for the same code, which can be noisy and confusing 

### The Solution

This PR introduces a new VS Code setting:
`"civet.langServer.includeTypescript"`

This setting allows users to control whether the Civet Language Server should activate for JavaScript and TypeScript files.

*   It defaults to `true` to maintain the existing behavior for all current users, ensuring no breaking changes.
*   When a user sets it to `false`, the language client will only register itself for `.civet` files, resolving the duplicate diagnostics issue.

### DIFF
--- a/lsp/package.json
+++ b/lsp/package.json
@@ -65,7 +65,18 @@
         "title": "Restart Civet Language Server",
         "category": "Civet"
       }
-    ]
+    ],
+    "configuration": {
+      "type": "object",
+      "title": "Civet",
+      "properties": {
+        "civet.langServer.includeTypescript": {
+          "type": "boolean",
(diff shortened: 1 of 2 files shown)


STRd6 (approved):
Nice addition, thanks!

PR #1774, opened 2025-08-14
fix(lsp): simple updates for tests for async TSService + removed hardcoded paths

adam2am:
Hey, got some simple positive changes here

- **Support Async `TSService`**: The tests have been updated with `async/await` to handle this correctly. A longer test timeout was also added to prevent premature failures.
- **Remove Hardcoded Paths**: The service tests used a hardcoded absolute file path (`/home/daniel/...`), which caused them to fail on any other machine.
- **Correct Sourcemap Access**: The utility tests were updated to access sourcemap lines via `sourceMap.data.lines` instead of `sourceMap.lines`


STRd6 (approved):
Thanks!

PR #1775, opened 2025-08-14
lsp(refactor): rearchitect update queue for atomic, project-based changes to handle conflicts and race conditions

adam2am:
#### The Motivation, Problem

While implementing the `renameProvider`, I discovered a limitation. When an operation would affect multiple files at once, the old `executeQueue` would process each file change sequentially.

This created a race condition where the TypeScript service would analyze the project in an inconsistent state (e.g., with file A updated but file B not yet updated), leading to temporary but blocking linter errors and incorrect diagnostics. This made cross-file operations not robust at all

#### The Solution

This PR refactors the document update mechanism to be **project-aware and atomic**.

Instead of a simple sequential queue, the new `executeQueue` works in two phases:
1.  **Group & Stage:** It first groups all pending document changes by their parent project. Then, for each project, it stages *all* relevant file changes in the TypeScript service's memory without triggering analysis.
2.  **Analyze:** Once a project's state is fully consistent with all changes, it triggers a single analysis pass to generate correct diagnostics for all affected files.

This ensures that the TypeScript service always operates on a complete and consistent view of the project during multi-file changes.

#### Benefits
-   **Fixes Cross-File Operations:** Unblocks the implementation of `renameProvider` and other complex, multi-file features.
-   **Improves Performance:** A single, batched analysis per project is performed instead of N redundant analyses for N changed files.
-   **Increases Stability:** Eliminates diagnostic flickering and incorrect errors during rapid or large-scale file modifications (e.g., git branch switching, find-and-replace).

Happy to hear any thoughts on this

### DIFF
--- a/lsp/source/server.mts
+++ b/lsp/source/server.mts
@@ -566,29 +566,47 @@ let changeQueue = new Set<TextDocument>()
 let executeTimeout: Promise<void> | undefined
 const documentUpdateStatus = new Map<string, WithResolvers<boolean>>()
 async function executeQueue() {
-  // Cancel updating any other documents while running queue of primary changes
+  // Cancel any in-flight project-wide diagnostics update to avoid conflicts
   if (runningDiagnosticsUpdate) {
     runningDiagnosticsUpdate.isCanceled = true
   }
-  // Reset queue to allow accumulating jobs while this queue runs
-  const changed = changeQueue
-  changeQueue = new Set
-  logger.log("executeQueue " + changed.size)
-  // Run all jobs in queue (preventing livelock).
-  for (const document of changed) {
-    await updateDiagnosticsForDoc(document)
-    documentUpdateStatus.get(document.uri)?.resolve(true)
-    Promise.resolve()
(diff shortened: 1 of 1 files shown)


lsp/source/server.mts
```diff
@@ -566,29 +566,47 @@ let changeQueue = new Set<TextDocument>()
...
+
+  // Group documents by their project path to ensure atomic updates per project.
+  for (const doc of changed) {
+    const sourcePath = documentToSourcePath(doc);
+    const projPath = getProjectPathFromSourcePath(sourcePath);
+    if (!docsByProject.has(projPath)) {
+      const service = await ensureServiceForSourcePath(sourcePath);
+      if (service) docsByProject.set(projPath, { service, docs: [] });
+    }
+    docsByProject.get(projPath)?.docs.push(doc);
```
edemaine, line 591: Would it help to deduplicate documents here? Or were they already deduplicated by `changeQueue`?

adam2am: `changeQueue` is being a Set, so I don't think there could be an issue

PR #1779, opened 2025-08-17
LSP: fix(diagnostics), ensure immediate updates for opened dependent files on change

adam2am:
Hey, quick fix for my previous refactor

### **The Problem (Bug)**

Currently, if a user modifies an exported member in `export.civet`, any other open file (`import.civet`) that imports that member does not instantly receive updated diagnostic errors even when files are open in editor. It appears only after direct edit in the `import.civet`

### **How to Replicate**

1.  **Create two files:**

    *   `export.civet`:
        ```
        abc := 321
        export {abc}
        ```

    *   `import.civet`:
        ```
        import {abc} from './export.civet'
        ```

2.  **Open both files** in the editor. Observe that there are no errors.

3.  **In `export.civet`**, change the exported variable name:
    ```civet
    // From
    abc := 321
    export {abc}

    // To
    a := 321
    export {a}
    ```

4.  **Observe (The Bug):** Switch to the `import.civet` file. No diagnostic error is shown. The linter is silent, even though the import of `abc` is now invalid.

5.  **Expected Behavior:** The moment `export.civet` is saved or changed, an error should immediately appear in `import.civet`, highlighting that `abc` is not an exported member.

### **The Solution**

The root cause of this bug was a path mismatch within the `updateProjectDiagnostics` function. The function iterates through all files in the TypeScript program using `program.getSourceFiles()`. 

The fix is to use **`service.getSourceFileName(sourceFile.fileName)`**. This method correctly resolves the path known by the TypeScript service back to its original source file path. This ensures we can reliably find the corresponding open document in our `documents` map and push the updated diagnostics to the editor immediately after a dependency changes.

Happy to hear any thoughts, its a small, but important fix imo

### DIFF
--- a/lsp/source/server.mts
+++ b/lsp/source/server.mts
@@ -725,12 +725,14 @@ const updateProjectDiagnostics = async (
   // A single, short delay before starting the update for a project.
   await setTimeout(diagnosticsPropagationDelay);
 
-  for (const sourceFile of program.getSourceFiles()) {
+  const projectFiles = program.getSourceFiles();
+  for (const sourceFile of projectFiles) {
     if (status.isCanceled) return;
 
-    // We only send diagnostics for files the user actually has open,
-    // even though we're checking every file in the project for correctness.
-    const docUri = pathToFileURL(sourceFile.fileName).toString();
+    // We only need to send diagnostics for files the user has open
+    // TypeScript is analyzing all files in memory anyway
+    const rawFileName = service.getSourceFileName(sourceFile.fileName);
+    const docUri = pathToFileURL(rawFileName).toString();
     const doc = documents.get(docUri);
     if (doc) {
       await updateDiagnosticsForDoc(doc, service);


adam2am:
By the way, I think 
```ts
if (doc) {
      await updateDiagnosticsForDoc(doc, service);
```
also have a room for improvement. But this not as urgent as this fix

lsp/source/lib/typescript-service.mts
```diff
@@ -455,7 +451,7 @@ async function TSService(projectURL = "./", logger: Console | RemoteConsole = co
   logger.info("CIVET VSCODE PLUGIN " + version)
   logger.info("TYPESCRIPT " + typescriptVersion)
 
-  const projectPath = fileURLToPath(projectURL)
+  const projectPath = projectURL.startsWith("file:") ? URI.parse(projectURL).fsPath : path.resolve(projectURL)
```
seanstrom, line 454: Is `projectURL` usually an absolute path? Is there a chance that a relative path would be passed in that wouldn't correctly resolve with `path.resolve`?

Not sure if this is a real concern, but this was the main thing that jumped out to me, let me know what you think.

edemaine: Maybe I'm missing something here, but wouldn't `path.resolve` correctly resolve a relative path? For example, `path.resolve('.')` returns `process.cwd()`.

Also, I'm guessing the answer to your first question is "yes". I checked and `fileURLToPath` crashes when given a relative path like `"./"` (the argument default!), so the old behavior would have just been crashing.

```node
> require('url').fileURLToPath('./')
Uncaught TypeError: Invalid URL
    at new URL (node:internal/url:828:25)
    at Object.fileURLToPath (node:internal/url:1609:12) {
  code: 'ERR_INVALID_URL',
  input: './'
}
```

seanstrom: Yeah `path.resolve` would treat `"."` as `process.cwd()`, I just wasn't sure which directory the LSP process starts from, perhaps it's the workspace directory (or whichever directory the editor was initially opened with)

and I was wondering if there was a chance where a relative path is passed in but expects a different base directory (a mismatch between the LSP cwd and the file path base directory)

edemaine: Oh, I see — fair question. I don't think the LSP could interpret a relative path another way though; the only other meaning would be relative to project root, but that's what this is specifying. The LSP needs to be able to read the files, so it can't just deal with abstract relative paths.

seanstrom: Yeah good point, I agree there shouldn't be abstract relative paths here, just stuff relative to the project root, if that's the case then I don't see an issue with this code.

PR #1786, opened 2025-09-04
Allow implicit object literal to end by `)`/`]`/`}`

### LINKED ISSUE #1781: Multi-line hash treated as multiple depending on last bracket position
Bit of a strange one, discovered under `civet coffeeCompat`, but applies to plain civet too.

The following multi-line single hash gets split into two hashes unless final bracket is on next line. I presume this is a bug?

```
console.log(
  401: "four zero one",
  422: "four two two")
```
compiles to:
```
console.log(
  {
    401: "four zero one",
  },
  { 422: "four two two" },
);
```

Moving the final bracket down a line works as expected:
```
console.log(
  401: "four zero one",
  422: "four two two"
)
```
compiles to:
```
console.log({
  401: "four zero one",
  422: "four two two",
});
```

edemaine:
Fixes #1781

Also removes some (but not all) added commas at the end of object literals, bringing us closer to the source.

@greghuc If you get a chance, feel free to test this as well. 😺

### DIFF
--- a/source/parser.hera
+++ b/source/parser.hera
@@ -2016,7 +2016,7 @@ ParameterElement
 
 ParameterElementDelimiter
   _? Comma
-  &( __ [)}] )
+  &( __ [)\]}] )
   &EOS InsertComma -> $2
 
 # https://262.ecma-international.org/#prod-BindingIdentifier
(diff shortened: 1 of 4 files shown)


STRd6 (approved):
👏

greghuc:
@edemaine your fix looks good!

I ran this branch against the code where I first saw the bug, and it now runs without error at runtime. Eyeballing the transpiled js, I can see that where there were previously two `{}` hashes, there is now (the intended) single `{}` hash.

Nice one, and thanks.

PR #1811, opened 2025-10-20
Limit implicit JSX fragments to desired scenarios

### LINKED ISSUE #1801: JSX auto-fragment combines lines it shouldn't
Input:
```coffee
function MyComponent()
  foo := <foo-element/>

  <want-to-return-this/>
```

Expected:
```typescript
function MyComponent() {
  const foo = <foo-element/>

  return <want-to-return-this/>
}
```

Actual:
```typescript
function MyComponent() {
  const foo = <>
  <foo-element/>

  <want-to-return-this/>
  </>;return foo
}
```

This problem happens in the top level, too. I just wrapped it in a function to illustrate a situation where the behavior is likely to cause confusion.

edemaine:
Fixes #1801

It seems like we needed a new flag for this, because we have some scenarios where we want to allow a nested binary operator but not a nested JSX tag (though also re-used some existing flags).

I might have missed some contexts where implicit JSX fragments should be forbidden, but it's easy to add them now.

### DIFF
--- a/source/parser.hera
+++ b/source/parser.hera
@@ -91,6 +91,7 @@ export const state = {    // parser state
   forbidTrailingMemberProperty: [false],
   forbidNestedBinaryOp: [false],
   forbidNewlineBinaryOp: [false],
+  forbidImplicitFragment: [false],
   forbidPipeline: [false],
   JSXTagStack: [undefined],
 }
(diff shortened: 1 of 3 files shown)


STRd6:
Thanks!

PR #1812, opened 2025-10-21
Spread `...x` makes `yield*` in `for*` loops, fix `for*` in one-line arrow functions

edemaine:
@bbrk24 noticed that there wasn't a way to `yield*` from a `for*` loop.  My intuition was to write `...x` in the body, which currently generates `yield ...x`, which is illegal JavaScript.  So this seems like a natural place to generate `yield* x`. We could instead offer `*x` (and still could in the future), but that would require bigger changes in the grammar.

I also noticed that one-line arrow functions didn't work with `for*` loops because of a typo in the grammar, so I fixed that as well.  This also caused some semicolons to disappear in a test, which seems fine.

### DIFF
--- a/source/parser.hera
+++ b/source/parser.hera
@@ -4559,7 +4559,7 @@ NoCommaStatement
   KeywordStatement
   VariableStatement
   IfStatement
-  IterationStatement
+  IterationActualStatement
   SwitchStatement
   TryStatement
   EmptyStatement
--- a/test/for.civet
+++ b/test/for.civet
@@ -1172,6 +1172,25 @@ describe "for", ->
   """
 
   testCase """
+    for generator with one-line body
+    ---
+    function* f()
+      for* x of y then x
+    ---
+    function* f() {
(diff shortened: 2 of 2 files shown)


source/parser/function.civet
```diff
@@ -1175,12 +1175,20 @@ function expressionizeIteration(exp: IterationExpression): void
     iterationDefaultBody statement
 
     assignResults block, (node) =>
-      type: "YieldExpression"
-      expression: node
-      children:
-        . type: "Yield"
-          token: "yield "
-        . node
+      let star: ASTNode
```
bbrk24, line 1178: Shouldn't this be `ASTNode | undefined`, since it's never set for non-spread expressions?

edemaine: [`ASTNode` includes `undefined`](https://github.com/DanielXMoore/Civet/blob/43abdd790e04e1144251dbfd7ac522f42ee3671c/source/parser/types.civet#L1)

PR #1814, opened 2025-11-02
Fix complex splice assignment by adding parentheses

### LINKED ISSUE #1813: Splice assignment bug
`x[i-1..i] = []` produces `x.splice(i-1, i + 1 - i-1, ...[])` but the `i-1` needs to be wrapped in parentheses!

STRd6:
Fixes #1813

### DIFF
--- a/source/parser/lib.civet
+++ b/source/parser/lib.civet
@@ -1141,7 +1141,7 @@ function processAssignments(statements): void
             c[1] = start
             c[2] = ", "
             if (end)
-              c[3] = [end, " - ", makeLeftHandSideExpression(start)]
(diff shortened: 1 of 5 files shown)


source/parser/lib.civet
```diff
@@ -1141,7 +1141,11 @@ function processAssignments(statements): void
             c[1] = start
             c[2] = ", "
             if (end)
-              c[3] = [end, " - ", start]
+              startExpr := if Array.isArray(start)
+                makeNode { type: "ParenthesizedExpression", children: ["(", start, ")"], expression: start }
+              else
+                start
+              c[3] = [end, " - ", startExpr]
```
edemaine: I think it's better to use `makeLeftHandSideExpression` which has a bunch more cases where parentheses aren't necesasry.

PR #1838, opened 2026-01-15
Infer partial object/array types from initializers

### LINKED ISSUE #1744: Default values in `::` object argument types are `unknown` when they could be inferred
When mixing `::` types and default values in function arguments `spacing` is type `unknown` when we could probably infer it as `number`.

```civet
export function DisplayOptions({
  stage,
  position:: Vec2,
  options:: {
    text: string
    onSelect: () => void
  }[],
  spacing = 60
}) { }
```

edemaine:
Fixes #1744

We could generalize to infer types of initializers like `= { foo: "bar" }`.  Sadly we can't write `typeof { foo: "bar" }`, but for object and array literals we could recurse. But maybe leave that for another PR? Actually I'm not sure what you'd really want to infer for e.g. `= {}`.

### DIFF
--- a/integration/unplugin-examples/vite/src/main.civet
+++ b/integration/unplugin-examples/vite/src/main.civet
@@ -1,3 +1,10 @@
 {a} from "./module"
+MyWorker from "./worker.civet?worker"
 
 console.log a
+
+worker := new MyWorker()
+worker.onmessage = (event) ->
+  console.log "Worker says:", event.data
+  document.body.appendChild document.createTextNode "Worker says: " + event.data + "\n"
+worker.postMessage "ping"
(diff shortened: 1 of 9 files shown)


edemaine:
@greptileai There actually is a test for initializer-only declarations (specifically `[ { x: other, y = init, ...r, }, z = 0, ...rest ] := list`). It's actually intentional that they don't get type definitions in this case. The idea is to only add type declarations when the user has used `::` explicitly somewhere, and stick to JavaScript otherwise.

STRd6:
> (In case it's not clear, I believe the AI is wrong, but I'm going to stop arguing with it.)

I think we can tweak it's settings to emphasize where it's more helpful or provide additional context. And they say it can also gain context over time so hopefully it improves. Otherwise we'll can put it on timeout :P

STRd6 (approved):

PR #1843, opened 2026-01-24
Fix REPL continuation prompt to `...`, clarifying indentation

### LINKED ISSUE #1768: CLI uses `|` instead of `...` for continuation
Node seems to have recently changed the continuation prompt from `...` to `:`. This means our prompts are no longer vertically aligned, which makes it hard to type indentation manually. We should find a way to restore, possibly by overriding [displayPrompt](https://nodejs.org/api/repl.html#replserverdisplaypromptpreservecursor).

## Old Node
```sh
⟫ civet -c
Civet 0.9.0 REPL.  Enter a blank line to transpile code.
🐈> a^1
...
a
🐈>
```

## New Node
```sh
⟫ civet -c
Civet 0.10.5 REPL.  Enter a blank line to transpile code.
🐈> a^1
|
a
🐈>
```

edemaine:
Fixes #1768

This hack seems to work on older Node (I tested 20 and 22) before the continuation prompt changed, as well as modern Node (24) where this hack is necessary. Demo:

image

One catch here is that I didn't invent this hack. I actually thought it was impossible! But I thought I'd give GPT-5.2 Codex + OpenCode a try at it, and it figured out a way to get at those internal symbols that I thought were impenetrable.

We can revisit this when Node finally makes the continuation prompt reconfigurable (there have been various PR attempts, though none merged yet). Though we probably will need the hack for a while, until patched versions of Node become commonplace.

### DIFF
--- a/source/cli.civet
+++ b/source/cli.civet
@@ -267,12 +267,13 @@ export function repl(args: string[], options: Options)
   } code.`
   global.quit = global.exit = => process.exit 0
   import * as nodeRepl from node:repl
-  r := nodeRepl.start
-    prompt:
-      switch
-        when options.ast then '🌲> '
-        when options.compile then '🐈> '
-        else '🐱> '
+  prompt :=
+    switch
+      when options.ast then '🌲> '
+      when options.compile then '🐈> '
+      else '🐱> '
+  r := nodeRepl.start {}
+    prompt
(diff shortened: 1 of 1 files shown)


STRd6 (approved):
Neat!

PR #1849, opened 2026-02-12
Spreads and multiple items in parenthesized postfix loops, postfixes in declarations

### LINKED ISSUE #1756: Fancy array comprehensions with `for` at end don't always work
While exploring [a related proposal for Python](https://discuss.python.org/t/pre-pep-unpacking-in-comprehensions/96362/75), @adqm and I noticed that some forms of fancy array comprehensions with `...rest` bodies don't work with `for` at the end.

1. `...xs for xs of [[1,2,3], [], [4,5]]` works (produces `[1, 2, 3, 4, 5]`)
2. `[...xs for xs of [[1,2,3], [], [4,5]]]` produces `[ [ 1, 2, 3 ], [], [ 4, 5 ] ]`. This might be correct actually; `[...foo]` has a special meaning.
3. `z = ...xs for xs of [[1,2,3], [], [4,5]]` **fails to parse**
4. `(...xs for xs of [[1,2,3], [], [4,5]])` **fails to parse**
5. Ditto for above two with `xs...` instead of `...xs`
6. `z = for xs of [[1,2,3], [], [4,5]] then ...xs` (or the indented form) works fine, so this is specific to the `for`-at-end form

edemaine:
Fixes #1756. In fact, some of those examples weren't supposed to work, because (I forgot) assignments go *inside* postfix `for` loops (same as CoffeeScript). This is now documented behavior.

The main issue was that `PostfixedExpression` didn't allow spreads. Now it allows spreads ~~as well as comma operators to indicate multiple items to add to the array~~.

I also noticed that `PostfixedCommaExpression` was broken. Previously it allowed a postfix followed by comma operators, which wasn't the intent. Now `PostfixedCommaExpression` does actually allow a form of mixing comma and postfix now, but only commas before postfix, and with a specific meaning: multiple items per iteration. We allow this in contexts where comma doesn't have another meaning.

I also changed `Initializer` to allow for `PostfixedExpression` and declaration shorthand to allow for `PostfixedCommaExpression`, so that you can write nice one-liner for loops (just using declarations instead of assignments).

### DIFF
--- a/source/parser.hera
+++ b/source/parser.hera
@@ -4562,6 +4562,8 @@ PostfixedExpression
   Expression
 
 # Expression with postfixes *or* comma operators (mixing seems ambiguous)
+# Note that postfixed loops do allow commas though, for accumulating multiple
+# items per iteration
 PostfixedCommaExpression
   # Loops allow commas and spreads before the postfix
   CommaExpressionSpread:expression _?:ws !IfClause PostfixStatement:post ->


civet.dev/reference.md
```diff
@@ -1990,6 +1990,32 @@ function concatIter(iters)
...
+
+If you want to assign the loop expression to a variable,
+wrap it in parentheses or use a non-postfix loop:
+
+<Playground>
+keys = (key for key in object)
+keys = for key in object then key
+</Playground>
+
+Declaration initializers do not need to be wrapper in parentheses:
```
STRd6: `wrapper` should be `wrapped`

STRd6:
I like the expanded syntax for spreads and commas in a postfix context but I think we may need to restrict it to the top level or inside parenthesized expressions and disallow it in array contexts.

With what the detected I think keeping the behavior inside arrays as it was previously is more correct.

Disallowing spreads and comma for if/unless postfixed expressions makes sense. Our current parsing seems to be somewhat incompatible with our ES compatibility goals and CoffeeScript doesn't allow it at all.

```
a, b, c if x
---
// Current behavior seems incorrect for ES compatibility perspective
if (x) {
  a, b, c
}
---
// Seems like a more sensible ES compatible parse
a, b, x ? c : void 0
```

I think in an array context the spreads and commas of postfix expressions should belong to the containing array and not to the postfix expression.

So in a situation like:

```
[ a, b, x.value for x of xs]
```

The `[a, b, ` should always be parsed as the array. If there were a spread it seems like it would also need to be part of the array otherwise the postfix would steal it and make it challenging to combine spreads with postifx in an array context.


```
[ a, b, ...x.values for x of xs]
```

The array should keep the `...` as in `[a, b, ...(result of postfix loop)]`.

To use the expanded syntax support for spreads and commas in an array context we should require wrapping in parens so we can keep explicit which pieces are part of the postifx and which are part of the containing array:


```
[ a, b, ...(item.x, item.y for item of items) ] // The parens allow us to use the expanded postfix context within the array.
```

We'll may also want to add a few more tests for this inside array contexts but everything at the top level grammar wise looks good to me.

PR #1853, opened 2026-02-25
Postfix support in parenthesized or indented function arguments

edemaine:
As mentioned in https://github.com/DanielXMoore/Civet/pull/1849#issuecomment-3928953069, we can safely support postfixes (like `... if ...`) in function arguments provided the arguments are

* parenthesized: `f(x if y)`, or
* indented:
  ```js
  f
    x if y
  ```
  or
  ```js
  f a,
    x if y
  ```
  These two are changed behavior. The latter [matches CoffeeScript](https://coffeescript.org/#try:f%20a%2C%0A%20%20x%20if%20y).
* but not implicit: `f x if y` has an existing meaning: `if (y) f(x)`

This PR adds that support. It brings parenthesized function arguments into alignment with array literals. Notably:

```js
f(a, b, c if d)
// means:
f(a, b, d ? c : undefined)
// NOT:
f(d ? (a, b, c) : undefined)
```

BREAKING CHANGE: Function application with a postfix on the last indented argument now applies the postfix to the argument, not the whole function application.

### DIFF
--- a/civet.dev/reference.md
+++ b/civet.dev/reference.md
@@ -1,4 +1,4 @@
----
+-:--
 title: Reference
 aside: false
 ---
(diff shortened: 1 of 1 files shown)


civet.dev/reference.md
```diff
@@ -1,4 +1,4 @@
----
+-:--
```
vendethiel, line 1: is that added on purpose? (or I mid-`:w` insert?)

edemaine: Oops, thanks for spotting!

PR #1855, opened 2026-02-28
LSP file completion from empty `import/from` via robust parsing; restrict to single-line module names

edemaine:
This is a followup to #1854 that allows for the LSP to autocomplete the list of modules when in these states (where `|` denotes cursor location):

```js
import |
import '|
import "|
import { foo } from |
import { foo } from "|
```

If the user adds a quote, the autocomplete will include a closing quote.

How this works:

* Civet notices invalid `import` or `from` clauses and generates a special `"Error"` node for them, and makes them "valid" by adding either an empty string `""` (if there wasn't anything) or a closing quote (if there was an open-but-not-closed string). (This was the approach of my previously failed #1519, which I'll revisit.)
* The LSP (already) keeps the transpilation for these types of errors (as opposed to full `ParseError`) because it gets an AST, and gives it to TypeScript.
* For this to work, I needed to forbid `import`s of the following form:
  ```coffee
  import "hello
  world"
  ```
  This previously transpiled to `import "hello\nworld"` which does actually work (at least in Node). I just hope people don't have newlines in their filenames. If they do, they could write `import "hello\nworld"` I guess.
* For this to work, I also needed to clean up a bunch of `__`. Previously this was valid:
  ```coffee
  import
  foo
  ```
  It transpiled to
  ```js
  import
  "foo"
  ```
  which in JS is equivalent to `import "foo"`. But I think it falls under "improperly indented code" so it's good to forbid it.
* Note: I actually ended up changing all `__` to `SameLineOrIndentedFurther` which might not always be perfect for multiline `import`s; once we have at least one strict indentation the rest could be `NotDedented`. But I didn't want to spend a lot of time on this; perhaps we can fix it later? I'd actually like to consider an indented form of `import` which might clean this up.

I couldn't easily get `from |` to work, because by itself, `from` is valid Civet code that isn't `import` related.

BREAKING CHANGE: `import`s must now be properly indented, and module names can no longer have actual newlines in them.

### DIFF
--- a/lsp/source/server.mts
+++ b/lsp/source/server.mts
@@ -298,13 +298,12 @@ function createCivetFileCompletions(
   files: string[],
   sourcePath: string,
   position: Position,
-  closingQuoteSuffix: "'" | '"' | '' = '',
 ): CompletionItem[] {
   return files.map((file) => {
     return {
       label: file,
       kind: CompletionItemKind.File,
-      insertText: `${file}${closingQuoteSuffix}`,
+      insertText: file,
       sortText: `${0}_${file}`,
       data: {
         sourcePath,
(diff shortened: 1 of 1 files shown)


seanstrom:
Seems to work pretty well for me, nice stuff!

STRd6 (approved):
Looks good! Changing `__` to stricter indentation makes sense. A lot of the import code was from before we had some of the parser rules. If someone does run into a case that should be valid it will hopefully be simple enough to correct it as it comes up.

lsp/source/server.mts
```diff
@@ -512,27 +544,35 @@ connection.onCompletion(async ({ textDocument, position, context }) => {
   //  0: Default, when sourcemap cursor position is already perfect.
   // -1: Gets inside closing quotes for import file completion.
   p += heuristics.cursorOffsetAdjustment
+  if (isSpaceImport) {
+    // If we're at `import ` followed by EOF, sourcemapping is a bit wonky,
+    // and we start on the left of the implicit "" instead of the right.
+    // So now we've gone one step left when we need to go one step right.
+    if (transpiledDoc.getText().slice(p, p + 3) === ' ""') p += 2
```
seanstrom, line 551: I'm a little surprised with source-mapping behaviour here, but the workaround makes sense to me.

Maybe we could pass a `Range` to `getText` instead?

Something like 
```
transpiledDoc.getText(new Range(p, p + 3)) === ' ""'
```

PR #1856, opened 2026-03-04
LSP completion at end of file

edemaine:
This started as an implementation of the optimization mentioned in https://github.com/DanielXMoore/Civet/pull/1855#discussion_r2868067387

But then I noticed that, at end of file, `import "` and `import '` failed to produce completions (while `import ` worked fine). This PR makes a few more tweaks to fix that.

@seanstrom would you mind testing once more?


seanstrom:
Hmm odd, I re-tested inside the Civet project and I'm getting the results that you mention, but when I ran the extension (in vscode debug mode) inside of another project I wasn't getting the package results. Probably has something to do with my project configuration around typescript (I think)

PR #1861, opened 2026-03-11
Allow class fields with the same name as modifiers

### LINKED ISSUE #1860: Can't have a class field named `readonly`
```
class X
  readonly: boolean;
```

https://civet.dev/playground?code=Y2xhc3MgWAogIHJlYWRvbmx5OiBib29sZWFuOwo%3D

Currently fails to parse but should parse as a non-readonly field named `readonly`.

STRd6:
Fixes #1860

### DIFF
--- a/test/class.civet
+++ b/test/class.civet
@@ -41,7 +41,7 @@ describe "class", ->
   """
 
   testCase """
-    field names are not reserved words
+    fields with same name as modifiers can be parsed
     ---
     class X
       abstract = false
--- a/source/parser.hera
+++ b/source/parser.hera
@@ -1421,20 +1421,20 @@ ClassSignatureElement
   Static ClassSignatureBody
 
 AbstractModifier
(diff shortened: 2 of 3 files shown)


test/class.civet
```diff
@@ -40,6 +40,31 @@ describe "class", ->
...
+      declare = false
+      override = false
+      private = false
+      protected = false
+      public = false
+      readonly = false
+      static = false
+    }
+  """
+
```
edemaine, line 67: Should we add a version with `: boolean` instead of initialization to `test/types/class.civet`?

test/class.civet
```diff
@@ -40,6 +40,31 @@ describe "class", ->
     }
   """
 
+  testCase """
+    field names are not reserved words
```
edemaine: Instead of "reserved" maybe "semi-reserved" or "special"?

Issue #1868, opened 2026-03-26
rewrite-civet-imports doesn't support dynamic imports


edemaine:
## Input
```js
"civet rewrite-civet-imports=.js"
await import "./foo.civet"
```
## Current Output
```js
await import("./foo.civet")
```
## Expected Output
```js
await import("./foo.js")
```

Obviously we can't rewrite all dynamic imports, but when it's a string literal, I believe we should. Otherwise it's difficult to typecheck the original file, while still loading the correct file after a build.

Hmm, `rewrite-civet-imports` is also running when typechecking (in VSCode and `civet --typecheck`). I wonder if we should disable it? Otherwise we'd need to turn it on only during build.

PR #1869, opened 2026-03-27
`eslint.config.civet` support via `civetlint`, ESM query support

edemaine:
Adds a `civetlint` binary to the eslint plugin that lets you:

* Just run `civetlint` with no configuration to lint with reasonable defaults (assuming eslint and optionally typescript-eslint is installed)
* Manually config via `eslint.config.civet` instead of `eslint.config.{js,mjs,cjs}`, via ESM plugin
* Or config via `eslint.config.{js,mjs,cjs}` as usual, but then you could just use `eslint`

This required updating the ESM plugin to support queries in the URLs. For some reason, eslint opens the config as `file:///C:/Users/edemaine/Projects/Civet/integration/eslint/example/eslint.config.civet?mtime=1774512142638`.

I also fixed up types and cleaned up the code. Type checking is almost perfect, except I'm a bit hampered by #1868.

### DIFF
--- a/source/esm.civet
+++ b/source/esm.civet
@@ -51,7 +51,7 @@ export type RegisterOptions
   parseOptions?: ParseOptions
 
 baseURL := pathToFileURL(process.cwd() + '/').href
-extensionsRegex := /\.civet$/
+extensionsRegex := /\.civet(?:$|[?#])/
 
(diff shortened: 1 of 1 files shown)


STRd6 (approved):
👍

PR #1871, opened 2026-03-27
Fix emitDeclaration builds in ESM unplugin output.

STRd6:
Provide a Node require shim for the bundled ESM unplugin build and add an integration regression test covering esbuild + emitDeclaration output.


edemaine (approved):
Wild. I'll try to report upstream to `@typescript/vfs`, but maybe they're constrained somehow... Anyway, this is a good workaround.

PR #1872, opened 2026-03-27
Sourcemap sources should be relative to outdir

STRd6:


source/unplugin/unplugin.civet
```diff
@@ -548,9 +533,16 @@ export const rawPlugin: Parameters<typeof createUnplugin<PluginOptions>>[0] =
         if sourceMap <? "string"
           JSON.parse(sourceMap)
         else
+          outDir := if meta.framework is 'esbuild'
+            esbuildOptions.outdir or
+            esbuildOptions.outfile!.split('/').slice(0, -1).join('/')
```
edemaine: Splitting on `/` is maybe fragile for Windows. Is there a reason not to use `path.dirname`? Are paths normalized to `/` already here? (Then we could still use `path.posix.dirname`.) FWIW, `yarn test` is working on Windows.

source/unplugin/unplugin.civet
```diff
@@ -548,9 +533,16 @@ export const rawPlugin: Parameters<typeof createUnplugin<PluginOptions>>[0] =
...
+          else
+            // TODO: Handle other frameworks
+            rootDir
```
edemaine: Maybe, similar to how `rootDir` works today, we could have an `outDir` "global" that gets set by each architecture in the config callback? For example, for Vite, I think it's [`build.outDir`](https://vite.dev/config/build-options#build-outdir).

Happy to take a stab at this refactor if you'd like.

STRd6: Thanks, good feedback!

PR #1873, opened 2026-03-27
Fix implicit object with arrow function body parsed as typed const

### LINKED ISSUE #1870: Implicit object is being parsed as a TS type then failing to parse
```civet
fn: =>
  x := y
  return
```

I think this used to work not too long ago.

Also:

```civet
fn: =>
  x : y
  return
```

Causes a compiler crash.

STRd6:
Fixes #1870. Two bugs when an implicit object property has an arrow function body containing declarations:

1. `fn: => x := y` was parsed as a typed const (`const fn = y`) instead of an implicit object. `TypeSuffix` would consume `: =>` as a type annotation because `TypeFunction` allows empty parameters, making `=> T` a valid type. Fix: add a negative lookahead in the `LexicalDeclaration` shorthand rule to reject `: =>`, `: ->`, `: async =>`, `: abstract new =>`, etc. before attempting TypeSuffix. Named captures replace `$0.slice(1)` since adding the lookahead shifts positional indices.

2. `fn: => x : y` crashed with a TypeError in `insertReturn` because `NonSingleBracedBlock` can store an `ObjectExpression` node directly in `expressions` (not wrapped in a `StatementTuple` array). Fix: handle non-array elements in the `.some()` check.


source/parser.hera
```diff
@@ -6369,10 +6369,13 @@ LexicalDeclaration
   # NOTE: Added const/let shorthands
   # NOTE: Using Expression to allow for If/SwitchExpressions
   # NOTE: Using PostfixExpression to allow postfix if/unless/for
-  Loc:loc ( BindingPattern / BindingIdentifier ) TypeSuffix? __ ( ConstAssignment / LetAssignment ):assign MaybeNestedPostfixedCommaExpression ->
+  # NOTE: Forbid bare TypeFunctionArrow (=> or ->) in TypeSuffix to avoid
+  # treating `fn: => body` as typed const; prefer implicit object interpretation.
+  # Also forbid `async =>`, `abstract new =>`, etc. (TypeFunction prefixes).
```
edemaine: Unfortunately, unbare type arrows are [also an issue](<https://civet.dev/playground?code=ZjogKHg6IFQpID0%2BIHkgOj0geg%3D%3D>):

```js
f: (x: T) => y := z
```

Maybe we should add something like `!":="` at the end of TypeArrow?

source/parser/function.civet
```diff
@@ -519,7 +519,8 @@ function insertReturn(node: ASTNode): void
   switch node.type
     when "BlockStatement"
       if node.expressions#
-        return if node.expressions.some ([, exp]) => isExit exp
+        return if node.expressions.some (item) =>
+          isExit if Array.isArray(item) then item[1] else item
```
edemaine: I think it's probably bad for `expressions` to have something other than a statament tuple. I worry about it becoming an array that we then treat as a tuple incorrectly. Perhaps we should add the `["", &]` wrapper?

edemaine: I'd conjecture that [this bug](https://civet.dev/playground?code=PT4geyBmb28gfSBmcm9tIC4veC5jaXZldA%3D%3D) is a side effect of the missing wrapper.

source/parser.hera
```diff
@@ -8869,8 +8872,10 @@ TypeBinaryOp
     return { $loc, token: "&" }
 
 TypeFunction
-  ( Abstract _? )?:abstract ( Async _? )?:async ( New _? )?:new_ Parameters __ TypeFunctionArrow ( ReturnType / Loc ):returnType ->
-    const children = [ abstract, ...$0.slice(2) ] // omit async
+  # NOTE: Forbid `:=` after return type to avoid treating `f: (x: T) => y := z`
+  # as a typed const; prefer implicit object/arrow function interpretation.
+  ( Abstract _? )?:abstract ( Async _? )?:async ( New _? )?:new_ Parameters __ TypeFunctionArrow ( ReturnType / Loc ):returnType !( _? ":=" ) ->
```
edemaine: We need to handle `.=` too. Probably there's an existing rule for lexical declaration shorthands. (Is there anything else we need to forbid?)

source/parser.hera
```diff
@@ -6369,10 +6369,13 @@ LexicalDeclaration
   # NOTE: Added const/let shorthands
   # NOTE: Using Expression to allow for If/SwitchExpressions
   # NOTE: Using PostfixExpression to allow postfix if/unless/for
-  Loc:loc ( BindingPattern / BindingIdentifier ) TypeSuffix? __ ( ConstAssignment / LetAssignment ):assign MaybeNestedPostfixedCommaExpression ->
+  # NOTE: Forbid bare TypeFunctionArrow (=> or ->) in TypeSuffix to avoid
+  # treating `fn: => body` as typed const; prefer implicit object interpretation.
+  # Also forbid `async =>`, `abstract new =>`, etc. (TypeFunction prefixes).
+  Loc:loc ( BindingPattern / BindingIdentifier ):pattern !( _? Colon _? ( Abstract _? )? ( Async _? )? ( New _? )? TypeFunctionArrow ) TypeSuffix?:typeSuffix NotDedented:ws ( ConstAssignment / LetAssignment ):assign MaybeNestedPostfixedCommaExpression:e ->
```
edemaine: My hope is that we no longer need these changes given the added negative lookahead.

PR #1876, opened 2026-03-30
Using Civet in parser.hera

STRd6:
Converts the code inside parser.hera to use Civet syntax.
                                                                                                                                                                                                 
  - parser.hera: Kept as mainly JS format for now but fixed a few edge cases where it was incompatible with Civet syntax                                                                                                                                
  - build/esbuild.civet: Configures the Hera esbuild plugin with language: "civet" to compile .hera files' embedded code as Civet                                                                
  - .mocharc.json / .mocharc-self.json: Switches from @danielx/hera/register to @danielx/hera/register/civet and drops the coffee extension
  - package.json: Bumps @danielx/hera to 0.8.19-pre.0 which adds Civet language support
  - source/main.civet: Fixes a bug where nested parse() calls with different input would pollute the top-level parse cache — adds initialInput tracking and skips cache reads/writes when the input doesn't match
  
Build and test are a little bit slower now since there is more Civet parsing in the largest part of the source. We can likely find some speedups as a followup in both the parser itself and in what work we do when building and testing.

See https://github.com/DanielXMoore/Hera/pull/51

### DIFF
--- a/package.json
+++ b/package.json
@@ -127,7 +127,7 @@
     "prettier": "^3.2.5",
     "shiki": "^3.0.0",
     "terser": "^5.16.1",
-"tslib": "^2.4.0",
+    "tslib": "^2.4.0",
     "typescript": "^5.5.2",
     "vite": "^4.4.12",
     "vitepress": "^1.0.0-alpha.35",
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -102,4 +102,5 @@
     "source",
     "test/**/*.civet",
     "types",
-  ]}
+  ]
+}


build/cache-utils.js
```diff
@@ -0,0 +1,66 @@
...
+const fs = require('fs');
+const path = require('path');
+
+const { compile: heraCompile } = require('@danielx/hera/dist/main.js');
+const heraVersion = require('@danielx/hera/package.json').version;
+
+const civetSourceRaw = process.env.CIVET_SOURCE ?? './node_modules/@danielx/civet';
+const civetSourceResolved = civetSourceRaw.startsWith('.')
+  ? require.resolve(path.resolve(process.cwd(), civetSourceRaw))
+  : require.resolve(civetSourceRaw);
```
edemaine: I'm pretty sure `require.resolve(path.resolve(civetSourceRaw))` would suffice in both cases, given [path.resolve](https://nodejs.org/api/path.html#pathresolvepaths)'s behavior.

build/cache-utils.js
```diff
@@ -0,0 +1,66 @@
...
+
+function findPackageVersion(resolvedPath) {
+  let dir = path.dirname(resolvedPath);
+  while (true) {
+    try { return require(path.join(dir, 'package.json')).version; } catch {}
+    const parent = path.dirname(dir);
+    if (parent === dir) return 'unknown';
+    dir = parent;
+  }
+}
```
edemaine: Why would `package.json` not be at the top level?

build/esm-hook.mjs
```diff
@@ -0,0 +1,43 @@
+import { readFileSync } from 'node:fs';
+import { createRequire } from 'node:module';
+import { fileURLToPath, pathToFileURL } from 'node:url';
+
+const _require = createRequire(import.meta.url);
+const { getCachePath, readCache, writeCache, civetCompile, heraCompile } = _require('./cache-utils.js');
```
edemaine: Why not `import { getCachePath, readCache, writeCache, civetCompile, heraCompile } from './cache-utils.js'`? Ah, because `cache-utils` is CJS instead of ESM. Perhaps worth rewriting to ESM and `cache-utils.mjs`? Ignore this comment if `require.js` really needs to override CJS require in addition to ESM register.

build/register.js
```diff
@@ -0,0 +1,46 @@
...
+  require.extensions['.hera'] = function(module, filename) {
+    const source = fs.readFileSync(filename, 'utf8');
+    module._compile(compileWithCache('hera-cjs', source, filename), filename);
+  };
+
+  require.extensions['.civet'] = function(module, filename) {
+    const source = fs.readFileSync(filename, 'utf8');
+    module._compile(compileWithCache('civet', source, filename), filename);
+  };
+}
```
edemaine, line 16: Do we need CJS support in the build step? I would have thought ESM sufficed.

If we do need it, it'd be nice for `compileWithCache` to be factored out and used in `esm-hook.mjs`. Perhaps it could be part of `cache-utils.js`?

docs/superpowers/plans/2026-04-01-unified-build-loader.md
```diff
@@ -0,0 +1,363 @@
+# Unified Build Loader Implementation Plan
```
edemaine: Did you mean to include this and the associated spec? It's an interesting peek behind the curtains (I personally rarely use planning and just go to straight development, but I also use GPT-5.4 which needs plans less), but it's repetitive and verbose.

source/esm.civet
```diff
@@ -15,13 +15,6 @@ In older Node, use `--loader` to install this ESM loader:
 node --loader @danielx/civet/esm source.civet
 ```
 
-Previously depended on ts-node esm loader being downstream:
-
-@example
-```bash
-node --loader ts-node/esm --loader @danielx/civet/esm source.civet
-```
-
```
edemaine, line 24: Should we remove the ts-node tests too? We could then remove ts-node as a dev dependency.

source/esm.civet
```diff
@@ -82,45 +74,25 @@ export async function load(url: string, context: any, next: any)
...
       throw e
 
-    // NOTE: Append .tsx to URL (before query) so ts-node treats as TypeScript
-    transpiledUrl :=
-      new URL url
-      ||> .pathname += ".tsx"
-      |> .href
-
-    // NOTE: Assuming ts-node hook follows load hook
-    result := await next transpiledUrl,
```
edemaine, line 108: Are we sure we want to avoid the `next` pipeline here? What if there was a legit use of another stage, e.g., JSX processing? (I don't know whether such an ESM override exists, but it seems plausible...)

source/main.civet
```diff
@@ -292,6 +295,9 @@ makeCache := ({hits, trace}: CacheOptions = {}): CacheEvents ->
 
       return if uncacheable.has(ruleName)
 
+      // Skip cache for nested parses with different input
+      return unless state.input is initialInput
```
edemaine: I like the idea here, but this is expensive: I believe it will compare the entire string, not the identity/pointer of the string. Doing this for every rule will likely slow things down.

A speedier alternative would be for everyone (`state` and cache) to keep a counter for the input, and just compare that.

But perhaps simpler would be for `state` to store the cache? And nested parses push the `state` onto a stack and pop after parses? Then every parse would get to use their own cache.

Related, I'm worried about nested parses not using a cache; that could cause them to go really slow. If this is about the nested civet option parsing, I thought that already didn't use the cache somehow? Actually, as far as I can see, every `parseProgram` makes its own cache? What's the problem here?

package.json
```diff
@@ -103,7 +103,7 @@
     "prepublishOnly": "pnpm build && pnpm test:coverage && pnpm changelog --release",
     "test": "bash ./build/test.sh",
     "test:coverage": "CIVET_COVERAGE=1 bash ./build/test.sh",
-    "test:self": "pnpm build && mocha --config .mocharc-self.json",
+    "test:self": "pnpm build && CIVET_SOURCE=./dist/main.js pnpm test",
```
edemaine: Nice, much simpler!

source/parser.hera
```diff
@@ -7376,6 +7376,7 @@ PushJSXOpeningElement
 PopJSXStack
   "" ->
     state.JSXTagStack.pop()
+    return
```
edemaine, line 7379: This makes me worry about various other implicit returns... at least all the `Pop` rules.

source/parser.hera
```diff
@@ -9385,6 +9388,7 @@ Reset
         ]) {
           config[option] = b
         }
+        return
```
edemaine, line 9391: Personally I prefer `: void` annotation over the explicit `return`s that are all over this PR, but that's a style thing. (Mentioning in case it was AI's preference, and your preference matches mine.)

source/esm.civet
STRd6: I think longer term we would put esbuild in here if we wanted to boost the TS compatibility of the esm loader.

source/parser.hera
STRd6: I think the `: void` annotation is what we'll use when rewriting the whole file in Civet style. This was just a minimal fix to prove it out

build/esm-hook.mjs
```diff
@@ -25,19 +24,6 @@ export async function load(url, context, next) {
...
-  let js;
-  if (format === 'hera') {
-    const civetOutput = heraCompile(source, { filename, module: true });
-    js = await civetCompile(civetOutput, { filename, js: true, inlineMap: true });
-  } else {
-    js = await civetCompile(source, { filename, js: true, inlineMap: true });
-  }
-
-  writeCache(p, js);
+  const js = await compileWithCache(source, filename);
```
edemaine, line 27: `await` seems unnecessary, though I'm not against it.

tsconfig.json
```diff
@@ -102,11 +102,4 @@
...
     "types",
-  ],
-  "ts-node": {
-    "transpileOnly": true,
-    "compilerOptions": {
-      "module": "ESNext"
-    }
-  }
-}
+  ]}
```
edemaine: Prefer `}` on its own line

source/esm.civet
edemaine: I don't think esbuild covers all use cases. For example, the Solid JSX transform isn't supported. Not that I know of anyone using Solid without a build step, but I feel like shortcutting discards some possibilities that we ideally shouldn't...

PR #1934, opened 2026-04-08
Fix ternary with trailing ? and indented values

### LINKED ISSUE #1933: Ternary with trailing symbols and indented values fails to parse
```civet
typeof token is "string" ?
  token :
  undefined
```

This throws a parse error when it should parse as a ternary.

STRd6:
Fixes #1933

Add a new `TernaryRest` alternative to handle the form:
```civet
condition ?
  then :
  else
```

Previously, `MaybeNestedExpression` at the `then` position would incorrectly parse `then :\n  else` as an implicit object literal `{then: else}`, consuming the ternary colon. The fix uses `PushIndent` to enter the indented context first, then parses `Expression _ Colon` followed by another `Nested Expression` for the else branch.

### DIFF
--- a/source/parser.hera
+++ b/source/parser.hera
@@ -975,7 +975,6 @@ ConditionalExpression
 
 TernaryRest
   NestedTernaryRest
-  # NOTE: Ternary `a ? b : c` is disabled if CoffeeScript binary existential `a ? b` is enabled
   # Trailing ?: condition ?
   #   then :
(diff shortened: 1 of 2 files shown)


STRd6:
@claude address greptile comments

STRd6:
@claude keep the comment that was removed and remove the other one instead

PR #1937, opened 2026-04-11
LSP E2E test coverage

STRd6:
Merge LSP E2E test coverage with unit test coverage.

PR #1939, opened 2026-04-12
LSP fixes for path casing (e.g. Windows) and ts documents

edemaine:
~~This is currently built on top of #1937 for testing purposes. Should wait for #1937 to merge first.~~

- Fixed `lsp` test failures on Windows by canonicalizing TypeScript host paths consistently before using them as document, snapshot, and metadata keys. This should fix any non-case-sensitive system.
- Updated the host to replace stored TypeScript / non-transpiled document entries and refresh `transpiledDoc` metadata on recompiles, preventing stale snapshots and metadata from being reused.
- Verified the fix by running #4309's `pnpm test` in `lsp`, with all tests passing on my Windows 11 machine.


STRd6 (approved):
I squash merged so you may need to rebase / cherry-pick.

Thanks for the fixes!

PR #1940, opened 2026-04-12
CI tests on Windows too

edemaine:
~~This is currently built on top of #1937 for testing purposes. Should wait for #1937 to merge first.~~ (now merged)

~~The LSP tests are also expected to fail on Windows until #1939 gets merged.~~ (now merged)

* Run CI tests on Windows in addition to Linux, to prevent future regressions
* Force LF line endings so tests work correctly
* Increase timeout for slow Windows
* Fix path handling in merge-coverage script

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -46,9 +46,9 @@ jobs:
         id: cache
         with:
           path: .cache
-          key: ${{ runner.os }}-${{ github.ref_name }}-${{ steps.month.outputs.ym }}
+          key: ${{ github.ref_name }}-${{ steps.month.outputs.ym }}
           restore-keys: |
-            ${{ runner.os }}-main-${{ steps.month.outputs.ym }}
+            main-${{ steps.month.outputs.ym }}
 
       - name: Cache VS Code download
         uses: actions/cache@v4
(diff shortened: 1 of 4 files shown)


.github/workflows/build.yml
```diff
@@ -35,9 +46,9 @@ jobs:
         id: cache
         with:
           path: .cache
-          key: ${{ github.ref_name }}-${{ steps.month.outputs.ym }}
+          key: ${{ runner.os }}-${{ github.ref_name }}-${{ steps.month.outputs.ym }}
```
STRd6: It should be fine to exclude `runner.os` from this cache since it is only compiled civet files which should be os independent.

.github/workflows/build.yml
```diff
@@ -64,21 +75,34 @@ jobs:
       - uses: actions/cache/save@v4
         with:
           path: .cache
-          key: ${{ github.ref_name }}-${{ steps.month.outputs.ym }}
+          key: ${{ runner.os }}-${{ github.ref_name }}-${{ steps.month.outputs.ym }}
```
STRd6: Same here

STRd6 (commented):
Happy to see Windows no longer neglected 👏

PR #1947, opened 2026-04-12
Dev: Fix eslint typecheck

STRd6:
Found when running `pnpm -r test` at root this typecheck was broken.

Added tsconfig and some expect-errors for dynamic imports.


edemaine:
@STRd6 should we put the `-r` in CI?

STRd6:
> @STRd6 should we put the `-r` in CI?

Probably or be more explicit about running all the tests

PR #1950, opened 2026-04-13
Extract LSP server into standalone @danielx/civet-lsp-server package

### LINKED ISSUE #161: Separate LSP from VS Code implementation
Relates to #61 but another good reason (outside of support for other editors) is that you can have a playground interact with the LSP on the web.

STRd6:
Extracts the LSP server from the VS Code extension into a standalone npm package `@danielx/civet-lsp-server`, enabling other editors (Neovim, Helix, Emacs, etc.) to use it independently.

Closes #161 (Step 1)


STRd6:
@claude rename the old lsp to lsp-vscode

STRd6:
@claude update the lockfile and the gh workflows

edemaine:
> rename the old lsp to lsp-vscode

Would `vscode` be better? Simpler and maybe more accurate (it's an LSP and an extension).
On the other hand, I could see wanting to cluster under `lsp-*`.

edemaine:
I'm getting this warning during fresh `pnpm install`:

```
WARN  Failed to create bin at C:\Users\edemaine\Projects\Civet\lsp-vscode\node_modules\.bin\civet-lsp. ENOENT: no such file or directory, stat 'C:\Users\edemaine\Projects\Civet\lsp-server\dist\server.js.EXE'
```

I'm not sure the right way to fix this: it naturally wants the lsp-server to be built before it can link a binary under lsp-vscode... Does this mean we need to `pnpm install && pnpm build && pnpm install` from a fresh repo?

edemaine:
> I'm not sure the right way to fix this: it naturally wants the lsp-server to be built before it can link a binary under lsp-vscode... Does this mean we need to `pnpm install && pnpm build && pnpm install` from a fresh repo?

GPT 5.4 says the right way to fix this is to make lsp-server's `bin` point to a static wrapper script instead of something in `dist`. I've added `bin/civet-lsp.js` accordingly. Feel free to edit!

lsp-vscode/build/build.sh
```diff
@@ -0,0 +1,28 @@
...
+# lsp-server/source/ correctly (relative paths would resolve to lsp-vscode/source/).
+node -e "
+  const fs = require('fs'), path = require('path');
+  const src = '../lsp-server/dist/server.js.map';
+  if (!fs.existsSync(src)) process.exit(0);
+  const map = JSON.parse(fs.readFileSync(src, 'utf8'));
+  const base = path.resolve('../lsp-server/dist');
+  map.sources = map.sources.map(s => path.resolve(base, s));
+  fs.writeFileSync('dist/server.js.map', JSON.stringify(map));
+"
```
edemaine: Didn't we move this into `build.civet`? Ah, I see, timing is more awkward now.

Are these sourcemaps distributed as part of the extension? Would be awkward to include absolute paths there.

vscode/package.json
```diff
@@ -83,51 +83,24 @@
...
-      ".civet",
-      ".ts"
-    ],
-    "include": [
-      "source"
-    ],
-    "exclude": [
-      "source/extension.civet"
-    ]
+    "test:e2e:coverage": "pnpm build-dev && pnpm -C ../lsp-server build-dev && rm -rf coverage/e2e-tmp && NODE_V8_COVERAGE=coverage/e2e-tmp node e2e/dist/runTest"
```
edemaine, line 89: Now that former unit tests are in `lsp-server`, should we rename `e2e` to regular `test` here? (including moving the directory)

STRd6: I could imagine having some acutal unit tests if the VSCode specific stuff in the extension grows at some point in the future.

PR #1951, opened 2026-04-13
Pin properties and pin patterns in arrays use `typeof` typing

edemaine:
Followup to #1942 to handle the general case, not just top-level pins. Also handles mixed cases with `::` like `([^foo, bar:: Type]) =>`.


STRd6 (approved):
Thanks!

PR #1964, opened 2026-04-16
v0.11.7

STRd6:


edemaine (approved):
IMO these release bump commits could be just pushed to main, without PR

STRd6:
Eventually I'd like the option for a protected main and maybe an automated PR that can be merged to release

edemaine:
Fair enough. One issue with release PRs is that they will appear in the changelog, possibly in the next release which may be weird. But we can probably tweak the changelog script...

PR #1965, opened 2026-04-16
LSP: auto-reload TSService when tsconfig.json changes

### LINKED ISSUE #1962: LSP: auto-reload TSService when tsconfig.json changes
## Problem

The Civet LSP reads `tsconfig.json` **once** at `TSService` construction (`lsp/server/source/lib/typescript-service.civet:391`) and never re-parses it. The file-change handler at `lsp/server/source/server.civet:1088` just logs:

```civet
connection.onDidChangeWatchedFiles (_change) =>
  logger.log('We received a file change event')
```

So any edit to `tsconfig.json` (adding `types`, toggling `strict`, changing `include`, etc.) requires a manual LSP restart. Observed while fixing the "Cannot find name 'describe'" error — added `"types": ["mocha", "node"]` to the root tsconfig but had to restart the server for the editor to see `describe`.

## Goal

When a `tsconfig.json` belonging to a known project root changes (or is created/deleted), rebuild *just that project's* `TSService` and update subscribers, without a full LSP restart.

## Performance considerations

Rebuilding a TS language service is not cheap — it rescans `node_modules/@types`, resolves all program files, etc. Naive handling can cause:

- **Thrashing on rapid saves.** Debounce per project root (200–500ms). Editors often emit multiple events per save.
- **Blocking user-visible requests.** Keep the old `TSService` answering hover/diagnostics/etc. until the new one is ready, then atomically swap. Double-buffer the service; don't `.dispose()` the old one until all in-flight requests finish (refcount or simple grace period).
- **Wasted rebuilds on no-op writes.** Hash or mtime-compare the parsed tsconfig before rebuilding — editor "save" on an unchanged file shouldn't trigger work.
- **Concurrent rebuilds of the same project.** Coalesce: if a rebuild is already pending/in-progress, fold newer changes into it rather than queueing a second one.
- **Cross-project fan-out.** Only rebuild the project whose tsconfig changed. The project lookup map (`lsp/server/source/server.civet:84` — "abs file path → nearest project path") already scopes this; don't invalidate sibling projects.
- **Open document re-registration.** Documents open in the editor at rebuild time must continue to receive diagnostics. Either replay `didOpen` against the new service, or have `ensureServiceForSourcePath` lazily attach open docs on next access.
- **Watch registration.** Today the client decides what to watch. Register dynamic `workspace/didChangeWatchedFiles` for `**/tsconfig.json` (and arguably `**/package.json`, since installing/removing `@types/*` changes effective types — though that's probably out of scope for v1).

## Out of scope

- Watching `package.json` / `node_modules/@types/*` changes (separate follow-up).
- Reloading the LSP server's own config (initialization options).
- Multi-root workspace edge cases beyond what the existing project lookup already handles.

## Relevant code

- `lsp/server/source/server.civet:1088` — file-change handler stub.
- `lsp/server/source/server.civet:84` — source-path → project-root lookup.
- `lsp/server/source/server.civet:137` — `TSService(projPath, ...)` construction site.
- `lsp/server/source/lib/typescript-service.civet:386-425` — `TSService` factory, loads tsconfig + builds host/service.

## Acceptance

- Editing `tsconfig.json` in a project's root causes the LSP to pick up the new options (verifiable: add/remove `"types": ["mocha"]` in a test project and see `describe` resolve/unresolve within ~1s of save) without restart.
- Typing in an open `.civet` file during the rebuild continues to return hover/diagnostics from the old service (no visible latency spike or "request dropped" errors).
- Saving `tsconfig.json` multiple times in quick succession triggers at most one rebuild per project.
- Saving an unchanged `tsconfig.json` (no-op write) triggers zero rebuilds.

@claude

STRd6:
Implements auto-reload of TSService when tsconfig.json changes, per #1962.

## Changes

- 300ms debounce per project root to absorb rapid saves
- No-op detection: skips rebuild if tsconfig content is unchanged
- Double-buffering: old service stays active until new one is ready, then atomic swap
- Coalesced concurrent rebuilds: at most one extra rebuild queued per project
- Grace-period dispose of old service (1s) for in-flight requests
- Diagnostics re-triggered for open documents after service swap

Closes #1962


STRd6:
@claude can we add a simple e2e test that covers this code?

Use lsp/vscode/scripts/show-uncovered to see if we can get all this new code covered.

STRd6:
@claude run pnpm test:coverage then use the lsp/vscode/scripts/show-uncovered script. Address greptile feedback

PR #1987, opened 2026-04-19
Reforbid comma in `:=` shorthand

edemaine:
#1849 accidentally added `x := 1, y = 2` as shorthand for `const x = 1, y = 2`, but actually parsed as `const x = (1, y = 2)` which is not how it executes in JavaScript. This is at the least buggy: Civet doesn't understand that `y` gets declared. I think it's also safer to go back to forbidding this, as the comma could have different meanings.

This PR reforbids `x := 1, y = 2` while preserving `x := a, b for ...` notation from #1849. It also allows `a, b for ...` in a few more locations, as described in the tests.


STRd6 (approved):
Thanks!

PR #1989, opened 2026-04-19
Preserve quotes in method signature names

### LINKED ISSUE #1988: Interface/class method with hyphenated quoted name produces invalid TypeScript
@claude

## Summary

When an interface method signature (or class method) has a quoted string-literal name containing characters that aren't valid in an identifier (e.g. hyphens), Civet strips the quotes in the output, producing invalid TypeScript.

## Reproduction

Input (`.civet`):

```civet
interface Thing
  "do-work"(x: string): string
```

Actual output:

```ts
interface Thing {
  do-work(x: string): string
}
```

Expected output (quotes preserved):

```ts
interface Thing {
  "do-work"(x: string): string
}
```

`tsc` rejects the actual output:

```
error TS1131: Property or signature expected.
error TS1005: ',' expected.
error TS1005: ';' expected.
error TS1128: Declaration or statement expected.
```

The quoted form is valid TypeScript, so Civet should preserve the quotes when the inner content isn't a valid identifier.

## Root cause

`source/parser.hera:4160` in the `MethodSignature` rule strips the surrounding quotes unconditionally whenever the token starts with `"` or `'`:

```civet
name = name.token.match(/^(?:"|')/) ? name.token.slice(1, -1) : name.token
```

It should only strip when the unquoted result is still a valid identifier; otherwise keep the original quoted token.

## Workaround

None other than using a non-hyphenated method name.

STRd6:
Fixes #1988

When a method signature (interface or class) used a quoted string-literal name that wasn't a valid identifier (e.g. "do-work"), Civet unconditionally stripped the quotes, producing invalid TypeScript. Now the quotes are stripped only when the inner content is a valid identifier.

### DIFF
--- a/source/parser.hera
+++ b/source/parser.hera
@@ -4157,14 +4157,10 @@ MethodSignature
     if name.name
       name = name.name
     else if name.token
-      // Strip quotes only when the inner content is a valid identifier,
-      // so that hyphenated (or otherwise non-identifier) quoted names
-      // like "do-work" remain quoted in the output (valid TypeScript).
-      if name.token.match(/^(?:"|')/)
-        unquoted := name.token.slice(1, -1)
-        name = unquoted.match(/^[$_\p{ID_Start}][$_\u200C\u200D\p{ID_Continue}]*$/u) ? unquoted : name.token
-      else
-        name = name.token
+      // Preserve quotes as authored in source so that quoted names
+      // (including non-identifier names like "do-work") remain valid
+      // in the output.
+      name = name.token
 
     // TypeScript supports optional methods with bodies; remove ? from JS output
     if optional then optional = { ...optional, ts: true }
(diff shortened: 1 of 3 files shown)


STRd6:
@claude instead keep quotes in source as authored

source/parser.hera
```diff
@@ -4157,7 +4157,10 @@ MethodSignature
     if name.name
       name = name.name
     else if name.token
-      name = name.token.match(/^(?:"|')/) ? name.token.slice(1, -1) : name.token
+      // Preserve quotes as authored in source so that quoted names
+      // (including non-identifier names like "do-work") remain valid
+      // in the output.
+      name = name.token
```
edemaine: I feel like this could be problematic for things like autoVar, or maybe overload detection, where some uses are quoted and others are not. Perhaps we should unwrap quotes for the `name` property, but include the full leaf node in the AST.

edemaine: I confirmed that this would have been a bug, adding a test that fails. And then I fixed the bug. Should be good now!

edemaine (approved):

PR #1995, opened 2026-04-21
Add `pnpm typecheck` for `.civet` + `.hera` files; refactor `types.civet` onto shared bases

STRd6:
## Summary

- **`pnpm typecheck`**: new script (`scripts/typecheck.civet`) that feeds every `.civet` and `.hera` source through a TS `Program`, remapping diagnostics back to the original files via their civet sourcemap.  Wired as the new `tsc --noEmit` step in `build/test.sh` (the old one only saw one `.mjs` file).
- **Full-mode hera typecheck unlocked.**  Root cause: `@danielx/hera` ships `dist/machine.ts` (the 708-line runtime) alongside `machine.js` with no `.d.ts`, so TS's resolver finds the `.ts` first when following the `"./lib": "./dist/machine.js"` exports entry and pulls the whole recursive-generics runtime into the program — overflowing the analyzer's stack on parser.hera-sized consumers.  We stand a synthetic `machine.d.ts` in front of the real file via our patched `system.fileExists` / `readFile`.  Upstream fix filed: [DanielXMoore/Hera#68](https://github.com/DanielXMoore/Hera/issues/68).
- **`parser.hera` preamble typed**; expression-chain hubs annotated with `::ASTNode`; `ParserState` interface covers both the up-front arrays and the `Object.defineProperties`-installed getters.
- **`types.civet` refactor**: introduces `AstObject` (parent-only) and `AstParent = AstObject & { children }` bases; ~116 AST node types migrate onto them, dropping ~180 lines of repeated `parent?: Parent` / `children: Children` declarations.
- **Grammar-truth fixes** in `types.civet`: `ASTRef.names/children` (populateRefs), `Star.parent`, `BindingElement.parent + .initializer`, `WithClause.parent`, `ArrowFunction.generator`, `MethodDefinition/SpreadProperty.delim`, `ASTError.{delim,name,names,typeSuffix,value,initializer,binding,ref}` (undefined declarations so walker unions destructure uniformly), `BindingRestElement/ElisionElement.initializer?: undefined`, `AtBindingProperty/BindingRestProperty.value?: undefined`.
- **Walker narrowings**: `util.civet` walkers (`addParentPointers`, `removeParentPointers`, `startsWith`, `startsWithPredicate`) use the existing `isParent` / `isToken` type guards instead of ad-hoc `'field' in node` probes.  `generate.civet` gets similar guards.  `negateCondition` gets a `Condition` parameter type.

## Impact

- Typecheck diagnostics went from effectively invisible (old `tsc --noEmit` checked one file) to **687 reported problems**, all visible and actionable.  Nothing suppressed that a real type contract would cover.
- No public-type surface change: ASTNode / ASTNodeObject / StatementNode / ExpressionNode / OtherNode all unchanged externally.
- `types.civet` is ~180 lines shorter and easier to extend (add a new AST node type ⇒ `export type X = AstParent & { type: "X"; ...fields }`).

## Test plan

- [x] `pnpm build` clean
- [x] `pnpm test` — 3664 passing / 0 failing
- [x] `pnpm typecheck` — reports 687 problems (previously effectively 0 reports because no files were actually checked); this PR's goal is that the tool exists and is wired in, not that the count is 0.  Follow-up PRs can drive the baseline down.
- [x] `pnpm build:self` path exercised via `bash build/build.sh` with the rebuilt parser.hera
- [ ] Decide CI posture for the `pnpm typecheck` step:  right now it fails test.sh when errors > 0.  Option is to gate at `errors <= 687` (no-regression) until the baseline is worked down.


edemaine:
It'd be nice to use `civet --typecheck` for this, but it doesn't support `.hera` files. I was wondering if maybe it could support ESM loaders...

STRd6:
> It'd be nice to use `civet --typecheck` for this, but it doesn't support `.hera` files. I was wondering if maybe it could support ESM loaders...

Eventually I'd like to get an integrated TypeScript VFS wrapper that can be shared across CLI, LSP, and Unplugin. In the meantime I'll just keep copying new ones :upside_down_face: 

Once we've got all the requirements and tests we should be able to unify them hopefully.

source/parser/types.civet
```diff
@@ -145,13 +145,26 @@ export type Parent = (ASTNodeObject & IsParent)?
 export type Children = ASTNode[] & (type?: undefined) & (token?: undefined) & (children?: undefined)
 export type ASTString = string & (type?: undefined) & (token?: undefined) & (children?: undefined)
 
+// Shared base for object-shaped AST nodes: every node in the tree may carry a
+// parent pointer installed by `addParentPointers`.  Concrete node types mix
+// this in via intersection (`= AstObject &`) instead of repeating
+// `parent?: Parent` by hand on each of the ~120 object types.
+export type AstObject =
```
edemaine: Other types like `ASTNode` use all-caps `AST`. Perhaps `AstObject` and `AstParent` should do the same?

source/parser/types.civet
```diff
@@ -267,65 +283,48 @@ export type AssignmentExpression
...
+export type UnaryExpression = AstParent &
   type: "UnaryExpression"
-  children: Children
-  parent?: Parent
   pre?: ASTNode[]
   expression: ASTNode
   post?: ASTNode
 
-export type Await
+export type Await = AstObject &
```
edemaine: Would it be worth defining `AstTokenOrParent` with `{ token?: string, children?: Children }`? This is a common pattern currently.

source/parser/types.civet
```diff
@@ -1032,34 +898,34 @@ export type PipelineExpression =
...
   type: "ArrowFunction"
-  children: Children
-  parent?: Parent
   name: string
   async: ASTNode[]
+  // Arrow functions can't grammatically be generators, but the shared
+  // FunctionNode machinery reads `.generator` uniformly (and errors when
+  // yield is found inside an arrow function).  Keeping the field optional
+  // here matches FunctionExpression / MethodDefinition.
+  generator?: ASTNode[]
```
edemaine: Would `generator?: never` capture this? I'm not sure.

source/generate.civet
```diff
@@ -75,9 +75,9 @@ function gen(root: ASTNode, options: Options): string
         //console.log 'set', node, options.sourceMap.srcLine, options.sourceMap.srcColumn if options?.sourceMap?
         return token
 
-      unless node.children
+      unless ("children" in node) and node.children
```
edemaine:
```suggestion
      unless isParent node
```

source/generate.civet
```diff
@@ -75,9 +75,9 @@ function gen(root: ASTNode, options: Options): string
...
         // Occasionally we create ASTLeaf nodes with no $loc field
-        if node.token?
+        if "token" in node and node.token?
```
edemaine:
```suggestion
        if isToken node
```

source/parser.hera
```diff
@@ -79,14 +79,52 @@ import {
...
+} from "./parser/types.civet"
 
 /**
  * Globals
  */
-let filename              // filename currently being parsed
-let initialConfig         // input for parser config
-let config                // current parser config after directives
-let sync                  // synchronous mode: as much as possible without await
+let filename: string | null | undefined        // filename currently being parsed
```
```suggestion
let filename: string??        // filename currently being parsed
```
Just a reminder that we have this shorthand.

source/parser.hera
```diff
@@ -98,7 +136,7 @@ export const state = {    // parser state
   forbidImplicitFragment: [false],
   forbidPipeline: [false],
   JSXTagStack: [undefined],
-}
+} as unknown as ParserState
```
```suggestion
} as! ParserState
```

source/parser.hera
```diff
@@ -1075,24 +1113,37 @@ PipelineTailItem
 
 # https://262.ecma-international.org/#prod-PrimaryExpression
 PrimaryExpression
-  ObjectLiteral
+  ObjectLiteral ::ASTNode ->
+    return $1
```
edemaine, line 1096: Is this a Hera limitation that we need `-> return $1` when we have a return type? Or can the `-> return $1` be omitted?

build/cache-utils.js
```diff
@@ -78,4 +79,88 @@ function compileWithCache(source, filename, module = true) {
...
+      filename, js: false, sourceMap: true, sync: true,
+      parseOptions: { rewriteCivetImports: '.civet.tsx' },
+      upstreamSourceMap: heraResult.sourceMap,
+    });
+  } else {
+    result = civetCompile(source, {
+      filename, js: false, sourceMap: true, sync: true,
+      parseOptions: { rewriteCivetImports: '.civet.tsx' },
+    });
+  }
```
edemaine, line 111: Some code duplication with `compileWithCache`, though I'm not sure it's important.

scripts/typecheck.civet
```diff
@@ -0,0 +1,285 @@
...
+// handlers; both compile modes have legitimate reasons to silence certain
+// diagnostics that don't reflect user-actionable bugs:
+//   6133 / 6196 / 6192 — "declared but never read" for preamble-only mode
+//     (imports used inside skipped rule handlers look unused).
+//   7006 / 7031 — noImplicitAny on positional $1..$9 args and destructuring
+//     patterns inside hera-generated handlers; the grammar author didn't
+//     write those and can't annotate them without editing hera.
+heraSuppressedCodes := new Set<number> [6133, 6196, 6192, 7006, 7031]
+
+function remap(d: Diagnostic): Diagnostic?
```
edemaine, line 218: Definitely a bunch of overlap with `unplugin.civet`. Maybe we could add more shared helpers to `ts-diagnostics.civet`... Also worth checking whether we're doing things here that should be copied over to `unplugin.civet`. Ultimately it'd be great if the unplugin and CLI supported `.hera` typechecking too, somehow, but I'm not quite sure how. Maybe a `customCompilers` option? Adding a `suppressedCodes` option to the unplugin would also be reasonable. (This can also be a target for a future PR.)

PR #2006, opened 2026-04-24
Upgrade @danielx/hera 0.8.20 → 0.9.0

STRd6:
## Summary

- Hera 0.9.0 changes successful `&`/`!` assertions to return `true` instead of `undefined`. `true` leaks into `$0` child arrays and into positional destructuring of grammar rules that used assertions as slot fillers; filter/normalize it at the affected sites.
- Fix two Hera 0.9.0 regressions surfaced by our grammar: `-> # trailing comment` is now swallowed as a handler body, and delimiter/assertion-only rules emit `true` values into AST tuples. Both handled in `source/parser.hera`.
- Bump typecheck baseline 759 → 763 for residual diagnostics on the upgraded generated parser.

### Where `true` is handled
- `recurse` (generate.civet) and the three `gather*` walkers (traversal.civet) skip boolean nodes.
- `processForInOf` (for.civet) and the `JSXAttribute` grammar handler normalize the assertion slot back to `undefined`.
- `StatementDelimiter` / `SimpleStatementDelimiter` / `ClosingDelimiter` get explicit `-> return undefined` handlers so `insertSemicolon` still sees a falsy delim on the assertion path.

## Test plan

- [x] `pnpm build` — clean
- [x] `pnpm test` — 3821 passing, 0 failing, 19 pending
- [x] Typecheck within updated baseline (763 diagnostics, compile failures: 0)

### DIFF
--- a/source/generate.civet
+++ b/source/generate.civet
@@ -36,6 +36,10 @@ function gen(root: ASTNode, options: Options): string
     if Array.isArray(node)
       return node.map(recurse).join('')
 
+    // Hera 0.9.0 emits `true` for successful `&`/`!` assertions; skip them.
+    if node <? "boolean"
+      return ""
+
     if node <? "object"
       if options.js and ("ts" in node) and node.ts
         return ""
--- a/source/generate.civet
+++ b/source/generate.civet
@@ -36,10 +36,6 @@ function gen(root: ASTNode, options: Options): string
     if Array.isArray(node)
       return node.map(recurse).join('')
 
-    // Hera 0.9.0 emits `true` for successful `&`/`!` assertions; skip them.
-    if node <? "boolean"
-      return ""
-
     if node <? "object"
       if options.js and ("ts" in node) and node.ts
         return ""
--- a/source/parser.hera
+++ b/source/parser.hera
@@ -1105,9 +1105,11 @@ PrimaryExpression
   # interpolated strings get checked first before StringLiteral.
   Literal ::ASTNode ->
     return $1
-  FunctionExpression ::ASTNode -> # NOTE: Must be before IdentiferExpression so `async function ...` isn't parsed as `async(function ...)`
+  # NOTE: Must be before IdentiferExpression so `async function ...` isn't parsed as `async(function ...)`
+  FunctionExpression ::ASTNode ->
     return $1
-  IdentifierReference ::ASTNode -> # NOTE: Must be below ObjectLiteral for inline objects `a: 1, b: 2` to not be shadowed by matching the first identifier
+  # NOTE: Must be below ObjectLiteral for inline objects `a: 1, b: 2` to not be shadowed by matching the first identifier
+  IdentifierReference ::ASTNode ->
     return $1
   ClassExpression ::ASTNode ->
     return $1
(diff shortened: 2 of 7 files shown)


source/parser/for.civet
edemaine: Type should also be updated to include `boolean`. This applies to all modified functions in this PR I think.

source/generate.civet
```diff
@@ -36,6 +36,10 @@ function gen(root: ASTNode, options: Options): string
     if Array.isArray(node)
       return node.map(recurse).join('')
 
+    // Hera 0.9.0 emits `true` for successful `&`/`!` assertions; skip them.
+    if node <? "boolean"
+      return ""
```
edemaine: If this is necessary, it's a sign that we didn't finish eradicating them when parsing.

source/parser.hera
```diff
@@ -6844,17 +6846,22 @@ ExpressionDelimiter
   &EOS InsertComma -> $2
 
 SimpleStatementDelimiter
-  &EOS
+  # Hera 0.9.0 returns `true` for assertions; normalize to undefined here
+  # so downstream delimiter checks (e.g. insertSemicolon) treat missing delim as falsy.
+  &EOS ->
+    return undefined
```
```suggestion
  &EOS -> undefined
```

Also below. Just style, feel free to ignore.

edemaine (dismissed):
I went ahead and updated according to the comments above. Feel free to overwrite if you disagree with anything. 

In particular, I removed the `<? "boolean"` test in `gen`, so if a `boolean` ever makes it to the AST, it will fail. I think that's good, but I guess it's also slightly dangerous if we don't have full coverage. If we put it back, we should add `boolean` to `ASTNode` type, like it is now for `TraversableNode`.

EDIT: Oops, I forgot to `pnpm install`, so I wasn't testing correctly. I'll see if I can fix the remaining `boolean`s quickly; otherwise, I'll revert.

edemaine:
OK, 2.5M GPT tokens later, I seem to have eliminated the remaining `true`s in the AST. I think this is the right approach?

(I do wonder if we did the right thing in Hera... perhaps we should have instead made `?` upgrade `undefined` into `true`? I'm not sure.)

STRd6:
I think `true` for the assertions is the better long term approach. It's only a hassle for us now because we built on the older expectation. And with this fix it looks like it's already behind us.

We can also consider allowing `$&Rule` in Hera to convert `true` -> `''` but that's fairly esoteric.

And filtering `undefined`/`true` in a pruning step is always an option if we wanted to allow the grammar to be sloppier.

Thanks for the cleanup!

source/parser/for.civet
```diff
@@ -246,7 +246,7 @@ function forRange(
 
 function processForInOf($0: [
   awaits: ASTNode
-  eachOwn: [ASTLeaf, ASTNode] | undefined
+  eachOwn: [ASTLeaf, ASTNode] | boolean | undefined
```
STRd6, line 249: Should we remove `| boolean` here?

source/parser/traversal.civet
```diff
@@ -7,6 +7,7 @@ import type {
 import { isFunction } from ./util.civet
 
 export type Predicate<T extends ASTNodeObject> = (arg: ASTNodeObject) => arg is T
+type TraversableNode = ASTNode | boolean
```
STRd6, line 10: Is this `| boolean` still necessary?

PR #2007, opened 2026-04-24
CLI lazy loads unplugin, fix race, allow symlink failure on Windows

edemaine:
Fix some testing for Windows, again...

* CLI had a race where it declared being done before actually emitting files. This caused the "wrote `.d.ts`" test to fail rarely. (This didn't actually happen in practice because the Node process would have waited... but it's good for API access to the CLI.)
* CLI now lazy-loads unplugin only when necessary. Speeds up tests that don't need it.
* Symlinks only work as administrator on Windows. This is fine for CLI but not for normal users. Just skip the tests in those cases.
* Better failure diagnostics

### DIFF
--- a/source/cli.civet
+++ b/source/cli.civet
@@ -2,8 +2,6 @@
 type { ASTError, BlockStatement, ParseError, ParseErrors } from ./main.civet
 { findConfig, loadConfig } from ./config.civet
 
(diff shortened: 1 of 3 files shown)


STRd6 (approved):
Looks good!

I think you can use `@skip()` inside the test to mark it as skipped.

PR #2009, opened 2026-04-25
Typescript cleanup

STRd6:
Claude is going ham on this, I still need to review.


STRd6:
Closed in favor of #2090 and #2038

PR #2011, opened 2026-04-25
Semantic tokens for comments, fixing `coffeeComment` syntax highlighting

edemaine:
TypeScript's semantic tokens, which #1881 recently added support for, sadly do not include comments. This PR adds such semantic tokens manually.  As a result, `coffeeComment` comments are finally highlighted correctly:

image

It works only on Civet 0.9.4+. Before that, the `SourceMap` interface was different, and I didn't feel like backporting further. But at least the VSCode plugin doesn't crash in this scenario; it falls back to not looking for `{ type: "Comment" }` nodes.

Minor LSP improvements, which surfaced when testing on an old repo of mine that had `coffeeComments` enabled:
* Old Civet didn't have `/config` module; add helpful logs in that case
* Output bundled Civet version number when it's used

### DIFF
--- a/lsp/server/package.json
+++ b/lsp/server/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@danielx/civet-language-server",
   "description": "Civet Language Server (standalone, editor-agnostic)",
-  "version": "0.3.32",
+  "version": "0.3.33",
   "repository": {
     "type": "git",
     "url": "https://github.com/DanielXMoore/civet.git"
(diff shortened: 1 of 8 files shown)


STRd6:
@claude add coverage to hit 100%

edemaine:
Oops, I was already working on coverage tests. @claude can you stop?

STRd6 (approved):
Thanks!

Sorry for throwing the robot into the middle of your work 🤖 😨

edemaine:
Haha, no worries! Eventually I figured out I can cancel the job to stop it. 

Should I bump the version number in this PR, or a separate one? I assume LSP release is automatic now?

edemaine:
Also, should I bump both server and vscode versions?

STRd6:
Release is automatic, bump in this PR is fine.

PR #2013, opened 2026-04-26
CI: separate typecheck step

STRd6:


edemaine (approved):
Can we get a `test:typecheck` (or whatever) NPM script too?

I've found it weird/unfamiliar that `pnpm test` outputs type errors — at least, as long as there are so many of them, it's annoying. But it'd be good to have a script to do it too, not just CI.

STRd6:
> Can we get a `test:typecheck` (or whatever) NPM script too?

`pnpm typecheck` should do it

PR #2015, opened 2026-04-28
Extract shared @danielx/civet/ts-service module

STRd6:
## Summary

Extracts the LSP server's TypeScript LanguageService machinery into a new shared `source/ts-service/` module exposed as `@danielx/civet/ts-service`, then routes the LSP, the build-time typecheck (now `civet --typecheck`), and unplugin's typecheck/declaration-emit path through one implementation instead of three parallel ones.  Bumps TypeScript across the workspace to 6.0.2 so consumers share the same `ts.LanguageService` types.

The unplugin's bundling fast path (esbuild/tsc/preserve modes) is unchanged — only the typecheck oracle moves.

## What's in `source/ts-service/`

```
types.civet         Plugin / Transpiler / FileMeta / TranspiledDoc / DocFactory / Logger
path.civet          getCanonicalFileName / getExtensionFromPath / removeExtension / remapFileName
snapshot.civet      Snap, fullDiffTextChangeRange (vue-style change ranges)
host.civet          TSHost factory + Host interface
service.civet       TSService factory + ResolvedService
doc.civet           createInMemoryDocFactory (LSP-free DocFactory for CLI / unplugin)
plugins/civet.civet makeCivetPlugin
plugins/hera.civet  makeHeraPlugin (Hera→Civet→TS, composed sourcemap)
```

`DocFactory<D>` abstracts the LSP-specific `vscode-uri` + `vscode-languageserver-textdocument` deps so non-LSP consumers can supply minimal in-memory docs.  Adds full `.hera` LSP support — typechecking, hover, go-to-def — via `makeHeraPlugin`, lifting the recipe previously buried in `build/cache-utils.js`.

## CLI: `civet --typecheck`

The build-time typecheck script (`scripts/typecheck.civet`, 276 lines) is gone — replaced by integrating typecheck mode into the existing CLI:

- `civet --typecheck` lints the whole project (`pnpm typecheck` calls this).
- `civet --typecheck path/to/file.civet [...]` scopes to specific files.
- `--max-errors N` (or `CIVET_TYPECHECK_MAX_ERRORS=N`) lets you carry a regression baseline above 0.
- `--tsbuildinfo PATH` persists incremental analysis across runs (used by CI to keep typecheck fast).

CI's `pnpm typecheck` step is now `./dist/civet --typecheck --max-errors 3671 --tsbuildinfo .cache/typecheck.tsbuildinfo`.

## Consumer impact

| Consumer | Before | After |
|---|---|---|
| `lsp/server/source/lib/typescript-service.civet` | 708 LOC | 128 LOC |
| `scripts/typecheck.civet` | 276 LOC | (removed) |
| `source/unplugin/unplugin.civet` typecheck path | hand-rolled VFS | thin wrapper around shared TSService |
| `source/cli.civet` | — | `--typecheck` / `--max-errors` / `--tsbuildinfo` flags |

## Trade-offs

- Diagnostic count drifts slightly from main (3645 → 3671 — TS 6.0.2 alignment + more precise host resolution); CI baseline updated.
- Typecheck wall time is in the same ballpark as before.  The disk-backed `compileToTS` cache the old hand-rolled script had can be reintroduced as a plugin-level cache later.
- `CIVET_TYPECHECK_TRACE` / `TRACE_RESOLUTION` are gone; if you want them back, plumb through `extraCompilerOptions`.

## Review feedback round-up

The first round of review (CHANGES_REQUESTED) flagged five concrete issues; each fixed in its own commit on top of the original extraction, exercised by a TDD test:

1. **TypeScript optional for plain transform consumers** — `unplugin.civet` no longer statically imports `@danielx/civet/ts-service`; the runtime module is loaded lazily inside `buildStart` behind the existing `transformTS` guard.  ([65e35203](https://github.com/DanielXMoore/Civet/pull/2015/commits/65e35203))
2. **Inline `options.tsConfig` reaches TSService** — TSService gains a `tsConfig?: any` option (the JSON shape `readConfigFile` returns); when supplied, the disk read is skipped and the inline config drives type-checking.  Bonus: TSService's mogrify-aware `configSys` adds *every* registered transpiler extension (`.civet` + `.hera`) so `.hera` files in `include` patterns now reach the typecheck program too — unplugin's own readDirectory wrapper only added `.civet`.  ([80434581](https://github.com/DanielXMoore/Civet/pull/2015/commits/80434581))
3. **`package.json#imports` rewriting restored** — TSHost overrides `readFile` to mogrify transpilable extensions in package.json's `imports` map, and `fileExists` to recognize synthetic `.tsx` siblings.  Lives in the shared host so LSP / CLI / unplugin all benefit.  ([c02dffb1](https://github.com/DanielXMoore/Civet/pull/2015/commits/c02dffb1))
4. **`buildEnd(useConfigFileNames=true)` honored** — the parameter was renamed `_useConfigFileNames` and ignored; now it pulls tsconfig-discovered files into the typecheck scope via a small new `host.addScriptFileName(p)` method.  ([6afebf3a](https://github.com/DanielXMoore/Civet/pull/2015/commits/6afebf3a))
5. **`node_modules` skip dropped** — defer to TS's `skipLibCheck` as the canonical filter; project's own typecheck baseline unchanged since our tsconfig sets `skipLibCheck: true`.  ([aaf193da](https://github.com/DanielXMoore/Civet/pull/2015/commits/aaf193da))

Bonus that fell out of 2 + 5: switched the diagnostic loop to `ts.getPreEmitDiagnostics` ([6c72f503](https://github.com/DanielXMoore/Civet/pull/2015/commits/6c72f503)) — folds in declaration-emit diagnostics that the manual `syn + sem + options + global` walk dropped, and stays in sync with whatever TS adds in future versions.

## Test plan

- [x] Main test suite: 3835 passing
- [x] LSP test suite: 220 passing
- [x] `pnpm typecheck` clean at 3671 baseline (with `.cache/typecheck.tsbuildinfo` for incremental fast paths)
- [x] CI green on the latest push: [run 25147945454](https://github.com/DanielXMoore/Civet/actions/runs/25147945454)
- [x] `source/ts-service/**` at 100/100/100/100 across the LSP coverage suite
- [x] `unplugin.civet` at 100/100/100/100 in the main coverage suite


source/ts-service/service.civet
```diff
@@ -0,0 +1,173 @@
...
+  //@ts-ignore
+  baseHost := createCompilerHost(parsedConfig)
+
+  host := TSHost(parsedConfig.options, parsedConfig.fileNames, baseHost, transpilers, docFactory, logger, libDir)
+  service := createLanguageService(host)
+
+  loadPlugins := async function (): Promise<void>
+    civetFolder := path.join(projectPath, "./.civet/")
+    civetFiles := sys.readDirectory(civetFolder)
+    pluginFiles := civetFiles.filter((file) => file.endsWith("plugin.mjs"))
```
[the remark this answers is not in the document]
STRd6: We should wrap in a try catch, there's no guarantee that the directory will still exist after calling `sys.directoryExists`

lsp/server/source/lib/typescript-service.civet
```diff
@@ -28,470 +25,49 @@ ts from typescript
...
+// Bridge the shared TSHost to vscode-languageserver-textdocument /
+// vscode-uri.  Other consumers (typecheck script, CLI --typecheck) supply
+// their own DocFactory implementations.
+docFactory: DocFactory<TranspiledDoc> := {
+  create: (uri, languageId, version, text) =>
+    TextDocument.create(uri, languageId, version, text)
+  update: (doc, text, version) =>
+    TextDocument.update(doc as TextDocument, [{ text }], version)
+  uriToPath: (uri) => URI.parse(uri).fsPath
+  pathToUri: (p) => URI.file(p).toString()
```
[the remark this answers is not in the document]
STRd6: False positive — `TextDocument.update` *does* mutate in place. Verified at runtime:

```js
const doc = TextDocument.create('file:///x', 'civet', 0, 'old');
const ret = TextDocument.update(doc, [{ text: 'new' }], 7);
ret === doc;        // → true (same instance returned for chaining convenience)
doc.version;        // → 7 (mutated)
doc.getText();      // → 'new' (mutated)
```

So discarding the return value is fine here — the original `transpiledDoc` reference in `pathMap` does pick up the new version/text. The 30s → 54s typecheck regression has a different cause (lost the disk-backed `compileToTS` cache from the old typecheck script's hand-rolled pipeline; can be reintroduced as a plugin-level cache later).

source/ts-service/service.civet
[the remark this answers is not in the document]
STRd6: The "crashes on missing folder" claim is a false positive — `ts.sys.readDirectory` returns `[]` for a missing directory in current TypeScript, doesn't throw:

```js
const ts = require('typescript');
ts.sys.readDirectory('/tmp/definitely-does-not-exist-xyzzy/');
// → []  (no throw)
```

Wrapping in try/catch is still worth doing as forward-compat insurance though, per @STRd6's separate suggestion below — a future TS version or alternative `System` impl might decide to throw, and there's no race-free way to pre-check existence between `directoryExists` and `readDirectory`. Done in [a9b08302](https://github.com/DanielXMoore/Civet/pull/2015/commits/a9b08302), c8-ignored since no current TS version triggers the catch.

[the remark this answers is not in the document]
STRd6: Done — wrapped in try/catch in [a9b08302](https://github.com/DanielXMoore/Civet/pull/2015/commits/a9b08302). c8-ignored since `ts.sys.readDirectory` returns `[]` rather than throwing in current TS versions, but the catch is in place for forward-compat / alternative `System` implementations.

edemaine:
@greptile-ai revise your review

source/ts-service/plugins/hera.civet
```diff
@@ -0,0 +1,89 @@
...
+ * gain `.hera` support — type checking, hover, go-to-def — for free.
+ *
+ * The composed sourcemap (Civet's `upstreamSourceMap`) means TS diagnostics
+ * on the synthetic `.hera.tsx` map all the way back to `.hera` line/column
+ * through the same `remapPosition` plumbing every other consumer uses.
+ */
+
+{ type CompileOptions, type ParseError } from @danielx/civet
+BundledCivetModule from @danielx/civet
+{ compile as bundledHeraCompile } from @danielx/hera
```
edemaine: This seems to require `@danielx/hera` to be a real dependency, but it's currently only a dev dependency. Perhaps we should aim to make it an optional peer dependency (importing this plugin only if `@danielx/hera` loads dynamically), as many Civet users might not need Hera support?

source/unplugin/unplugin.civet
edemaine, line 301: I'm surprised not to see all the above code moved to the shared package. Isn't it useful in all cases?

Furthermore, it seems that the tsconfig that gets mogrified here and then passed in is ignored, and the shared package just loads `tsconfig.json` again?

This means that all the magic `tsconfig` rewriting features, like handling `.civet` extensions in various circumstances, no longer works.

source/unplugin/unplugin.civet
```diff
@@ -251,110 +273,103 @@ export rawPlugin := (options: PluginOptions = {}, meta: UnpluginContextMeta) =>
...
-            sourceMap := sourceMaps.get file.fileName
-            if (!sourceMap) return diagnostic
-
-            sourcemapLines := sourceMap.lines
-            /* c8 ignore start -- \`|| 0\` / \`|| 1\` fallbacks only fire for diagnostics with undefined or zero start/length; TS always sets these when file is non-null */
+    // useConfigFileNames is preserved for backwards compat; previously
+    // toggled rootNames between fsMap.keys() and tsconfig's fileNames.  The
+    // TSService-driven flow walks the actual program built from registered
+    // documents + their transitive imports, so neither set drives scope
+    // anymore.  Underscore-prefix to silence noUnusedParameters.
```
edemaine: This seems to break a useful previous feature: CLI `civet --typecheck` used to check all files, by setting this parameter to `true`. Now `discoverProjectFiles` is set to `false` in all cases (see a few lines above this comment), so files to typecheck need to be listed explicitly.

source/unplugin/unplugin.civet
```diff
@@ -580,6 +595,23 @@ export rawPlugin := (options: PluginOptions = {}, meta: UnpluginContextMeta) =>
...
+        // drives buildEnd's diagnostics + declaration emit.
+        if service?
+          // Version is just a host-bookkeeping monotonic counter; unplugin
+          // isn't editor-driven, so a fixed 0 is fine — every load() call
+          // creates a fresh doc that addOrUpdateDocument swaps in, which
+          // invalidates the snapshot map and forces a recompile.
+          doc := docFactory.create(
+            pathToFileURL(filename).toString()
+            "civet"
+            0
```
edemaine: Fixing the version number to 0 will fail when using watching mode. Admittedly I've never tested typechecking in watch mode...

source/unplugin/unplugin.civet
```diff
@@ -251,110 +273,103 @@ export rawPlugin := (options: PluginOptions = {}, meta: UnpluginContextMeta) =>
...
+        // its own type errors.  Skip node_modules: their diagnostics aren't
+        // actionable for the user's build.
+        diagnostics: Diagnostic[] := []
+        for sourceFile of program.getSourceFiles()
+          fn := sourceFile.fileName
+          /* c8 ignore next 2 -- second clause covers Windows backslash paths; CI is Linux-only */
+          if fn.includes('/node_modules/') or fn.includes('\\node_modules\\')
+            continue
+          syn := service.getSyntacticDiagnostics fn
+          sem := service.getSemanticDiagnostics fn
```
edemaine: We used to use `getPreEmitDiagnostics`. New code drops global/options/declaration diagnostics.

source/unplugin/unplugin.civet
```diff
@@ -251,110 +273,103 @@ export rawPlugin := (options: PluginOptions = {}, meta: UnpluginContextMeta) =>
...
+        // This matches the previous `getPreEmitDiagnostics(program)` shape:
+        // a `.ts` file imported from a tracked .civet file should surface
+        // its own type errors.  Skip node_modules: their diagnostics aren't
+        // actionable for the user's build.
+        diagnostics: Diagnostic[] := []
+        for sourceFile of program.getSourceFiles()
+          fn := sourceFile.fileName
+          /* c8 ignore next 2 -- second clause covers Windows backslash paths; CI is Linux-only */
+          if fn.includes('/node_modules/') or fn.includes('\\node_modules\\')
+            continue
```
edemaine, line 517: Do we always want to skip `node_modules`? Maybe - just checking because we didn't before.

edemaine (commented):
I (or rather, GPT) found a bunch of regressions in the unplugin. Currently I think this would break a lot of the special `tsconfig` rewriting behaviors that I worked hard on before, and also breaks the feature `civet --typecheck` without specified filenames. (I guess this is what happens when we don't have tests for functionality — the unplugin is full of such things.) Also I think Hera should be an optional peer dependency, but the way the code's written now, it needs to be required, and it's not listed.

I tried to add specific comments where they're most relevant, but I'll also include the full GPT review below.

---

**Findings**
1. **Hera and TypeScript are now hard unplugin runtime deps.** `source/unplugin/unplugin.civet:11-21` statically imports `@danielx/civet/ts-service`; `ts-service` exports/imports `makeHeraPlugin`, which imports `@danielx/hera` at `source/ts-service/plugins/hera.civet:17`, and the shared module imports `typescript` broadly. `build/esbuild.civet:271-275` externalizes both, but `package.json:120-158` has Hera only in `devDependencies` and TypeScript as an optional peer. This likely breaks plain unplugin users without Hera/TS installed.

2. **`options.tsConfig` is parsed by unplugin but ignored by the actual TSService program.** Unplugin parses inline/disk config at `source/unplugin/unplugin.civet:215-260`, but then constructs `TSService(servicePath, ...)` at `281-296`; `TSService` rereads only `servicePath/tsconfig.json` at `source/ts-service/service.civet:126-145`. Inline `tsConfig`, custom `files`, `paths`, `strict`, `moduleResolution`, etc. can diverge from diagnostics/declaration emit.

3. **`buildEnd(useConfigFileNames=true)` behavior changed.** The parameter is now explicitly ignored at `source/unplugin/unplugin.civet:298-303`. `source/cli.civet:678-684` still passes `true` when no filenames are provided, which previously meant “typecheck tsconfig file list”; now the shared service is created with `discoverProjectFiles: false` at `source/unplugin/unplugin.civet:292-295`, so CLI/unplugin whole-project typecheck semantics are not preserved.

4. **Watch/rebuild typecheck can reuse stale transpiled text.** Unplugin always creates service docs with version `0` at `source/unplugin/unplugin.civet:607-613`. The host only recompiles when `sourceDocVersion > transpiledDoc.version` at `source/ts-service/host.civet:350-352`; after the first compile the transpiled doc is also version `0`, so a later changed doc can keep old diagnostics/declarations.

5. **Several old unplugin VFS behaviors are missing from shared service.** Main’s unplugin rewrote `package.json#imports` `.civet` targets and generated `.civet.tsx` on demand; the new host does neither. Current shared discovery only patches `readDirectory` at `source/ts-service/service.civet:96-122`, so tsconfig `"files": ["x.civet"]` is not equivalently virtualized.

6. **Unplugin diagnostics behavior changed.** Old code used `ts.getPreEmitDiagnostics(program)`; new code only walks source files and asks syntactic/semantic diagnostics at `source/unplugin/unplugin.civet:330-337`. That drops global/options/declaration diagnostics, while also skipping `node_modules` at `331-334`. Additionally, because `TSService` still seeds the host with normal tsconfig `.ts/.js` files, unplugin may now report unrelated tsconfig TS files even though comments say it is bundler-driven.

7. **Some Hera/typecheck ideas did not make it into shared service.** The Hera two-stage compile with composed sourcemap is present (`source/ts-service/plugins/hera.civet:46-76`), but the old typecheck shim for `@danielx/hera/dist/machine.ts` and the `sourceMap.lines ?? sourceMap.data?.lines` fallback were not carried over. Current host stores only `sourceMap?.lines` at `source/ts-service/host.civet:399`.

Overall: the shared module is a good extraction of the LSP host shape and Hera compile recipe, but I would not merge this as “behavior-preserving” for unplugin yet. The dependency regression and ignored `tsConfig`/`useConfigFileNames` paths are the biggest blockers.

source/ts-service/plugins/hera.civet
STRd6: Done in [7fa65b6b](https://github.com/DanielXMoore/Civet/pull/2015/commits/7fa65b6b) — switched to a lazy `createRequire`, declared `@danielx/hera` in `peerDependenciesMeta` as optional, and updated the doc on `HeraPluginOptions.Hera` to call out the lazy-load.  The createRequire base picks `__filename` in CJS and falls back to `import.meta.url` in ESM so the same source compiles cleanly to both `.js` and `.mjs` (and survives the LSP server bundling our `.mjs` files into its CJS dist, where esbuild's `define` rewrites `import.meta.url` to `""").

— Your pal Claude

source/unplugin/unplugin.civet
STRd6: Mostly intentional but worth unpacking the layers here:

* The unplugin in **TS-mode** (`tsc` or `civet`) still needs its own per-file pipeline — `compileOptions` / `compilerOptionsWithSourceMap` drive `civet.generate` and `tsTranspile` for the actual code emit downstream.  That's not type-checking work, so it doesn't belong in the shared `@danielx/civet/ts-service`.
* The type-checking pass IS now driven by TSService, and the consumer-supplied tsconfig flows through via the `extraCompilerOptions` parameter (added in [7247adc3](https://github.com/DanielXMoore/Civet/commit/7247adc3)).  TSService merges it on top of its parsed tsconfig, so unplugin / CLI can override compile options without rewriting tsconfig.json.
* For the `.civet`-extension handling: TSService's `configSys.readDirectory` wrapper rewrites `.civet` → `.civet.tsx` on glob discovery, and TSHost now also rewrites raw `.civet` entries from `tsconfig.files` ([7fa65b6b](https://github.com/DanielXMoore/Civet/pull/2015/commits/7fa65b6b)) — both paths covered now.

What's NOT yet shared: unplugin's per-file fsMap, declaration emit, source-map plumbing for the bundler output.  Those felt like a separate refactor from this PR's typecheck-unification scope; happy to file a follow-up if you want them moved.

— Your pal Claude

source/unplugin/unplugin.civet
STRd6: Already fixed in [afaa655a](https://github.com/DanielXMoore/Civet/commit/afaa655a) — the CLI now passes `discoverProjectFiles: filenames.length is 0` so the no-args invocation (the form `pnpm typecheck` and CI use) opts back into project-wide lint, while explicit-filenames invocations stay scoped to those files.  The default change to `false` was deliberate for bundler-driven flows where Vite/webpack/etc. own the file set.

— Your pal Claude

source/unplugin/unplugin.civet
STRd6: Good catch — fixed in [7fa65b6b](https://github.com/DanielXMoore/Civet/pull/2015/commits/7fa65b6b).  unplugin now bumps a `nextDocVersion` counter on every `load()` call, so the host's `sourceDocVersion > transpiledDoc.version` check actually advances after a watch-mode reload (the previous fixed `0` would equal the transpiledDoc's version after the first compile, leaving the second load's new content stuck on the stale transpile).

— Your pal Claude

source/unplugin/unplugin.civet
STRd6: Fixed in [7fa65b6b](https://github.com/DanielXMoore/Civet/pull/2015/commits/7fa65b6b) — the buildEnd loop now folds `program.getOptionsDiagnostics()` and `program.getGlobalDiagnostics()` in alongside the per-file syntactic + semantic batches.  Skipped `getDeclarationDiagnostics()` on purpose: it requires emit and we're running with `noEmit: true` (or only writing tsbuildinfo), so it'd return nothing useful.

— Your pal Claude

source/unplugin/unplugin.civet
STRd6: Intentional — matches what the deleted `scripts/typecheck.civet` did for the same reason: TS diagnostics inside `node_modules` are almost always `.d.ts` quirks the user can't act on (mismatched lib types, package-internal lints), and the same modern tsconfigs that opt into typechecking via this flow also tend to set `skipLibCheck: true` which already swallows them.  Keeping the skip avoids a noisy regression for projects whose diagnostic count was previously dominated by `node_modules` chatter.

If we want it user-configurable, easy follow-up — `includeNodeModulesDiagnostics: true` on PluginOptions — but let's wait for someone to actually need it before adding a knob.

— Your pal Claude

edemaine: Isn't this what `skipLibCheck` is for? Why would we have another flag when we can just respect this one?

source/unplugin/unplugin.civet
edemaine: `filenames.length is 0` seems like a reasonable approach. However, I'm opposed to keeping the `useConfigFileNames` parameter when it's completely ignored. Please remove.

source/unplugin/unplugin.civet
STRd6: Fair, you're right — dropped the hard-coded skip in [aaf193da](https://github.com/DanielXMoore/Civet/pull/2015/commits/aaf193da).  `skipLibCheck` does the right job and our skip was actually broader than that (covered `.ts`/`.js` under `node_modules` too, e.g. yarn-linked workspaces, occasionally hiding real diagnostics).  Civet's own typecheck baseline holds at 3671 — our tsconfig has `skipLibCheck: true` so removing the redundant filter didn't surface any new diagnostics.

— Your pal Claude

source/unplugin/unplugin.civet
edemaine: > Skipped `getDeclarationDiagnostics()` on purpose: it requires emit and we're running with `noEmit: true` (or only writing tsbuildinfo), so it'd return nothing useful.

Disagree. In fact, we emit declarations in some cases. Declaration diagnostics are “would declaration emit be valid?” diagnostics; they can be computed before deciding whether to write files.

I'd still lean toward using `getPreEmitDiagnostics` which includes everything, and feels more future proof in case more types of diagnostics get added. It should also resolve the `node_modules` issue; if `skipLibCheck` is true, it should skip those files.

edemaine (changes requested):
Regarding the changes in response to the last review (in addition to the comment threads where I  responded), I see these issues:

- **Still a blocker: unplugin now eagerly requires TypeScript.** `source/unplugin/unplugin.civet:11-21` imports `@danielx/civet/ts-service` at module load time, and `source/ts-service/service.civet:14-23` imports `typescript`. So even users who only use the normal unplugin transform path, with no `typecheck`, `emitDeclaration`, or `ts: "tsc"`, now need TypeScript installed. Hera was fixed; TypeScript was not.

- **Still a blocker: inline `options.tsConfig` is parsed but not used by TSService.** Unplugin parses `options.tsConfig` at `source/unplugin/unplugin.civet:271-316`, but `TSService` is created at `source/unplugin/unplugin.civet:352-369` and then rereads `servicePath/tsconfig.json` in `source/ts-service/service.civet:134-153`. The new `extraCompilerOptions` path only carries `tsBuildInfoFile` options at `source/unplugin/unplugin.civet:341-350`, not the parsed inline config/options/files. This means custom `tsConfig.compilerOptions`, inline `files`, and cases without an actual `tsconfig.json` are still not parity-preserving.

- **Still missing: old `package.json#imports` rewriting.** Main’s unplugin VFS rewrote package `imports` entries ending in `.civet` to `.civet.tsx`. I don’t see an equivalent in the shared host; `source/ts-service/host.civet:108-199` has a TODO for package.json directory resolution but no import-map rewriting. The current test at `test/infra/unplugin.civet:943-978` does not really cover it because the source imports `./foo.civet.tsx` directly and catches errors.

- **Partially fixed: `buildEnd(true)` / project-wide scope.** The CLI no-args case is fixed by setting `discoverProjectFiles: filenames.length is 0` at `source/cli.civet:565-569`. But the public `buildEnd(_useConfigFileNames = false)` parameter remains ignored at `source/unplugin/unplugin.civet:371-376`, so external callers relying on `buildEnd(true)` behavior still get a behavior change.

- **Still a behavior change: blanket `node_modules` diagnostics skip.** `source/unplugin/unplugin.civet:456-460` skips all per-file diagnostics in `node_modules`. The response says this is intentional, but your follow-up point stands: `skipLibCheck` is the established knob, and the old unplugin did not blanket-suppress this.

source/unplugin/unplugin.civet
edemaine: Sorry, I meant the `package.json` is mogrified (to rewrite `imports`) but it's ignored downstream.

STRd6:
Re: [the CHANGES_REQUESTED review](https://github.com/DanielXMoore/Civet/pull/2015#pullrequestreview-4200331441) — all five items addressed in the latest push.

1. **TypeScript no longer eagerly required.**  Fixed in [65e35203](https://github.com/DanielXMoore/Civet/pull/2015/commits/65e35203).  Dropped the static `import { TSService, ... } from '@danielx/civet/ts-service'` at the top of `unplugin.civet`; it's now a `typeof import(...)` for types only, with the runtime module lazy-loaded inside `buildStart` behind the existing `transformTS` guard (same shape as `tsPromise`).  TDD: a new test spawns a fresh node subprocess (so prior typecheck-mode tests don't pollute `require.cache`) and asserts both `require('@danielx/civet/unplugin')` and `rawPlugin({ ts: 'civet' }, ...)` leave `typescript` out of the cache.

2. **Inline `options.tsConfig` reaches TSService.**  Fixed in [80434581](https://github.com/DanielXMoore/Civet/pull/2015/commits/80434581).  TSService now accepts a `tsConfig?: any` option — the JSON shape `readConfigFile` returns, *before* `parseJsonConfigFileContent`.  When supplied, the disk read is skipped and TSService's own `configSys` mogrification + `parseJsonConfigFileContent` runs over the in-memory config.  Bonus: TSService's `configSys` adds *every* registered transpiler extension (`.civet` + `.hera`), so .hera files in `include` patterns now reach the typecheck program too — unplugin's own readDirectory wrapper only added `.civet`.  TDD: an inline-only `tsConfig.compilerOptions` with `declaration: true, isolatedDeclarations: true` produces TS9010 (which can only fire when the compilerOptions actually drove the program).

3. **`package.json#imports` rewriting restored.**  Fixed in [c02dffb1](https://github.com/DanielXMoore/Civet/pull/2015/commits/c02dffb1).  Lives in `TSHost` now (so the LSP, CLI typecheck, and unplugin all get it) rather than the unplugin-only VFS.  Two host overrides: `readFile` rewrites `.civet`/`.hera` values inside `package.json`'s `imports` field to their `.tsx` siblings (walking nested condition objects too), and `fileExists` returns `true` for synthetic targets when the underlying source exists — without that, TS's resolver still bottoms out on a missing-file check after reading the rewritten map.  TDD: a fixture with `"imports": { "#dep": "./dep.civet" }` plus a type-mismatch through the import — passes only when both the rewrite and the file-existence override land.

4. **`buildEnd(useConfigFileNames=true)` honored.**  Fixed in [6afebf3a](https://github.com/DanielXMoore/Civet/pull/2015/commits/6afebf3a).  The parameter was renamed `_useConfigFileNames` and ignored; now `buildStart` captures `parsedConfig.fileNames` (already mogrified) into a closure-level `configFileNames`, and `buildEnd(true)` walks them through a small new `host.addScriptFileName(p)` method that registers paths as program roots without requiring a `TranspiledDoc`.  TDD: two `.civet` files listed in `tsConfig.files` but only one loaded via `load()` — `buildEnd(false)` reports one diagnostic, `buildEnd(true)` reports both.

5. **`node_modules` skip dropped.**  Fixed in [aaf193da](https://github.com/DanielXMoore/Civet/pull/2015/commits/aaf193da).  Defer to TS's `skipLibCheck` as the canonical filter; project's own typecheck baseline holds at 3671 since our tsconfig sets `skipLibCheck: true`.

Bonus that fell out of #2 and #5 together: switched the diagnostic loop to `ts.getPreEmitDiagnostics` ([6c72f503](https://github.com/DanielXMoore/Civet/pull/2015/commits/6c72f503)) — folds in declaration-emit diagnostics that the manual `syn + sem + options + global` walk dropped, and stays in sync with whatever TS adds in future versions.

CI: green on the latest push ([run 25147945454](https://github.com/DanielXMoore/Civet/actions/runs/25147945454)).  Coverage: `source/ts-service/**` at 100 across the LSP suite; `unplugin.civet` at 100 in the main suite.

— Your pal Claude

edemaine:
One more thing:

- Legacy sourcemap fallback is still missing: `source/ts-service/host.civet:494` uses only `sourceMap?.lines`, not `sourceMap?.data?.lines`. If legacy/project Civet versions can return only `data.lines`, diagnostics won’t remap.

In most contexts, we don't care about legacy, but the LSP server needs to care, because the user's Civet installation may be old. So I think it's worth fixing this.

STRd6:
> One more thing:
> 
> * Legacy sourcemap fallback is still missing: `source/ts-service/host.civet:494` uses only `sourceMap?.lines`, not `sourceMap?.data?.lines`. If legacy/project Civet versions can return only `data.lines`, diagnostics won’t remap.
> 
> In most contexts, we don't care about legacy, but the LSP server needs to care, because the user's Civet installation may be old. So I think it's worth fixing this.

I think removing legacy sourcemap is fine. If someone raises an issue I'd rather backport the sourcemap format than maintain both indefinitely.

edemaine:
> I think removing legacy sourcemap is fine. If someone raises an issue I'd rather backport the sourcemap format than maintain both indefinitely.

I think this would affect me. Just looking [at my uses of Civet](https://github.com/search?q=owner%3Aedemaine+%22%40danielx%2Fcivet%22%3A&type=code), about half are <0.9.4. Admittedly, it would just disable source mapping in VSCode, so it's not the end of the world — but it would probably be confusing for a while until I figured out what was going on and then upgraded Civet. Upgrading across major versions is not always trivial though...

So I added the one-line change. We can reconsider in the future, but 0.9.4 still feels relatively new to me.

build/esbuild.civet
```diff
@@ -245,3 +245,44 @@ for format of ["esm", "cjs"]
...
+// extension.  External: typescript / @danielx/civet, like ts-diagnostic.
+//
+// Sourcemaps are enabled (overriding the file-level default) so c8 / Codecov
+// can attribute coverage hits from consumers (LSP, unplugin, typecheck) back
+// to the original .civet sources rather than the compiled .mjs.
+for format of ["esm", "cjs"]
+  build({
+    entryPoints: [
+      ...(for name of ['index', 'types', 'snapshot', 'path', 'host', 'service', 'doc']
+        `source/ts-service/${name}.civet`)
```
edemaine: Is there a reason to keep these separate, instead of bundling them together? (I can think of a couple of possible reasons: emitting types, or tree shaking. But not sure they apply.)

(On the other hand, definitely makes sense for plugins to be separate.)

.github/workflows/build.yml
```diff
@@ -75,7 +75,7 @@ jobs:
       - name: Typecheck
         if: runner.os == 'Linux'
         run: |
-          CIVET_TYPECHECK_MAX_ERRORS=3667 node_modules/.bin/civet scripts/typecheck.civet
+          ./dist/civet --typecheck --max-errors 3671 --tsbuildinfo .cache/typecheck.tsbuildinfo
```
edemaine: I'm not sure this will work in CI... I'm guessing we cache `.cache` already, but it's still not guaranteed to persist. Probably we should remove this option from CI and just use it in `pnpm typecheck`?

STRd6:
> So I added the one-line change. We can reconsider in the future, but 0.9.4 still feels relatively new to me.

That's fair, thanks for the fix!

build/esbuild.civet
STRd6: Good call — collapsed to a single `source/ts-service/index.civet` entry in [2e75dae4](https://github.com/DanielXMoore/Civet/pull/2015/commits/2e75dae4).  The package only exposes `./ts-service` in its `exports` map, so per-file subpaths were never reachable to consumers anyway; ~22 output files (.js + .mjs + .maps + per-file dups) collapse to 4.  Source maps still attribute coverage to each contributing `.civet` source independently — LSP suite reports 100/100/100/100 across host / service / both plugins / doc / path / snapshot / types.

The `rewriteCivetImports` plugin dropped out as well — with a single entry there are no relative .civet imports between separately-bundled outputs to rewrite.

— Your pal Claude

.github/workflows/build.yml
STRd6: Done in [52ed07fa](https://github.com/DanielXMoore/Civet/pull/2015/commits/52ed07fa).  Switched the CI step from a literal `./dist/civet ...` invocation to `pnpm typecheck --max-errors 2942` so the `--tsbuildinfo` path lives in the `package.json` script (single source of truth) and the CI-only baseline gets passed through.  The workflow's existing restore/save-on-`.cache` block carries the tsbuildinfo across runs when the bucket key matches; cold runs just pay the full typecheck cost as before.

— Your pal Claude

edemaine:
I'm finally back to reviewing this; sorry for the delay.

I tested the new unplugin in some of my Vite projects. HMR works. Build works. So that's a good sign!

I asked GPT for another review, and it flagged these issues, mostly with the incremental checking (which I'm still understanding):

1. `source/unplugin/unplugin.civet:425-430` omits declaration diagnostics in the new incremental path.
   The previous path used `ts.getPreEmitDiagnostics`, which includes declaration diagnostics such as `isolatedDeclarations` TS9007/TS9010. The new `createIncrementalProgram` collection only gathers config/options/global/syntactic/semantic diagnostics. For `civet --typecheck` with project discovery, declaration-related errors can disappear again.

2. `source/ts-service/typecheck.civet:143-159` loses `package.json#imports` support in incremental typecheck.
   `TSHost` handles this by overriding `readFile` for `package.json` and rewriting `.civet`/`.hera` import targets to synthetic `.tsx` siblings. The new incremental host falls through to `ts.resolveModuleName(..., baseHost)` without that `readFile` mogrification, so no-args `civet --typecheck` can regress on projects using:
   `imports: { "#dep": "./dep.civet" }`.

3. `source/unplugin/unplugin.civet:425` plus `source/unplugin/unplugin.civet:578` can duplicate tsconfig parse errors in the incremental path.
   `rawDiagnostics` already includes `builder.getConfigFileParsingDiagnostics()`, then `configErrors` is also unshifted into `diagnostics`. This can double-print config errors and inflate `--max-errors` counts.

4. `package.json:128` vs `package.json:152` has an inconsistent Hera version contract.
   The devDependency is pinned to `@danielx/hera@0.9.2`, but the optional peer dependency still requires `^0.9.4`. If 0.9.2 is intentionally supported or preferred, the published peer range should allow it; otherwise consumers get unnecessary peer warnings despite matching the tested version.

PR #2020, opened 2026-05-03
Fix implicit return of `abstract class` mixin body

### LINKED ISSUE #2019: Bug: mixin function whose body is a single `abstract class` emits invalid `return abstract class`
When a mixin function's entire body is `abstract class X extends S ...`, Civet emits the class as the function's return *expression*:

```ts
return abstract class Mixin extends S { ... }
```

That is not valid TypeScript — `abstract` is only allowed on class *declarations*, not class expressions. TypeScript reports:

- `TS2304` Cannot find name 'abstract'
- `TS1005` ';' expected
- `TS2797` A mixin class that extends from a type variable containing an abstract construct signature must also be declared 'abstract'

### Repro

```civet
type Constructor<T = {}> = abstract new (...args: any[]) => T

export Mixin := <T extends Constructor>(S: T) ->
  abstract class Mixin extends S
    mix: string = "mix"
```

Compiles to:

```ts
export const Mixin = function<T extends Constructor>(S: T) {
  return abstract class Mixin extends S {  // ← invalid TS
    mix: string = "mix"
  }
}
```

### Workaround

Adding an explicit trailing `return Mixin` makes Civet emit the class as a statement followed by a return, which is valid:

```civet
export Mixin := <T extends Constructor>(S: T) ->
  abstract class Mixin extends S
    mix: string = "mix"
  return Mixin
```

```ts
export const Mixin = function<T extends Constructor>(S: T) {
  abstract class Mixin extends S {
    mix: string = "mix"
  }
  return Mixin
}
```

### Expected

The implicit-return form should also emit the named-declaration-then-return shape (or any other valid TS) so `abstract class` mixins don't require a manual workaround.

### Notes

- Plain (non-`abstract`) `class X extends S` does not hit this — `class` expressions can carry a binding name and are valid in expression position.
- Civet version: 0.11.2

STRd6:
Closes #2019

When a function's implicit-return body was a single named `abstract class`, Civet emitted `return abstract class X { ... }`, which is invalid TypeScript (`abstract` is only allowed on class *declarations*, not class expressions).

The fix mirrors the existing `FunctionExpression`-with-id handling in `insertReturn`: when the implicit-return expression is a `ClassExpression` with `abstract` set and a name, emit the class as a statement and splice in a separate `;return ClassName`. Non-abstract class expressions are unchanged.

### DIFF
--- a/source/parser/function.civet
+++ b/source/parser/function.civet
@@ -12,6 +12,7 @@ import type {
   BreakStatement
   CallExpression
   Children
+  ClassExpression
   Declaration
   ForStatement
   FunctionNode
@@ -83,6 +84,7 @@ import {
   isFunction
   isLoopStatement
   isStatement
+  isToken
   isWhitespaceOrEmpty
   makeLeftHandSideExpression
   makeNode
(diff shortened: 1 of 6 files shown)


STRd6:
@claude

- should handle anonymous functions with a `ref`
- Also needs equivalent code in assignResults

edemaine:

edemaine (approved):
LGTM, other than new type errors requiring a bump. (Maybe worth trying to reduce some?)

TIL you can return an `abstract class`. I guess I need to learn more about mixins.

source/parser/function.civet
```diff
@@ -381,6 +381,17 @@ function patternBindings(pattern: ASTNodeObject): BindingIdentifier[]
       when "Identifier", "AtBinding"
         bindings.push pattern
 
+// Ensure an `abstract class` expression has a name, generating a ref if it
+// doesn't, so it can be emitted as a class declaration. Returns the
+// identifier/ref to use for referring to the class afterwards.
+function nameAbstractClass(exp: ASTNodeObject): ASTNode
+  return exp.name if exp.name
```
edemaine: I agree this should probably be:

```suggestion
  return exp.id if exp.id
```

PR #2030, opened 2026-05-07
[bisect probe H] workflow change from 088a38b3 only on top of d5c52eac

STRd6:
Definitive test: source code = d5c52eac (good anchor, 3/3 GREEN). Only the .github/workflows/build.yml change from 088a38b3 (the Windows test step rewrite) is applied. If RED, the workflow change is the deterministic regressor; the source commits in the bisect window are just flake-rate noise.


STRd6:
Bisect complete — see BISECT_VERDICT.md. Closing draft.

PR #2032, opened 2026-05-07
Fix LSP `@` shorthand incorrectly completing as `@this.member`

### LINKED ISSUE #825: vscode extension incorrectly autocomplete '@this.'
When writing for a class in vscode using `@`, and accepting an autocomplete suggestion regarding class members vscode adds the redundant `this.` making the expression into an error.

vsc_civet_this

edemaine:
I took Claude's branch and merged into the modern LSP setup (`main`), and added an e2e test.

UPDATE: merged again. And fixing code coverage issues.

Original description:

---

Root cause: When user types `@` in a Civet class method, it compiles to `this` (without dot) in TypeScript. TypeScript's "ThisProperty" completions return `this.member` as insertText. When applied in Civet at position after `@`, this produced `@this.member` instead of `@member`.

Fixes:
1. In convertCompletions: strip `this.` prefix from insertText for ThisProperty/ completion entries in Civet files, so accepted completions produce `@member` not `@this.member`.
2. When @ is the trigger character: detect cursor inside `this` after forwardMap and advance to after `this`/`this.` for better TS completions; also change trigger from @ to . so TypeScript gives member completions.
3. Return isIncomplete: true when @ is the trigger, so VS Code re-fetches completions as user types more chars rather than filtering the initial list.

Tests:
- Add forwardMap test for @ shorthand in test/util.civet
- Add member completion test for @ shorthand in test/service.civet

Fixes #825


STRd6 (approved):

edemaine:
@greptileai revise you review

PR #2034, opened 2026-05-08
LSP: fix `@` completion in implicit return context

edemaine:
I noticed that `@` wasn't completing to attributes at all when used at the beginning of a line at the end of a method, i.e., in an implicit `return` context. This PR fixes that by broadening the search-for-`@` heuristic slightly.


edemaine:
Oops forgot a file. 
@greptileai please review again

STRd6 (approved):

PR #2035, opened 2026-05-08
Fix unplugin + esbuild sourcemap source paths when using library

### LINKED ISSUE #1784: Sourcemap has incorrect `source` attr for civet file in `node_modules` package
As discussed on Discord..

The esbuild+unplugin generated sourcemap has an incorrect `source` attribute for a civet file imported from a `node_modules` package:
* Bad: `"../node_modules/civet-pkg/lib/node_modules/civet-pkg/lib/math.civet"`
* Should be: `"../node_modules/civet-pkg/lib/math.civet"`

As requested, I put together a demo repo: https://github.com/greghuc/civet-sourcemaps-bug

Check out the current source map in `build/index.js.map`. I also added a non-civet package as a dependency, so you can see how a vanilla js file appears in sourcemap `sources`

edemaine:
Fixes #1784

This PR makes the unplugin distinguish Civet modules resolved by the unplugin itself from `.civet` package exports resolved by esbuild (for libraries), avoiding duplicated package paths like `node_modules/civet-pkg/node_modules/civet-pkg/...` while preserving correct output-relative paths for project imports.

This seems to only affect esbuild. Other systems (based on the rollup API) seem to be happy with relative paths, and they get rewritten by the system itself. So I think the bug was introduced in #1726.

Includes esbuild and Vite regression coverage for package imports, nested project entries, and relative project imports.

### DIFF
--- a/source/unplugin/unplugin.civet
+++ b/source/unplugin/unplugin.civet
@@ -241,6 +241,7 @@ export rawPlugin := (options: PluginOptions = {}, meta: UnpluginContextMeta) =>
     enforce: 'pre'
 
     buildStart(): Promise<void>
+      pluginResolvedIds.clear()
       civetConfigPath .= options.config
       if civetConfigPath is undefined
         civetConfigPath = await findInDir process.cwd()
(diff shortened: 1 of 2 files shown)


edemaine:

STRd6 (approved):

Issue #1382, opened 2024-08-28
Non-identifier operators


bbrk24:
I've talked about this before but I couldn't find an existing issue about it.

Swift requires operators to consist of *non*-alphanumeric characters, so they aren't valid identifiers. This allows custom operators like `infix operator .*` for dot multiplication, `postfix operator %`/`postfix operator °` for unit conversion, and more. I've occasionally wanted this myself in Civet code, such as `×` for Cartesian product or `∘` for function composition.

Given the limitations of JavaScript, this would necessarily involve name mangling. As long as the mangling is consistent and depends only on the source symbol, though, this shouldn't be a problem and should even allow for importing/exporting such operators. (One such operation could be replacing illegal characters with something of the form <code>_u<em>XXXX</em></code>, like `△` becoming `_u25B3`.)

edemaine:
Somewhat related is operator overloading: https://github.com/DanielXMoore/Civet/discussions/1081
(This used to be an issue but moved to a discussion.)

Related to name mangling is #555's mention of the idea to allow emoji in all Civet identifiers. If we extended this to mathematical operators and much of Unicode space, this would let a user immediately define a lot of custom operators (including e.g. `°` and `△`) using the existing `operator` functionality.

I think we could extend Civet to allow defining new ASCII operators (that don't conflict with existing operators). Something like:

```js
operator (.*) (a, b)  // inner product
  a.x * b.x + a.y * b.y
```

We'd need to define a set of ASCII letters that we'd like to allow in such binary operators, then (similar to how custom operators work today) check for them and see if they're defined.

I *think* we could even allow overriding built-in binary operators, at least to a limited extent, if we checked for this *first* in the grammar. This would be nice for when we inevitably add more binary operators to Civet.

I see you've also maybe suggested custom prefix and postfix operators. That's in principle doable too...

bbrk24:
> We'd need to define a set of ASCII letters that we'd like to allow in such binary operators, then (similar to how custom operators work today) check for them and see if they're defined.

Swift's set is `./=-+!*%<>&|^?~`, for the record.

bbrk24:
Oops, this is #1034

PR #2036, opened 2026-05-08
Allow non-identifier Unicode codepoints as identifiers

STRd6:
## Summary

Identifiers can now mix standard JS identifier chars (`\p{ID_Start}` / `\p{ID_Continue}`, `_`, `$`, ZWNJ, ZWJ) with non-identifier Unicode codepoints (emoji, dingbats, math symbols not claimed as operators). Non-id-unicode codepoints are escaped to a JS-safe `u$<HEX>$` form (mirrors the `\u{XXXX}` escape syntax) so the resulting name is a valid JS identifier:

- `△` → `u$25B3$`
- `😊` → `u$1F60A$`  *(astral codepoint — surrogate pair handled correctly)*
- `😊😔` → `u$1F60A$u$1F614$`  *(multi-codepoint identifier)*
- `pre😊post` → `preu$1F60A$post`  *(mixed with ASCII)*

The trailing `$` terminates each codepoint so adjacent ASCII parts stay unambiguous. ID_Continue letters (`é`, `ő`, `ω`) are already valid in JS and pass through unchanged — `pokémon` and `Erdős` are unaffected.

This is the smallest viable foothold for the design space in:
- discussion #1034 — Unicode non-letter functions/operators
- #555 — Unicode operator support (mentions emoji-in-identifiers)
- #1382 — Non-identifier operators (Swift-style with name mangling)

Anchors a concrete implementation for further design discussion in those threads — not closing them.

## Design choices

- **Multi-codepoint identifiers** allowed (e.g. `😊😔`) — IdentifierName extended to admit non-id-unicode chars in both start and continue positions, mirroring how JS already admits `\p{ID_Start}` / `\p{ID_Continue}`.
- **Mangling: `u$<HEX>$`** by codepoint (not UTF-16 code unit), no zero-padding. Trailing `$` terminates the codepoint so codepoints can sit adjacent to ASCII id chars without ambiguity. Surrogate pairs handled via codepoint-iterating `for…of` and `u`-flag regex.
- **`escapeIdentifier`** lives in `source/parser/identifier.civet`.
- **Operator-table wins**: codepoints already claimed by the operator/punctuation grammar (`÷—‖’•‥…⁇→⇒∈∉∋∌≔≠≡≢≣≤≥≪≫⋙▷⧺⩵⩶`) are excluded from the non-id-unicode char class, so `a ≤ b` still parses as `a <= b`. A drift-guard test scans `parser.hera` and asserts every non-ASCII grammar codepoint is in the operator-exclusion list — future operator additions can't silently regress to identifiers.
- **`ApplicationStart` unchanged**: implicit application still requires whitespace. `∀N` is one identifier (`u$2200$N`), not `∀(N)`. To call, write `∀ N`.
- **`\p{ID_Continue}` precision**: standalone combining marks (`̀`), non-ASCII digits (`٠`), and modifier letters reject at identifier start (since they're in ID_Continue but not ID_Start, and excluded from non-id-unicode).
- **`ReservedWord` keywords sorted longest-first** within each alternation so prefix conflicts (`instanceof` / `interface` before `in`) match correctly, letting the boundary check delegate to a `NonIdContinue` rule call instead of an inline `(?!...)` per regex.
- **Hand-written `u$25B3$` keeps working** as a normal `$`-prefixed identifier. We accept that hand-written and escape-mangled forms collide; demangling in hover/stack traces is best-effort.

## Deferred follow-ups

- LSP / stack-trace demangling so users see `△` instead of `u$25B3$`.

## Test plan
- [x] New test cases in `test/identifier.civet`: BMP symbol, astral emoji, multi-codepoint, ASCII-prefix-with-unicode, unicode-prefix-with-ASCII, ASCII-surrounding-unicode, reference, hand-written escape form, ID_Continue letters (`pokémon` / `Erdős`), ∀-as-function with space, operator codepoints stay as operators, adjoined-without-whitespace as one identifier, combining-mark / non-ASCII-digit rejection, drift-guard.
- [x] Full test suite: 3886 passing.
- [x] Coverage gate at 100% statements / branches / functions / lines.
- [x] Typecheck within baseline (888).
- [x] Smoke test: `△`, `😊`, `😊😔`, `pre😊post`, `pokémon`, `Erdős`, `imports`, `import.meta.url`, `a ≤ b`, `∀ N` all behave correctly.

### DIFF
--- a/civet.dev/reference.md
+++ b/civet.dev/reference.md
@@ -1070,6 +1070,43 @@ N := [1, 2, 3]
 ∀ N
 </Playground>
 
+#### Property keys
+
+When a non-id-unicode identifier appears as a *property key* — in an
+object literal, a destructure, a class member, an `@`-shorthand, or
+a member access — the source name is the emitted property name (so
+JS lookup hits the right key) while the variable side uses the
+mangled form. Member access switches from dot to bracket, since
+`obj.😊` isn't valid JS.
+
+<Playground>
+{ 😊 } := data
+x := { 😊 }
+x := { 😊: 1 }
+data.😊
+data?.😊
+class Foo
+  😊 = 1
+  😊() 2
+function foo(@😊)
+  @😊
+</Playground>
+
+#### Imports and exports
+
+Module-facing names (the side passed to / from another module) use the
+source form; the local-variable side keeps the mangled form. The
+combined `export X := value` declaration splits into a regular
+declaration plus an explicit re-export, since JS doesn't allow
+string-named exports inside `export const`.
+
+<Playground>
+import { 😊 } from "mod"
+import { 😊 as x } from "mod"
+export { 😊 }
+export 😊 := 3
+</Playground>
+
 ## Functions
 
 ### Function Calls
--- a/civet.dev/reference.md
+++ b/civet.dev/reference.md
@@ -1037,6 +1037,39 @@ Here is a table of all currently supported:
 
 </div>
 
+### Unicode Identifiers
+
+Identifiers can include non-identifier Unicode codepoints (emoji, dingbats,
+math symbols not in the [Unicode Operators](#unicode-operators) table)
+anywhere in the name, mixed freely with regular identifier characters.
+Each non-id-unicode codepoint is escaped to `u$<HEX>$` (mirroring JS's
+`\u{XXXX}` escape syntax) so the compiled name is a valid JS identifier.
+
+<Playground>
+let △ = 3
+let 😊 = 1
+let pre😊post = 2
+let 😊😔 = 4
+</Playground>
+
+Codepoints already in `\p{ID_Continue}` (letters with diacritics like `é`,
+`ő`, `ω`; digits; combining marks) are valid JS identifier chars and pass
+through unchanged.
+
+<Playground>
+let pokémon = "Pikachu"
+let Erdős = 1
+</Playground>
+
+Implicit function application still requires whitespace, so `∀N` is one
+identifier — write `∀ N` to call.
+
+<Playground>
+function ∀(arr) arr.every (x) => !!x
+N := [1, 2, 3]
+∀ N
+</Playground>
+
 ## Functions
 
 ### Function Calls
--- a/source/parser.hera
+++ b/source/parser.hera
@@ -8386,13 +8386,17 @@ EnumDeclaration
           ["let ", id, " = {};\n"],
           ...block.properties.map((property, i) => {
             let init, isString
+            // Use the source string for the property key so the runtime
+            // enum is keyed by the user-visible name (mangled would be
+            // unreachable from `E.😊` style access).
+            propKey := property.name.unicode ?? property.name.name
             if property.initializer
               // Replace references to other enum members.
               // TS further restricts this to past enum members,
               // but we don't need to enforce that here.
               init = replaceNodes(deepCopy(property.initializer),
                 (n) => n.type is "Identifier" and names.has(n.name),
-                (n) => [id, '["', n.name, '"]'])
+                (n) => [id, '["', n.unicode ?? n.name, '"]'])
               value := init[init# - 1]
               isString = value.type is "TemplateLiteral" or
                 (value.type is "Literal" and value.subtype is "StringLiteral")
@@ -8400,17 +8404,18 @@ EnumDeclaration
               // Default initializer is previous property + 1, or 0 if first.
               // TS further restricts this to when previous property is
               // constant, but we don't need to enforce that here.
+              prev := block.properties[i-1]?.name
               init = i is 0 ? " = 0" :
-                [" = ", id, '["', block.properties[i-1].name, '"] + 1']
+                [" = ", id, '["', prev?.unicode ?? prev?.name, '"] + 1']
             // String enums do not get back references
             if isString
               return [
-                id, '["', property.name, '"]', init, ";\n",
+                id, '["', propKey, '"]', init, ";\n",
               ]
             else
               return [
-                id, "[", id, '["', property.name, '"]', init,
-                '] = "', property.name, '";\n',
+                id, "[", id, '["', propKey, '"]', init,
+                '] = "', propKey, '";\n',
               ]
           }),
         ],
@@ -8456,7 +8461,10 @@ EnumProperty
       type: "EnumProperty",
       name,
       initializer,
-      children: $0,
+      // Unicode member: emit `"name" = …` so the TS-form enum has a
+      // string-literal member; runtime emission (in EnumDeclaration) also
+      // uses the source string for the property key.
+      children: [propertyKey(name), ...$0.slice(1)],
     }
 
 TypeProperty
--- a/source/parser.hera
+++ b/source/parser.hera
@@ -1529,9 +1529,11 @@ ThisLiteral
   # NOTE: Added @identifier shorthand, also works for private identifiers
   # Converts 'IdentifierName' node to string so this won't interfere with refs
   AtThis:at $( Hash? IdentifierName ):id ->
-    // `@😊` → `this["😊"]` (matches the binding-side bracket form).
+    // `@😊` → `this["😊"]` (matches the binding-side bracket form), but
+    // private `@#😊` keeps dot access since private fields can't be
+    // string-keyed — emit `this.#u$1F60A$` instead.
     escaped := escapeIdentifier(id)
-    if escaped !== id
+    if escaped !== id and not id.startsWith("#")
       return {
         type: "MemberExpression",
         children: [at, {
@@ -1550,13 +1552,13 @@ ThisLiteral
       type: "MemberExpression",
       children: [at, {
         type: "PropertyAccess",
-        name: id,
+        name: escaped,
         children: [".", {
           $loc: {
             pos: $loc.pos + 1,
             length: $loc# - 1,
           },
-          token: id
+          token: escaped
         }],
       }],
       thisShorthand: true,
@@ -2501,7 +2503,9 @@ BindingProperty
       type: "BindingProperty",
       children,
       name: binding,
-      value: undefined,
+      // Setting value to the binding when expanded keeps the AST shape
+      // consistent (downstream code distinguishes shorthand by `value: undefined`).
+      value: binding.unicode ? binding : undefined,
       typeSuffix,
       initializer,
       names: binding.names,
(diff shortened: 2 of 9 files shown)


edemaine:
My main concern here is whether we want to allow multi-character Unicode strings to be used in an identifier. For example, maybe I want`😊😔` to give the happySad index.

This would mean we'd have to add spaces back for function applications, but maybe that's OK? `∀ x` also looks pretty good... 

An in-between solution would be that identifiers need to be normal IdStart/IdContinue OR entirely non-id Unicode. So `∀x` is a function application  but `😊😔` is an identifier.

Overall I'm not sure which of these three is best. My worry though is that, once we make this decision, we can't easily switch. Currently I'm leaning toward "identifiers can include any Unicode symbols/emoji" as a simple rule... But it may not be so simple.

STRd6:
> "identifiers can include any Unicode symbols/emoji" as a simple rule... But it may not be so simple.

This seems like a good rule, we can explore farther down that path before making a decision.

edemaine:
I guess a fourth option would be to make this a config option. Then the golfers could get their one-character symbols :) (though not sure they could afford `"civet unicodeSingle"`)

edemaine:
Thinking about this some more:

1. Single-Unicode-character design (like current PR) is like APL, where spaces between special symbols is optional. So you get some potentially crazy/cool/terrible/golfy syntax. On the other hand, the modern evolution J requires spaces, perhaps because the special symbols got mapped into multi-ASCII-character sequences.
2. Multiple-Unicode-character design definitely offers fun flexibility, like `◥▅◤` or `◕‿◕✿`. It might also be the only clean way to support emoji like `👨‍👩‍👧‍👦` or `👍🏾` that involve many UTF-8 characters.
3. ASCII + Unicode character design has the advantages of Option 2 and also feels most consistent with the design of JavaScript which *already allows* identifiers like `Erdős` or `pokémon` (Unicode letters are already part of `ID_Start`/`ID_Continue`).

So I'm leaning toward Option 3, which is the one you seemed to like above. I could still see Option 1 as a compiler flag (and maybe Option 2, but perhaps that's too many options). We could even call it `"civet apl"`, although perhaps that should be reserved for when we support all of the APL built-ins. 😉 

Another possible issue with the current PR is that `AfterNonAscii` seems to treat "all non-ASCII characters" as "Unicode symbols". This *might* break existing support for JavaScript's identifiers `Erdős` and `pokémon` which end with a non-ASCII character — although perhaps not given the greedy nature of matching.

In the future, I look forward to supporting something like `¯\_(ツ)_/¯` — I think the idea would be that a bunch of ASCII symbols can be treated as an identifier, when they don't already have a meaning like `--` or `++` or `>=`. But that can be left for a future extension.

STRd6:
@edemaine Just pushed and updated the PR description to add unicode code points directly into Identifiers with minimal other changes.

edemaine:
@greptileai review again please?

source/parser.hera
```diff
@@ -1490,7 +1492,7 @@ FieldDefinition
         return {
           type: "FieldDefinition",
           id,
-          children: [id, " = ", exp],
+          children: [propertyKey(id), " = ", exp],
         }
```
edemaine, line 1497: Should this be done inside `ClassElementName`?

source/parser.hera
```diff
@@ -1931,6 +1951,18 @@ PropertyAccess
     }
 
   AccessStart:dot InlineComment*:comments ( IdentifierName / PrivateIdentifier / LengthShorthand ):id ->
+    // Escaped Unicode property `data.😊` becomes `data["😊"]` since the
+    // mangled `u$1F60A$` is a valid JS identifier but not the actual property name.
+    if id.unicode
```
edemaine: I don't understand this (also above). Isn't it the same to use property access? Seems like we should stay closer to the source.

source/parser.hera
```diff
@@ -2218,7 +2250,11 @@ AtIdentifierRef
   ReservedWord:r ->
     return makeRef(`_${r}`, r)
   IdentifierName:id ->
-    return makeRef(id.name)
+    ref := makeRef(id.name)
+    // Carry the source string so binding-side `this.X = X` can emit
+    // bracket access matching the expression-side `this["X"]`.
+    if id.unicode then ref.unicode = id.unicode
+    return ref
```
edemaine, line 2257: Should this be part of `makeRef`?

source/parser.hera
```diff
@@ -2403,7 +2439,7 @@ BindingProperty
       }
     return {
       type: "BindingProperty",
-      children: [ws1, name, ws2, colon, ws3, value, initializer],  // omit typeSuffix
+      children: [ws1, propertyKey(name), ws2, colon, ws3, value, initializer],  // omit typeSuffix
       name,
```
edemaine, line 2443: Should this be in `PropertyName`?

civet.dev/reference.md
```diff
@@ -1037,6 +1037,39 @@ Here is a table of all currently supported:
...
+
+Codepoints already in `\p{ID_Continue}` (letters with diacritics like `é`,
+`ő`, `ω`; digits; combining marks) are valid JS identifier chars and pass
+through unchanged.
+
+<Playground>
+let pokémon = "Pikachu"
+let Erdős = 1
+</Playground>
+
```
edemaine, line 1063: Should explain how property access works. It took me a while to realize the difference. Maybe also import/export demo.

test/identifier.civet
```diff
@@ -20,3 +22,330 @@ describe "identifier", ->
...
+      let u$1F60A$happy = 3
+    """
+
+    testCase """
+      ASCII surrounding unicode
+      ---
+      let pre😊post = 3
+      ---
+      let preu$1F60A$post = 3
+    """
```
edemaine, line 82: Should test Unicode characters with many UTF-16 symbols (family, skin color modifiers)

I wonder if we need to canonicalize, because there are multiple ways to express the same character (I think).

source/parser/identifier.civet
```diff
@@ -0,0 +1,47 @@
+// Escape non-identifier Unicode codepoints to `u$<HEX>$` so the result is
+// a valid JS identifier. The trailing `$` terminates each codepoint so
+// `pre😊post` → `preu$1F60A$post` is unambiguous. `\p{ID_Continue}` chars
+// (letters / digits / combining marks like `é`, `ő`) pass through unchanged.
+function escapeIdentifier(raw: string): string
+  result .= ""
+  for char of raw
+    cp := char.codePointAt(0)!
+    if cp > 0x7F and not /\p{ID_Continue}/u.test(char)
```
edemaine, line 12: Does this correctly handle `ID_Start` at beginning? Extra complicated because currently `raw` may start with `#`.

source/parser/identifier.civet
```diff
@@ -0,0 +1,47 @@
...
+      result += `u$${cp.toString(16).toUpperCase()}$`
+    else
+      result += char
+  result
+
+import { gatherRecursive } from ./traversal.civet
+import type { Identifier } from ./types.civet
```
edemaine: Move imports to top of file?

source/parser.hera
```diff
@@ -2459,6 +2496,11 @@ BindingProperty
         names: [],
       }
 
+    // Escaped Unicode binding shorthand `{ 😊 }` becomes `{ "😊": u$1F60A$ }`
+    // so the lookup uses the user-visible property name.
+    if binding.unicode
+      children = [ws, propertyKey(binding), ": ", binding, initializer]
```
edemaine, line 2500: Shouldn't we change `value` too?

source/parser.hera
```diff
@@ -3277,20 +3319,28 @@ SymbolElement
   SymbolLiteral -> [ "[", $1, "]" ]
 
 # https://262.ecma-international.org/#prod-Identifier
+# Extended: identifiers may include non-identifier Unicode (emoji, math
+# symbols not claimed as operators); escapeIdentifier converts them to a
+# JS-valid form.
 Identifier
   # perf: assertion to exit early
-  /(?=\p{ID_Start}|[_$])/ !ReservedWord IdentifierName:id -> id
+  /(?=\p{ID_Start}|[_$]|[^\u0000-\u007F\p{ID_Continue}\s\u00F7\u2014\u2016\u2019\u2022\u2025\u2026\u2047\u2192\u21D2\u2208\u2209\u220B\u220C\u2254\u2260\u2261\u2262\u2263\u2264\u2265\u226A\u226B\u22D9\u25B7\u29FA\u2A75\u2A76])/ !ReservedWord IdentifierName:id -> id
```
edemaine: Are we OK with this set increasing as we add more Unicode operators? Alternatively, we might aim to allow identifiers to contain the symbols, so long as they're longer than one character? I guess this would prevent tight use of Unicode operators though.

source/parser.hera
```diff
@@ -3277,20 +3319,28 @@ SymbolElement
...
 
 # https://262.ecma-international.org/#prod-IdentifierName
 IdentifierName
-  /(?:\p{ID_Start}|[_$])(?:\p{ID_Continue}|[\u200C\u200D$])*/ ->
+  /(?:\p{ID_Start}|[_$]|[^\u0000-\u007F\p{ID_Continue}\s\u00F7\u2014\u2016\u2019\u2022\u2025\u2026\u2047\u2192\u21D2\u2208\u2209\u220B\u220C\u2254\u2260\u2261\u2262\u2263\u2264\u2265\u226A\u226B\u22D9\u25B7\u29FA\u2A75\u2A76])(?:\p{ID_Continue}|[\u200C\u200D$]|[^\u0000-\u007F\p{ID_Continue}\s\u00F7\u2014\u2016\u2019\u2022\u2025\u2026\u2047\u2192\u21D2\u2208\u2209\u220B\u220C\u2254\u2260\u2261\u2262\u2263\u2264\u2265\u226A\u226B\u22D9\u25B7\u29FA\u2A75\u2A76])*/ ->
```
edemaine: Hera feature request to support repeated patterns in regular expressions. Maybe `///` with substitutions?

source/parser.hera
```diff
@@ -6878,7 +6974,7 @@ SemicolonDelimiter
     }
 
 NonIdContinue
-  /(?!\p{ID_Continue})/ -> undefined
+  /(?!\p{ID_Continue}|[^\u0000-\u007F\p{ID_Continue}\s÷—‖’•‥…⁇→⇒∈∉∋∌≔≠≡≢≣≤≥≪≫⋙▷⧺⩵⩶])/ -> undefined
```
edemaine, line 6985: Looks nicer in Unicode - should we use this instead of the long `\u` sequences above?

edemaine:
I'm glad you fixed the import/export Unicode naming because I was otherwise going to complain that property access non-mangling is treated too differently from identifiers. 😸  Now it's looking great!

PR #2038, opened 2026-05-09
Dev: Update @danielx/hera 0.9.2 → 0.9.7

STRd6:
## Summary

- Bumps `@danielx/hera` to 0.9.7 so TS gets real return types instead of `MaybeResult<any>`.
- Adds `::Type` annotations on parser.hera rules (`Literal`, `NumericLiteral`, `StringLiteral`, `TypeArguments`, `TypeTuple`, `Keyword`, and many more) and a `Keyword` type in `types.civet` to break inference cycles and flow specific shapes through the grammar.
- Tightens types in `parser/` helpers (`lib`, `util`, `function`, `auto-dec`, `binding`, `pipe`, `op`, `for`, `block`) and tests; assertion non-null where the grammar guarantees presence.
- Sets `CI_TYPECHECK_MAX_ERRORS` baseline reflecting the new hera-exposed errors.


source/parser.hera
```diff
@@ -2224,37 +2224,37 @@ PinPattern
...
-  Caret SingleLineExpressionWithIndentedApplicationForbidden:expression ->
+  Caret SingleLineExpressionWithIndentedApplicationForbidden:expression ::unknown ->
     return {
       type: "PinPattern",
       children: [expression],
       expression,
       typeSuffix: typeSuffixForExpression(expression),
     }
-  ActualMemberExpression:expression ->
+  ActualMemberExpression:expression ::unknown ->
```
edemaine: A bunch of the unknowns could probably be actual types (like `PinPattern`). But can be dealt with in the future.

edemaine (approved):

source/parser.hera
STRd6: Yes, definitely! I tried previously a few times to annotate the parser more heavily but it Claude and I always ended up spiraling out of control with it. I'm trying to do this one much more incrementally.

STRd6:
I think we can merge this in since it includes the Hera bump. The Hera bump exposes many more errors in `parser.hera` but they were previously hidden by being `any`. It's also fixed many errors. We should be able to continue with smaller PRs now that we have the better typecheck CI integration as well.

source/parser/function.civet
```diff
@@ -1134,7 +1135,7 @@ function processParams(f: FunctionNode): void
     prefix.push makeNode {
       type: "Declaration"
       children: ["const ", subbindings[1..]]
-      names: subbindings.flatMap .names ?? []
+      names: subbindings.flatMap((s: any) => s.names)
```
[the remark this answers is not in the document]
edemaine: I think this might be worth putting back in, for safety... but up to you

.github/workflows/build.yml
```diff
@@ -11,7 +11,7 @@ permissions:
 
 env:
   API_TOKEN: ${{ secrets.API_TOKEN }}
-  CI_TYPECHECK_MAX_ERRORS: 826
+  CI_TYPECHECK_MAX_ERRORS: 956
```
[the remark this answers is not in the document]
edemaine: New count could be reduced to 941. I'm impressed you got it down to +115!

scripts/audit-node.civet
```diff
@@ -0,0 +1,195 @@
+#!/usr/bin/env civet
+/**
+ * audit-node.civet
+ *
+ * Print producer + consumer sites for a given AST node type, with the
+ * `children: [...]` literal extracted at every producer so children-layout
+ * drift across the tree is visible at a glance.
```
edemaine, line 7: Cool!

source/parser/auto-dec.civet
```diff
@@ -31,25 +33,25 @@ function findDecs(statements: ASTNode)
...
     .filter .type is "Declaration"
     .flatMap .names
     new Set declarationNames
 
   declaredIdentifiers .= findVarDecs statements
 
-  function hasDec(name)
-    declaredIdentifiers.has(name) or scopes.some .has name
+  function hasDec(name: string): boolean
+    declaredIdentifiers.has(name) or scopes.some((s) => s.has name)
```
edemaine: Unnecessary removal of Civet syntax.

source/parser/function.civet
```diff
@@ -1200,7 +1201,7 @@ function processSignature(f: FunctionNode): void
       addGenerator = true
     else
       for each y of gatherRecursiveWithinFunction block, .type is "YieldExpression"
-        i := y.children.findIndex .type is "Yield"
+        i := y.children.findIndex((c: any) => c.type is "Yield")
```
edemaine: Can use `(&: any).type is "Yield"`

source/parser/types.civet
```diff
@@ -227,6 +229,86 @@ export type CommentNode =
   parent?: Parent
   children?: never
 
+// A pre-parsed source keyword.  Distinct from raw strings the grammar
```
edemaine: `/**` comments would be nicer for type comments

source/parser.hera
```diff
@@ -1079,25 +1129,27 @@ PipelineExpressionBodySameLine
...
     }
-  NWTypePostfix TypePostfix* ->
+  NWTypePostfix TypePostfix* ::ASTNode ->
     return makeAmpersandFunction({
       body: [" ", $1, ...$2],
     })
-  ForbidNewlineBinaryOp PipelineHeadItem? RestoreNewlineBinaryOp ->
-    return $skip unless $2
+  ForbidNewlineBinaryOp PipelineHeadItem? RestoreNewlineBinaryOp ::ASTNode ->
+    return $skip as never unless $2
```
edemaine: I thought Hera now types `$skip` as `never`, so these casts can be removed.

source/parser.hera
```diff
@@ -3074,21 +3126,21 @@ BracedThenClause
 
 # A block that must include braces (function body, try/catch/finally)
 BracedOrEmptyBlock
-  BracedBlock
-  EmptyBlock
+  BracedBlock ::BlockStatement
+  EmptyBlock ::BlockStatement
```
edemaine, line 3130: Feature request for Hera: allow `::Type` on the root, as in `BracedOrEmptyBlock ::BlockStatement`

source/parser.hera
```diff
@@ -7154,27 +7204,27 @@ Comma
 
 Comptime
   "comptime" NonIdContinue !":" ->
-    return { $loc, token: $1 }
+    return { type: "Keyword" as const, $loc, token: $1 }
```
edemaine, line 7207: I wonder if we need an `as const` shorthand in Civet...

That said, I wonder if these would all be cleaner via `::Keyword`?

source/parser/function.civet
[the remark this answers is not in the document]
STRd6: I'm more in favor of offensive `asserts` than defensive conditionals.

Long term being aggressive with asserts and fixing the types at the root should lead to simpler code in the end state that is easier to cover with tests because it has many fewer conditionals and is explicit about the assumptions. That plus our large test suite should give us a good level of confidence.

Though I also acknowledge that this strategy lead to some bugs in the 0.11.8 release but they were fixed quickly thanks to the loud erroring.

notes/ast-cleanup.md
```diff
@@ -0,0 +1,100 @@
+# AST Cleanup Plan
```
edemaine, line 1: Still in use? Should it be in NOTES or omitted?

edemaine (approved):
Looking good! A few small comments. Also, GPT caught these:

Reviewed PR #2038 at head `017b844`.

- **[P2] `skipIfOnlyWS` no longer strips empty token leaves**: `source/parser/util.civet:772` now uses `isToken(target)`, but `isToken` requires a truthy `node.token`, so `{ token: "" }` is no longer treated as whitespace. The grammar intentionally creates empty tokens, so this can preserve nodes that were previously dropped. Use the old `target.token != null` style or adjust `isToken` for this case.
- **[P2] Pipeline `import` tail builds an invalid `Identifier`**: `source/parser.hera:1143` sets `name: $1` and `names: [$1]`, but `Import` returns a `Keyword` node, not a string. `Identifier.name` / `names` should contain `"import"` or this should avoid claiming to be a full `Identifier`.

source/parser.hera
STRd6: Emoji possibilities beckon.

source/parser.hera
STRd6: It does but that hasn't released yet.

STRd6:
`isToken({token: ""})` returns `true` — Civet's `?` postfix compiles to `!= null`, not a truthy check. Empty-token nodes still reach the `.trim() === ''` branch and get stripped as before.

PR #2040, opened 2026-05-11
Lift `&` past `for` body in `when` clause

### LINKED ISSUE #1788: `&` placeholder inside `when` clause puts `continue` inside arrow function
Civet code:
```
for some i in[0..&] when 0==&%i
```
Expected:
```js
($) => {
  let results = false;
  for (const i in range(0, $ + 1)) {
    if (!(0 == $ % i)) continue;
    results = true;
    break;
  }
  return results;
}
```
Actual:
```js
($) => {
  let results = false;
  for (const i in range(0, $ + 1)) {
    ($1) => {
      if (!(0 == $1 % i)) continue;
      return;
    };
    results = true;
    break;
  }
  return results;
};
```
This results in a syntax error as the `continue` is not considered to be inside the loop. Even if it did run, the inner arrow function is never evaluated, so this becomes equivalent to `& >= 0`.

STRd6:
## Summary
- `for x of arr when &.test(x)` was wrapping the synthetic `if (!…) continue` in an arrow function, putting `continue` outside the loop and producing a syntax error.
- The arrow came from the `&` placeholder walker treating the for-body `BlockStatement` as the nearest lift target. Mark the synthetic `IfStatement` with `implicitLift: true` so the walker skips the for-body and lifts to the enclosing scope — same place an `&` in the for clause already lifts to.
- Fixes #1788.

## Test plan
- [x] New regression test in `test/for.civet` ("when with & placeholder")
- [x] `pnpm test` — 3925 passing
- [x] `pnpm typecheck --max-errors 888` — within baseline
- [x] Manually verified the issue's exact input produces the expected output

### DIFF
--- a/source/parser.hera
+++ b/source/parser.hera
@@ -5112,6 +5112,10 @@ ForStatementControlWithWhen
           type: "IfStatement",
           then: block,
           children: ["if (!", makeLeftHandSideExpression(trimFirstSpace(condition)), ") ", block],
+          // & placeholders in the condition came from outside the for-body,
+          // so they should lift past it (otherwise `continue` ends up trapped
+          // inside an arrow function — issue #1788).
+          implicitLift: true,
         }, ";"]
       ],
     }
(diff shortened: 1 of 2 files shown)


STRd6:
:+1:

edemaine (approved):

PR #2041, opened 2026-05-11
Unicode escape sequences `\uXXXX` in identifiers

### LINKED ISSUE #2039: Unicode escape sequences are valid in variable names in JS, they should be valid in Civet
```javascript
let \u0061 = 10;
```

Currently this valid JS fails to parse in Civet.

@claude Implement a fix.

Commit and push to branch at every step.

STRd6:
Per ECMAScript, identifiers may include \uXXXX and \u{X+} escape sequences. Extend the IdentifierName regex to accept them and decode in escapeIdentifier so the escape and direct-character spellings resolve to the same binding.

Fixes #2039

### DIFF
--- a/source/parser.hera
+++ b/source/parser.hera
@@ -1531,9 +1531,10 @@ ThisLiteral
   AtThis:at $( Hash? IdentifierName ):id ->
     // `@😊` → `this["😊"]` (matches the binding-side bracket form), but
     // private `@#😊` keeps dot access since private fields can't be
-    // string-keyed — emit `this.#u$1F60A$` instead.
-    escaped := escapeIdentifier(id)
-    if escaped !== id and not id.startsWith("#")
+    // string-keyed — emit `this.#u$1F60A$` instead. IdentifierName's
+    // own action already filtered out-of-range `\u` escapes via $skip.
+    escaped := escapeIdentifier(id)!
+    if escaped.token !== id and not id.startsWith("#")
       return {
         type: "MemberExpression",
         children: [at, {
@@ -1552,13 +1553,13 @@ ThisLiteral
       type: "MemberExpression",
       children: [at, {
         type: "PropertyAccess",
-        name: escaped,
+        name: escaped.name,
         children: [".", {
           $loc: {
             pos: $loc.pos + 1,
             length: $loc# - 1,
           },
-          token: escaped
+          token: escaped.token
         }],
       }],
       thisShorthand: true,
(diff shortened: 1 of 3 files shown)


source/parser.hera
```diff
@@ -3324,11 +3324,11 @@ SymbolElement
 # JS-valid form.
 Identifier
   # perf: assertion to exit early
-  /(?=\p{ID_Start}|[_$]|[^\u0000-\u007F\p{ID_Continue}\s÷—‖’•‥…⁇→⇒∈∉∋∌≔≠≡≢≣≤≥≪≫⋙▷⧺⩵⩶])/ !ReservedWord IdentifierName:id -> id
+  /(?=\p{ID_Start}|[_$\\]|[^\u0000-\u007F\p{ID_Continue}\s÷—‖’•‥…⁇→⇒∈∉∋∌≔≠≡≢≣≤≥≪≫⋙▷⧺⩵⩶])/ !ReservedWord IdentifierName:id -> id
```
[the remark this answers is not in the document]
STRd6: Delegating to JS to handle this.

STRd6:
:+1:

edemaine:

source/parser/identifier.civet
```diff
@@ -1,19 +1,31 @@
...
+// Non-id-continue codepoints (emoji, non-operator math symbols) mangle to
+// `u$<HEX>$` in both — trailing `$` keeps `pre😊post` → `preu$1F60A$post`
+// unambiguous. `\uXXXX` / `\u{X+}` escapes decode for `name` (so `a`
+// binds to `a`); `token` keeps them verbatim when they spell a valid id
+// char (JS treats `a` ≡ `a`), mangling otherwise.
+// IdentifierName bounds `\u{X}` to ≤ 0x10FFFF, so `fromCodePoint` is safe.
+function escapeIdentifier(raw: string): { name: string, token: string }
+  name .= ""
+  token .= ""
+  for m of raw.matchAll /\\u\{([\dA-Fa-f]+)\}|\\u([\dA-Fa-f]{4})|[\s\S]/gu
```
edemaine: Took me a while to realize `[\s\S]` is everything. Could we use `.` instead? I assume identifiers can't have newlines. If they can, `[^]` is a more common way to write "everything".

source/parser/identifier.civet
```diff
@@ -1,19 +1,31 @@
...
+    [unit, brace, simple] := m
+    cp := if brace? or simple?
+      parseInt (brace ?? simple)!, 16
+    else
+      unit.codePointAt(0)!
+    char := String.fromCodePoint cp
     if cp > 0x7F and not /\p{ID_Continue}/u.test(char)
-      result += `u$${cp.toString(16).toUpperCase()}$`
+      mangled := `u$${cp.toString(16).toUpperCase()}$`
```
edemaine, line 22: I didn't realize before: this isn't normalized to four characters. For possible decoding, should we `padStart('0', 4)`? I think an example where this would occur is `©` (U+00A9). (All of U+0XXX block seems to be within ID_Continue.)

edemaine (approved):

source/parser/identifier.civet
STRd6: JS pads it: `console.log("\u{A9}")`

edemaine: I meant the Unicode character case:
https://civet.dev/playground?code=wqk%3D

STRd6: I think it's fine having them implicitly padded. We never decode them and if we did the implicit padding is fine.

STRd6: `String.fromCharCode(parseInt("A9", 16))` => `'©'`

STRd6: And the whole range of valid sequences is 2-6 hex characters.

edemaine: I was more thinking about distinguishing between `©A9` (U+A9 followed by "A9") and `ꦩ` (U+A9A9).
I know Unicode hex characters can go up to 6 hex characters, but `charCodeAt` only gives 4 hex characters at a time (I think via UTF-16). `\uXXXX` is similarly limited. So I still think it would be natural and somewhat helpful to pad. I can imagine decoding `$uXXXX` in identifiers for debugging purposes.

STRd6: The encoding is `©` => `u$A9$` so `©A9` => `u$A9$A9`. The trailing `$` delimits the ending of the hex run (it's different from the very first proposed iteration which lacked the trailing delimiter).

`😊` => `u$1F60A$` has 5 hex. Padding to 4 wouldn't cover 5-6, but handling 2-6 without padding and using `codePointAt` avoids the `charCodeAt` and UTF-16 pitfalls.

edemaine: Oops, I missed the trailing `$`! 🤦 I agree, that solves everything.

PR #2042, opened 2026-05-11
Honor autoConst/autoLet for `export x = …`, allow `export` destructuring

### LINKED ISSUE #1321: `autoConst` doesn’t work with destructuring
For example:
~~~ coffee
"civet autoConst"

str = "string"

export {length} = str # does not produce export const {length} = str
~~~

STRd6:
Fixes #1321.

## Cause

Two separate bugs under `"civet autoConst"`:

- `export {a} = obj` failed to parse — `TypeAndNamedExports` matched `{a}` as a re-export clause before `ExportVarDec` could try the initializer form.
- `export func = =>` parsed but always emitted `export var func = …`, ignoring autoConst/autoLet entirely (only the `ExportVarDec` shorthand was affected; `export x := …` already worked).

## Approach

- `TypeAndNamedExports` gets a `!( __ Equals )` lookahead so `export {a} = obj` falls through to `ExportVarDec`. The only other caller (`ExportFromClause`) is always followed by `from`, so the constraint is safe.
- `ExportVarDec` reads `config.autoConst` / `config.autoLet` and inserts `const ` / `let ` / `var ` accordingly, and tags the declaration with `autoExport: true`.
- In `auto-dec.civet`, when an `autoExport` declaration's names are already declared in scope, `splitAutoExport` rewrites the statement to `<assign>; export { names }`. This avoids the TDZ / redeclaration errors that would otherwise result from `let a = 1; export let a = 2`. Unicode-name handling reuses `buildExportSplitClause`.

The two previously-disabled tests in `auto-const.civet` and `auto-let.civet` (which documented the buggy `export var` shortcut) are updated to assert the new correct output, with separate tests added for the split path.

## Test plan

- [x] `export {a} = obj` and `export func = =>` produce `export const …` under autoConst
- [x] Same forms produce `export let …` under autoLet
- [x] No autoConst/autoLet: `export var {a, b} = obj` (previously a parse error) now compiles
- [x] `let a = 1; export a = 2` (autoConst) → `let a = 1; a = 2; export { a }` via split
- [x] Existing `export {a}`, `export {a} from "…"`, `export a, b`, unicode `export 🦀 := 3` all unchanged

### DIFF
--- a/source/parser.hera
+++ b/source/parser.hera
@@ -6374,7 +6374,8 @@ ExportDeclaration
   # NOTE: Declaration and VariableStatement should come before NamedExports
   # so that NamedExports doesn't grab function, async, type, var, etc.
   # NOTE: TS decorators come before `export` keyword but TC39 stage 3 decorators proposal come after
-  Decorators?:dec Export __ ( Declaration / VariableStatement / TypeAndNamedExports / ExportVarDec ):declaration ->
+  # NOTE: NamedExportsNotAssigned defers `export {a} = obj` to ExportVarDec (#1321)
+  Decorators?:dec Export __ ( Declaration / VariableStatement / NamedExportsNotAssigned / ExportVarDec ):declaration ->
     // `export 😊 := 3` → `const u$…$ = 3; export { u$…$ as "😊" }`
     // (JS forbids string-named exports inside `export const`).
     splitClause := buildExportSplitClause(declaration)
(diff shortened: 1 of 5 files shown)


STRd6:
@claude make sure newly added code doesn't increase type errors https://github.com/DanielXMoore/Civet/actions/runs/25684911814/job/75405946149?pr=2042

STRd6:
:+1: Reviewed (GH doesn't let me add a review to my own PR so I'm adding my own comment when reviewing Claude co-authored PRs for tracking)

edemaine (approved):

PR #2044, opened 2026-05-12
Monaco and browser LSP support; LSP in the Playground

edemaine:
This PR adds browser-side Civet LSP support and wires it into the docs Playground through Monaco. The full-page Playground now supports richer editor behavior (type errors, completions, definitions, etc.) while keeping the existing textarea fallback and compile/output flow.

Major changes:

- New `@danielx/civet-monaco` package: Monaco helper package
  - Registers the Civet language with Monaco
  - Provides Monarch/TextMate-friendly token setup
  - Adapts Civet LSP responses for Monaco semantic tokens, hover, definitions, completions, and completion resolve
  - README documents usage, also linked from civet.dev/integrations

- Browser LSP package entry points
  - Split the LSP server into shared `startServer` logic plus explicit runtime entries
  - `@danielx/civet-language-server/node` entry point replaces old server
  - New `@danielx/civet-language-server/browser` entry point for browser appliactions
  - New `@danielx/civet-language-server/worker` entry point to launch and manage browser Worker that runs above
    - Queues messages until the server is ready
    - Supports diagnostics, semantic tokens, hover, definition, completion, completion resolve, change notifications, request rejection, dispose, and termination behavior
  - Default package entry is conditional: Node resolves to the Node server, browser builds resolve to the browser server
  - Executable has changed names from `dist/server.js` to `dist/node.js`, leaving room for other runtimes in the future

- Playground full-page editor (civet.dev/playground)
  - Async loads Monaco which replaces the textarea fallback once ready
  - Monaco uses the browser Civet LSP worker
  - Defaults to ES2025 lib, but adds DOM lib if requested; would be easy to extend to more libs (Node perhaps?)
  - Shows LSP diagnostics as Monaco markers
  - Allow toggling TypeScript diagnostics to skip annoying red squiggles (but Civet compile errors always render)
  - If Civet gives non-fatal "Error" nodes, display them all (like in VSCode) and also show compile output on right
  - Fun output cursor/highlight mapping from Monaco cursor/hover positions to generated code
    - Made somewhat annoying because Prettier doesn't provide a sourcemap, but a fast character diff seems to suffice

- Source map utilities
  - Export source mapping types
  - Move `forwardMap` for source-to-generated position mapping from LSP to ts-diagnostic
    - Now used in Playground output cursor mapping

- TS service fixes (which came up in type checking)
  - Preserve original script filename casing while still canonicalizing lookup keys, avoiding TypeScript complaining about multiple uses of a filename with different cases
  - Fix inherited `rootDir` handling from extended tsconfigs
  - Allow for `system` override instead of forcing `ts.sys` everywhere
  - Add browser-specific TS service wrapper
  - Add virtual/open-file lookup behavior for browser-style resolution

- Browser shims and shared utilities
  - Add a small `assert-shim`
  - Expand browser shim tests
  - Clean up path normalization helpers
  - Export shared source mapping/util types where they are used across packages

- Build, docs, and package updates
  - Add `lsp/monaco` to the workspace
  - Update docs and integration notes for the Monaco/LSP setup
  - Update LSP README with the new package entries
  - Copy TypeScript lib files needed by browser/Node LSP builds
  - Avoid including the Monaco TypeScript worker from the docs build

- Coverage and CI
  - Add Monaco tests and coverage
  - Add browser worker LSP tests
  - Add TS service browser/path/config coverage
  - Add worker request/error/dispose coverage
  - Include Monaco tests in CI alongside the other LSP checks
  - Fix merged coverage handling for browser-only/server-only files and Windows sourcemap paths
  - Keep the merged coverage gate at 100%

## Payload Sizes

Measured from `pnpm docs:build`. Sizes are raw / gzip / brotli.

| Asset group | Raw | gzip | brotli |
|---|---:|---:|---:|
| Monaco site payload: editor core + editor worker + codicon font | 3983.4 KiB | 1065.2 KiB | 797.0 KiB |
| `@danielx/civet-monaco` helper package dist | 15.7 KiB | 3.7 KiB | 3.3 KiB |
| Civet Playground worker wrapper + Civet compiler asset | 978.9 KiB | 141.9 KiB | 109.9 KiB |
| Civet browser compiler asset alone, `browser.min.js` | 973.7 KiB | 140.2 KiB | 108.4 KiB |
| Civet browser LSP worker, `lsp/server/dist/browser.js` | 10832.6 KiB | 1715.9 KiB | 1228.2 KiB |
| Optional DOM lib, `lib.dom.d.ts` | 2294.4 KiB | 346.9 KiB | 251.9 KiB |

Notes:

- The Monaco editor payload is about 4x the raw size of the current Civet Playground worker+compiler path, and about 7.5x gzip / brotli.
- Compared with the Civet browser LSP worker, the Monaco editor payload is smaller: about 37% of the raw size, 65% of gzip / brotli.

BREAKING CHANGE: Neovim (and other direct users of the LSP server) need to use `dist/node.js` instead of `dist/server.js`


lsp/monaco/source/monaco.civet
```diff
@@ -0,0 +1,553 @@
...
+      ]
+      tripleSingleString: [
+        [/'''/, 'string', '@pop']
+        [/./, 'string']
+      ]
+      bracketCounting: [
+        [/\{/, 'delimiter.bracket', '@bracketCounting']
+        [/\}/, 'delimiter.bracket', '@pop']
+        include: 'root'
+      ]
```
STRd6, line 128: Should triple backtick be in here too?

source/config.civet
```diff
@@ -26,6 +26,7 @@ export function findInDir(dirPath: string): Promise<string | undefined>
   pathFor := (name: string) => path.join dirPath, name
 
   // Check for config file in .config directory
+  /* c8 ignore next -- stat-failure branch is covered by symlink tests where the platform permits symlinks. */
```
STRd6: Is this due to windows differences?

lsp/server/integration/test.civet
```diff
@@ -1,19 +1,21 @@
...
-    }, (args) ->
+    }, (args: OnLoadArgs) ->
       return readFile args.path, 'utf8'
-      .then (source) ->
+      .then (source: string) ->
         filename := path.relative(process.cwd(), args.path)
         return
           contents: compile source, { filename, js: true }
-      .catch (e) ->
+      .catch (e: Error) ->
```
STRd6: `e` is likely `unknown` here

lsp/server/source/browser.civet
```diff
@@ -0,0 +1,71 @@
...
+  BrowserMessageWriter
+  createConnection
+  ProposedFeatures
+} from vscode-languageserver/browser
+{ startServer } from ./server.civet
+{ browserDefaultTsConfig } from ./lib/defaults.civet
+{ createBrowserTSService } from ./lib/typescript-service-browser.civet
+{ createSystem } from @typescript/vfs
+path from node:path
+// @ts-ignore Virtual TypeScript lib files provided by esbuild at build time; see build/build.civet
```
STRd6: I generally prefer `@ts-expect-error` over `@ts-ignore`

STRd6 (approved):
Thanks! Looks like a valuable addition to the docs site. I just added a few small comments.

source/config.civet
edemaine: Yes, we [skip the symlink tests on Windows](https://github.com/DanielXMoore/Civet/blob/6e7e6caed718855f09462ffc953293d13382b044/test/infra/config.civet#L73) because symlinks require Administrator privileges on Windows. 😿  So this was to get coverage 100% when I test locally.

But I found a better way: add a stub for `fs.stat` that simulates an `EACCES` error. Then even Windows can exercise the `catch` branch. So now we don't need the `c8 ignore`, even for Windows testing. Thanks for the nudge. 😸

edemaine:
Thanks! I'm hoping this change highlights one of the major benefits of Civet, which is its support for TypeScript tooling.

PR #2054, opened 2026-05-13
Fix: unwrap if/then/else assignment to ternary without ref

### LINKED ISSUE #2051: Unnecessary ref use in if / then conditional expression
```civet
k : string := "a"
key := if k == " " then "Space" else k


k : string := "a"
key := (if k == " " then "Space" else k)
```

```typescript
const k : string = "a"
let ref;if (k == " ") { ref = "Space"} else ref = k;const key =ref


const k : string = "a"
const key = ((k == " "? "Space" : k))
```
Second version with parens correctly unwraps to ternary without ref. First version also should.

@claude TDD it

STRd6:
Closes #2051.

When the initializer of a const/let declaration is a bare if/then/else expression with simple-expression branches, skip the `let ref` hoisting in `prependStatementExpressionBlock` and let `processStatementExpressions` convert the StatementExpression to an `IfExpression` (ternary) — matching the behavior of the parenthesized form.

### DIFF
--- a/source/parser/lib.civet
+++ b/source/parser/lib.civet
@@ -364,9 +364,8 @@ function canExpressionizeIfAsTernary(statement: IfStatement): boolean
 // Mirror of `expressionizeBlock`'s success condition — keep in sync.
 function canExpressionizeBlock(blockOrExpression: ASTNode): boolean
   if { expressions } := blockOrExpression
-    for [, s] of expressions
-      return false unless isExpression(s)
-  true
+    for every [, s] of expressions
+      isExpression(s)
 
 function expressionizeIfStatement(statement: IfStatement): ASTNode
   { condition, then: b, else: e } := statement
--- a/source/parser/lib.civet
+++ b/source/parser/lib.civet
@@ -352,10 +352,9 @@ True iff `expressionizeIfStatement` would produce a clean ternary
 */
 function canExpressionizeIfAsTernary(statement: IfStatement): boolean
   { then: b, else: e } := statement
-  return false unless canExpressionizeBlock b
-  if e
-    return false unless canExpressionizeBlock e.block
-  true
+  (and)
+    canExpressionizeBlock b
+    e ? canExpressionizeBlock e.block : true
 
 function canExpressionizeBlock(blockOrExpression: ASTNode): boolean
   if { expressions } := blockOrExpression
--- a/source/parser/lib.civet
+++ b/source/parser/lib.civet
@@ -346,6 +346,23 @@ function expressionizeBlock(blockOrExpression: ASTNode)
   else
     blockOrExpression
 
+/**
+True iff `expressionizeIfStatement` would produce a clean ternary
+(`IfExpression`) for this statement, rather than wrapping it in an IIFE.
+*/
+function canExpressionizeIfAsTernary(statement: IfStatement): boolean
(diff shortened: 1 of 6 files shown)


source/parser/lib.civet
```diff
@@ -346,6 +346,23 @@ function expressionizeBlock(blockOrExpression: ASTNode)
   else
     blockOrExpression
 
+/**
+True iff `expressionizeIfStatement` would produce a clean ternary
+(`IfExpression`) for this statement, rather than wrapping it in an IIFE.
+*/
+function canExpressionizeIfAsTernary(statement: IfStatement): boolean
+  { then: b, else: e } := statement
+  return false unless canExpressionizeBlock b
```
STRd6: @claude use implicit return with nested `(and)` form like this:
```
(and)
  canExpressionizeBlock b
  e ? canExpressionizeBlock e : true
```

STRd6: May need more parens

test/if.civet
```diff
@@ -977,20 +965,40 @@ describe "if", ->
...
+    // #2051
+    testCase """
+      if/then/else expression unwraps to ternary same as parenthesized
+      ---
+      key1 := if k == " " then "Space" else k
+      key2 := (if k == " " then "Space" else k)
+      ---
+      const key1 = (k == " "? "Space" : k)
+      const key2 = ((k == " "? "Space" : k))
     """
```
edemaine, line 1002: Should add tests for when `then` or `else` clause cannot be expressionized; or do those tests already exist?

source/parser/lib.civet
```diff
@@ -346,6 +346,28 @@ function expressionizeBlock(blockOrExpression: ASTNode)
...
+  (and)
+    canExpressionizeBlock b
+    e ? canExpressionizeBlock e.block : true
+
+// Mirror of `expressionizeBlock`'s success condition — keep in sync.
+function canExpressionizeBlock(blockOrExpression: ASTNode): boolean
+  if { expressions } := blockOrExpression
+    for [, s] of expressions
+      return false unless isExpression(s)
+  true
```
edemaine:
```suggestion
    for every [, s] of expressions
      isExpression(s)
```

source/parser/lib.civet
```diff
@@ -346,6 +346,28 @@ function expressionizeBlock(blockOrExpression: ASTNode)
...
+Mirrors `expressionizeIfStatement` + `expressionizeBlock` truthiness: must
+return false for the same inputs that would force the IIFE fallback. If
+`expressionizeBlock` gains a new early-return condition, update
+`canExpressionizeBlock` to match.
+*/
+function canExpressionizeIfAsTernary(statement: IfStatement): boolean
+  { then: b, else: e } := statement
+  (and)
+    canExpressionizeBlock b
+    e ? canExpressionizeBlock e.block : true
```
edemaine, line 362: I guess an alternative would be to call `expressionizeIfStatement` and detect an IIFE wrapper in the output. But I'm not sure that's cleaner — detection is probably a bit messy.

Another alternative would be to add a `useIIFE` parameter to `expressionizeIfStatement` that says whether to use `wrapIIFE` (default) or just return `undefined`.

edemaine (approved):
Nice, this does seem to make for cleaner output when possible.

PR #2055, opened 2026-05-13
Fix Unicode property names in `@`-bind, `::` prototype access, `:` symbol shorthand

### LINKED ISSUE #2052: Inconsistency of emoji identifiers with non-`.` access
These examples throw TypeErrors at runtime. I would expect them to print `3`.

```
🧑 := { 🍊() { return 3 } }
f := 🧑@🍊
console.log f()
```

```
'civet coffeePrototype'

class 🧑 { 🍊() { return 3 } }
f := 🧑::🍊
console.log f()
```

A workaround for now is to use quoted bracket access, e.g.
```
'civet coffeePrototype'

class 🧑 { 🍊() { return 3 } }
f := 🧑::["🍊"]
console.log f()
```

STRd6:
Fixes #2052

`x@🍊` and `x::🍊` used the mangled binding name (`u$1F34A$`) as a JS property, which doesn't match the user-visible key the property was stored under. They now emit bracket access with the source name so that `{ 🍊() ... }@🍊` and `class { 🍊() ... }::🍊` resolve at runtime, matching how `data.🍊` already compiled to `data["🍊"]`.

Non-null `x!@🍊` drops the dot (`x!["🍊"]`) since TS only pairs `.[` with optional `?.`; optional `x?@🍊` keeps it as `x?.["🍊"]`.

### DIFF
--- a/source/parser.hera
+++ b/source/parser.hera
@@ -2003,11 +2003,15 @@ PropertyAccess
       optional: modifier?.token === "?",
     }
     if id
+      // `x::😊` → `x.prototype["😊"]`: unicode names need bracket access.
+      access := id.unicode
+        ? [ "prototype[", propertyKey(id), "]" ]
+        : [ "prototype.", id ]
       return {
         type: "PropertyAccess",
         name: id.name,
         dot: start,
-        children: [ start, "prototype.", id ],
+        children: [ start, ...access ],
       }
     else
       return {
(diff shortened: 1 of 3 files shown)


STRd6:
@claude

Also fix symbol shorthand:

```civet
:🍊
```

Should become:

```typescript
Symbol.for("🍊")
```

STRd6:
@claude address greptile feedback

PR #2060, opened 2026-05-14
Coverage: eliminate flakiness from parallel mocha + dual-compile

STRd6:
## Summary

Recent CI runs (and local `pnpm coverage` runs) have been failing the 100% coverage gate non-deterministically — different files missing each run, percentages bouncing between ~99.56% and ~99.99%. Two independent races were tangling together:

### 1. Parallel mocha + v8 coverage race
The existing `v8.takeCoverage()` in mocha's `afterAll` flushes coverage data up to that point, but mocha's IPC-disconnect of parallel workers can preempt subsequent libuv writes — hits generated during mocha's own teardown, reporters, etc. are dropped. Direct evidence: `unplugin.civet` measured 87.6% under parallel vs. 100% under sequential, with the same source and same tests.

This matches a documented industry-wide issue ([vitest #2591](https://github.com/vitest-dev/vitest/pull/2591), [jest #15360](https://github.com/jestjs/jest/issues/15360), [mochista](https://github.com/laggingreflex/mochista), [nodejs/node #46378](https://github.com/nodejs/node/issues/46378)) — the consistent fix across projects is to serialize when collecting coverage.

**Fix:** `build/test.sh` forces `threads=0` when `CIVET_COVERAGE=1`, and bumps mocha's default timeout to 10s (sequential mode loses the parallelism that some subprocess-spawning tests rely on to stay under 2s).

### 2. Dual-compilation of `source/ts-service`
`test/infra/cache.civet` was importing factories from `../../source/ts-service/index.civet` directly. That triggered root mocha's `.civet` require-extensions hook to compile a second copy of those files alongside the esbuild-bundled `@danielx/civet/ts-service` that lsp/server tests (and production consumers via unplugin) load. v8 emitted separate coverage records for each compiled JS, both source-mapping back to the same `.civet` files. c8's merge couldn't reconcile the two fnMaps cleanly:

- For `service.civet`: root's compiled output had a function range (82–153) that lsp/server's bundle didn't. The union added a phantom 32nd "TSService" with 0 hits, producing `functions 31/32` even though every function was actually covered.
- For `hera.civet`'s `getHeraVersion`: root had 5 hits on the function, lsp/server had 0; c8 took lsp/server's fnMap as authoritative and discarded root's hits as misaligned. Under parallel mode the race occasionally let root's data win the merge; under sequential it never did.

**Fix:** Switch the import in `test/infra/cache.civet` to `@danielx/civet/ts-service` (bundle path). Now both suites load the same compiled bytes — single fnMap, no merge artifacts.

### Genuine uncovered code (separate from the artifacts)
With the races removed, two real coverage gaps surfaced that the parallel-mode flake had intermittently masked:
- `source/browser.civet` lines 49–56: browser-only bootstrap inside `unless window <? "undefined"`. Node tests can't reach it. `c8 ignore start/stop` with rationale.
- `source/ts-service/plugins/civet.civet:92`: defensive `catch` for a corrupt cache entry. `c8 ignore next 2` per the existing project convention for defensive arms.

### Verification
Three back-to-back `pnpm coverage` runs all hit `Coverage gate passed: 100% statements / branches / functions / lines.` deterministically.

## Test plan

- [x] `pnpm coverage` passes the gate 3× in a row on this branch
- [x] `pnpm test` still passes (no behavior change in the test suites themselves)
- [x] No production code touched outside of two narrowly-scoped c8-ignore directives


build/test.sh
```diff
@@ -24,11 +33,23 @@ export CIVET_THREADS=
...
 if [[ "$OSTYPE" == msys* || "$OSTYPE" == cygwin* || "$OSTYPE" == win* ]]; then
   args="--timeout 5000 $args"
 fi
 
+# Sequential mode + c8 instrumentation slows subprocess-spawning tests
+# (comptime, cli, etc.) past the default 2s.  This appears AFTER the Windows
+# block so the higher timeout ends up later in the arg list and wins.
+if [ "${CIVET_COVERAGE:-0}" = "1" ]; then
+  args="$args --timeout 10000"
+fi
```
edemaine, line 51: Moving Windows into `elif` here (so only one `--timeout` argument) would be cleaner.

PR #2062, opened 2026-05-14
diag(coverage): re-enable parallel + log SIGTERM'd workers

STRd6:
## Purpose

**Diagnostic-only.** Empirically validate the workerpool-SIGTERM hypothesis from `MOCHA-PARALLEL-COVERAGE-RACE.md` — that the flaky 99.x% coverage gate is caused by parallel mocha workers occasionally taking longer than workerpool's 1 s default `workerTerminateTimeout` to drain, getting SIGTERM'd by the parent, and silently losing their V8 coverage data.

Not for merge. Revert once we have data.

## What the branch does

- Reverts the `--no-parallel`/`threads=0` workarounds from #2060 so coverage runs in parallel again.
- Adds `build/sigterm-trace.js`, a `--require` shim gated on `MOCHA_WORKER_ID`. On `SIGTERM` it appends the worker PID to `/tmp/civet-sigterm-trace.log`, then `SIGKILL`s itself — preserving the coverage-loss symptom so we can correlate "log has PIDs" with "coverage gate failed".
- Wires the shim into root `.mocharc.json` and `lsp/server/.mocharc.json`.
- CI workflow: clears the log before `pnpm coverage`, marks the coverage step `continue-on-error: true`, and adds a dedicated **SIGTERM trace** step that `exit 1`s when the log has any PIDs.

## What outcomes mean

| `pnpm coverage` | SIGTERM trace step | Interpretation |
|---|---|---|
| pass | pass (no PIDs) | This run didn't trigger the bug; rerun. |
| fail | **fail (PIDs logged)** | **Hypothesis confirmed.** |
| pass | fail (PIDs logged) | SIGTERMs happen but didn't cause measurable coverage loss this run. Still confirms the mechanism. |
| fail | pass (no PIDs) | Hypothesis wrong; bug is somewhere else. |

Plan: rerun a few times to collect data, then decide on the upstream fix shape (e.g. bumping `workerTerminateTimeout` via the new mocha option we drafted).

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -65,25 +65,38 @@ jobs:
           pnpm install --frozen-lockfile
           pnpm build
 
-      - name: Test (with coverage)
+      - name: Test (with coverage, forced-low worker timeout to PROVE the bug)
         if: runner.os == 'Linux'
-        # Allow coverage gate failure here — the SIGTERM-trace check runs
-        # next regardless, and is the diagnostic signal we're after.
+        # Allow coverage gate failure here — the trace check runs next.
+        # CIVET_WORKER_TERMINATE_TIMEOUT=1 means workers have 1 ms to
+        # respond to the IPC terminate before the parent SIGTERMs them.
+        # If the hypothesis is right, this run should leave SIGTERM entries
+        # in the trace log AND tank coverage.
+        env:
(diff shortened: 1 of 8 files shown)


STRd6:
`mocha --parallel` + v8 coverage wasn't actually flaky. It was cli calling `register` which imported civet files with a different compiler and the source maps collided causing an apparent loss of coverage. Fixed in #2064

PR #2064, opened 2026-05-14
Coverage: re-enable `--parallel` for test runs

STRd6:
## Summary

Reverts the `--no-parallel` half of PR #2060.  The flaky 99.x% gate that motivated that workaround no longer reproduces on the current toolchain — see `notes/mocha-parallel-coverage-diagnosis.md` for the full investigation.

Empirical basis: **35 consecutive parallel coverage runs (5 CI + 30 local), zero failures, zero SIGTERMs.** Rule-of-three puts the upper bound on the per-run failure rate at ~9% at 95% CI, far below the rate that historically tripped #2052/#2055/#2057.

Kept intact from PR #2060: the dual-compile fix in `test/infra/cache.civet` (the bundled-import change).  Reverting that locally and rerunning 5 more times also produced no failures, so it's not load-bearing either — but it's harmless belt-and-suspenders and the comment block is useful future context.

## What changed

- `build/test.sh` — drop the `CIVET_COVERAGE=1 ⇒ threads=0` override.
- `lsp/server/package.json` — drop `--no-parallel` from `test:coverage`.
- `notes/mocha-parallel-coverage-race.md` — original symptom + hypothesis writeup, with a pointer to the follow-up.
- `notes/mocha-parallel-coverage-diagnosis.md` — investigation results: reproducer matrix, mocha-source findings, the empirical disproof of the SIGTERM theory, and the resolution recommendation.

## Expected impact

About **3 minutes of CI time** reclaimed per coverage run.

## Test plan

- [ ] Coverage gate passes on CI
- [ ] Watch the next several main builds; if a flake ever returns, the diagnosis doc has the cheap playbook for re-confirming or refining the hypothesis


notes/mocha-parallel-coverage-diagnosis.md
```diff
@@ -0,0 +1,258 @@
+# Mocha parallel + V8 coverage race — diagnosis update
+
+Follow-up to `mocha-parallel-coverage-race.md` (same folder).  That
```
edemaine: These are the first note files to be checked in. I don't think they're needed in the repo... though nice to have in the PR description. But if you like them we could keep them.

edemaine (approved):

PR #2066, opened 2026-05-16
Sourcemap: walk back in `remapPosition` when line has no mapping

STRd6:
## Summary

- `remapPosition` now walks back through preceding lines when the queried TS line has no usable 4-tuple mapping (standard sourcemap "nearest preceding mapping wins" semantics, like `source-map`'s `originalPositionFor`).
- Existing "no preceding mapping anywhere → return position as-is" behavior is preserved.

## Why

The Hera → Civet → TS pipeline composes two sourcemaps.  Stage 1 (Hera → Civet) is sparse — only ~17% of generated Civet lines carry a 4-tuple back to `.hera` source.  After composition with the dense stage-2 (Civet → TS) map, the final TS → `.hera` map inherits stage 1's sparsity: ~82% of generated TS lines have no source position.

When TS reported a diagnostic on one of those 82% of lines, the old `remapPosition` walked the entries on that line, found no 4-tuple, and fell through to `return position` — handing back a raw TS line number.  On `source/parser.hera` (9,696 lines, compiles to a ~32,000-line .tsx), most diagnostics ended up thousands of lines past EOF.  LSP clients either crashed (e.g. `mcp-language-server` panicked on `lines[10035]` when reading `.hera` of length 9696) or pointed at garbage lines.

After the fix, every diagnostic lands on a valid `.hera` line — the nearest enclosing rule.  Character offset doesn't interpolate across line boundaries; the source column is the mapping's anchor.

## Test plan

- [ ] `pnpm build`
- [ ] `pnpm test` — 3,992 passing locally (6 new tests for walk-back paths)
- [ ] LSP `.hera` diagnostics return valid positions

### DIFF
--- a/source/ts-diagnostic.civet
+++ b/source/ts-diagnostic.civet
@@ -21,7 +21,16 @@ export type SourcemapLines = SourceMap['lines']
  * Take a position in generated code and map it into a position in source code.
  * Reverse mapping.
  *
- * Return position as-is if no sourcemap is available.
+ * When the queried line has no usable 4-tuple mapping, walks back through
+ * preceding lines until one is found — matching the standard sourcemap
+ * "nearest preceding mapping wins" behavior (e.g. source-map's
+ * `originalPositionFor`).  Required for chained / sparse maps like the
+ * Hera→Civet→TS pipeline, where only ~18% of generated TS lines have a
+ * direct source mapping; without walk-back, the other 82% return raw TS
+ * line numbers and land far past EOF in the .hera source.
+ *
+ * Return position as-is if no sourcemap is available or no mapping exists
+ * anywhere at or before the queried position.
  */
 export function remapPosition(
   position: Position,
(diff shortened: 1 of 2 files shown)


edemaine:

source/ts-diagnostic.civet
```diff
@@ -32,41 +41,55 @@ export function remapPosition(
...
+      // (e.g. `import` keyword) can produce negative interpolation
+      return
+        line: srcLine
+        character: Math.max(0, newChar)
+
+  // No usable mapping on the queried line — walk back through preceding
+  // lines for the nearest 4-tuple.  Character offset does not interpolate
+  // across line boundaries; the source column is the mapping's anchor.
+  let l = Math.min(line - 1, sourcemapLines.length - 1)
+  while l >= 0
```
edemaine: `for l of [Math.min line, sourcemapLines# > .. 0]`

source/ts-diagnostic.civet
```diff
@@ -32,41 +41,55 @@ export function remapPosition(
...
+    prevLine := sourcemapLines[l]
+    if prevLine?.length
+      lastMapping .= undefined
+      for mapping of prevLine
+        if mapping.length === 4
+          lastMapping = mapping
```
edemaine: Reverse order would be faster... something like
```js
lastMapping :=
  for first i of [prevLine# > .. 0] when mapping[i]# is 4
    mapping[i]
```

edemaine (commented):
Minor: This is going to conflict with #2070

PR #2069, opened 2026-05-16
LSP handlers for `signatureHelp`, `prepareRename`, `documentHighlight`, `typeDefinition`, `implementation`, `foldingRange`, `selectionRange`, `linkedEditingRange`

STRd6:
## Summary

Adds eight LSP handlers that delegate to existing TS language service methods, each following the same sourcemap-remapping pattern as the existing handlers in `lsp/server/source/server.civet`.

| LSP method | TS service method |
|---|---|
| `textDocument/signatureHelp` | `getSignatureHelpItems` |
| `textDocument/prepareRename` | `getRenameInfo` |
| `textDocument/documentHighlight` | `getDocumentHighlights` |
| `textDocument/typeDefinition` | `getTypeDefinitionAtPosition` |
| `textDocument/implementation` | `getImplementationAtPosition` |
| `textDocument/foldingRange` | `getOutliningSpans` |
| `textDocument/selectionRange` | `getSmartSelectionRange` |
| `textDocument/linkedEditingRange` | `getLinkedEditingRangeAtPosition` |

User-visible payoff: parameter tooltips inside calls, smart selection expansion, go-to-type/implementation, document-highlight on identifiers, code folding, prepare-rename validation, and JSX linked-edit ranges — all features that editors expect by default.

## Notes

- `linkedEditingRange` registration is guarded with `if connection.languages?.onLinkedEditingRange` because the lightweight connection used by some unit tests doesn't expose `connection.languages`. The capability is advertised regardless.
- `renameProvider` capability changes from `true` to `{ prepareProvider: true }` to enable `prepareRename`. Existing rename tests still pass.
- For `typeDefinition` and `implementation`, the shared `locationLookup` helper avoids the literal copy-paste that `onDefinition`/`onReferences` use today; if you'd prefer those two stay inlined for symmetry, easy to revert.

## Test plan

- [x] `pnpm test` under `lsp/server` — 268 passing (260 existing + 8 new)
- [ ] Manual smoke in VS Code with the Civet extension built against this branch (signature help on a Civet function call, fold a class body, jump to interface implementation)
- [ ] CI green

### DIFF
--- a/lsp/server/source/server.civet
+++ b/lsp/server/source/server.civet
@@ -28,7 +28,8 @@
   type FoldingRange
   type SelectionRange
   type LinkedEditingRanges
-  type DefinitionLink
+  type SignatureHelpContext
+  SignatureHelpTriggerKind
   type Range
 } from vscode-languageserver
 
@@ -93,6 +94,24 @@ semanticTokenModifiers := [
   "defaultLibrary" // 4
   "local"          // 5
 ] as const
+
+// Map LSP SignatureHelpContext → TS SignatureHelpItemsOptions so that
+// retriggers and character-typed invocations preserve activeSignature
+// instead of resetting to 0 on every overload cycle / cursor move.
+function toTsSignatureHelpOptions(
+  context: SignatureHelpContext | undefined
+): ts.SignatureHelpItemsOptions | undefined
+  return undefined unless context
+  if context.isRetrigger
+    triggerReason:
+      kind: 'retrigger'
+      triggerCharacter: context.triggerCharacter as ts.SignatureHelpRetriggerCharacter | undefined
+  else if context.triggerKind is SignatureHelpTriggerKind.TriggerCharacter and context.triggerCharacter
+    triggerReason:
+      kind: 'characterTyped'
+      triggerCharacter: context.triggerCharacter as ts.SignatureHelpTriggerCharacter
+  else
+    triggerReason: { kind: 'invoked' }
 
 export interface ServerDependencies
   TSService: typeof import("./lib/typescript-service.civet").default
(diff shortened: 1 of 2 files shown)


edemaine:
Do any of these need to get explicitly wired into the vscode plugin? I imagine they do for Monaco.

STRd6:
Good question — checked both:

**VS Code**: no wiring needed. `vscode-languageclient/node` auto-negotiates standard LSP capabilities from `serverCapabilities`, so the extension just constructs a `LanguageClient` and the new requests route through transparently.

**Monaco**: you're right that it needs explicit `register*Provider` calls. The wrapper at `lsp/monaco/source/monaco.civet` currently only ships providers for hover, definition, completion, and semantic tokens — none of the eight new features have a Monaco-side adapter, so they won't surface in Monaco-hosted editors using this wrapper.

I'd rather keep that work in its own PR — it roughly doubles the wrapper's surface area and the adapter pattern (client interfaces, type conversions, dispose semantics, coverage tests) deserves its own focused review independent of the server-side delegation here. Filed #2073 to track it.

Also addressed the three Greptile P2s in 2a80c48:
- Dropped the unused `DefinitionLink` import.
- Forwarded `SignatureHelpContext` through to TS as `SignatureHelpItemsOptions` so retriggers and character-typed invocations no longer reset `activeSignature` on every overload cycle.
- Mapped `HighlightSpanKind.definition` (function/class/type/interface declaration name sites) to `DocumentHighlightKind.Write` instead of falling through to `Text`. Added a test on a function declaration to exercise it and dropped the c8-ignore.

edemaine:

edemaine (commented):
In addition to the latest Greptile comment, GPT found this:

- [P2] `server.civet:248`: `signatureHelpProvider.triggerCharacters` omits `" "`, so LSP clients will not auto-trigger signature help for Civet’s common whitespace-call form, e.g. `greet "world"`. I verified the handler can answer a manual `textDocument/signatureHelp` request at that position, so this is a registration/trigger mapping gap. Add space as a trigger, and avoid casting non-TS trigger characters directly into `ts.SignatureHelpTriggerCharacter`; treat space-triggered requests as `invoked` or otherwise sanitize before calling TS.

STRd6:
Addressed both open items in 6edc096e:

- **[Greptile P1] selectionRange alignment** — `onSelectionRanges` now pushes a fallback unit range `{ start: pos, end: pos }` when `getSmartSelectionRange` throws, keeping the response array index-aligned with `positions` per the LSP spec. Multi-position test added with a deliberately out-of-range entry to assert the contract.
- **[GPT P2] signatureHelp space trigger** — added `' '` to `signatureHelpProvider.triggerCharacters` for Civet's whitespace-call form. `toTsSignatureHelpOptions` now validates the trigger char against TS's allowed set (`(`, `,`, `<`, plus `)` for retrigger) and falls through to `{ kind: 'invoked' }` for anything else, so we never cast a non-TS char into `ts.SignatureHelpTriggerCharacter`. Test contexts extended to 6 entries covering all branches.

Coverage gate still 100%.

PR #2070, opened 2026-05-16
LSP: skip semantic tokens in unmapped TS prelude code

### LINKED ISSUE #2067: LSP: phantom `type` semantic token lands on Civet bareword imports
The Civet LSP server emits single-char `type` (and sometimes `typeParameter`) semantic tokens that land on the trailing letter of bareword import identifiers like `assert from assert`, but only when the file is large/complex enough.

Observed via `textDocument/semanticTokens/full` on `test/cli.civet`: the server returns a `type` token at `(line:0, char:5, len:1)` and a `typeParameter` token at `(line:0, char:17, len:1)` on line `assert from assert`. Both are degenerate single-char ranges on the trailing `t` of each `assert`. There is nothing typed on this line — it's a pure value import.

### Reproduction — failing test

In `lsp/server/test/lsp-features.test.civet`, inside the shared-project describe block:

```civet
it "semantic tokens don't emit phantom type tokens on Civet bareword imports", ->
  // The bug reproduces against the real test/cli.civet but resists
  // shrinking to a small synthetic fixture — it depends on the cumulative
  // structure of the file. Bisecting confirmed the minimum-known
  // triggering subset is lines 1-17 + 793-794 + 800-833 of cli.civet; any
  // smaller slice or any synthetic fixture we tried (with `<?`, `as
  // const`, regex literals, type-only imports, etc.) does not trigger.
  // Load the real file so this test guards regressions even if the
  // underlying trigger remains mysterious.
  testDir := path.dirname new URL(import.meta.url).pathname
  cliPath := path.resolve testDir, "../../../test/cli.civet"
  fs2 := await import "node:fs/promises"
  text := await fs2.readFile cliPath, "utf8"

  uri := docUri path.join(projectRoot, "semtok-bareword-import.civet")
  client.notify "textDocument/didOpen", {
    textDocument: { uri, languageId: "civet", version: 1, text }
  }
  await waitForDiag(client, (p) => p.uri is uri)

  tokens := await client.request "textDocument/semanticTokens/full", {
    textDocument: { uri }
  }
  assert tokens, "expected semantic tokens response"

  legend := init.capabilities.semanticTokensProvider.legend.tokenTypes
  typeIdx := legend.indexOf "type"
  typeParamIdx := legend.indexOf "typeParameter"

  decoded: { line: number, character: number, length: number, tokenType: number }[] := []
  line .= 0
  char .= 0
  for i of [0...tokens.data.length] by 5
    deltaLine := tokens.data[i]
    deltaChar := tokens.data[i + 1]
    length := tokens.data[i + 2]
    tokenType := tokens.data[i + 3]
    if deltaLine > 0
      line += deltaLine
      char = deltaChar
    else
      char += deltaChar
    decoded.push { line, character: char, length, tokenType }

  // `assert from assert` is a Civet bareword import — pure value, no types.
  // Observed bug: server emits two single-char tokens on the import line,
  // one `type` and one `typeParameter`, landing on the trailing letter of
  // each `assert`.
  phantom := decoded.find (t) =>
    t.line is 0 and (t.tokenType is typeIdx or t.tokenType is typeParamIdx)
  assert.equal phantom, undefined,
    `expected no type/typeParameter tokens on bareword import line; got ${JSON.stringify phantom}`
```

Run with `pnpm test` in `lsp/server/`. Fails:

```
AssertionError [ERR_ASSERTION]: expected no type/typeParameter tokens on
  bareword import line; got {"line":0,"character":5,"length":1,"tokenType":5}
```

### Bisect notes

I tried to minimise the trigger and could not produce a small synthetic repro:

- `assert from assert` alone — passes.
- `assert from assert` + every Civet construct I tried in isolation (`<?`, `as const`, type-only imports, regex literals, async functions, `for…of`) — passes.
- Real `test/cli.civet` — fails.
- Minimum reproducing subset of `test/cli.civet`: lines 1-17 + 793-794 + 800-833 (1-indexed). Any smaller slice passes.
- The boundaries are non-monotonic — e.g. line 17 fails, 18 passes, 19 fails — suggesting a cumulative-threshold or sourcemap-budget issue rather than a single-construct trigger.

### Hypothesis

A TS-side `type` declaration (probably one Civet synthesises in the compiled output) sourcemaps back to a position that collapses to `(0, last-char-of-first-identifier)` when the sourcemap can't find a real source location. The pattern — single-char range landing on the trailing letter of an identifier — matches a "no source location → fallback to nearest mapped position" recovery.

STRd6:
Closes #2067

`remapPosition` falls back to identity when the generated line has no 4-tuple source mapping at or before the requested column. For compiler-inserted prelude lines (e.g. Civet's `type AutoPromise<T> = Promise<Awaited<T>>;` helper, emitted when an async function needs Promise-wrapping), TS still classifies the `AutoPromise` and `T` identifiers as `type` / `typeParameter` tokens, and that identity fallback collapses them onto the corresponding TS-line-0 column in the Civet source — landing on the trailing letter of a bareword-import identifier when one exists.

This PR introduces `remapPositionStrict`, which returns `undefined` when no mapping exists, and has the semantic-tokens path use it: skipping a token that has no source location is correct, whereas the identity fallback emits phantom tokens. Other callers (hover, definition, etc.) keep the existing best-effort fallback.

Includes the regression test from the issue, which fails at `main` and passes here.

### DIFF
--- a/lsp/server/source/lib/util.civet
+++ b/lsp/server/source/lib/util.civet
@@ -25,6 +25,7 @@ export {
   flattenDiagnosticMessageText
   forwardMap
   remapPosition
+  remapPositionStrict
   remapRange
   type SourceMapping
   type SourcemapLines
--- a/lsp/server/source/server.civet
+++ b/lsp/server/source/server.civet
@@ -27,7 +27,7 @@
   type Position
 } from vscode-languageserver-textdocument
 * as Previewer from ./lib/previewer.civet
-{ collectCommentTokens, convertNavTree, forwardMap, getCanonicalFileName, convertDiagnostic, getSymbolKind, remapPosition, logTiming, type WithResolvers, withResolvers, type SourcemapLines } from ./lib/util.civet
+{ collectCommentTokens, convertNavTree, forwardMap, getCanonicalFileName, convertDiagnostic, getSymbolKind, remapPosition, remapPositionStrict, logTiming, type WithResolvers, withResolvers, type SourcemapLines } from ./lib/util.civet
 { asPlainTextWithLinks, tagsToMarkdown } from ./lib/textRendering.civet
 {
   adjustCursorForEmptyQuotes
@@ -1295,9 +1295,17 @@ export function startServer(connection: Connection, dependencies: ServerDependen
         tokenModifiers := (encoded & 0xFF)
         continue if tokenType < 0 or tokenType >= semanticTokenTypes.length
 
-        // Map position from transpiled TS back to Civet source
+        // Map position from transpiled TS back to Civet source.  Use
+        // strict remapping when we have a sourcemap so that tokens in
+        // compiler-inserted prelude code (e.g. `type AutoPromise<T> = ...`)
+        // are skipped rather than collapsed onto unrelated source
+        // identifiers via the identity fallback.
         tsPos := tsDoc.positionAt(tsStart)
-        pos := remapPosition(tsPos, sourcemapLines)
+        pos := if sourcemapLines
+          remapPositionStrict(tsPos, sourcemapLines)
+        else
+          tsPos
+        continue unless pos
         // Skip tokens with invalid remapped positions
         continue if pos.line < 0 or pos.character < 0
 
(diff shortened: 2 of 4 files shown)


STRd6:
@claude

Extract a minimal self contained fixture.

Fix windows CI error:

```
  1) LSP features (shared project server)
       semantic tokens don't emit phantom type tokens on Civet bareword imports:
     Error: ENOENT: no such file or directory, open 'D:\D:\a\Civet\Civet\test\cli.civet'
      at async open (node:internal/fs/promises:640:25)
      at async Module.readFile (node:internal/fs/promises:1287:14)
      at async Context.<anonymous> (file:///D:/a/Civet/Civet/lsp/server/test/lsp-features.test.civet:216:18)
```

edemaine (approved):

PR #2074, opened 2026-05-17
Fix `x.literal` shorthand

### LINKED ISSUE #2065: Compiler crashes on `{a.1`
Input:

```
{a.1
```

[On playground](https://civet.dev/playground?code=e2EuMQ%3D%3D) (Civet 0.11.9)

Error:

```
Compiler crashed; please report as a compiler bug.TypeError: Cannot read properties of undefined (reading 'type')
    at needsRef (https://civet.dev/assets/browser.min.Dri6HQ9-.js:1:27030)
    at maybeRef (https://civet.dev/assets/browser.min.Dri6HQ9-.js:1:27156)
    at maybeRefAssignment (https://civet.dev/assets/browser.min.Dri6HQ9-.js:1:27427)
    at https://civet.dev/assets/browser.min.Dri6HQ9-.js:1:460984
    at PropertyDefinition$5 (https://civet.dev/assets/browser.min.Dri6HQ9-.js:1:461663)
    at Array.PropertyDefinition (https://civet.dev/assets/browser.min.Dri6HQ9-.js:1:462060)
    at https://civet.dev/assets/browser.min.Dri6HQ9-.js:1:2045
    at https://civet.dev/assets/browser.min.Dri6HQ9-.js:1:2350
    at Array.SingleLineObjectProperties (https://civet.dev/assets/browser.min.Dri6HQ9-.js:1:446471)
    at https://civet.dev/assets/browser.min.Dri6HQ9-.js:1:2045
```

Similar inputs that make it crash (each line individually causes a crash:

```
{a.1}
{a.1x
{a.11
{a.🍊
x := {a.1
```

STRd6:
`x.<literal>` and `x.<unicode>` PropertyAccess shorthands produced an Index AST node with no `expression` field. The PropertyDefinition object-shorthand path (`{x.1}`, `{x."b"}`, `{x.😊}`, etc.) feeds `expression` through `maybeRefAssignment` → `needsRef`, which crashed on the missing property.

Wrap simple literals (StringLiteral, IntegerLiteral) in a `Literal` node so `needsRef` recognises them as no-ref values, matching the bracket-form `{x[1]}`/`{x["b"]}` behavior. TemplateLiteral and SymbolLiteral keep their own types so refs are still introduced for their side-effecting evaluation. Index children layout is shifted to `[open, expression, ws, close]` so `last.children.slice(-2)` continues to yield `[ws, close]` when refs are needed.

Fixes #2065.

### DIFF
--- a/source/parser.hera
+++ b/source/parser.hera
@@ -1923,11 +1923,27 @@ PropertyAccess
   # NOTE: Added shorthand x.3 -> x[3]
   # NOTE: Added shorthand x.:symbol -> x[Symbol.for("symbol")]
   AccessStart:dot ( TemplateLiteral / StringLiteral / IntegerLiteral / SymbolLiteral ):literal ->
+    // Wrap simple literals so consumers (e.g. PropertyDefinition shorthand, postfix
+    // update extraction) that read `expression` and feed it through `needsRef` see
+    // a Literal — matching the bracket-form `{a[1]}`/`{a["b"]}` no-ref behavior.
+    // TemplateLiteral/SymbolLiteral keep their own types so refs are introduced.
+    expression .= literal as ASTNode
+    if literal.type is "StringLiteral" or not literal.type
+      expression = {
+        type: "Literal",
+        subtype: literal.type ?? "NumericLiteral",
+        children: [literal],
+        raw: literal.token,
+      }
     return {
       type: "Index",
+      expression,
+      // children layout mirrors the bracket-form Index `[open, expression, ws, close]`
+      // so `last.children.slice(-2)` still yields `[ws, close]`.
       children: [
         adjustIndexAccess(dot),
-        literal,
+        expression,
+        "",
         "]",
       ]
     }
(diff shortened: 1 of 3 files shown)


source/parser.hera
```diff
@@ -1923,11 +1923,27 @@ PropertyAccess
   # NOTE: Added shorthand x.3 -> x[3]
   # NOTE: Added shorthand x.:symbol -> x[Symbol.for("symbol")]
   AccessStart:dot ( TemplateLiteral / StringLiteral / IntegerLiteral / SymbolLiteral ):literal ->
```
edemaine: Could we add `::Index` to enforce `expression` too? Er, and add `expression` to `Index` type. 😅

source/parser.hera
```diff
@@ -1923,11 +1923,27 @@ PropertyAccess
...
+    // a Literal — matching the bracket-form `{a[1]}`/`{a["b"]}` no-ref behavior.
+    // TemplateLiteral/SymbolLiteral keep their own types so refs are introduced.
+    expression .= literal as ASTNode
+    if literal.type is "StringLiteral" or not literal.type
+      expression = {
+        type: "Literal",
+        subtype: literal.type ?? "NumericLiteral",
+        children: [literal],
+        raw: literal.token,
+      }
```
edemaine: I guess the cleaner way to do this would be to make `AccessLiteral` with the same wrapper as `Literal`, ideally factored out into a helper function, that matches `StringLiteral` and `IntegerLiteral`. Or maybe even simpler:

```coffee
AccessLiteral
  &(StringLiteral / IntegerLiteral) Literal -> $2
```

Then we wouldn't need to duplicate any wrapping code here, I believe.

edemaine (approved):

PR #2075, opened 2026-05-17
LSP: fix missing line/column info by unwrapping ParseErrors

### LINKED ISSUE #2072: parse errors from LSP always point to the first line
Using helix with the civet-lsp, and it seems any error where the diagnostic comes from the parser points to the wrong location.

The same things in VSCode or the playground point to the correct line.

The message of the LSP actually says the correct line but the JSON the LSP emits does not, 

```json
helix_lsp::transport [INFO] civet-lsp <- {
  "jsonrpc": "2.0",
  "method": "textDocument/publishDiagnostics",
  "params": {
    "uri": "file:///home/user/project/src/test.civet",
    "diagnostics": [
      {
        "severity": 1,
        "range": {
       // wrong line here
          "start": {
            "line": 0,
            "character": 0
          },
          "end": {
            "line": 0,
            "character": 10
          }
        },
       // correct line info here  --------------------V
        "message": "/home/user/project/src/test.civet:4:1 Failed to parse\nExpected:\n\tNonNewlineWhitespace /[ \\t]+/\n\t [...]",
        "source": "civet"
      }
    ]
  }
}
```

(output from the LSP from running helix with `hx -v --log log.txt .`)

the test code was a basic

```coffee
some := 0
other := 10
later := 
```

but this happens for every parser error in my experience

STRd6:
Fixes #2072.

### Root cause

`lsp/server/source/server.civet` iterated `parseErrors` and expected each entry to be a single `ParseError` with own `line`/`column`. When civet's `compile` threw a `ParseErrors` (plural) wrapper — which has only an `errors[]` array — the inline check `'line' in e and 'column' in e` returned false, and every diagnostic collapsed onto the defensive `(0,0)–(0,10)` fallback. The inner `ParseError.message` already carried the `filename:line:column` prefix so the text was correct even though the range wasn't.

### Fix

Extracted a `parseErrorsToDiagnostics` helper into `lib/completions.civet` that recurses into any entry carrying an `errors[]` array, so inner `ParseError`s contribute their own ranges.

### Tests

- Unit: 3 new cases in `test/completions.civet` (single ParseError, unwrapped wrapper, fallback)
- Protocol: new case in `test/lsp-features.test.civet` using the issue's exact source, asserts the diagnostic range falls past line 0
- Full LSP suite: 264 passing, 0 failing

### DIFF
--- a/lsp/server/source/lib/completions.civet
+++ b/lsp/server/source/lib/completions.civet
@@ -316,23 +316,19 @@ export function parseErrorsToDiagnostics(
       diagnostics.push ...parseErrorsToDiagnostics(inner.errors)
       continue
     pe := e as ParseError
-    if 'line' in pe and 'column' in pe and pe.line? and pe.column?
-      { start, end, message } := parseErrorRange(pe)
-      diagnostics.push {
-        severity: DiagnosticSeverity.Error
-        range: { start, end }
-        message
-        source: 'civet'
-      }
-    else
-      diagnostics.push {
-        severity: DiagnosticSeverity.Error
-        range:
-          start: { line: 0, character: 0 }
-          end:   { line: 0, character: 10 }
-        message: (e as Error).message
-        source: 'civet'
-      }
+    { start, end, message } :=
+      if pe.line? and pe.column?
+        parseErrorRange(pe)
+      else
+        start: { line: 0, character: 0 }
+        end:   { line: 0, character: 10 }
+        message: pe.message
+    diagnostics.push {
+      severity: DiagnosticSeverity.Error
+      range: { start, end }
+      message
+      source: 'civet'
+    }
   diagnostics
 
 // -- semantic token length measurement --------------------------------------
--- a/lsp/server/source/lib/completions.civet
+++ b/lsp/server/source/lib/completions.civet
@@ -12,8 +12,10 @@
 
 {
   type CompletionItem
+  type Diagnostic
   CompletionItemKind
   CompletionItemTag
+  DiagnosticSeverity
 } from vscode-languageserver
 
 ts from typescript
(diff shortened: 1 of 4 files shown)


lsp/server/source/lib/completions.civet
```diff
@@ -300,6 +302,39 @@ export function parseErrorRange(e: ParseError): {
...
+  parseErrors: readonly (Error | ParseError | { errors: (Error | ParseError)[] })[]
+): Diagnostic[]
+  diagnostics: Diagnostic[] := []
+  for e of parseErrors
+    inner := e as { errors?: (Error | ParseError)[] }
+    if Array.isArray(inner.errors)
+      diagnostics.push ...parseErrorsToDiagnostics(inner.errors)
+      continue
+    pe := e as ParseError
+    if 'line' in pe and 'column' in pe and pe.line? and pe.column?
```
edemaine:
```suggestion
    if pe.line? and pe.column?
```

At least I think the type is narrowed enough here.

lsp/server/source/lib/completions.civet
```diff
@@ -300,6 +302,39 @@ export function parseErrorRange(e: ParseError): {
...
+      }
+    else
+      diagnostics.push {
+        severity: DiagnosticSeverity.Error
+        range:
+          start: { line: 0, character: 0 }
+          end:   { line: 0, character: 10 }
+        message: (e as Error).message
+        source: 'civet'
+      }
```
edemaine: This code could probably be refactored to share the `push` and object construction, just need to set `start` and `end` differently. Maybe:

```coffee
{ line, column, message } .= e as ParseError
let start, end
if line? and column?
  { start, end, message } = ? parseErrorRange(e as ParseError) :
else
  start = { line: 0, character: 0 }
  end =   { line: 0, character: 10 }
diagnostics.push {
  severity: DiagnosticSeverity.Error
  range: { start, end }
  message
  source: 'civet'
}
```

edemaine (approved):

PR #2076, opened 2026-05-17
Fix reserved word ref creation

edemaine:
Fix `ReservedWord` grammar alternatives to return the matched keyword string instead of the regex capture tuple.

This PR fixes a sourcemap crash for reserved-word `@` bindings such as `@default`, exposed by [`unlessgames/mumux`](https://gitlab.com/unlessgames/mumux) when building with latest Civet, with error `outputStr.split is not a function`.

The regression was introduced in #2036, where `ReservedWord` changed from predicate-style matching to returning `$1`. It should have returned `$1[0]` (the string) instead of the entire `match` array (`$1`).

Related, the Hera documentation about the behavior here was incorrect. Now fixed here: https://github.com/DanielXMoore/Hera/pull/90

### DIFF
--- a/source/parser.hera
+++ b/source/parser.hera
@@ -6905,11 +6905,11 @@ TemplateBlockCharacters
 # before `in`) match the longest match first; boundary check via NonIdContinue.
 # `let` added since Civet assumes strict mode.
 ReservedWord
-  /(?:off|yes|on|no)/ NonIdContinue CoffeeBooleansEnabled -> $1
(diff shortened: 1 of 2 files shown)


STRd6 (approved):
Can also use `$` text prefix operator in Hera like `$/of/`

edemaine:
> Can also use `$` text prefix operator in Hera like `$/of/`

Yeah, I was wondering about that. Is one better than the other? Does `$` avoid constructing the array?

UPDATE: Apparently not, so no difference. But I think `$` is easier to read, so I'll change it.

PR #2077, opened 2026-05-17
Coffee classes: backtrack on `:symbol` LHS so `[Symbol.x]` wraps

### LINKED ISSUE #2053: Symbol shorthand breaks with coffeeClasses
```
"civet coffeeClasses"
class X
  @:iterator = 5

// vvv

class X {
  static Symbol.iterator = 5
}
```

STRd6:
## Summary
- Fixes #2053 — `@:iterator = 5` under `coffeeClasses` now compiles to `static [Symbol.iterator] = 5` instead of invalid `static Symbol.iterator = 5`.
- Root cause: `CoffeeClassesEnabled` ClassElement rule captured the LHS as a generic `ActualAssignment`, in which `:iterator` lowers to bare `Symbol.iterator`. The standard ClassElement rule (which wraps `:symbol` via `SymbolElement`) was never tried because the coffeeClasses rule matched first.
- Fix: `$skip` (backtrack) when `assignment.assigned.type === "SymbolLiteral"`, letting the standard rule wrap it as a computed `[Symbol.x]` field name.
- Bonus: also repairs the non-static case `:iterator = 5` under `coffeeClasses`, which previously emitted `Symbol.iterator = 5` inside an IIFE.

## Test plan
- [x] Added two `testCase` entries to `test/compat/coffee-classes.civet` (`@:iterator = 5` and `:iterator = 5`)
- [x] `pnpm test` (3987 passing)
- [x] `pnpm test:self` (3987 passing)

### DIFF
--- a/source/parser.hera
+++ b/source/parser.hera
@@ -85,6 +85,7 @@ import {
 } from "./parser/lib.civet"
 import type {
   ArrayBindingPattern,
+  AssignmentExpression,
   ASTNode,
   BlockStatement,
   ObjectBindingPattern,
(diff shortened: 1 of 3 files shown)


edemaine (approved):
Looks good, aside from failing typecheck. Maybe adding `::AssignmentExpression` to `ActualAssignment` could fix it? Or bump the error limit.

PR #2080, opened 2026-05-18
CLI: parallelize compile loop

STRd6:
## Summary
- Replace the sequential `for await {filename,...} of readFiles` loop in `cli.civet` with `await.all filenames.map(...)` so reads, worker IPC, and worker compiles overlap.
- Workers already spawned at `CIVET_THREADS>0` but sat idle awaiting each `compile()` before the next read. End-to-end self-compile (55 files): **2.43× at `THREADS=4`**, **2.54× at `THREADS=8`**.
- libuv throttles `fs.readFile` to its threadpool size, so in-flight content stays bounded (~50 MB for 10k files at 5 KB avg) — no full pre-buffering.
- Mixed inputs (`--eval` or stdin `-`) keep going through the existing `readFiles` generator since stdin can't be parallelized. Two new tests exercise the non-stdin arm via `--eval` + filename combos.

## Benchmark

| Mode | Wall | CPU% | Speedup |
|---|---|---|---|
| `CIVET_THREADS=0` | 4.70 s | 140 % | 1.00× |
| `CIVET_THREADS=4` | 1.93 s | 477 % | **2.43×** |
| `CIVET_THREADS=8` | 1.85 s | 727 % | **2.54×** |

Per-file output is byte-identical to main (only difference: the rebuilt `cli.civet.tsx` itself).

## Test plan
- [x] `pnpm test` — 3996 passing / 19 pending / 0 failing (vs 3994 / 19 / 0 on main; same baseline + 2 new tests)
- [x] `pnpm coverage:check` — 100% statements / branches / functions / lines
- [x] `pnpm typecheck --max-errors 888` — 887 diagnostics, within baseline (no new TS regressions)
- [x] Multi-file output ordering preserved (single-thread + 4/8 workers, 5 → 20 files)
- [x] `--ast` with multiple files emits ASTs in input order
- [x] Load + transpile errors report correct per-file filename in stderr; exit code aggregates
- [x] Run mode works: CJS file, ESM file (top-level await / import / export), `-e` eval, piped stdin `-`
- [x] Multi-file `--typecheck` stable across repeated runs (no parallel `.load()` flakiness on the unplugin TSService path)


edemaine:
This is a great improvement, but I found the implementation a bit messy. So I refactored some:
* Now `readFiles` is actually useful: it returns an array of Promises instead of an async generator.
* Forbade multiple stdin inputs because this doesn't really make sense (how would you separate them in the output?).
* Output is now in parallel too; just stdout and errors need to be sequentialized.
* Didn't really need `await.all`; we just needed to await in order.

In more detail, here's how the PR stands now:

`readFiles` changed from an async generator to an array of already-started promises:

```diff
-function readFiles(...): AsyncGenerator<ReadFile>
+function readFiles(...): Promise<ReadFile>[]
+  if (for count filename of filenames when filename is '-') > 1
+    return [Promise.resolve { filename: '<stdin>', error: new Error "...", stdin: true }]
+
+  reads: Promise<ReadFile>[] .= []
```

`yield` sites became promise pushes / return values:

```diff
-    yield {filename, content, stdin}
+    {filename, content, stdin}

-    yield {filename, error, stdin}
+    {filename, error, stdin}
```

File reads now explicitly use the shared encoding:

```diff
-        content = await fs.readFile filename
+        content = await fs.readFile filename, {encoding}
```

The old sequential compile/output loop:

```diff
-  errors .= 0
-  for await let {filename, error, content, stdin} of readFiles filenames, options.eval
-    ... load errors, compile, write output, increment errors ...
```

is now a parallel pipeline:

```civet
errorCount .= 0

readFiles(filenames, options.eval)
|> .map (readFile: Promise<ReadFile>) => async do
  ...
  {errors, stdout}
|> for result of &
  {errors, stdout} := await result
  // Each result has either stdout or errors, not both.
  process.stdout.write stdout if stdout?
  console.error error for error of errors
  errorCount++ if errors#
```

Important behavior change: reads, compiles, and file writes can overlap, but console output is still emitted in filename order by awaiting each promise sequentially in the final loop.

Error handling changed from immediate logging to collecting per-file ordered output:

```diff
-      console.error `${filename} failed to load:`
-      console.error error
-      errors++
+      errors.push `${filename} failed to load:`, loadError
```

```diff
-          console.error `${outputFilename} failed to write:`
-          console.error error
-          errors++
+          errors.push `${outputFilename} failed to write:`, error
```

The aggregate error counter was renamed to avoid collision with per-file `errors`:

```diff
-  process.exitCode = Math.min 255, errors
+  process.exitCode = Math.min 255, errorCount
```

---

Hopefully you like these changes; if not, feel free to revert or modify further.

STRd6:
@edemaine Nice, I like it even better that way!

STRd6:
Only minor concern is that all the file reads kick off immediately but I haven't profiled so it may not even be an issue.

PR #2082, opened 2026-05-18
LSP: add `--version` / `--help` flags, default to `--stdio`

STRd6:
## Summary

`civet-lsp` previously had no CLI handling — the bin script just required the server module, which threw `Connection input stream is not set` when run interactively without `--stdio`.

- `-v` / `--version`: print package version
- `-h` / `--help`: print a short usage blurb
- Inject `--stdio` when no connection-mode flag (`--stdio` / `--node-ipc` / `--socket` / `--pipe`) is passed, so bare `civet-lsp` works (matching what the help text says).

## Test plan
- [x] `civet-lsp --version` prints `0.3.35`
- [x] `civet-lsp --help` prints usage
- [x] `civet-lsp` (no args) starts the server cleanly on stdio
- [x] `lsp/server/test/bin.test.civet` covers all three paths (5 new tests, all passing)

### DIFF
--- a/lsp/server/bin/civet-lsp.js
+++ b/lsp/server/bin/civet-lsp.js
@@ -3,44 +3,11 @@
 const fs = require('node:fs')
 const path = require('node:path')
 
-const args = process.argv.slice(2)
-const pkg = require('../package.json')
-
-if (args.includes('--version') || args.includes('-v')) {
-  console.log(pkg.version)
-  process.exit(0)
-}
-
-if (args.includes('--help') || args.includes('-h')) {
-  console.log(`${pkg.name} ${pkg.version}`)
-  console.log(pkg.description)
-  console.log()
-  console.log('Usage: civet-lsp [options]')
-  console.log()
-  console.log('Options:')
-  console.log('  -v, --version  Print version and exit')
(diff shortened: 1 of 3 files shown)


lsp/server/bin/civet-lsp.js
```diff
@@ -3,11 +3,44 @@
...
 }
 
+// vscode-languageserver demands an explicit connection mode; default to --stdio
+// when none of --stdio / --node-ipc / --socket / --pipe is present.
+const hasConnectionMode = args.some(a =>
+  a === '--stdio' || a === '--node-ipc' ||
+  a === '--socket' || a.startsWith('--socket=') ||
+  a === '--pipe' || a.startsWith('--pipe=')
+)
+if (!hasConnectionMode) process.argv.push('--stdio')
```
edemaine: This seems weird to put here instead of `node.civet`. Why not set a default when parsing the arguments? This approach duplicates logic.

Actually, I wonder whether `node.civet` should process `--help`/`--version`/usage too...

STRd6:
@claude

Use triple backtick block template literals for help message.

Make sure branch coverage is at 100%

PR #2084, opened 2026-05-19
Add comptime Float16Array support

bbrk24:
[`Float16Array`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Float16Array) is a new kind of TypedArray introduced in ES2025. node.green doesn't list it, so I can't say for certain what node versions do and don't have it, but given how new it is I used the same trick we use for `Buffer` to make the comptime serializer not error out when it doesn't exist.


STRd6:
Thanks!

PR #2085, opened 2026-05-20
CI typecheck: rebuild base in its own worktree

STRd6:
## Summary

The Typecheck-diff job symlinks HEAD's `node_modules` and `dist` into the base worktree so the base run can reuse the freshly built CLI. That assumes HEAD and base agree on every runtime dependency — which breaks the moment a PR bumps a dep loaded by the CLI (e.g. `@danielx/hera` 0.9.2 → 0.9.6 in #2038).

When that happens, the base typecheck is really HEAD's compiler/parser running against main's `.civet` source. In #2038 that combination crashes TypeScript 6.0.2:

```
RangeError: Maximum call stack size exceeded
    at checkExpression / checkAssertionWorker / checkAssertion …
```

The base log captures only the Node stack trace — zero TS diagnostics. The diff script then sees `Old: 0 → New: 1730` and reports every HEAD error as a regression. [Failing run](https://github.com/DanielXMoore/Civet/actions/runs/26128880281/job/76849202537).

## Fix

- `pnpm install --frozen-lockfile && pnpm build` inside the base worktree so main's source is typechecked with main's own compiler (no version skew).
- Guard step: if the base log contains no `error TS…` lines at all, fail with a clear notice instead of silently treating the whole HEAD error set as new.

This reverts the speed optimization from e1f1f0b8 / 4ff21657 — the symlink-HEAD-CLI approach is correctness-fragile, and the extra install+build cost (~30s install, ~30s build locally) is the price of an apples-to-apples diff.

## Test plan

- [x] Local repro: with `pnpm install + pnpm build` in a base worktree at the same SHA CI used (`8c7794c`), `./dist/civet --typecheck` produces 883 TS diagnostics (matches main's natural error count under `pnpm typecheck`).
- [x] Local repro: with HEAD's `dist`/`node_modules` symlinked in (old recipe), the same step crashes with `RangeError` and writes 0 TS-error lines — guard correctly fires.
- [ ] CI: PR's own Typecheck-diff job runs against `main` as base. Since this PR doesn't change source, expected diff is `0 / 0 / 0`.
- [ ] Verify on #2038 (rebase or merge to pick up this fix) that the base step now produces non-zero diagnostics and the diff reflects only genuinely-new errors.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -172,7 +172,6 @@ jobs:
           base_dir="$RUNNER_TEMP/typecheck-base"
           base_log="$RUNNER_TEMP/typecheck-before.log"
           head_log="$RUNNER_TEMP/typecheck-after.log"
-          typecheck_bin="$PWD/dist/civet"
 
           trap 'git worktree remove --force "$base_dir" >/dev/null 2>&1 || true' EXIT
           rm -rf "$base_dir"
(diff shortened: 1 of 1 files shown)


edemaine:
@greptileai revise review

edemaine (approved):

PR #2087, opened 2026-05-20
CI: Clarify typecheck-diff output

edemaine:
- Clarify failing typecheck CI output by making the final failing step emit a specific GitHub error annotation based on the actual failure mode: new regressions, exceeded error limit, or both.
- Reprint the saved plain-text typecheck diff output in the failing step, filtering out the internal `::notice` line, so the red CI step includes the actionable diagnostics instead of only showing `exit 1`.

## [Example CI run](https://github.com/DanielXMoore/Civet/actions/runs/26144208433/job/76895842185?pr=2087)

image


edemaine:
@greptileai revise review

STRd6 (approved):

PR #2090, opened 2026-05-21
Dev: Property children false-leak + literal-leaf type tags

STRd6:
Two cherry-picks from #2009 that still apply on top of #2038.

- **lib**: `processCallMemberExpression` was leaking `false` into the synth Property's children via `cond and x` short-circuits.  Harmless today (no predicate uses strict `'foo' in node`), but ce990c88 removed the traversal-level boolean filter that hid it.  Replaced with `... or ""` so the failing arm becomes ASTString.
- **types**: parser emits `BooleanLiteral` / `NullLiteral` leaves but `LiteralContentNode` only listed `NumericLiteral` / `StringLiteral`.  Added both to the union and exported named aliases.

### DIFF
--- a/source/parser/lib.civet
+++ b/source/parser/lib.civet
@@ -742,9 +742,9 @@ function processCallMemberExpression(node: CallExpression | MemberExpression): A
             delim: part.delim
             names: part.names
             children: [
-              isWhitespaceOrEmpty(part.children[0]) and part.children[0]
+              isWhitespaceOrEmpty(part.children[0]) ? part.children[0] : ""
               name
-              isWhitespaceOrEmpty(part.children[2]) and part.children[2]
+              isWhitespaceOrEmpty(part.children[2]) ? part.children[2] : ""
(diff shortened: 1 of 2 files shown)


source/parser/lib.civet
```diff
@@ -742,9 +742,9 @@ function processCallMemberExpression(node: CallExpression | MemberExpression): A
             delim: part.delim
             names: part.names
             children: [
-              isWhitespaceOrEmpty(part.children[0]) and part.children[0]
+              isWhitespaceOrEmpty(part.children[0]) and part.children[0] or ""
               name
-              isWhitespaceOrEmpty(part.children[2]) and part.children[2]
+              isWhitespaceOrEmpty(part.children[2]) and part.children[2] or ""
```
edemaine: Slight preference for:

```suggestion
              isWhitespaceOrEmpty(part.children[2]) ? part.children[2] : ""
```

(and similarly above)

edemaine (approved):

PR #2092, opened 2026-05-22
Support destructuring-style imports with rest and nested patterns

### LINKED ISSUE #175: Unify `imports` with object destructuring syntax
It's a shame that ESM `import` statements don't have the same syntax as destructuring. Let's add an optional alternative syntax so they can be unified.

```typescript
import {x: y} from "./x"
import {a: a1, b: b1, ...rest} from "./y"
---
import {x as y} from "./x"
import * as ref from "./y"
const {a: a1, b: b1, ...rest} = ref
```

Nested destructuring should also work in a similar manner.

STRd6:
Closes #175.

## Summary

`import {a, ...rest, b} from "./y"` and `import {a: {b, c}} from "./y"` desugar to a namespace import plus a const destructure, since native ESM can't express rest or nested destructuring inside `import {…}`:

```civet
import {a: a1, b: b1, ...rest} from "./y"
↓
import * as imports from "./y"
const {a:a1,b:b1, ...rest} = imports
```

Plain `import {x, y}` and `import {x: y}` still emit native ESM — the desugar only fires when a rest element or pattern-binding is present.

Rest can appear in any position (`{...rest, a}`, `{a, ...rest, b}`); the parser reuses Civet's existing `reorderBindingRestProperty` to normalize it to the tail, matching the destructuring convention.

## Scope

Implements the rest and nested cases. Deferred (would be follow-ups):
- `rest.T` rewriting in type position
- `type {T} = module` syntax
- `import "./x" as {…}` (overlaps with #320)

## Test plan

- [x] `pnpm test` — 4032 passing (12 new cases in `test/import.civet` under `describe "destructuring-style"`)
- [x] `pnpm test:self` — 4032 passing (Civet compiles itself)
- [x] `pnpm typecheck` — net +0 vs main

### DIFF
--- a/source/parser.hera
+++ b/source/parser.hera
@@ -6231,9 +6231,7 @@ ImpliedImport
 ImportClause
   ImportedBinding:binding ( __ Comma __ ( NameSpaceImport / NamedImports ) )?:rest ->
     if rest
-      // Spread `named` so its optional fields (specifiers / rest / needsDesugar
-      // from NamedImports; star from NameSpaceImport) flow through without
-      // tripping exactOptionalPropertyTypes on explicit `: undefined` sets.
+      // Spread to forward NamedImports/NameSpaceImport optional fields.
       named := rest[3]
       return {
         ...named,
@@ -6266,16 +6264,10 @@ NameSpaceImport
 # https://262.ecma-international.org/#prod-NamedImports
 NamedImports
   OpenBrace ( NamedImportRest / TypeAndImportSpecifier )*:items ( __ Comma )? __ CloseBrace ->
-    // Accept `...rest` anywhere (matching Civet object destructuring), then
-    // reorderBindingRestProperty swaps delimiters and moves rest to the tail.
-    // We ignore the returned `names` and compute our own below from each
-    // specifier's binding (item.names isn't populated for ImportSpecifiers).
+    // Rest can appear anywhere; reorder moves it to the tail.
     reordered := (reorderBindingRestProperty items).children as (NamedImportElement | { type: "Error", message: string })[]
-    // reorderBindingRestProperty appends an Error node when restCount > 1.
-    // We exclude it from `specifiers` (it has no `binding`) so the parser can
-    // backtrack cleanly out of speculative implied-import matches like
-    // `{...c, ...d} = e`; we still surface it through `children` so real
-    // multi-rest imports report the error.
+    // reorder appends an Error on multi-rest; surface via children, skip in
+    // specifiers so speculative `{...c, ...d} = e` matches can backtrack.
     error := reordered.find .type is "Error"
     specifiers := reordered.filter .type is undefined
     rest := reordered.find .type is "BindingRestProperty"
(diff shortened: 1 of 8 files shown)


edemaine (commented):
Does this work in the dynamic context? 

```js
=>
  import {x, ...rest} from foo
```

source/parser.hera
```diff
@@ -6258,14 +6263,37 @@ NameSpaceImport
...
 # https://262.ecma-international.org/#prod-NamedImports
 NamedImports
-  OpenBrace TypeAndImportSpecifier*:specifiers ( __ Comma )? __ CloseBrace ->
-    names := specifiers.flatMap(({binding}) => binding.names)
+  OpenBrace ( NamedImportRest / TypeAndImportSpecifier )*:items ( __ Comma )? __ CloseBrace ->
+    // Rest can appear anywhere; reorder moves it to the tail.
+    reordered := (reorderBindingRestProperty items).children as (NamedImportElement | { type: "Error", message: string })[]
+    // reorder appends an Error on multi-rest; surface via children, skip in
+    // specifiers so speculative `{...c, ...d} = e` matches can backtrack.
+    error := reordered.find .type is "Error"
```
edemaine: Could there be multiple Error nodes? Looking at the code, I guess not, but more future-proof to filter them all and splat them in. Or perhaps simpler, filter out all but a found `rest` to compute `children`.

source/parser.hera
```diff
@@ -6258,14 +6263,37 @@ NameSpaceImport
...
+    specifiers := reordered.filter .type is undefined
+    rest := reordered.find .type is "BindingRestProperty"
+    names := specifiers.flatMap .binding.names
+    if rest
+      names.push(...rest.binding.names)
+    needsDesugar := !!rest || specifiers.some .pattern
```
edemaine: Would it be simpler to move this check to the desugar helper? All the same information is there, and it would put it next to where it handles `rest` and `pattern`.

source/parser/declaration.civet
```diff
@@ -485,6 +487,92 @@ function dynamizeFromClause(from: any)
     from.push ", {", assert.keyword, ":", assert.object, "}"
   ["(", ...from, ")"]
 
+// Rewrite `import {…rest|nested}` to namespace import + const destructure.
+// Per-spec `type` imports get pulled into a separate native `import {type X}`
+// since a const-destructure can't bind types.
+function desugarStaticImport(decl: ImportDeclaration)
```
edemaine: Possible alternate name: `processStaticImport`. (A little more consistent with other names.)

source/parser.hera
```diff
@@ -6325,27 +6353,35 @@ ImportAssertion
 # https://devblogs.microsoft.com/typescript/announcing-typescript-3-8-beta/#type-only-imports-exports
 TypeAndImportSpecifier
   __ TypeKeyword ImportSpecifier ->
-    return { ts: true, children: $0, binding: $3.binding }
+    // Flatten inner ImportSpecifier's children so reorderBindingRestProperty's
+    // last-child-is-delim invariant holds when `...rest` shuffles past us.
+    return { ts: true, children: [$1, $2, ...$3.children], binding: $3.binding, pattern: $3.pattern, delim: $3.delim }
```
```suggestion
    return { ts: true, children: [$1, $2, ...$3.children], $3.{binding,pattern.delim} }
```

edemaine (approved):
This is super cool! I love both features.

Maybe we could add `export { a: {b,c}, x: {y,z} }` too in the future?

Do you want to add docs, or should I?

PR #2098, opened 2026-05-22
Drop results accumulator from postfix-for when value is discarded

### LINKED ISSUE #565: postfix loop implicitly assigns results = [] variable
with no explicit declaration, eg: `stuff := for x of loop` the results variable is still assigned

```ts
((push shard for shard of bucket) for bucket of buckets)

// ↓↓↓ Civet 0.6.10

(() => {
  const results = [];
  for (const bucket of buckets) {
    results.push(
      (() => {
        const results1 = [];
        for (const shard of bucket) {
          results1.push(push(shard));
        }
        return results1;
      })()
    );
  }
  return results;
})();
```

STRd6:
Closes #565.

Postfix-for expressions like `(foo(x) for x of arr)` always allocated and
populated a `results` array, even when nothing observed the value:

```ts
((push shard for shard of bucket) for bucket of buckets)
// →
const results=[]; for (const bucket of buckets) {
  const results1=[]; for (const shard of bucket) { results1.push(push(shard)) }
  results.push(results1)
} results
```

When the iteration expression sits at the top of its containing block and a
walk up the parent chain shows the value will never be consumed (followed by
another statement, inside a for/while body, or at module root in non-IIFE
mode), `expressionizeIteration` now emits just the bare for-statement —
skipping the `const results = []` declaration, the `results.push(...)` body
wrapping, and the trailing reference.

Reductions (`for some|every|sum|... x of y`) and empty-body comprehensions
(`(for x of y)`) keep their accumulator: those are explicit value-computing
syntactic forms whose test cases asserted the current behavior. `--iife` /
`--repl` mode also opts out, since the root block becomes the implicit
return of the wrapping IIFE.

### DIFF
--- a/scripts/typecheck-diff.civet
+++ b/scripts/typecheck-diff.civet
@@ -109,15 +109,39 @@ function parse(logPath: string): Diag[]
   out.push current if current
   out
 
-// Sort union members within short quoted literals so that
-// reorderings don't masquerade as new errors.  Skip literals with `<`
-// (generic type parameters) where naive splitting on ` | ` would corrupt.
+// Sort union members within short quoted literals so that reorderings don't
+// masquerade as new errors.  Sort both parenthesized sub-unions (so
+// `'(B | A)[]'` matches `'(A | B)[]'`) and top-level unions.  Skip literals
+// with `<` (generic type parameters) where naive splitting on ` | ` would
+// corrupt nested arguments.
 function sortQuotedUnions(s: string): string
   s.replace /'([^']+)'/g, (full, inner) =>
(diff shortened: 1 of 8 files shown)


edemaine (commented):
Did you consider the "opposite" approach? Marking the IIFE and accumulation as optional in some sense, and after all implicit returns and `civet iife` processing, stripping out anything marked optional in this way? I'm not sure which is cleaner, though this opposite approach would seem to duplicate less logic. But maybe it's too messy to convert IIFE + accumulation back to normal loop?

PR #2108, opened 2026-05-24
LSP: accept inline tsConfig in civetconfig (#379)

### LINKED ISSUE #379: LSP wont exclude dist folder without tsconfig.json
### How the problem occurred ?
I started with a basic civet install with esbuild. But because the src folder didn;t contain ant ts files the TS LSP complaint as follows:

`No inputs were found in config file 'e:/Projects/test/ts-civet/tsconfig.json'. Specified 'include' paths were '[]' and 'exclude' paths were '["dist"]'.`

So following my instinct i renamed the tsconfig.json to tsconfig.civet.json, Also supplying this new file to esbuild config. 
But this lead to a new issue where Civet LSP would complain now because it's not ignoring the .js files compiled in the dist folder.

image

STRd6:
Fixes #379.

Civet-only projects can now put their tsconfig under a `tsConfig` key in civetconfig instead of shipping a real `tsconfig.json`.

```json
{
  "tsConfig": {
    "compilerOptions": { "target": "es2020", "strict": true },
    "include": ["src"],
    "exclude": ["dist"]
  }
}
```

Without a `tsconfig.json` on disk, VSCode's built-in TypeScript LSP no longer fires on a Civet-only project (it was complaining that `include: ["src"]` matched zero `.ts` files), while the Civet LSP still picks up the same include/exclude/compilerOptions.

## Approach

- New optional `tsConfig` field on `CompileOptions`.
- LSP `TSService` destructures it out of the loaded civetconfig and forwards it to the shared service's already-existing `tsConfig` option (which short-circuits `<projectPath>/tsconfig.json` on disk). Stripped from the per-plugin config so cache keys aren't perturbed for users who don't adopt this.
- `civet --typecheck` picks it up for free: `Options extends CompileOptions`, the config spread already lands the field in `options.tsConfig`, and unplugin already forwards that to `makeIncrementalTypecheckProgram`.

## Test plan

- [x] New LSP test: civetconfig.json with inline `tsConfig.include`/`exclude`, no `tsconfig.json` on disk → `configFound` true, `src/a.civet.tsx` in scope, `dist/a.js` excluded.

### DIFF
--- a/civet.dev/config.md
+++ b/civet.dev/config.md
@@ -275,3 +275,4 @@ top-level options (above `parseOptions`):
       "exclude": ["dist"]
     }
   }
+  ```
--- a/civet.dev/config.md
+++ b/civet.dev/config.md
@@ -264,7 +264,7 @@ top-level options (above `parseOptions`):
 - `threads`: Use specified number of Node worker threads to compile Civet files faster. Default: `0` (don't use threads), or `CIVET_THREADS` environment variable if set.
 - `tsConfig`: Inline `tsconfig.json` contents (same shape — `compilerOptions`, `include`, `exclude`, …) used by the Civet LSP and `civet --typecheck`.  When present, takes precedence over `tsconfig.json` on disk.  Lets a Civet-only project skip having a `tsconfig.json` at all, which keeps VSCode's built-in TypeScript LSP from complaining about an empty `include` (since it doesn't see your `.civet` files).  Example:
 
-  ```js
+  ```json
   {
     "tsConfig": {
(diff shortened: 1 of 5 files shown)


edemaine:
Sounds handy! Should it be called `tsconfig` instead of `tsConfig` to match the filename?

STRd6:
> Sounds handy! Should it be called `tsconfig` instead of `tsConfig` to match the filename?

In unplugin already we call it `tsConfig`. I think camel case is generally more common in JS even if the filename is all lower case.

PR #2114, opened 2026-05-25
Omit anonymous rest from destructure when unreferenced

### LINKED ISSUE #2112: Unused ref in anonymous `...`
## Input

```coffee
if [x,...] := a
```

## [Current Output](https://civet.dev/playground?code=aWYgW3gsLi4uXSA6PSBh) 

```js
if ((a) && Array.isArray(a) && a.length >= 1) {const [x,...ref] = a;}
```

## Desired Output

```js
if ((a) && Array.isArray(a) && a.length >= 1) {const [x] = a;}
```

Otherwise, TypeScript complains about unused `ref` variable. Alternatively, we could use `_ref`.

STRd6:
Closes #2112.

`[x, ...] := a` now emits `[x] = a` (matching the issue's desired output) instead of synthesizing an unused binding. `adjustBindingElements` drops the anonymous rest from the emitted `children` and strips the trailing comma on the preceding element, but keeps the rest in a separate `elements` field so:
- pattern matching still sees `rest: true` and emits `length >= n` rather than an exact-length check
- `gatherBindingPatternTypeSuffix` can still reach the rest's typeSuffix when one is present

Cases where the synthesized name is actually referenced (`[a, ..., b]` uses it for the post-rest splice) keep the default `ref` name. Top-level function rest parameters like `(a, ...) =>` and `function f(x, ...)` keep the rest (removing it would change the type signature's arity) and use `_ref` so TypeScript doesn't flag the unused name.

```
# before
if ((a) && Array.isArray(a) && a.length >= 1) {const [x,...ref] = a;}
# after
if ((a) && Array.isArray(a) && a.length >= 1) {const [x] = a;}
```

### DIFF
--- a/source/parser.hera
+++ b/source/parser.hera
@@ -2731,10 +2731,12 @@ BindingRestElement
       rest: true,
     }
 
-# NOTE: Allows for empty binding rest pattern like in CoffeeScript
+# NOTE: Allows for empty binding rest pattern like in CoffeeScript.
+# Uses `_`-prefixed base so TypeScript doesn't flag the synthesized name as
+# an unused variable when it isn't referenced by a post-rest splice.
 EmptyBindingPattern
   "" ->
-    ref := makeRef()
+    ref := makeRef("_ref")
     return {
       type: "EmptyBinding",
       children: [ref],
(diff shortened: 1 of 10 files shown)


edemaine (approved):

edemaine:
Thanks for the quick fix! I imagine detecting and removing unused `..._ref` is more annoying, but maybe worth optimizing later?

PR #2119, opened 2026-05-25
Snug ternaries `a?b:c` and `a ?b :c`

### LINKED ISSUE #1507: ternary requires even more space now
`a ?b :c` used to work, with #1498, it no longer parse, have to do `a ?b : c` now

STRd6:
Closes #1507.

`a ?b :c` regressed after #1498 (`b :c` parsed as `b(Symbol.for("c"))`); this restores it and adds fully snug `a?b:c`.

## Approach

- `UnaryPostfix` `?` and the snug-`?` `AccessStart` shorthand get a `!SnugTernaryColonAhead` lookahead, so `a?b:c` falls through to the ternary instead of existential `a?` / optional chain `a?.b`.
- New `forbidColonArgument` flag, set in a ternary "then": stops `b :c` from parsing as the implicit symbol call `b(Symbol.for("c"))`.
- New `forbidInlineObjectLiteral` flag, set only in the snug form: stops `b:c` from parsing as `{b:c}`. Spaced ternaries keep inline objects, so `x ? a: b : c: d` still works.

## Disambiguation

| Input | Output |
|---|---|
| `a?b:c` | `a?b:c` (snug ternary, new) |
| `a ?b :c` | `a ?b :c` (half-snug, the regression) |
| `a?b` | `a?.b` |
| `a?` | `(a != null)` |
| `f :foo` | `f(Symbol.for("foo"))` |
| `x ? a: b : c: d` | `x ? {a: b} : {c: d}` |
| `a ?b :c :d` | `a ?b :c(Symbol.for("d"))` |
| `a ? f :sym : c` | parse error — breaking change; a spaced `:x` in a ternary "then" no longer implicit-calls. Parenthesize: `a ? (f :sym) : c` |


edemaine:

civet.dev/comparison.md
```diff
@@ -106,8 +106,9 @@ symbol operators (`+`, `-`, etc.) need to be spaced consistently:
   the operator, but this is not guaranteed behavior.)
 * Comma operator is forbidden in some contexts, e.g.,
   array/object indexing.  If you run into this limitation, wrap in parentheses.
-* Ternary operator `x ?y :z` needs space before both `?` and `:`
-  (in particular to distinguish from unary postfix `x?`).
+* Ternary operator: `x?y:z` works snug, `x ?y :z` half-snug, and `x ? y : z`
+  spaced. Snug `x?y` alone is still optional chain `x?.y`; `x?` alone is
+  still the existential check `x != null`.
```
edemaine: I think the two instances of "still" should be removed here. (This is AI's tendency to write with time before/after its own changes, not thinking about the context where someone would read this.)

source/parser.hera
```diff
@@ -545,6 +563,10 @@ ForbiddenImplicitCalls
   PostfixStatement NoBlock
   # `x ... y` is reserved for a range, but `x ...y` is an implicit call
   "... "
+  # Within a ternary "then" expression, `b :foo` would otherwise be parsed as a
+  # SymbolLiteral implicit-call argument `b(Symbol.for("foo"))`, swallowing the
+  # ternary colon.
+  TernaryColonForbidden ":"
```
edemaine: "ternary colon forbidden" doesn't quite sound like the right meaning here... It's more like "waiting for ternary colon". But maybe the name is good enough for parallelism.

edemaine (commented):
Two more state flags seems like a lot for snug ternaries... but I guess `forbidInlineObjectLiteral` might be useful for other things.

edemaine:
Wait, can't we use ForbidTernaryColon to prevent implicit object literals too?

STRd6:
On the two-flag question — I think they have to stay separate, given the existing braceless-object ternary test:

```
x ? a: b : c: d   →   x ? {a: b} : {c: d}
```

That's the spaced/half-snug path. We *do* need `ForbidTernaryColon` there (otherwise `a ?b :c` regresses again to `a?.b(Symbol.for("c"))`), but we *don't* want to forbid InlineObjectLiteral, since `a:b` and `c:d` rely on it. The full-snug `a?b:c` path is the opposite: it needs both forbidden, otherwise `b:c` becomes `{b:c}` and eats the ternary colon.

So the orthogonal axes are real, even though they coincide for the snug case.

I also took the wording note on `TernaryColonForbidden` — kept the name for parallelism, but added a comment in parser.hera explaining the intended reading ("this construct is forbidden in this context", not "ternary colons are forbidden") and why InlineObjectLiteral is a separate flag. And dropped the two `still`s in comparison.md.

(Pushed in 076e180.)

- your pal Claude

edemaine:
> `x ? a: b : c: d   →   x ? {a: b} : {c: d}`

While it's true that we need two flags to support that, I'm not sure we really *want*/need to... though it [does work now](https://civet.dev/playground?code=eCA%2FIGE6IGIgOiBjOiBk). Personally, I'd be OK with "once ternary `?` is active, the next top-level `:` is its `else` clause". But maybe I'm not being ambitious enough. 🐱

PR #2120, opened 2026-05-25
Keep trailing comments inside indented blocks

### LINKED ISSUE #832: Comments jump ahead
Hi, 

First a little background with a use case.

The authors of ESLint believe that a comment should be placed in an empty block. Otherwise, you cannot bypass the `no-empty` rule in try-catch with an empty catch block. See [examples](https://eslint.org/docs/latest/rules/no-empty#rule-details) of correct code that do not lead to a linter error.

Too bad this code can't be used with ESLint without disabling `no-empty` rule:
```js
try
  something
↓↓↓
try {
  something;
} catch (e) {}
```

Satisfying ESLint rules may not be the goal of Civet, but trying to solve this problem by specifying a redundant catch block results in strange behavior:

```js
try
  something
catch
  // ESLint please don't say it's an empty block
↓↓↓
try {
  something;
} catch {}
// ESLint please don't say it's an empty block
```
This dropout is seen with any empty blocks.

While writing this message, an even more interesting example was found:
```js
try
  stmt1 //comment 1
  stmt2 //comment 2
catch
  //comment 3
↓↓↓
try {
  stmt1; //comment 1
  stmt2;
} catch {
  //comment 2
}
//comment 3
```
Apparently I finally found a way to make a comment in an empty catch block.

This situation prevents the use of various kinds of “magic” comments, which should be located either strictly on the same line after the desired expression, or strictly before the desired line.

It would be nice if comments didn’t jump into subsequent blocks at least.

And I also have a question that is not related to the issue, but on a related topic. Should Civet be able to produce code that will pass all of the linter's checks (or at least as many of them as possible)? I noticed other errors when trying to use Civet and ESLint together, which cannot be eliminated without editing the compilation result, but it is not clear whether they are worth mentioning.

STRd6:
Fixes #832.

## Summary
- A trailing `//` comment as the last line of an indented block was being escaped out past the closing brace; the same was true for comment-only block bodies and trailing inline comments on the last statement.
- Cause: `EOS` in the next outer `Nested` greedily swallows trailing comment lines, so the block's expressions ended before the comment.

## Approach
- Add a `TrailingNestedComments?` slot to `NestedBlockStatements` (plus helpers `TrailingInlineComment` and `TrailingNestedCommentLine`) so the indented block captures its own trailing comments before the outer block can see them.
- Add a `PushIndent` fallback (`TrackCommentLine`) so a comment-only body still pushes an indent.
- Put the captured comments in `block.children`, NOT `block.expressions`. Downstream AST walkers (assignResults, insertReturn, pattern-matching's continue-switch check) keep seeing the last real statement at `expressions.-1`, so they need no per-walker skip logic.
- `braceBlock` (block.civet) checks `hasTrailingComment block.children` and uses `\n}` so the brace doesn't get consumed by a trailing comment.
- `WhenClause` prefixes the inserted `;break;` with `\n` when the body ends with a comment, for the same reason.

## Test plan
- [x] `pnpm test:grep "832"` — three new red→green tests in `test/try.civet`
- [x] `pnpm test` — full suite, only pre-existing failures (5× CLI --typecheck baseline + 1× unplugin timeout)
- [x] `pnpm test:self` — passes the same with the rebuilt parser
- [x] `pnpm typecheck` — 0 regressions, net -1 vs main

### DIFF
--- a/source/parser.hera
+++ b/source/parser.hera
@@ -9879,15 +9879,6 @@ TrackIndented
     state.indentLevels.push(indent)
     return $1
 
-# Like TrackIndented, but for a deeper-indent comment-only line. Used so an
-# indented block whose body is only comments still pushes an indent. (#832)
-TrackCommentLine
-  Indent:indent &SingleLineComment ->
-    {level} := indent
-    return $skip if level <= state.currentIndent.level
-    state.indentLevels.push(indent)
-    return $1
-
 # Indents one level deeper,
 # without consuming the indentation so it can be by Nested
 # Must be matched with PopIndent
@@ -9895,9 +9886,11 @@ PushIndent
   # TrackIndented pushes the indent level if it is deeper than the current level
   # it will skip if it is not deeper
   &( EOS TrackIndented )
-  # Fallback for comment-only block bodies (issue #832) so the body still has
-  # a pushed indent level even though it lacks a real statement to gate on.
-  &( EOL TrackCommentLine )
+  # Fallback for comment-only block bodies (issue #832): `EOS` would have
+  # consumed the comment, leaving TrackIndented with only the dedented line
+  # beyond.  `EOL` stops at the first newline, so TrackIndented sees the
+  # comment line's own indent.
+  &( EOL TrackIndented )
 
 PopIndent
   "" ->
--- a/source/parser.hera
+++ b/source/parser.hera
@@ -3350,16 +3350,21 @@ PostfixedSingleLineNoCommaStatements
     }
 
 NestedBlockStatements
-  PushIndent NestedBlockStatement*:statements PopIndent ->
-    return $skip unless statements#
+  PushIndent NestedBlockStatement*:statements TrailingNestedComments?:trailing PopIndent ->
+    return $skip unless statements# or trailing?#
 
     // Each element of statements is a list of same-line statements. Flatten.
-    statements = statements.flat()
+    expressions: StatementTuple[] := statements.flat()
+    // Trailing same-indent comments (issue #832) live in children, NOT in
+    // expressions, so downstream walkers (assignResults / insertReturn / etc.)
+    // still see the last real statement as expressions.-1.
+    children: ASTNode[] := [expressions]
+    children.push(trailing) if trailing?#
 
     return {
       type: "BlockStatement",
-      expressions: statements,
-      children: [statements],
+      expressions,
+      children,
       bare: true,
     }
 
@@ -3369,6 +3374,27 @@ NestedBlockStatement
       [nested, ...statements[0]],
       ...statements.slice(1).map (s) => ["", ...s],
     ]
+
+# Captures trailing `// ...` comments at the end of an indented block: an
+# optional inline comment on the last statement's line, plus any number of
+# standalone same-indent comment lines. The lookahead `&EOL` keeps these from
+# eating into content that belongs to the surrounding block. (issue #832)
+TrailingNestedComments
+  TrailingInlineComment?:inline TrailingNestedCommentLine*:lines ->
+    return $skip unless inline or lines?#
+    out := []
+    out.push(inline) if inline
+    out.push(...lines) if lines?#
+    return out
+
+TrailingInlineComment
+  _?:ws SingleLineComment:c &EOL ->
+    return [ws, c]
+
+TrailingNestedCommentLine
+  EOL:eol Indent:indent SingleLineComment:c &EOL ->
+    return $skip unless indent.level is state.currentIndent.level
+    return [eol, indent, c]
 
 BlockStatementPart
   # NOTE: !EOS forces semicolon after all but last statement, forbids leading __
@@ -5625,6 +5651,12 @@ CaseClause
   }
   # NOTE: Added "when" from CoffeeScript. `when` always inserts `break;`.
   When CaseExpressionList:cases IgnoreColon InsertOpenBrace ( ThenClause / BareBlock ):block InsertBreak:b InsertNewline InsertIndent InsertCloseBrace ::WhenClause ->
+    // If the block ends with a `// ...` comment (issue #832), prefix the
+    // inserted `;break;` with a newline so it isn't swallowed by the comment.
+    if hasTrailingComment block
+      bIdx := $0.indexOf b
+      b = { ...b, token: "\n" + b.token }
+      $0[bIdx] = b
     return {
       type: "WhenClause",
       children: $0,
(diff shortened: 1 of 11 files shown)


edemaine:

source/parser.hera
```diff
@@ -3350,16 +3350,21 @@ PostfixedSingleLineNoCommaStatements
     }
 
 NestedBlockStatements
-  PushIndent NestedBlockStatement*:statements PopIndent ->
-    return $skip unless statements#
+  PushIndent NestedBlockStatement*:statements TrailingNestedComments?:trailing PopIndent ->
+    return $skip unless statements# or trailing?#
```
edemaine: `?`s seem unnecessary; `TrailingNestedComments` always matches. Unless you intended otherwise.

source/parser.hera
```diff
@@ -3370,6 +3375,27 @@ NestedBlockStatement
...
+    out.push(inline) if inline
+    out.push(...lines) if lines?#
+    return out
+
+TrailingInlineComment
+  _?:ws SingleLineComment:c &EOL ->
+    return [ws, c]
+
+TrailingNestedCommentLine
+  EOL:eol Indent:indent SingleLineComment:c &EOL ->
```
edemaine: Maybe a `#` comment that we can't use `Nested` because that uses `EOS` which consumes comments.

source/parser.hera
```diff
@@ -3370,6 +3375,27 @@ NestedBlockStatement
       ...statements.slice(1).map (s) => ["", ...s],
     ]
 
+# Captures trailing `// ...` comments at the end of an indented block: an
+# optional inline comment on the last statement's line, plus any number of
+# standalone same-indent comment lines. The lookahead `&EOL` keeps these from
+# eating into content that belongs to the surrounding block. (issue #832)
```
edemaine: I actually don't understand what the `EOL`s are doing. Doesn't `SingleLineComment` always eat all non-`\r`/`\n` characters so is guaranteed to be followed by an `EOL`?

source/parser.hera
```diff
@@ -3370,6 +3375,27 @@ NestedBlockStatement
...
+    out.push(...lines) if lines?#
+    return out
+
+TrailingInlineComment
+  _?:ws SingleLineComment:c &EOL ->
+    return [ws, c]
+
+TrailingNestedCommentLine
+  EOL:eol Indent:indent SingleLineComment:c &EOL ->
+    return $skip unless indent.level is state.currentIndent.level
```
edemaine: I wonder whether `>=` would make sense? Deeper comments should also be part of the block?

source/parser.hera
```diff
@@ -9838,13 +9870,25 @@ TrackIndented
     state.indentLevels.push(indent)
     return $1
 
+# Like TrackIndented, but for a deeper-indent comment-only line. Used so an
+# indented block whose body is only comments still pushes an indent. (#832)
+TrackCommentLine
```
edemaine: Does this need to be not-cached?

source/parser.hera
```diff
@@ -5625,6 +5651,12 @@ CaseClause
   }
   # NOTE: Added "when" from CoffeeScript. `when` always inserts `break;`.
   When CaseExpressionList:cases IgnoreColon InsertOpenBrace ( ThenClause / BareBlock ):block InsertBreak:b InsertNewline InsertIndent InsertCloseBrace ::WhenClause ->
+    // If the block ends with a `// ...` comment (issue #832), prefix the
+    // inserted `;break;` with a newline so it isn't swallowed by the comment.
+    if hasTrailingComment block
+      bIdx := $0.indexOf b
+      b = { ...b, token: "\n" + b.token }
+      $0[bIdx] = b
```
edemaine: For caching, I don't think we should modify `$0` in place. Or is the array created every time?

Also, this makes me a little worried that there might be other cases where we add content on the next line. For example, when adding `else` branches to `if/unless`.

I guess we can't just add a newline to the trailing comments because then we might get a blank line, unless there were a way to *remove* the newline from the next EOL, but that seems hard...

An ugly but safe solution would be to add a node that we later check against its neighbor and it renders as `"\n"` or `""` depending. Or e.g. emit a special character like `\0` that means "newline if not followed by one". I used a similar (opposite) trick [here](https://github.com/search?q=repo%3Aedemaine%2Ftexlish%20%5C0&type=code).

test/try.civet
```diff
@@ -151,6 +151,59 @@ describe "try", ->
...
+    }
+    catch {
+      y()
+      // trailing inside catch
+    }
+  """
+
   testCase """
     catch with then
     ---
```
edemaine, line 209: Missing `switch ... when` tests.

edemaine (commented):
I tried some other examples and found some bugs:

```js
z = if x
  x // foo
```

`// foo` doesn't appear in the output (but it [does currently](https://civet.dev/playground?code=eiA9IGlmIHgNCiAgeCAvLyBmb28%3D)).

```js
class Foo
  // foo
```

Comment stays outside class block.

source/parser.hera
edemaine: I asked GPT to look for places where we append code that might be susceptible to bugs in this way. It didn't find any inputs that failed, but found a bunch of potential risk surfaces:

---

I found a few spots worth auditing. The main ones are direct suffix appends to existing `children`, because those can land after a moved trailing `//` comment:

- `source/parser/declaration.civet:417` and `source/parser/declaration.civet:424` append `";"` to bare `else`/`then` blocks. This is the closest analogue to the fixed `braceBlock` case. I couldn’t trigger bad output with current `unless` declaration-condition cases, but if either block can now have trailing comments in `children`, the semicolon should use the same `hasTrailingComment` newline treatment.

- `source/parser/function.civet:783` and `source/parser/function.civet:785` append collected result / `;return results;` to iteration statement children. Tested indented and `then` loop bodies with trailing `//`; current output parses, but this is a high-risk append point if comments ever attach to the statement rather than its block.

- `source/parser/function.civet:493` appends an implicit `else { ... }` after an `if` with no else. This currently relies on `braceBlock(exp.then)`; with the new guard in `braceBlock`, my probes were OK.

- `source/parser/function.civet:554` checks `hasTrailingComment node.block.expressions`, but this branch now stores trailing nested comments in `block.children`, not `expressions`. That may be logically stale. It did not produce same-line-after-`//` output in my probes, but it is probably the wrong place to look now.

- Delimiter appends via `append(...)` in `source/parser.hera:3689`, `source/parser.hera:3720`, `source/parser.hera:3822`, and the type-list equivalents around `source/parser.hera:9024`. I tested arrays/tuples/objects with trailing `//`; commas currently land before the comment, so these look safe.

I also verified the already-covered `;break;` path with an implicit-returning `switch` inside a result-collecting loop. It emitted a newline before `;break;`, so that fix is doing the right thing.

---

So I'm still concerned, and would prefer a safer solution, but I'm happy that it didn't find anything actually broken. `hasTrailingComment node.block.expressions` sounds like it might be incorrect though.

PR #2122, opened 2026-05-25
Docs: Wrap top-level yield in a generator before Prettier in the playground

### LINKED ISSUE #1673: `yield` with array literals compiles incorrectly
Input:
```js
yield [2,3]
```

Actual output:
```js
yield[(2, 3)];
```

Expected output:
```js
yield[2, 3];
```

STRd6:
Closes #1673.

Prettier misparses `yield` outside a generator (it isn't legal there), turning `yield [a, b]` into `yield[(a, b)]` and `yield await [foo, bar]` into `yield[(await foo, await bar)]`. The underlying upstream issue is [prettier/prettier#16167](https://github.com/prettier/prettier/issues/16167).

Both the in-browser playground worker and the SSR `<Playground>` code-block path now wrap the TS output in `function*\$civetYieldWrap(){…}` when `ast.topLevelYield` is set, format, and strip the wrapper (function header, trailing `}`, one indent level). Non-yield snippets bypass the wrap and format exactly as before.

### DIFF
--- a/civet.dev/.vitepress/utils/compileCivet.ts
+++ b/civet.dev/.vitepress/utils/compileCivet.ts
@@ -6,15 +6,37 @@ export function compileCivet(
   parseOptions: any
 ) {
   // This SSR rendering code must be synchronous because of Markdown API
-  let tsCode = civetInstance.compile(code, { parseOptions, sync: true });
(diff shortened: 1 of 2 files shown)


edemaine:

PR #2123, opened 2026-05-26
Fix implicit extension and directory resolution when typechecking

edemaine:
@vendethiel reported that implicit extensions weren't resolving when typechecking.

This PR updates the Civet TS service and incremental `civet --typecheck` resolver so that unresolved imports are checked against registered transpiler extensions after normal TypeScript resolution fails. This makes Civet typechecking behave more like bundler resolution for `.civet` sources.

Effective improvements:

- Supports implicit Civet extensions, e.g. `import "./dep"` resolving to `./dep.civet`
- Supports directory imports (with implicit `index.civet`) containing dots — previously this was under a "no extension" guard for no apparent reason
- Adds directory import support in the incremental typecheck path, e.g. `import "./dep"` resolving to `./dep/index.civet` — previously this was only in the host path
- Preserves TypeScript’s failed resolution result when all resolution fails, keeping better failure metadata and diagnostics

While here, I factored out the resolver into a separate shared module between the incremental/host paths. We should probably factor out more in the future, to reduce duplication.

### DIFF
--- a/source/ts-service/typecheck.civet
+++ b/source/ts-service/typecheck.civet
@@ -25,6 +25,7 @@ ts from typescript
   mogrifyPackageJsonImports
   parseTsConfigForCivet
 } from ./config.civet
+{ resolveTranspiledModuleName } from ./resolve.civet
 
 export interface IncrementalTypecheckOptions
   /** Plugins providing transpilers for `.civet`, `.hera`, etc. */
@@ -59,7 +60,6 @@ export function makeIncrementalTypecheckProgram(
   options: IncrementalTypecheckOptions
 ): IncrementalTypecheckResult
   transpilers := buildTranspilers options.plugins
-  extraExtensions := Array.from transpilers.keys()
   parsed := parseTsConfigForCivet projectPath, transpilers, { tsConfig: options.tsConfig }
 
   compilerOptions: ts.CompilerOptions := {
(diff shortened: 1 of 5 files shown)


source/ts-service/typecheck.civet
```diff
@@ -127,6 +127,8 @@ export function makeIncrementalTypecheckProgram(
     sf
 
   origFileExists := baseHost.fileExists.bind baseHost
+  /* c8 ignore next -- modern TS hosts provide directoryExists; fallback supports older/custom hosts. */
```
vendethiel, line 130: Fallback?

edemaine: `?? => false`

STRd6 (approved):
Thanks! 

Greptile may have some actionable feedback.

We could consider putting this behind the `moduleResolution: "bundler" | "node"` option eventually.

edemaine:
> We could consider putting this behind the `moduleResolution: "bundler" | "node"` option eventually.

Good idea! It turns out to be fairly messy: summarizing https://www.typescriptlang.org/docs/handbook/modules/reference.html

moduleResolution Setting | Context | Extensionless Paths Support | Directory Modules (index.js) Support
-- | -- | -- | --
`node16` / `nodenext` | ECMAScript `import` syntax | ❌ | ❌
  | CommonJS `require` syntax | ✅ | ✅
`bundler` | Both (`import` & `require`) | ✅ | ✅
`node10` a.k.a. `node` | Both (import & require) | ✅ | ✅

Anyway, this is now implemented, even with `classic` support. Also addressed Greptile's helpful comment.

@greptileai Revise your review

PR #2125, opened 2026-05-27
Grammar cleanup; `////`+ is a comment; fix empty heregex

edemaine:
This PR updates Civet’s Hera dependency to 0.9.8 and applies the parser grammar cleanups enabled by the newer Hera syntax. It also ports over some of the heregex cleanup done in Hera 0.9.8, including fixing empties (`/// ///` for example) and changing 4+ `/`s into a comment.

- Bump `@danielx/hera` from `0.9.7` to `0.9.8`.
- Factor out repeated `::Type` annotations up to the rule level in `source/parser.hera`.
- Use Hera `///...///` regex syntax to make Civet’s long identifier regexes more readable.
- Extract shared identifier regex fragments for Unicode escapes and non-ASCII identifier/operator handling.
- Fixes empty heregex output so `/// ///` compiles to `/(?:)/` instead of invalid `//`, like CoffeeScript and Hera.
- Treats `////` and longer slash runs as comments while preserving exactly `///` for heregexes/directives. I think this is useful because I so presumably others like writing `/////////////////////` as a separator in files. And in the current state, `//////` (empty regex) and `/////` (heregex opening followed by the start of a comment) are extremely confusing and hard to read.
- Cleans up Civet’s heregex grammar implementation using a shared escape table and simpler rule actions, modeled after how I cleaned up the grammar in Hera.
- Adds tests for empty heregexes, slash-run comments, escaped tab/line/paragraph separators, and unescaped whitespace stripping.

## Addendum

- Fixed a parser ambiguity where consecutive heregex literals could be misread as binary division plus a `//` comment. For example, after parsing `///x///`, the next `///y///` could previously be split as `/` followed by `//y///`, so every other heregex was treated as a comment.
- Binary division now rejects a following `/`, preserving `///...///` for heregex parsing while keeping `//` and `////`+ comment behavior intact.

### DIFF
--- a/package.json
+++ b/package.json
@@ -134,7 +134,7 @@
     "@babel/core": "^7.29.0",
     "@babel/parser": "^7.29.2",
     "@danielx/civet": "0.11.6",
-    "@danielx/hera": "0.9.7",
+    "@danielx/hera": "0.9.8",
     "@types/assert": "^1.5.6",
     "@types/babel__core": "7.20.5",
     "@types/mocha": "^10.0.8",
@@ -153,7 +153,7 @@
     "vscode-languageserver-textdocument": "^1.0.8"
   },
   "peerDependencies": {
-    "@danielx/hera": "^0.9.7",
+    "@danielx/hera": "^0.9.8",
     "typescript": ">=4.5",
     "yaml": "^2.4.5"
   },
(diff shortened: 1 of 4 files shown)


STRd6 (approved):

edemaine:

PR #2126, opened 2026-05-27
Drop type-only named imports entirely in JS mode

STRd6:
## Summary

`import { type S, type T } from "foo"` in `--js` (strip-types) mode was emitting `import {} from "foo"` after per-specifier type stripping. esbuild faithfully preserves that as a runtime side-effect import — which broke the LSP browser worker's data-URL eval (and any other consumer that doesn't want a phantom side-effect import).

## Cause

`generate.civet` omits nodes marked `ts: true` in JS mode. Each `TypeAndImportSpecifier` already gets `ts: true`, so the inner specifiers vanished — but the surrounding `import { … } from "x"` framing has no such flag, leaving the empty wrapper.

## Fix

`markAllTypeImports` runs from the early-return paths of `processStaticImport`:

- Standalone NamedImports (no default / namespace binding) with all specifiers `type` → mark the entire `ImportDeclaration` `ts: true` so the whole statement is dropped in JS.
- Default + named (`import D, { type S } from "foo"`) → wrap just the trailing `, { type S }` portion in a `ts: true` node, leaving `import D from "foo"`.

TS mode output is unchanged in every case.

## Test plan

- [x] 3 new `testCase.js` cases in `test/types/js.civet` (standalone, shorthand, default+named)
- [x] Full suite: 4172 passing
- [x] `typecheck-diff` vs main: 0 regressions, 0 fixed

### DIFF
--- a/source/parser/declaration.civet
+++ b/source/parser/declaration.civet
@@ -504,10 +504,9 @@ function dynamizeFromClause(from: any)
   ["(", ...from, ")"]
 
 // True iff `node` is a NamedImports Declaration whose specifiers all carry
-// a `type` modifier (so they vanish in JS mode) and has no `...rest`.
+// a `type` modifier (so they vanish in JS mode). Returns false on empty
+// specifiers so `import {} from "x"` survives as a side-effect import.
 function isAllTypeNamed(node: any): boolean
-  return false unless node?.specifiers
-  return false if node.rest
   specs := node.specifiers
   return false unless specs.length
   specs.every (s: any) => s.ts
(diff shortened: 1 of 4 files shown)


source/parser/declaration.civet
```diff
@@ -503,18 +503,58 @@ function dynamizeFromClause(from: any)
     from.push ", {", assert.keyword, ":", assert.object, "}"
   ["(", ...from, ")"]
 
+// True iff `node` is a NamedImports Declaration whose specifiers all carry
+// a `type` modifier (so they vanish in JS mode). Returns false on empty
+// specifiers so `import {} from "x"` survives as a side-effect import.
+function isAllTypeNamed(node: any): boolean
```
edemaine: `any` should probably be `ImportDeclaration["imports"]` or a new type reflecting that.

Also not sure why "named". Maybe "imports" should be mentioned in the function name?

Actually, maybe this should be inlined: it's just `node.specifiers.every .length`

source/parser/declaration.civet
```diff
@@ -503,18 +503,58 @@ function dynamizeFromClause(from: any)
...
+  specs := node.specifiers
+  return false unless specs.length
```
edemaine: Seems unnecessary; `every` will handle empty case

source/parser/declaration.civet
```diff
@@ -503,18 +503,58 @@ function dynamizeFromClause(from: any)
...
+
+// Drop the named-imports portion in JS mode when every specifier is
+// `type`-prefixed (otherwise per-spec stripping leaves `import {} from "x"`,
+// which is a runtime side-effect import esbuild faithfully preserves).
+function markAllTypeImports(decl: ImportDeclaration): ImportDeclaration
+  return decl if decl.ts
+  { imports } := decl
+  return decl unless imports
+
+  i := imports as any
```
edemaine: Not a fan of the `any` casting

source/parser/declaration.civet
```diff
@@ -503,18 +503,58 @@ function dynamizeFromClause(from: any)
...
+    c?.type is "Declaration" and c?.specifiers and c is not i.binding
+  return decl unless namedIdx > 0 and isAllTypeNamed children[namedIdx]
+
+  // ImportClause assembles children as [binding, ...rest] where rest is the
+  // 4-tuple [ws, ",", ws, named] — so the trailing slice is namedIdx-3..end.
+  tail := children.slice(namedIdx - 3)
+  newImportsChildren := [
+    ...children.slice(0, namedIdx - 3)
+    { ts: true, children: tail }
+  ]
```
edemaine: I imagine a little restructuring at construction time could make this cleaner.

source/parser/declaration.civet
STRd6: `[].every(…)` is `true`, so without it `import {} from "x"` looks all-type and gets dropped — but it's a side-effect import we should keep.

edemaine: Oops, my bad. Indeed, this is important for preserving side effects, and I'm glad you tested it.

PR #2127, opened 2026-05-27
LSP: Hera-aware grammar-level features for `.hera`

STRd6:
LSP grammar-level features for `.hera` (definition, references, document symbols, completions, hover) plus syntax highlighting in Zed and VS Code.

## Cause
The LSP transpiles `.hera` → Hera → Civet → TSX and answers requests through the composed sourcemap. For grammar tokens (rule references) that map is coarse, so ctrl-click / hover land at the wrong rule, the wrong column, or inside `@danielx/hera/lib`.

## Approach
- **LSP server** — new `hera-analyzer.civet` parses `.hera` with the *project's* `@danielx/hera` (resolved like the project's Civet) and answers grammar-level requests directly; handler-body positions fall through to the TS service. A cheap indentation/fence text scan gates grammar vs. handler (no AST, so it works mid-parse).
- **tree-sitter-hera** — new grammar classifying the Hera surface; embedded handler/code regions are injected into the Civet grammar (Zed highlighting).
- **Zed** — adds a `Hera` language route + grammar.
- **VS Code** — registers the `hera` language with a TextMate grammar (handler/```​code blocks embed `source.civet`) and routes `.hera` to the LSP.

## Deferred (follow-ups, tracked in `notes/`)
- Rule doc-comments in hover, pending a Hera AST change — `notes/hera-ast-comments.md`.
- Re-pin the Zed grammar `rev` to the merge commit.


STRd6:
Re: double `await updating()` on the grammar-level fall-through — it's effectively a no-op, not a doubled round-trip. `executeQueue` deletes the update-status entry right after resolving it, so the second `updating(doc)` is `undefined?.promise` → `await undefined` (resolves immediately). The in-block await is intentional: it refreshes the Hera analyzer cache before querying it. Leaving as-is.

STRd6:
@greptile please update your review — the P1 (hover/completion now show rule productions), the weak completion-resolve assertion, the indent=3 gate, and the Zed grammar rev are all addressed; the double-`updating()` note is answered above.

PR #2128, opened 2026-05-27
Don't indent next line after keyword inside an Object

kutu:
This will prevent unnecessary indent on new line when creating an Object with keys that are keyword
```js
qwe :=
    asd: 1
    class: 2| // <- cursor here, press enter
        | // <-- cursor after pressing enter with old indentation rules, shouldn't be indent here
```

this works in sublime, please test in vscode

### DIFF
--- a/lsp/vscode/syntaxes/civet-configuration.json
+++ b/lsp/vscode/syntaxes/civet-configuration.json
@@ -28,7 +28,7 @@
     ["`", "`"]
   ],
   "indentationRules": {
-    "increaseIndentPattern": "^(?:.*(?:[({\\[]|=>>|->>|=>|->|\\bdo\\b|\\btry\\b|\\belse\\b|\\bcatch\\b|\\bfinally\\b|\\bloop\\b|\\bthen\\b)\\s*$|\\s*(?:export\\s+|async\\s+|default\\s+){0,3}(?:if|unless|for|while|until|switch|class|interface|enum|function|namespace|module|when|case|else|catch|finally)\\b(?!.*\\bthen\\b).*$)",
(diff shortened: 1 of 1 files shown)


edemaine (approved):
Thanks!

FWIW, implicit object literals can't have whitespace before the `:`, so I thought the `\s*` could be omitted... but braced object literals do allow `class : 2`, so it's probably better to leave it as is.

I added an E2E test, and verified that it works for me on VSCode.

edemaine:
Oops, apparently I can't push. Can you add this commit?

```patch
diff --git a/lsp/vscode/e2e/indentation.test.civet b/lsp/vscode/e2e/indentation.test.civet
index df59afa0..cd6cfce2 100644
--- a/lsp/vscode/e2e/indentation.test.civet
+++ b/lsp/vscode/e2e/indentation.test.civet
@@ -68,3 +68,7 @@ suite 'Auto-indent on Enter (issue #170)', =>
   test 'does NOT indent after an inline-then conditional', async =>
     indent := await indentAfterEnter 'if cond then x else y'
     assert indent is 0, `expected no indent after inline-then, got ${indent}`
+
+  test 'does NOT increase indent after a keyword object key', async =>
+    indent := await indentAfterEnter 'qwe :=\n    asd: 1\n    class: 2'
+    assert indent is 4, `expected object-key indent to stay at 4, got ${indent}`
```

kutu:
spaces work in playground
image
and LSP in sublime doesn't complain too

maybe i didn't understand the case when it won't work

kutu:
> Oops, apparently I can't push. Can you add this commit?

the change is small, just do it on your own, and delete my pull request

edemaine:
OK, created #2129 as replacement.

PR #2133, opened 2026-05-30
LSP: pull-model diagnostics (`textDocument/diagnostic`)

STRd6:
## Cause

The Civet LSP only *pushes* diagnostics (`publishDiagnostics`). Clients that prefer the LSP 3.17 **pull model** — notably the `mcp-language-server` bridge used for agent workflows — get no `diagnosticProvider`, so they fall back to a cache of the last push. That cache lags the debounced push pipeline (and isn't republished for unopened consumer files), which is why MCP diagnostics didn't refresh on edits.

## Approach

- Advertise a `diagnosticProvider` capability (`interFileDependencies: true` — our results depend on other files via cross-file TS types).
- Handle `textDocument/diagnostic` by computing against the **current** document at request time, so an edit is reflected on the very next pull with no push race. Push still coexists for live editors (VS Code).
- Extract `computeDiagnosticsForDoc`, shared by the push path and the pull handler.

Tests cover the full report, synchronous edit reflection, the unopened-document case, and an empty-document case (which exercises the no-snapshot guard for real rather than relying on it being unreachable).

Bridge-side counterpart: STRd6/mcp-language-server#1 consumes the pull results.

### DIFF
--- a/civet.dev/integrations.md
+++ b/civet.dev/integrations.md
@@ -163,3 +163,4 @@ hx --health civet
 ## Tooling
 
 - [YavaScript](https://github.com/suchipi/yavascript) is a standalone script runner supporting Civet (no Node required)
+- [mcp-language-server](https://github.com/STRd6/mcp-language-server) exposes `civet-lsp` to AI agents over the Model Context Protocol (diagnostics, hover, go-to-definition, references, edits)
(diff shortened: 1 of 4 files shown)


edemaine:
We should link to https://github.com/STRd6/mcp-language-server somewhere - maybe in Integrations?

STRd6:
Addressed review:
- Fixed the stale `computeDiagnosticsForDoc` JSDoc (it returns `Diagnostic[]`, never `undefined`, after the refactor) — 904551d.
- Added `mcp-language-server` under Integrations → Tooling, per @edemaine's suggestion — 883d8fb.

PR #2134, opened 2026-05-30
parser: `$loc` for postfix-conditional consequent braces

STRd6:
## Cause

A postfix `stmt unless cond` (or `if`/`while`/`for`) compiles to a single-line `if (!cond) { stmt }`. The synthetic opening brace was a bare string with no `$loc`, so it produced no sourcemap entry. Branch-coverage tools (c8 / v8-to-istanbul, via `originalPositionFor`) resolve the consequent block's start to the nearest mapped position — which, with no brace mapping, is the condition. The never-taken arm gets attributed to the condition, silently **hiding untested branches** (e.g. a `return [] unless x` guard whose `x`-falsy arm no test exercises reads as covered).

## Approach

Give the synthetic `{` the consequent statement's leading `$loc` so the block maps to its own source position, distinct from the condition. Falls back to a bare string when the statement has no located leaf.

Generated output **text is unchanged** (only sourcemap metadata changes) — the full suite passes. New test asserts the brace maps to the consequent via `originalPositionFor`, the exact lookup c8 uses; it's red without the fix (the brace has no mapping at all).

Surfaced while investigating why an LSP guard's branch read as covered (DanielXMoore/Civet#2133).

### DIFF
--- a/source/parser/lib.civet
+++ b/source/parser/lib.civet
@@ -199,9 +199,22 @@ function addPostfixStatement(statement: ASTNode, ws: ASTNode, post: IterationSta
     ["", statement]
   ]
 
+  // Give the synthetic opening brace the consequent's leading $loc. The
+  // generated `if (!c) { stmt }` is single-line, so without a located `{`
+  // the brace has no source mapping and collapses onto the condition —
+  // which makes branch-coverage tools (c8) attribute the consequent arm to
+  // the condition, hiding the gap. Falls back to a bare string when the
+  // statement has no located leaf.
+  located := gatherRecursive(statement, (n) => "$loc" in n).0
+  open: ASTNode :=
+    if located and "$loc" in located and located.$loc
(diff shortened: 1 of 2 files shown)


STRd6:
Closing: the premise is wrong. Verified end-to-end (compile a postfix-guard fixture with old vs fixed `dist/civet`, run c8 over the output — also through the register hook via `CIVET_SOURCE=./dist/main.js`) that this change produces **byte-identical** c8 coverage. c8 already reports the untested consequent arm as missed for single-line `if (!c) { stmt }`; v8-to-istanbul derives the branch from the consequent's inner statement, not the synthetic brace's $loc. The 'hiding' I'd attributed to the brace was actually statement/line coverage of the *multi-line* block form, unrelated to $loc. The change is at most a minor sourcemap-accuracy tweak and isn't worth carrying on a false premise.

PR #2135, opened 2026-05-30
docs: coverage mechanics guide for agents

STRd6:
Adds `notes/coverage-mechanics.md` (linked from CONTRIBUTING.md → Coverage) documenting how c8 actually sees Civet code, to prevent the misdiagnosis that just happened in a session:

- Branch coverage of single-line conditionals already works; a synthetic brace's `$loc` does **not** change c8 output (verified byte-identical). The "hiding" people notice is *statement/line* coverage attributing a single-line consequent to the condition line — a different thing.
- The register hook compiles `.civet` with the **installed** `@danielx/civet`, not `dist/` (use `CIVET_SOURCE=./dist/main.js` to test a local compiler); `lsp/server` is the exception (uses `../../dist/civet`).
- c8 reads maps via `originalPositionFor` (greatest genCol ≤ query, no interpolation), which differs from Civet's interpolating `remapPosition` — don't use the latter to predict c8.
- To check whether a compiler change moves coverage, diff c8 output, not sourcemap segments.

### DIFF
--- a/notes/coverage-mechanics.md
+++ b/notes/coverage-mechanics.md
@@ -20,9 +20,11 @@ reporting the branch correctly all along.
 - The **register hook** (`build/register.js` → `build/cache-utils.js`) compiles
   `.civet`/`.hera` at test time with the package resolved from
   `./node_modules/@danielx/civet` — a pnpm symlink to the **installed previous
-  release**, NOT your `dist/` build. So a change to `source/parser/**` does
-  **not** affect the sourcemaps c8 sees for `source/**` in root coverage until
-  it's released and the dep is bumped.
+  release** of the compiler, NOT your `dist/` build. Coverage still measures
+  your **current source** (that's what gets compiled and run); what lags is the
+  *compiler*. So a change to how `source/parser/**` emits code or sourcemaps
+  won't affect the maps c8 sees for `source/**` in root coverage until that
+  change is released and the dep is bumped.
 - `lsp/server` is the exception: its build uses `../../dist/civet` (your build),
   so compiler changes *do* affect lsp/server coverage.
 - To run your **local** compiler through the register hook (e.g. to test a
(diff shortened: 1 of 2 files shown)


notes/coverage-mechanics.md
```diff
@@ -0,0 +1,75 @@
...
+  directly (per-file `s`/`b`/`f` hit maps) rather than the aggregate pass/fail.
+
+## Which compiler compiles what (this is the subtle one)
+
+- The **register hook** (`build/register.js` → `build/cache-utils.js`) compiles
+  `.civet`/`.hera` at test time with the package resolved from
+  `./node_modules/@danielx/civet` — a pnpm symlink to the **installed previous
+  release**, NOT your `dist/` build. So a change to `source/parser/**` does
+  **not** affect the sourcemaps c8 sees for `source/**` in root coverage until
+  it's released and the dep is bumped.
```
edemaine: Is this how we should be measuring coverage? Shouldn't we be measuring the current code for coverage purposes (like `self:test`)?

edemaine (approved):

notes/coverage-mechanics.md
STRd6: It's the coverage of the current code built with the packaged compiler. The comment is about how if there is an issue with sourcemapping then that won't get included until we bump the packaged civet version.

edemaine: Oh, I see. That makes sense. Maybe could be clarified? Or maybe I misread.

STRd6:
Addressed review (50e6a00):
- Pointed the `c8 ignore` convention reference at CONTRIBUTING.md → Coverage (CLAUDE.md isn't tracked).
- Clarified the register-hook section per @STRd6: coverage measures the **current source**; only the *compiler* lags — a sourcemap/emit change isn't reflected until the packaged civet is bumped.

PR #2139, opened 2026-05-31
feat: infer @-param types from class field declarations

### LINKED ISSUE #1745: Use types from class declaration if present when using `@vars` in class instance methods / constructor
```civet
class Type
  name: string = "unknown"
  emoji: string = "❓"
  description: string = ""

  @({@name, @emoji, @description})
```

It would be neat to use the types of the instance properties automatically when using `@var` parameters.

STRd6:
Closes #1745

## Summary
When a class instance method or constructor uses `@var` parameters, the parameter now inherits its type from the matching class field declaration (if the field has an explicit type and the parameter does not). This works for simple `@(@name)`, destructured `@({@name})`, and array `@([@name])` forms; explicit parameter types still take precedence.

## Test plan
- [x] New test describe block "automatic @-param typing from class fields" in `test/types/class.civet` (6 cases)
- [x] Full test suite (`pnpm test`): 4259 passing, 0 failing

### DIFF
--- a/source/parser/function.civet
+++ b/source/parser/function.civet
@@ -7,6 +7,7 @@ import type {
   ASTNode
   ASTNodeObject
   ASTRef
+  AtBindingProperty
   Binding
   BindingElement
   BindingIdentifier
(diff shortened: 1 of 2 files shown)


STRd6:
@greptile Update review

PR #2143, opened 2026-06-04
feat: allow yield in fat-arrow functions (#2140)

### LINKED ISSUE #2140: Suggestion: allow yield in arrow functions
Apparently, there's a request for this in ECMAScript (TC39), but it's been languishing for a long time. I'm wondering if it's possible for civet to generate code that, in effect, does this.

The main reason I want this is so that a new 'this' isn't created, i.e. inside my object methods, if I create and use a generator, I can refer to properties of the object without using the "self = this" hack where I have to use self in place of @.

Something to think about, at least.

STRd6:
Closes #2140.

## Summary

Allows `yield` inside `=>` arrow functions by compiling to a generator function expression bound to the enclosing `this`:

```civet
=> yield @x
###
(function*() {return  yield this.x}).bind(this)
```

When the arrow body doesn't reference `this`, the `.bind(this)` wrapper is omitted.

Implements the feature @edemaine sketched in #2140, matching CoffeeScript 1's `=> yield @` behavior.

## Test plan

- [x] Full test suite passes (4293 passing)
- [x] New tests in `test/yield.civet` cover simple, @-property, parameters, multi-yield, assignment, async, yield*, and class-method cases
- [x] Existing "yield forbidden in arrow" tests updated to expect the new compiled output

### DIFF
--- a/source/parser/function.civet
+++ b/source/parser/function.civet
@@ -1233,6 +1233,70 @@ function findSuperCall(block: BlockStatement): number
   else
     -1
 
+/**
+ * Convert an arrow function whose body contains `yield` into a generator
+ * function expression, wrapped with `.bind(this)` so that `this` references
+ * inside the body still refer to the enclosing scope's `this` (matching the
+ * lexical-`this` semantics of the original arrow function).
+ *
+ * Issue #2140 / TC39 proposal for arrow generators.
+ *
+ *   => yield @x
+ *   ↓↓↓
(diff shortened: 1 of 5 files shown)


STRd6:
@claude make code comments much more succinct

STRd6:
@claude ensure typecheck passes

edemaine:
Awesome, thanks!

PR #2148, opened 2026-06-11
unplugin: support SolidStart / vinxi filesystem routers

### LINKED ISSUE #1319: Solid Start file router
Civet does work with Solid Start's filesystem router. 
Apparently `/src/routes/example.civet` loads but without transpilation.

Current theory is that the file is loaded with a query string like `?pick=...` and that might break the load matching in the unplugin.

STRd6:
Fixes #1319 (same report as [solid-start#1549](https://github.com/solidjs/solid-start/issues/1549)).

**Cause:** two independent breaks.
1. vinxi's filesystem routers ([`analyzeModule`](https://github.com/nksaraf/vinxi/blob/main/packages/vinxi/lib/fs-router.js)) enumerate a route module's exports by reading the file raw off disk and parsing it with esbuild's tsx loader. Civet source fails to parse, `addRoute` swallows the error, and the route is silently dropped — before any vite plugin runs.
2. The plugin's `.civet` → `.civet.tsx` id rewrite breaks vinxi's manifest correspondence: route chunks are looked up in vite's build manifest by the original `.civet?pick=...` path, crashing prod SSR with `Cannot read properties of undefined (reading 'file')`.

**Approach:**
- Wrap each vinxi fs-router's `toRoute` to compile `.civet` routes to TSX and substitute `fs.readFileSync` (exact-path match, restored in `finally`) for the duration of the analysis. Hooked from the vite `config` hook (`enforce: 'pre'` runs before vinxi's hook that first materializes routes) *and* from a construction-time microtask via `globalThis.app`, because `vinxi build` builds each router in a subprocess while the parent also materializes routes without running any vite hook.
- Skip the output-extension append when a vinxi router is detected (same shape as the existing esbuild carve-out). JSX still compiles because SolidStart forwards `extensions` to vite-plugin-solid.
- Warn on `ts: "preserve"` under vinxi: with `.civet` ids, neither vite core nor vite-plugin-solid strips TypeScript syntax, so output must be type-free (`civet` / `esbuild` / `tsc`).

Includes a SolidStart example (`integration/unplugin-examples/solid-start`) with a Civet page route and API route — verified with `vinxi build` + serving the output, and `vinxi dev` — plus unit tests for the router patching and id behavior, and README/docs updates.

Longer term this wants an upstream analysis hook (sketched in the linked solid-start issue discussion); this fix works without one and could feature-detect it later.

### DIFF
--- a/source/unplugin/unplugin.civet
+++ b/source/unplugin/unplugin.civet
@@ -8,10 +8,11 @@ import {
   // using ts-ignore because the version of @danielx/civet typescript is checking against
   // is the one published to npm, not the one in the repo
 } from '@danielx/civet/ts-diagnostic'
-import * as fs from 'fs'
-// Mutable fs module object (the `import * as` namespace is read-only in ESM);
-// patchVinxiFsRouter temporarily swaps readFileSync on it.
-import nodeFs from 'fs'
+// Default import gives the mutable fs module object (an `import * as`
+// namespace is read-only in ESM); vinxi route analysis temporarily swaps
+// readFileSync on it.
+import fs from 'fs'
+import type { Stats } from 'fs'
 import path from 'path'
 import { pathToFileURL } from 'url'
 import type {
@@ -107,24 +108,30 @@ type VinxiViteConfig
 
 patchedVinxiRouters := new WeakSet<VinxiFsRouter>
 
-// Active analysis substitutions: route file path → compiled TSX.  A single
-// permanent readFileSync intercept consults this map rather than each
-// analysis swapping the function pointer — concurrent toRoute calls (e.g.
-// two routers analyzing in dev) would race a save/restore pattern, since
-// restoring brackets an await.  set/delete are synchronous around the
-// analyzer's synchronous read, so entries can't leak across calls.
+// Active analysis substitutions: route file path → compiled TSX.  An
+// intercept on fs.readFileSync consults this map rather than each analysis
+// swapping the function pointer — concurrent toRoute calls (e.g. two
+// routers analyzing in dev) would race a save/restore pattern, since
+// restoring brackets an await.  The intercept is installed only while
+// analyses are in flight, counted so the original function is restored
+// when the last one finishes.
 vinxiAnalysisSubstitutions := new Map<string, string>()
-vinxiReadFileSyncPatched .= false
-function installVinxiReadFileSyncIntercept(): void
-  return if vinxiReadFileSyncPatched
-  vinxiReadFileSyncPatched = true
-  realReadFileSync := nodeFs.readFileSync
-  nodeFs.readFileSync = ((file: unknown, ...args: unknown[]) =>
-    if vinxiAnalysisSubstitutions.has file as string
-      vinxiAnalysisSubstitutions.get(file as string)!
-    else
-      (realReadFileSync as any).call nodeFs, file, ...args
-  ) as typeof nodeFs.readFileSync
+vinxiActiveAnalyses .= 0
+let vinxiRealReadFileSync: typeof fs.readFileSync
+function beginVinxiAnalysis(src: string, compiled: string): void
+  if vinxiActiveAnalyses++ is 0
+    vinxiRealReadFileSync = fs.readFileSync
+    fs.readFileSync = ((file: unknown, ...args: unknown[]) =>
+      if vinxiAnalysisSubstitutions.has file as string
+        vinxiAnalysisSubstitutions.get(file as string)!
+      else
+        (vinxiRealReadFileSync as any).call fs, file, ...args
+    ) as typeof fs.readFileSync
+  vinxiAnalysisSubstitutions.set src, compiled
+function endVinxiAnalysis(src: string): void
+  vinxiAnalysisSubstitutions.delete src
+  unless --vinxiActiveAnalyses
+    fs.readFileSync = vinxiRealReadFileSync
 
 queryPostfixRE := /\?.*$/s
 escapedHashRE := /\0#/g
@@ -165,7 +172,7 @@ function extractCivetFilename(id: string, outputExtension: string): {filename: s
       filename = filename[< -outputExtension#]
   {filename, postfix}
 
-function tryStatSync(file: string): fs.Stats?
+function tryStatSync(file: string): Stats?
   try
     // The "throwIfNoEntry" is a performance optimization for cases where the file does not exist
     return fs.statSync file, throwIfNoEntry: false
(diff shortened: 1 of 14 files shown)


source/unplugin/unplugin.civet
```diff
@@ -236,22 +253,77 @@ export rawPlugin := (options: PluginOptions = {}, meta: UnpluginContextMeta) =>
...
+      // Substitute the compiled TSX while the original toRoute runs its
+      // readFileSync + esbuild analysis on `src`.
+      realReadFileSync := nodeFs.readFileSync
+      nodeFs.readFileSync = ((file: unknown, ...args: unknown[]) =>
+        file is src ? compiled : (realReadFileSync as any).call(nodeFs, file, ...args)
+      ) as typeof nodeFs.readFileSync
+      try
+        await originalToRoute.call routes, src
+      finally
+        nodeFs.readFileSync = realReadFileSync
```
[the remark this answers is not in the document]
STRd6: Good catch — fixed in e0e3b109 using the suggested approach: a single permanent intercept consulting a `Map<string, string>` of active substitutions, with synchronous `set`/`delete` bracketing the analyzer's synchronous read. Added a concurrent-analysis regression test.

source/unplugin/unplugin.civet
```diff
@@ -236,22 +253,77 @@ export rawPlugin := (options: PluginOptions = {}, meta: UnpluginContextMeta) =>
...
+    patchedVinxiRouters.add routes
+    originalToRoute := routes.toRoute
+    routes.toRoute = async (src: string) =>
+      return originalToRoute.call routes, src unless src.endsWith civetSuffix
+      rawSource := decode await fs.promises.readFile src
+      vinxiCompileOptions ??= loadCompileOptions()
+      compiled: string := await civet.compile rawSource, {
+        ...await vinxiCompileOptions
+        filename: src
+        js: false
```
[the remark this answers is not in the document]
STRd6: Declining this one: vinxi's `addRoute` already does `console.error(e)` on anything `toRoute` throws (that's how the original esbuild parse errors surfaced in the issue repro), so a Civet compile error in a route is logged with its filename and location — identical to how a TSX syntax error in a route behaves. An extra warn here would double-log.

source/unplugin/unplugin.civet
```diff
@@ -236,22 +272,75 @@ export rawPlugin := (options: PluginOptions = {}, meta: UnpluginContextMeta) =>
...
+  function patchVinxiFsRouter(router: VinxiViteConfig['router']): void
+    routes := router?.internals?.routes
+    return unless routes? and typeof routes.toRoute is 'function'
+    return if patchedVinxiRouters.has routes
+    patchedVinxiRouters.add routes
+    originalToRoute := routes.toRoute
+    routes.toRoute = async (src: string) =>
+      return originalToRoute.call routes, src unless src.endsWith civetSuffix
+      rawSource := decode await fs.promises.readFile src
+      vinxiCompileOptions ??= loadCompileOptions()
```
edemaine: I don't understand why we need to `loadCompileOptions()` twice. Can we just use `compileOptions ?= loadCompileOptions()` in both places, so that whoever runs first loads the options one time? (and drop the `vinxiCompileOptions` variable)
We'd just need to drop the initializer from `compileOptions`'s declaration.

source/unplugin/unplugin.civet
```diff
@@ -236,22 +272,75 @@ export rawPlugin := (options: PluginOptions = {}, meta: UnpluginContextMeta) =>
...
-        civetConfigPath = await findInDir process.cwd()
-      if civetConfigPath
-        compileOptions = await loadConfig civetConfigPath
-      // Merge parseOptions, with plugin options taking priority
-      compileOptions.parseOptions = {
-        ...compileOptions.parseOptions
-        ...options.parseOptions
-      }
+      compileOptions = await loadCompileOptions()
       compileOptions.threads = options.threads if options.threads?
```
edemaine: Shouldn't this line move into the function too?

source/unplugin/unplugin.civet
```diff
@@ -9,6 +9,9 @@ import {
   // is the one published to npm, not the one in the repo
 } from '@danielx/civet/ts-diagnostic'
 import * as fs from 'fs'
+// Mutable fs module object (the `import * as` namespace is read-only in ESM);
+// patchVinxiFsRouter temporarily swaps readFileSync on it.
+import nodeFs from 'fs'
```
edemaine: Why do we need both imports? Can we just do the latter?

source/unplugin/unplugin.civet
```diff
@@ -90,6 +93,39 @@ type CacheEntry
...
+function installVinxiReadFileSyncIntercept(): void
+  return if vinxiReadFileSyncPatched
+  vinxiReadFileSyncPatched = true
+  realReadFileSync := nodeFs.readFileSync
+  nodeFs.readFileSync = ((file: unknown, ...args: unknown[]) =>
+    if vinxiAnalysisSubstitutions.has file as string
+      vinxiAnalysisSubstitutions.get(file as string)!
+    else
+      (realReadFileSync as any).call nodeFs, file, ...args
+  ) as typeof nodeFs.readFileSync
```
edemaine: I don't like the (revised) idea of permanently modifying `fs.readFileSync` in this way.

To handle concurrent calls, can we use a counter, and reset when it returns to zero?

edemaine (commented):
Thanks for taking this on!  Are you working on an upstream patch too?

I thought Tanstack Start used Vinxi, but apparently they migrated away last year. Nonetheless, we should probably test Tanstack Start with the unplugin at some point.

STRd6:
> Are you working on an upstream patch too?

Yes: https://github.com/solidjs/solid-start/pull/2156

source/unplugin/unplugin.civet
STRd6: Done in f3033abe: the intercept is now installed on the 0→1 transition of an in-flight-analysis counter and the original function is restored when it returns to zero (kept the substitution map within the window so overlapping analyses still can't race the pointer swap). Added a test that deterministically overlaps two analyses and asserts the original `readFileSync` identity is restored at idle.

source/unplugin/unplugin.civet
STRd6: Done in f3033abe — `compileOptions` now holds the shared `Promise<CompileOptions>` with `compileOptions ?= loadCompileOptions()` at both sites, and `vinxiCompileOptions` is gone. Consumers await it where needed (one local in the TSService setup, `...await compileOptions` in `load()`). One behavior note: buildStart previously re-loaded the civet config on every build; with `?=` it loads once per plugin instance, which seems more correct anyway.

source/unplugin/unplugin.civet
STRd6: Done — the threads merge moved into `loadCompileOptions()` along with the parseOptions merge.

source/unplugin/unplugin.civet
STRd6: Done — single default import now (the namespace import was only still needed for `fs.Stats` in a type position, which a type-only `Stats` import covers).

edemaine:

PR #2152, opened 2026-06-11
Object semantics for `:` in export specifiers; object-construction exports

### LINKED ISSUE #1092: Extending import renaming shortcut `exportedName: importedName` to `export from`
### Input code
```ts
export { s: t } from ./stuff
```
### Expected output
```ts
export { s as t } from "./stuff"
```
[playground link](https://civet.dev/playground?code=ZXhwb3J0IHsgczogdCB9IGZyb20gLi9zdHVmZg%3D%3D)

STRd6:
Closes #1092

Per the discussion here and on the issue: in export braces, `:` follows **object semantics** — the key is the public (exported) name, the value is where it comes from, mirroring `module.exports = { publicName: localValue }`. Imports are unchanged (destructuring semantics: `import { s: t }` is still `s as t`). The unified reading: keys are always the module-interface names.

**Phase 1 — rename shorthand (inverted):**
```coffee
export { s: t }                    → export { t as s }
export { s: t } from "./m"         → export { t as s } from "./m"
export { type S: T }               → export { type T as S }
export { "😊": x }                 → export { x as "😊" }
```

**Phase 2 — object-construction exports:**
```coffee
export { a: {b, c} }
# → const a = {b, c};export { a as a }

export { a: {b, c} } from "./m"
# → import { b as b1, c as c1 } from "./m";const a = {b: b1, c: c1};export { a as a }

export { a: {b}, d } from "./m"   # simple specifiers stay static re-exports
# → import { b as b1 } from "./m";const a = {b: b1};export { a as a };export { d } from "./m"
```
With `from`, leaf identifier values name exports of the source module (deduped named imports under collision-safe refs); nesting works; import assertions carry over to the synthesized import. Values that aren't identifiers or nested object literals are compile errors, as are type-only construction exports.

This replaces the accidental (untested) dynamize behavior where `export { a: {b,c} } from "./m"` compiled to `export const {a:{b,c}} = await import("./m")` — under object semantics that desugar read the keys in the wrong direction. The explicit form still works. `export { s: t } = obj` (#1321) is genuine destructuring and is unchanged — same expression/pattern duality as `{x: y}` elsewhere in the language.

Array-literal values (`export { a: [b, c] } from "./m"`) still fall through to the old dynamize path; reserving/extending those is left for a follow-up.

### DIFF
--- a/civet.dev/reference.md
+++ b/civet.dev/reference.md
@@ -3203,7 +3203,11 @@ fs from fs/promises
 metadata from ./package.json with type: 'json'
 </Playground>
 
-### Import-Like Object Destructuring
+### Object-Destructuring Import
+
+Import braces act like object destructuring: the key is the name in the
+module and the value is the local binding (the dual of
+[Object-Constructing Export](#object-constructing-export)).
 
 <Playground>
 import {X: LocalX, Y: LocalY} from "./util"
@@ -3272,11 +3276,12 @@ export a, b, c from "./cool.js"
 export x = 3
 </Playground>
 
-### Export Renaming
+### Object-Constructing Export
 
-In export braces, `:` renames with object semantics: the key is the
-public (exported) name and the value is the local binding it refers to,
-mirroring `module.exports = {publicName: localValue}`.
+The dual of [Object-Destructuring Import](#object-destructuring-import):
+export braces act like constructing the module's interface object. `:`
+renames, where the key is the public (exported) name and the value is the
+local binding, mirroring `module.exports = {publicName: localValue}`.
 With `from`, the value is the name in the source module instead:
 
 <Playground>
@@ -3286,18 +3291,8 @@ export { type S: T }
 export { "string name": x }
 </Playground>
 
-::: info
-Imports and exports are duals: import braces act like *destructuring*
-(`import { s: t }` binds local `t` from the module's `s`), while export
-braces act like *constructing* the module's interface object.
-Either way, the key is the name in the module being imported from or
-exported, and the value is the local side.
-:::
-
-### Object-Construction Exports
-
-Extending the object analogy, a value in export braces can be an object
-literal built from local bindings, exported under the key's name:
+Extending the object analogy, a value can be an object literal built from
+local bindings, exported under the key's name:
 
 <Playground>
 export { a: {b, c} }
--- a/civet.dev/reference.md
+++ b/civet.dev/reference.md
@@ -3272,6 +3272,51 @@ export a, b, c from "./cool.js"
 export x = 3
 </Playground>
 
+### Export Renaming
+
+In export braces, `:` renames with object semantics: the key is the
+public (exported) name and the value is the local binding it refers to,
+mirroring `module.exports = {publicName: localValue}`.
+With `from`, the value is the name in the source module instead:
+
+<Playground>
+export { s: t }
+export { s: t } from "./stuff"
+export { type S: T }
+export { "string name": x }
+</Playground>
+
+::: info
+Imports and exports are duals: import braces act like *destructuring*
+(`import { s: t }` binds local `t` from the module's `s`), while export
+braces act like *constructing* the module's interface object.
+Either way, the key is the name in the module being imported from or
+exported, and the value is the local side.
+:::
+
+### Object-Construction Exports
+
+Extending the object analogy, a value in export braces can be an object
+literal built from local bindings, exported under the key's name:
+
+<Playground>
+export { a: {b, c} }
+</Playground>
+
+With `from`, the values name exports of the source module,
+which get imported first (nesting works too):
+
+<Playground>
+export { a: {b, c} } from "./m"
+export { a: {b}, d } from "./m"
+export { a: {b: {c}} } from "./m"
+</Playground>
+
+::: info
+Constructed exports are built once at module load, so unlike plain
+re-exports they are snapshots rather than live bindings.
+:::
+
 ### Export Default Shorthand
 
 Most declarations can also be `export default`:
--- a/source/parser.hera
+++ b/source/parser.hera
@@ -6878,13 +6878,14 @@ NamedExports
     return [open, first, ...rest, close]
 
 # https://262.ecma-international.org/#prod-ExportSpecifier
+# ImportAsToken allows `:` as shorthand for `as` (#1092), matching imports.
 ExportSpecifier
-  __:ws1 ( TypeKeyword __ )?:ts ModuleExportName:local ( __ As __ ModuleExportName )?:rename ObjectPropertyDelimiter:delim ->
+  __:ws1 ( TypeKeyword __ )?:ts ModuleExportName:local ( ImportAsToken __ ModuleExportName )?:rename ObjectPropertyDelimiter:delim ->
     // Local stays mangled (JS variable ref); external side uses propertyKey
     // so the exported name is the user-visible source.
     children :=
       if rename
-        [ws1, ts, local, rename[0], rename[1], rename[2], propertyKey(rename[3]), delim]
+        [ws1, ts, local, rename[0], rename[1], propertyKey(rename[2]), delim]
       else if local.unicode
         [ws1, ts, local, " as ", propertyKey(local), delim]
       else
--- a/source/parser.hera
+++ b/source/parser.hera
@@ -2476,11 +2476,8 @@ NWBindingIdentifier
     return { children: [$1], names: [] }
 
 AtIdentifierRef ::ASTRef
-  # Explicit `_` prefix even for Civet-only reserved words (`is`, `loop`, ...)
-  # which are valid JS so makeRef leaves them bare; a bare binding could
-  # shadow helpers like `var is`.
-  ReservedWord:r ->
-    return makeRef(`_${r}`, r)
+  # `@is`, `@loop`, etc. fall through to IdentifierName; makeRef +
+  # populateRefs keep the binding valid and unshadowed (see #1092 discussion).
   IdentifierName:id ->
     ref := makeRef(id.name)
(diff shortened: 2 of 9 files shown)


STRd6:
We may want to invert this to match object semantics.

test/export.civet
```diff
@@ -258,6 +258,46 @@ describe "export", ->
...
+    ---
+    export { s as t } from "./stuff"
+  """
+
+  testCase """
+    nested destructuring from still dynamizes
+    ---
+    export { a: {b, c} } from "./stuff"
+    ---
+    export const {a:{b, c}} = await import ("./stuff")
```
edemaine: This doesn't seem right. On the other hand, I'm not totally sure what it should mean. One possibility that seems natural:
```js
import { b, c } from "./stuff"
const a = {b, c}
export { a }
```
I'm surprised to see that `export { a: {b, c} }` doesn't already parse to the latter two lines. I think you added this to `import` but not `export` yet. So maybe both are for a future PR.

Another possibility for the test code would be a glob, but probably better spelled as `export { a{b,c} } from "./stuff"`:
```js
import { a } from "./stuff"
export a.{b,c}
// i.e.
export { a: a.b, a: a.c }
```

test/export.civet
```diff
@@ -258,6 +258,46 @@ describe "export", ->
...
+    ---
+    export { s as t } from "./stuff"
+  """
+
+  testCase """
+    colon rename shorthand without from
+    ---
+    export { s: t }
+    ---
+    export { s as t }
```
edemaine: I do think reversal makes sense. As mentioned in a comment below, we surely want to allow `export { a: {b, c} }` in the future, and this would only generalize
```js
export { s: t }
---
export { t as s }
```

STRd6:
Reworked to full inversion per the discussion: `:` in export braces now reads as object semantics (key = public name), and object-construction exports (`export { a: {b, c} }`, with and without `from`) are implemented per @edemaine's sketch. The pattern-dynamize fallthrough for object literals in `export … from` is gone. Note for @esthedebeste: the issue's example now compiles to `export { t as s } from "./stuff"` — the key is the re-exported name.

civet.dev/reference.md
```diff
@@ -3272,6 +3272,51 @@ export a, b, c from "./cool.js"
 export x = 3
 </Playground>
 
+### Export Renaming
```
edemaine: I wonder whether the two Export sections would be better combined into one section "Object-Like Export", analogous to [Import-Like Object Destructuring](https://civet.dev/reference#import-like-object-destructuring) (which should maybe be retitled "Object-Destructuring Import"? I guess it depends how you look at it).

civet.dev/reference.md
```diff
@@ -3272,6 +3272,51 @@ export a, b, c from "./cool.js"
...
+
+In export braces, `:` renames with object semantics: the key is the
+public (exported) name and the value is the local binding it refers to,
+mirroring `module.exports = {publicName: localValue}`.
```
edemaine: I hadn't thought of the `module.exports` mirror until Greptile mentioned it. Great analogy!

test/export.civet
```diff
@@ -258,6 +258,177 @@ describe "export", ->
...
+    ---
+    export { type S: T }
+    ---
+    export { type T as S }
+  """
+
+  testCase """
+    colon rename to string name
+    ---
+    export { "😊": x }
```
edemaine, line 304: What about without quotes? Ah, more complex case of that comes later.

test/export.civet
```diff
@@ -258,6 +258,177 @@ describe "export", ->
...
+    export var { s: t } = obj
+  """
+
+  describe "object-construction exports", ->
+    testCase """
+      local construction
+      ---
+      export { a: {b, c} }
+      ---
+      const a = {b, c};export { a as a }
```
edemaine, line 323: Too bad we can't write `export { a }` in this case, but it's tough the way refs work.

edemaine (commented):
Overall looks great! GPT found a couple bugs:

1. **[P1] Reserved export names compile to invalid bindings**

   `source/parser/declaration.civet:792` uses the public export name as the temp ref base. Since `ModuleExportName` accepts keyword-like `IdentifierName`s, valid inputs like:

   ```civet
   export { default: {b} }
   ```

   compile to invalid JS:

   ```js
   const default = {b};export { default as default }
   ```

   `esbuild` rejects this with `Expected identifier but found "default"`. Same issue hits `class`, `for`, etc. The temp ref needs a bindable/safe base for reserved words.

   [Erik] Maybe we should fix this for refs in general? So they always add 1 or 2 or ... if on  a list of reserved words?

2. **[P2] Comments around constructed specifiers are dropped**

   `source/parser.hera:6923` captures whitespace/comments around the key, colon, and value, but the constructed spec stores only `ws1`; then `source/parser/declaration.civet:821` and `source/parser/declaration.civet:853` rebuild output without those nodes. For example:

   ```civet
   export { a /*key*/ : {b} }
   ```

   compiles to:

   ```js
   const a = {b};export { a as a }
   ```

   Normal export specifier tests explicitly preserve comments, so this new path should either preserve them in the generated `const`/`export` or add coverage if the intended behavior differs.

   [Erik] Not mission critical.

Some of Greptile's comments seem good too.

source/parser.hera
```diff
@@ -2471,6 +2476,9 @@ NWBindingIdentifier
     return { children: [$1], names: [] }
 
 AtIdentifierRef ::ASTRef
+  # Explicit `_` prefix even for Civet-only reserved words (`is`, `loop`, ...)
+  # which are valid JS so makeRef leaves them bare; a bare binding could
+  # shadow helpers like `var is`.
```
edemaine: I can't say I understand this, but I might be missing something. Aren't helpers also refs, so they won't be duplicated? I don't think this special rule is needed anymore, with your ref extension. If it were, why wouldn't we need it in export too?

source/parser/ref.civet
```diff
@@ -16,7 +16,33 @@ import {
...
+    pos: 0,
+    input: name,
+  })
+
+// Words that can't be `const`-bound in JS output: reserved words
+// (including strict-mode ones, since Civet assumes strict mode) plus
+// `eval`/`arguments` which strict mode forbids as binding targets.
+// Broader in places than the Civet ReservedWord rule, which lacks
+// `implements`, `package`, `eval`, and `arguments`.
+jsReservedWord := /^(?:arguments|await|break|case|catch|class|const|continue|debugger|default|delete|do|else|enum|eval|export|extends|false|finally|for|function|if|implements|import|in|instanceof|interface|let|new|null|package|private|protected|public|return|static|super|switch|this|throw|true|typeof|var|void|while|with|yield)$/
```
edemaine: A Set would be faster I think.

source/parser.hera
STRd6: Confirmed — you're right, it was redundant. Removed in 5bd7c90d.

Traced it: the `_${r}` prefix was added long ago in 9875623f, predating makeRef's reserved-word handling (added this PR in 8e639841). Two mechanisms already cover it:

- **JS reserved words** (`@return`, `@var`) — makeRef still prefixes these, so they're unchanged.
- **Civet-only-but-valid-JS words** (`is`, `loop`, ...) — populateRefs collision-renaming handles them. The prelude (helper defs) is spliced into `statements` at lib.civet:2094, *before* populateRefs runs at :2109, so the `is` helper ref is in scope when `@is` resolves.

The "could shadow `var is`" worry doesn't materialize: with the prefix gone, `(@is) -> @is is 3` under `objectIs` compiles to `var is = Object.is; (function(is1){ this.is = is1; return is(this.is, 3 as const) })` — populateRefs renamed the binding to `is1`, helper call intact. And as you guessed, the construction-export path already lacks any such rule and behaves identically (`export { is: {b} }` under objectIs → `const is1 = {b}; export { is1 as is }`), which is the direct answer to "why wouldn't we need it in export too?"

Added regression tests for both the plain Civet-only case and the Object.is collision.

source/parser/ref.civet
STRd6: Done in 60a2d23d. Benchmarked it to be sure: the Set is ~1.5x the anchored regex on the realistic ~0%-reserved distribution (≈150M vs ≈95M ops/s), up to ~2x when more inputs hit. Absolute impact is tiny since makeRef isn't hot, but it's a free win and the single word list reads better than the `|`-wall.

source/parser/declaration.civet
```diff
@@ -776,13 +787,130 @@ function convertWithClause(withClause: WithClause, extendsClause?: [ASTLeaf, WSN
...
+    statements.push {
+      type: "Error"
+      message: "Object-construction exports cannot be type-only"
+    }
+  for each spec of constructed
+    ref := constructedExportRef spec
+    statements.push "const ", ref, " = ", spec.value, ";"
+    spec.children = [spec.ws, ref, " as ", exportedName(spec.exported), spec.delim] as Children
+  statements.push { type: "ExportDeclaration", children } as ExportDeclaration
+  statements
```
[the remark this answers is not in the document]
STRd6: Intentionally not short-circuiting. The LSP compiles with `errors: []` (non-throwing) and runs tsserver on the generated TSX, not the Civet AST — so desugaring past the error is what gives tsserver usable symbols. Verified: current output is `const a = {b}; export … { a as a }` (hover/completion work, Error node still flags it); an early return blanks the line to `""`. Added a comment (5d03e4ab) so it's not re-flagged.

civet.dev/reference.md
STRd6: Done in 08ad4ba7 — went with the fully-parallel pair: retitled the import section **Object-Destructuring Import** and merged the two export sections into **Object-Constructing Export**, cross-linked as duals.

PR #2153, opened 2026-06-11
Docs: One-line functions don't require `do`

### LINKED ISSUE #723: documentation nit, on one-line function definition syntax
The (relatively) new reference page shows three examples of one-line function definitions:
```
function do console.log "Anonymous"
function f do console.log "Named"
function f(x) do console.log x
```
strongly implying the `do` is necessary. But all three examples work perfectly fine without `do`.

STRd6:
Fixes #723

The One-Line Functions section showed only `do`-forms, implying `do` is required. It's only needed for anonymous one-liners whose body starts with an identifier (e.g. `function console.log x` parses `console` as the function name); named and parenthesized forms work without it, and tests lock that in. Lead with the plain forms and explain when `do` matters.

### DIFF
--- a/civet.dev/reference.md
+++ b/civet.dev/reference.md
@@ -1193,6 +1193,17 @@ function count()
 ### One-Line Functions
 
 <Playground>
+function f(x) console.log x
+function greet() console.log "Hello"
+function f sideEffect()
+</Playground>
+
+An identifier immediately after `function` is treated as the
+function's name, so an anonymous one-liner whose body starts with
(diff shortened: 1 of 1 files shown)


edemaine (commented):
We've debated in the past whether parameters' parentheses can be optional in some cases: https://github.com/DanielXMoore/Civet/discussions/407

I still think it's a good idea, but it would break the parameterless example: `function f sideEffect()`.

This has been the blocker in my mind for resolving this documentation issue.

PR #2158, opened 2026-06-15
Fix one-line then/else swallowing surrounding comma

### LINKED ISSUE #2157: Commas in one-line if/then/else
## Input

```coffee
[if counterRef then [counterRef, "++"], close]
```

## [Current Output](https://civet.dev/playground?code=W2lmIGNvdW50ZXJSZWYgdGhlbiBbY291bnRlclJlZiwgIisrIl0sIGNsb3NlXQ%3D%3D)

```ts
[(counterRef? ([counterRef, "++"], close):void 0)]
```

## Expected Output

```ts
[(counterRef? [counterRef, "++"]:void 0), close]
```

Curiously, [CoffeeScript just crashes](https://coffeescript.org/#try:%5Bif%20counterRef%20then%20%5BcounterRef%2C%20%22%2B%2B%22%5D%2C%20close%5D%0A).

## Another example

[Same issue](https://civet.dev/playground?code=W2lmIGNvdW50ZXJSZWYgdGhlbiBbY291bnRlclJlZiwgIisrIl0gZWxzZSBbXSwgY2xvc2Vd) with explicit `else` branch:

```coffee
[if counterRef then [counterRef, "++"] else [], close]
```

## Proposal

Especially in an array context, comma operator is counterintuitive. Probably we should forbid comma operators in all one-line then/else blocks? I imagine function call arguments are similarly broken.

STRd6:
Fixes #2157.

## Cause

A one-line `then`/`else`/block body parses through `DeclarationOrStatement` → `Statement` → `CommaExpressionStatement`, which greedily consumes a top-level comma operator that belongs to the enclosing array/call/argument list:

```coffee
[if counterRef then [counterRef, "++"], close]
# was: [(counterRef? ([counterRef, "++"], close):void 0)]   -- close swallowed
# now: [(counterRef? [counterRef, "++"]:void 0), close]
```

## Approach

`DeclarationOrStatement` is used only in single-line block bodies, so route its statement alternative to the existing `NoCommaStatement` (the no-comma variant already used by arrow bodies) instead of `Statement`. `Declaration` still allows commas, so loop-accumulator declarations (`coords := x, y for {x, y} of points`) keep working, and indented multi-line blocks are unaffected.

This is per @edemaine's diagnosis in the issue. One fix covers every same-line block context — if/else, `unless`, `for`/`while`/`loop`/`do` bodies, object-literal values — in both array and call-argument positions. Explicit parens remain an escape hatch for a deliberate comma body: `[if c then (a, b), d]`.

Statement-level `if x then a, b` now errors instead of emitting `if (x) a, b`, matching the issue's intent to forbid comma operators in one-line then/else.

A before/after diff over ~20 constructs showed exactly the bug fixes plus that one error; pipes, binary ops, trailing-call/switch/try expressionization, and for/while/loop/do one-liners are byte-identical. Typecheck diff: 799→799 (0 regressions).

### DIFF
--- a/source/parser.hera
+++ b/source/parser.hera
@@ -3406,9 +3406,13 @@ NonSingleBracedBlock ::BlockStatement
       children: [o, expressions, c],
     }
 
+# NOTE: NoCommaStatement (not Statement) so a single-line `then`/`else`/block
+# body doesn't swallow a top-level comma operator that belongs to a surrounding
+# array/call/argument list. Declaration still allows commas (loop accumulators
+# like `coords := x, y for ...`). (#2157)
 DeclarationOrStatement
   Declaration
-  Statement
+  NoCommaStatement
 
 # NOTE: SingleLineStatements includes the empty case
 SingleLineStatements
--- a/source/parser.hera
+++ b/source/parser.hera
@@ -3423,8 +3423,11 @@ SingleLineStatements
   # NOTE: Statement can start with __ via AssignmentExpression.
   # Force staying on the same line via !/\n/ assertion.
   ForbidNewlineBinaryOp ( _? SingleLineDeclarationOrStatement SemicolonDelimiter )*:stmts ( _? SingleLineDeclarationOrStatement SemicolonDelimiter? )?:last RestoreNewlineBinaryOp ::BlockStatement ->
-    expressions := [...stmts]
-    expressions.push(last) if last
+    // NOTE: the .map keeps `expressions` a loose union array (not a strict
+    // [prefix, statement, delim] tuple) so the optional final delimiter and the
+    // trailing-comment "\n" push below stay assignable.
+    expressions := stmts.map(([prefix, statement, delim]) => [prefix, statement, delim])
+    expressions.push([last[0], last[1], last[2]]) if last
 
     children := [expressions]
     children.push(["\n"]) if hasTrailingComment(expressions)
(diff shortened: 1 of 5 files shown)


source/parser.hera
```diff
@@ -3406,9 +3406,13 @@ NonSingleBracedBlock ::BlockStatement
       children: [o, expressions, c],
     }
 
+# NOTE: NoCommaStatement (not Statement) so a single-line `then`/`else`/block
```
edemaine, line 3414: Maybe worth explicitly mentioning that this rule is only used in a single-line context. (I'd even consider changing the name of the rule... But I'm not sure to what (something about SingleLine?). We're already not super consistent in whether we mention comma or not in rule names.)

edemaine (approved):

source/parser.hera
```diff
@@ -3215,7 +3215,7 @@ Block ::BlockStatement
 
   ThenClause
   # NOTE: !EOS prevents capturing a following unindented Statement
-  _?:ws !EOS DeclarationOrStatement:s ->
+  _?:ws !EOS SingleLineDeclarationOrStatement:s ->
```
edemaine: Hmm, now I wonder whether `!EOS` should be inside `SingleLineDeclarationOrStatement`... It's always used that way.

edemaine: Nice, that's a lot cleaner!

PR #2159, opened 2026-06-15
Fix dropped comments before later pattern/catch clauses

### LINKED ISSUE #2151: Comments before second and later `catch` clauses are dropped
A comment (or any leading trivia) before the second or later `catch` clause of a pattern-matching `try` is silently dropped from the output:

```civet
try
  foo()
catch null
  ;
// eslint-disable-next-line prefer-const
catch e
  bar()
```

compiles to

```js
try {
  foo()
}
catch(e1) {if(e1 === null) {
  ;
}
else  {const e = e1;
  bar()
}}
```

— the comment is gone. This breaks directive comments (`eslint-disable-next-line`, `@ts-expect-error`, istanbul ignores) as well as ordinary commentary. Noted originally in #1487, where it meant the lint error couldn't even be suppressed manually; the `prefer-const` side is fixed by #2150 but the comment loss is a separate bug.

Comments before a *single* `catch`, before the *first* clause of a multi-catch, before `finally`, and inside clause bodies are all preserved. The drop happens because `processTryBlock`'s pattern-matching transform rebuilds the merged catch clause from `cs[0].children[0]` (the first clause's leading whitespace/comments) plus each clause's patterns and block, discarding later clauses' leading trivia. A fix could prepend each clause's leading trivia into its corresponding switch clause (e.g. ahead of the `PatternClause`/`DefaultClause` children).

STRd6:
Fixes #2151.

## Cause
When a pattern-matching switch is rebuilt into an if/else chain in `processPatternMatching`, each clause's leading trivia is discarded — only `patterns` and `block` survive. So comments before the second-and-later clauses of a pattern-matching `try` (and a general `switch`) were silently dropped, breaking directive comments like `eslint-disable-next-line` and `@ts-expect-error`.

## Approach
- `processTryBlock` carries each later catch clause's leading trivia into the generated switch clause (the first clause's trivia is already consumed as the merged catch clause's leading whitespace).
- `processPatternMatching` gathers any `Comment` nodes from a clause's leading trivia and re-emits them before the `else` that introduces the clause.

This fixes both pattern-matching `catch` and general `switch` for second-and-later clauses.

### DIFF
--- a/source/parser/lib.civet
+++ b/source/parser/lib.civet
@@ -487,13 +487,18 @@ function processTryBlock($0: [ASTLeaf, undefined, BlockStatement, CatchClause[],
       children: [ "(", ref, ")" ]
       expression: ref
     defaultClause .= false
-    clauses: CaseBlock["clauses"] := cs.map (clause) =>
+    clauses: CaseBlock["clauses"] := cs.map (clause, i) =>
+      // The first clause's leading trivia becomes the merged catch clause's
+      // leading whitespace (below); carry later clauses' trivia into the
+      // generated switch clause so processPatternMatching can re-emit comments
+      // that would otherwise be dropped before the `else` (issue #2151).
+      leadingTrivia := i > 0 ? [clause.children[0]] : []
       if {type: "CatchPattern", patterns} := clause.binding?.parameter
         {
           type: "PatternClause"
           patterns
           block: clause.block
-          children: [ patterns, clause.block ]
+          children: [ ...leadingTrivia, patterns, clause.block ]
         }
       else
         defaultClause = true
@@ -526,7 +531,7 @@ function processTryBlock($0: [ASTLeaf, undefined, BlockStatement, CatchClause[],
 
         type: "DefaultClause"
         block: clause.block
-        children: [ "default: ", clause.block ]
+        children: [ ...leadingTrivia, "default: ", clause.block ]
 
     // If no default clause specified, add one that rethrows exception
     unless defaultClause
--- a/source/parser/lib.civet
+++ b/source/parser/lib.civet
@@ -488,17 +488,18 @@ function processTryBlock($0: [ASTLeaf, undefined, BlockStatement, CatchClause[],
       expression: ref
     defaultClause .= false
     clauses: CaseBlock["clauses"] := cs.map (clause, i) =>
-      // The first clause's leading trivia becomes the merged catch clause's
-      // leading whitespace (below); carry later clauses' trivia into the
-      // generated switch clause so processPatternMatching can re-emit comments
-      // that would otherwise be dropped before the `else` (issue #2151).
-      leadingTrivia := i > 0 ? [clause.children[0]] : []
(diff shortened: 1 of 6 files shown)


test/switch.civet
```diff
@@ -1675,6 +1675,29 @@ describe "switch", ->
...
     """
 
+    // #2151: leading comments before later pattern clauses were dropped
+    testCase """
+      preserves comments before later pattern clauses
+      ---
+      switch x
+        null
+          a()
+        // a comment here
```
edemaine, line 1685: What happens if the comment is indented inside the previous block? I assume it stays there, from previous PR, but maybe worth a test to ensure.

test/try.civet
```diff
@@ -475,6 +475,60 @@ describe "try", ->
       else  {throw e1}}
     """
 
+    // #2151: comments before second-and-later catch clauses were dropped
+    testCase """
+      preserves comments before later catch clauses
```
edemaine, line 505: Should test comment before first catch too. (It already worked but good for coverage/regression.)

source/parser/lib.civet
```diff
@@ -487,13 +487,18 @@ function processTryBlock($0: [ASTLeaf, undefined, BlockStatement, CatchClause[],
       children: [ "(", ref, ")" ]
       expression: ref
     defaultClause .= false
-    clauses: CaseBlock["clauses"] := cs.map (clause) =>
+    clauses: CaseBlock["clauses"] := cs.map (clause, i) =>
+      // The first clause's leading trivia becomes the merged catch clause's
+      // leading whitespace (below); carry later clauses' trivia into the
+      // generated switch clause so processPatternMatching can re-emit comments
+      // that would otherwise be dropped before the `else` (issue #2151).
+      leadingTrivia := i > 0 ? [clause.children[0]] : []
```
edemaine: I'm not a fan of relying on `children` layout like this. Could we instead add a property, maybe `leading` or `ws` or `leadingWs`, to `CatchClause`? ("Trivia" also doesn't seem normal wording here. We usually just call it "whitespace", admittedly a misnomer.)

source/parser/pattern-matching.civet
```diff
@@ -117,6 +117,16 @@ function processPatternMatching(statements: ASTNode): void
 
       l := clauses.length
       for each c, i of clauses
+        // Re-emit comments leading a second-or-later clause before the `else`
+        // that introduces it; otherwise the clause's leading trivia is dropped
+        // when the switch is rebuilt as an if/else chain (issue #2151).
+        if e?
+          comments := gatherRecursiveAll c.children[0], .type is "Comment"
```
edemaine: Again not a fan of assuming `children` layout here. I'm not even sure what's traversed in all cases, as Greptile comment alludes...

edemaine (commented):
Another case, possibly for a future PR: handling the first clause in `switch`:

```coffee
switch x
  // comment before first
  null
    a()
  undefined
    b()
```

The comment gets dropped, even with this PR. Maybe easy to add-on?

edemaine (approved):

PR #2161, opened 2026-06-16
Fix pattern-to-value conversion with anonymous `...`

edemaine:
This PR fixes pattern-to-value conversion for array destructuring with post-rest bindings, especially anonymous `...`.
~~It's currently based on #2160 because it touched similar code; should merge that first. Meanwhile, please look at just the last commit.~~ Rebased.

Previously, code that derived a value from an array binding pattern, such as implicit function returns and implicit `for` bodies, looked only at the lowered JS-compatible destructuring pattern. For post-rest patterns, that meant it could accidentally return the hidden rest array instead of the source-level trailing bindings.

This PR reconstructs the source-level array pattern when converting bindings back into values, while still emitting JS-compatible destructuring.

## Broken Examples Fixed

@STRd6 originally pointed out the first example in Discord.

| Civet | Before | After |
|---|---|---|
| `coordinates := for [..., x, y] of points` | implicit body was `[...]` / the hidden rest array, effectively collecting the prefix | implicit body is `[x, y]` |
| `pairs := for [a, ...rest, x, y] of points` | implicit body omitted trailing bindings, effectively `[a, ...rest]` | implicit body is `[a, ...rest, x, y]` |
| `pairs := for [...] of points` | implicit body could reference the hidden anonymous rest | implicit body is `[]` |
| `-> [..., x, y] := point` | implicit return was the hidden rest array / prefix | implicit return is `[x, y]` |
| `-> [a, ..., x] := point` | implicit return omitted `x` | implicit return is `[a, x]` |
| `-> [a,, ..., x] := point` | implicit return omitted `x` after an elision | implicit return is `[a,, x]` |
| `-> [...] := point` | implicit return could reference the hidden anonymous rest | implicit return is `[]` |
| `total := for sum [..., x] of y` | reduction used the wrong binding value | reduction uses `x` |
| `total := for sum [...] of y` | reduction could use an empty/hidden binding value | now reports `Empty binding pattern in loop with implicit body` |

## Implementation Notes

- New `arrayPatternValueElements` helper reconstructs the source-level array binding elements for value conversion.
- Track anonymous terminal `...` as `omittedRest` so emitted destructuring can omit it while pattern matching and type inference can still treat it as a rest pattern.
- `elements` now omits it too, not just `children`. Now `elements` appears in (actually is) `children`, which is usually assume.
- Update `patternAsValue` to use reconstructed array elements instead of the lowered JS destructuring shape.
- Keep post-rest pattern matching pointed at `blockPrefix.elements` directly instead of reaching through generated children.
- Tighten array binding pattern types around `ArrayBindingPatternElement` / `ArrayBindingPatternContent`.

## Tests

Added coverage for:

- implicit `for` bodies with anonymous and named post-rest array destructuring
- implicit function returns from anonymous post-rest array destructuring
- reductions over post-rest array destructuring
- empty anonymous rest behavior
- type inference for omitted rest

### DIFF
--- a/source/parser.hera
+++ b/source/parser.hera
@@ -86,6 +86,7 @@ import type {
   AccessStart,
   Argument,
   ArrayBindingPattern,
+  ArrayBindingPatternContent as ArrayBindingPatternContentType,
   ArrowFunction,
   AssignmentExpression,
   ASTError,
@@ -2608,12 +2609,9 @@ BindingPropertyList
 ArrayBindingPattern
   _?:ws1 OpenBracket:open ArrayBindingPatternContent:c __:ws2 CloseBracket:close ::ArrayBindingPattern ->
     return gatherBindingPatternTypeSuffix({
-      ...c, // names, blockPrefix, length
+      ...c, // names, blockPrefix, omittedRest, length
       type: "ArrayBindingPattern",
-      // `c.elements` may differ from `c.children` when adjustBindingElements
-      // omits an anonymous rest from the emitted destructure but still needs
-      // pattern matching / type generation to see it.
-      elements: c.elements ?? c.children,
+      elements: c.children,
       children: [ws1, open, c.children, ws2, close],
     })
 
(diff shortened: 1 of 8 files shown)


STRd6 (approved):
Thanks!

PR #2164, opened 2026-06-25
LSP: Fix sourcemapping of trimmed spaces and `#` length shorthand

edemaine:
This PR fixes source-map handling and thus semantic token drift for:
1. Zero-length generated output
   Empty generated tokens still update the source cursor for later error locations, but they should not emit a source-map mapping entry because there is no generated span to map. Emitting those zero-width mappings caused cases like implicit call whitespace to override the following real token mapping, making semantic highlighting start one character early.
2. Standalone Civet shorthands `@` (for `this`) and `#` (for `length`), 
   Ensure these measure as one-character semantic token targets instead of inheriting the generated identifier `this` or `length`.

Bonus: LSP rename for implicit-call arguments no longer eats the space in `foo data` when renaming `data`.

## Before screenshot

image

1. The five characters `#, sam` gets highlighted blue in the first line. This is because `length` has five characters. (It's actually the generated key `length:` that causes this coloring, and `length` maps to the `#`. But it should highlight just the `#`.)
2. The last character of `data` and `buffer` are not properly highlighted. This is because of the trimmed space in the implicit function call shifting things by 1.

## After screenshot

image

### DIFF
--- a/lsp/server/package.json
+++ b/lsp/server/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@danielx/civet-language-server",
   "description": "Civet Language Server (standalone, editor-agnostic)",
-  "version": "0.3.40",
+  "version": "0.3.41",
   "repository": {
     "type": "git",
     "url": "https://github.com/DanielXMoore/civet.git"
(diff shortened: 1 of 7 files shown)


STRd6 (approved):
Thanks, glad it was so simple!

edemaine:
Bonus: LSP rename for implicit-call arguments was eating the space in `foo data` when renaming `data`. This PR fixes that too, so I added a regression test.

PR #2171, opened 2026-07-07
Use `[]` fallback for `...if`/`...unless` inside array literals

### LINKED ISSUE #2170: More ergonomic `...if` in array literals
## Input

Recently I wanted to write a conditional splat in an array literal like so:

```js
      children: [
        "\\includegraphics"
        ...unless isEmpty trimContent altText
          ["[alt={", altText, "}]"]
        "{", url, "}"
      ]
```

## [Current Output](https://civet.dev/playground?code=ICAgICAgY2hpbGRyZW46IFsNCiAgICAgICAgIlxcaW5jbHVkZWdyYXBoaWNzIg0KICAgICAgICAuLi51bmxlc3MgaXNFbXB0eSB0cmltQ29udGVudCBhbHRUZXh0DQogICAgICAgICAgWyJbYWx0PXsiLCBhbHRUZXh0LCAifV0iXQ0KICAgICAgICAieyIsIHVybCwgIn0iDQogICAgICBd)

```js
      ({children: [
        "\\includegraphics",
        ...(!isEmpty(trimContent(altText))?
          ["[alt={", altText, "}]"]:void 0),
        "{", url, "}"
      ]})
```

But `[...void 0]` causes `Uncaught TypeError: (void 0) is not iterable`.

## Proposed Output

```js
      ({children: [
        "\\includegraphics",
        ...(!isEmpty(trimContent(altText))?
          ["[alt={", altText, "}]"]:[]]),
        "{", url, "}"
      ]})
```

In the case of an `if`/`unless` without `else` that is immediately inside an array literal splat, I propose we replace `void 0` with the more helpful `[]` fallback value. This makes it easier to add optional parts to arrays.

STRd6:
Closes #2170.

For a spread without an `else` clause, `expressionizeIfStatement` emits `:void 0`. That's a TypeError when spread into an array (`[...void 0]` throws), and makes it awkward to add optional parts to an array literal.

Mark auto-generated `void 0` fallbacks on the resulting `IfExpression` with `elseFallback: true` and, in `processArraySpreadIfFallback`, rewrite them to `:[]` when the `IfExpression` is the direct expression of a `SpreadElement` under an `ArrayExpression`. Handles both prefix (`[...if x then a]`) and postfix (`[...a if x]`) forms. Object spreads and non-array spreads keep `:void 0` since they don't throw.

### DIFF
--- a/source/parser/lib.civet
+++ b/source/parser/lib.civet
@@ -401,7 +401,7 @@ function expressionizeIfStatement(statement: IfStatement): ASTNode
   // to the operand they came from so following generated syntax stays live.
   if b.trailing?#
     children.push b.trailing
-  elseFallback .= false
+  implicitElse .= undefined
   if e
     e2 := expressionizeBlock(e.block)
     unless e2
@@ -415,14 +415,13 @@ function expressionizeIfStatement(statement: IfStatement): ASTNode
     children.push ...e.children[3..]
   else
     children.push "\n" if b.trailing?#
-    children.push ":void 0"
-    elseFallback = true
+    children.push implicitElse = ":void 0"
   children.push closeParen
 
   makeNode {
     type: "IfExpression"
     children
-    elseFallback
+    implicitElse
   }
 
 function expressionizeTypeIf([ifOp, condition, t, e]: [any, any, any, any]): TypeConditional
@@ -1784,11 +1783,8 @@ function processArraySpreadIfFallback(statements: StatementTuple[]): void
     exp .= spread.expression as any
     if exp?.type is "StatementExpression"
       exp = exp.statement
-    continue unless exp?.type is "IfExpression" and exp.elseFallback
-    for i of [0...exp.children#]
-      if exp.children[i] is ":void 0"
-        exp.children[i] = ":[]"
-        break
+    continue unless exp?.type is "IfExpression" and exp.implicitElse
+    replaceNode exp.implicitElse, ":[]", exp
 
 /**
 Wrap any remaining statement expressions in IIFE.
--- a/source/parser/lib.civet
+++ b/source/parser/lib.civet
@@ -401,6 +401,7 @@ function expressionizeIfStatement(statement: IfStatement): ASTNode
   // to the operand they came from so following generated syntax stays live.
   if b.trailing?#
     children.push b.trailing
+  elseFallback .= false
   if e
     e2 := expressionizeBlock(e.block)
     unless e2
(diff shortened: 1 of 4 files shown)


source/parser/lib.civet
```diff
@@ -1767,6 +1770,26 @@ function processTypes(node: ASTNode)
...
           suffixIndex--
 
+/**
+Rewrite auto-generated `:void 0` fallbacks to `:[]` in `IfExpression`s that
+are the direct expression of a `SpreadElement` inside an `ArrayExpression`,
+so that `[...if cond then arr]` yields `[]` (not a TypeError) when `cond`
+is falsy.
+*/
+function processArraySpreadIfFallback(statements: StatementTuple[]): void
+  for each spread of gatherRecursiveAll statements, .type is "SpreadElement"
```
edemaine, line 1779: Instead of a postprocessing phase, could we do this check when generating the `else` clause in `expressionizeIfStatement`? @claude

edemaine: @claude Oh, if it's done at parse time, then could we do it when parsing the array splat? Just check if content is an IfStatement with `elseFallback`?

edemaine: @STRd6 Which do you think is better? Current approach might be cleaner code, but time cost of an extra traversal... Maybe keep as is?

STRd6: I think keeping the extra traversal is fine for simplicity.

Eventually we should be able to collapse a lot of the traversals.

source/parser/lib.civet
```diff
@@ -1767,6 +1770,26 @@ function processTypes(node: ASTNode)
...
+    // Prefix `...if`/`...unless` wraps the IfExpression in a StatementExpression;
+    // postfix `...a if x` places the IfExpression directly under the spread.
+    exp .= spread.expression as any
+    if exp?.type is "StatementExpression"
+      exp = exp.statement
+    continue unless exp?.type is "IfExpression" and exp.elseFallback
+    for i of [0...exp.children#]
+      if exp.children[i] is ":void 0"
+        exp.children[i] = ":[]"
+        break
```
edemaine:
```suggestion
    replaceNode ":void 0", ":[]", exp
```

or with the proposed change above:

```suggestion
    replaceNode exp.implicitElse, ":[]", exp
```

source/parser/lib.civet
```diff
@@ -415,11 +416,13 @@ function expressionizeIfStatement(statement: IfStatement): ASTNode
   else
     children.push "\n" if b.trailing?#
     children.push ":void 0"
+    elseFallback = true
```
edemaine: Maybe `implicitElse` would be more consistent naming. Also it would be cleaner later to set it to the actual child, like so:
```suggestion
    children.push implicitElse = ":void 0"
```
and change default to `undefined` above.

edemaine (approved):
Looks good! Just some simple proposed cleanup which I think @claude could handle.

edemaine:
Oh, it would be nice to document this somewhere too.

edemaine:
@claude Fix type error regression. Add docs to reference, maybe after "Array items can have a postfix if, with the else case implicitly being undefined", at least for postfix.

edemaine (approved):

PR #2173, opened 2026-07-17
Fix named destructuring `name^` hoisting

### LINKED ISSUE #2162: Named destructuring hoisted outside IIFE
Today I tried to include the following code in `parser.hera`:

```js
if lastParameter^{ delim: { implicit: true } } := parameters.-1
  lastParameter.delim = undefined
  lastParameter.children.pop()
```

Hera generates Civet code that keeps the destructuring inside the parser action:

```js
const $$m = (function($loc, tp, p, parameters) {
  return $skip unless parameters#
  parameters = trimFirstSpace(parameters) if tp
  if lastParameter^{ delim: { implicit: true } } := parameters.-1
    lastParameter.delim = undefined
    lastParameter.children.pop()
  return {
    type: "Parameters",
    children: [tp, {p.$loc, token: "("}, parameters, ")"],
    tp,
    parameters,
    bare: true,
  }
})($$r.loc, $$value[0], $$value[1], $$value[2]);
```

But Civet compiles that to JavaScript with an extra destructuring declarator outside the IIFE:

```js
const $$m = (function($loc, tp, p, parameters) {
  if (!parameters.length) { return $skip }
  if (tp) { parameters = trimFirstSpace(parameters) }
  let ref1;
  if ((ref1 = parameters[parameters.length-1]) && typeof ref1 === 'object' && 'delim' in ref1 && typeof ref1.delim === 'object' && ref1.delim != null && 'implicit' in ref1.delim && ref1.delim.implicit === true) {
    const lastParameter = ref1, { delim: { } } = lastParameter;
    lastParameter.delim = undefined
    lastParameter.children.pop()
  }
  return {
    type: "Parameters",
    children: [tp, {$loc: p.$loc, token: "("}, parameters, ")"],
    tp,
    parameters,
    bare: true,
  }
})($$r.loc, $$value[0], $$value[1], $$value[2]), { delim: { } } = lastParameter;
```

The trailing part is invalid because lastParameter is scoped inside the if block:

```js
, { delim: { } } = lastParameter
```

At runtime this throws:

```
ReferenceError: lastParameter is not defined
```

This looks like a Civet lifting bug for named destructuring declaration shorthand in expression/IIFE contexts, rather than a Hera bug, because Hera’s generated Civet keeps the shorthand in the correct scope.

edemaine:
Fixes #2162

[Before behavior](https://civet.dev/playground?code=ICAgIHJlc3VsdCA6PSAoKCkgPT4NCiAgICAgIGlmIGl0ZW1ee2F9IDo9IHZhbHVlDQogICAgICAgIGl0ZW0uYSA9IHVuZGVmaW5lZA0KICAgICAgcmV0dXJuIHVuZGVmaW5lZA0KICAgICkoKQ%3D%3D) on the test:

```js
    const result = (() => {
      if ((value) && typeof value === 'object' && 'a' in value) {const item = value, {a} = item;
        item.a = undefined
      }
      return undefined
    }
    )(), {a} = item
```

### DIFF
--- a/source/parser/declaration.civet
+++ b/source/parser/declaration.civet
@@ -143,7 +143,7 @@ function processDeclarations(statements: StatementTuple[]): void
 
     for i of [bindings#>..0]
       binding := bindings[i]
-      subbindings := gatherSubbindings binding
+      subbindings := gatherSubbindings binding.pattern
       if subbindings#
         simplifyBindingProperties binding
         simplifyBindingProperties subbindings
--- a/test/assignment.civet
+++ b/test/assignment.civet
@@ -456,6 +456,24 @@ describe "assignment", ->
   """
 
   testCase """
+    named destructuring declaration inside IIFE stays local (#2162)
+    ---
+    result := (() =>
+      if item^{a} := value
+        item.a = undefined
+      return undefined
+    )()
+    ---
+    const result = (() => {
+      if ((value) && typeof value === 'object' && 'a' in value) {const item = value, {a} = item;
+        item.a = undefined
+      }
+      return undefined
+    }
+    )()
+  """
+
+  testCase """
     named destructuring declaration
     ---
     do const {a, b^: [c, d]} = y


STRd6 (approved):
Thanks!

PR #2179, opened 2026-07-24
Fix crash on empty-braces property definitions or invalid globs

### LINKED ISSUE #2178: Compiler crashes on `{\n  {}`
https://civet.dev/playground?code=ewogIHt9Cg%3D%3D

Image

The following all make the compiler crash:

```
{
 {}
```

```
{
 foo{}
```

```
{
  {
}
```

```
{{{{
  {}
```

STRd6:
Fixes #2178

## Summary

Compiler previously crashed on inputs like `{\n  {}` with `last.children is not iterable`, and on `foo{bar: baz if n}` with a circular-JSON error.

- `PropertyDefinition`'s `ObjectExpression` case returned `[undefined]` for an empty glob; now returns `[]`.
- `NestedPropertyDefinition` gained guards matching its `SingleLineObjectProperties` sibling.
- `processCallMemberExpression`'s glob-error `JSON.stringify` skips `parent` back-pointers.

## Test plan

- [x] All 4504 tests pass (5 new regression tests added in `test/object.civet`)
- [x] Verified every crashing example from the issue now produces a clean parse error instead

### DIFF
--- a/source/parser.hera
+++ b/source/parser.hera
@@ -4145,10 +4145,16 @@ NestedPropertyDefinition
     return inlineProps.flatMap( ([prop, delim]: [any, any], i: number) => {
       // Allow each prop to be a single Property object or an array of such
       if not Array.isArray(prop) then prop = [prop]
+      // Empty property (e.g. from `{}` glob) contributes nothing;
+      // matches guard in SingleLineObjectProperties.
+      return [] unless prop.some((p: any) => p?)
       if i is 0
         [first, ...rest] := prop
         prop = [ prepend(ws, first), ...rest ]
       last := prop[prop#-1]
+      // Skip attaching delimiter if last is missing or lacks children
+      // (e.g. Error node from an invalid glob).
+      return prop unless last?.children
       prop = [
         ...prop.slice(0, prop#-1),
         {
@@ -4309,7 +4315,10 @@ PropertyDefinition
         // (via `processCallMemberExpression`)
         case "ObjectExpression":
           first .= value!.properties[0]
-          if first
+          return [] unless first
+          // An Error part (e.g. from `foo{bar: baz if n}`) has no children;
+          // pass it through so the error surfaces.
+          if first.children
             first = {
               ...first,
               children: [ws, ...first.children],
--- a/source/parser.hera
+++ b/source/parser.hera
@@ -4079,6 +4079,9 @@ SingleLineObjectProperties
       prop = Array.isArray(prop) ? prop : [prop]
       last .= prop[prop#-1]
       return [] unless last
+      // Error nodes from invalid globs have no children; pass them through
+      // so generation can surface the parse error.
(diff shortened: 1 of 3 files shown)


edemaine (approved):
Claude’s original commit, 2b0db771, fixes the primary crashes reported in #2178:

- empty property-glob expansions no longer produce `[undefined]`;
- malformed nested braces now reach the normal parser-error path;
- invalid-glob diagnostics omit circular AST `parent` pointers during serialization.

While reviewing that commit, GPT found the following remaining issues.

#### 1. Invalid globs still crashed inside single-line objects

The new code passed childless parser `Error` nodes through `PropertyDefinition`, and `NestedPropertyDefinition` handled them. However, `SingleLineObjectProperties` still unconditionally spread `last.children`.

This meant the top-level case worked:

```civet
foo{bar: baz if n}
```

but wrapping it in an object still crashed:

```civet
{foo{bar: baz if n}}
```

with:

```text
TypeError: last.children is not iterable
```

#### 2. The coverage gate failed

The nested-property guard used:

```civet
return [] unless prop.some((p: any) => p?)
…
return prop unless last?.children
```

The optional-chain nullish branch was not exercised, leaving `source/parser.hera` at 987/988 branches and failing the required Ubuntu coverage check.

The guard also checked whether *any* property existed, even though the following code specifically consumes the final property.

#### 3. New tests did not verify the actual diagnostics

Several regression tests asserted only `ParseError` or `ParseErrors`. That confirmed the compiler no longer threw a JavaScript exception, but did not verify:

- the reported source location;
- the intended “Failed to parse” message;
- the invalid-glob diagnostic;
- that AST serialization retained useful details after removing parent pointers.

## Fixes I pushed

The follow-up commit, 22d37cf6, addresses all of these issues.

### Handle childless errors in single-line objects

`SingleLineObjectProperties` now passes an error node without `children` through unchanged:

```civet
return prop unless last.children
```

Generation can then surface the intended `ParseErrors` diagnostic instead of the grammar action crashing.

A regression test now covers:

```civet
{foo{bar: baz if n}}
```

### Simplify the nested-property invariant

The nested guard now checks the property that will actually be consumed:

```civet
return [] unless prop.-1
```

Once that guard succeeds, `last` is known to exist, so the unnecessary optional chain is removed:

```civet
return prop unless last.children
```

This preserves the necessary childless-error handling while removing the unreachable coverage branch.

A second regression test covers the multiline equivalent:

```civet
{
  foo{bar: baz if n}
}
```

### Assert complete error messages

All seven tests introduced by this PR and follow-up now assert the complete error string, including line, column, and diagnostic content.

This verifies that malformed braces produce the expected `Failed to parse` errors and that invalid globs produce the intended:

```text
Glob pattern must have call or member expression value, found …
```

diagnostic rather than merely throwing some `ParseErrors` instance.

PR #2184, opened 2026-08-10
Fix pattern destructuring in some `for` loops

### LINKED ISSUE #2183: `for each` pinned destructuring
## Input
```js
for each pair^[, attr] of $0
```

## [Current Output](https://civet.dev/playground?code=Zm9yIGVhY2ggcGFpcl5bLCBhdHRyXSBvZiAkMA%3D%3D)
```js
for (let i = 0, len = $0.length; i < len; i++) {
  const pair = $0[i];
}
```

## Expected Output
```js
for (let i = 0, len = $0.length; i < len; i++) {
  const pair = $0[i];
  const [, attr] = pair;
}
```

## Workaround: Dropping `each` Works
```js
for each pair^[, attr] of $0
```
[transpiles to](https://civet.dev/playground?code=Zm9yIHBhaXJeWywgYXR0cl0gb2YgJDA%3D)
```js
for (const pair of $0) {
  const [, attr] = pair;
}
```

edemaine:
Fixes #2183 by handling `ForDeclaration` pattern follow-ups alongside their corresponding loop declarations.

Previously, named destructuring and related pattern transformations were added later by `processDeclarations`. This missed some loop forms, such as `for each`, and could place range-loop follow-ups before the primary binding was initialized. For example:

```js
for string^[first] of ["a".."c"]
---
for (let i = 97; i <= 99; ++i) {
  const [first] = string;
  const string = String.fromCharCode(i);
}
```

This PR:

- Moves loop-specific pattern expansion into the `for` processing code.
- Uses `pushForBlockDeclaration` for regular, `for each`, zip, range, secondary `in`/`of` bindings, and explicit-declaration Coffee-compatible loops.
- This ends up cleaning up a lot of code, and combining some declarations to boot.
- Keeps primary bindings and their named destructuring, post-rest, and `@` follow-ups in the correct order.
- Merges compatible adjacent declarations generated from the same loop binding.
- Fixes missing pattern follow-ups for `for...in` values and `for...of` indices.
- Removes the late `ForStatement` scan from `processDeclarations`.
- Adds regression coverage across Civet loop forms and explicit-declaration Coffee loops.

Coffee-compatible loops without an explicit `const`, `let`, or `var` use a separate assignment path and remain outside the scope of this PR. I think it's doable, but looking like a 200+ line change, so I'm not sure it's worth it...

### DIFF
--- a/source/parser.hera
+++ b/source/parser.hera
@@ -28,6 +28,7 @@ import {
   escapeIdentifier,
   expressionizeTypeIf,
   declarationIterand,
+  forDeclarationPatternPrefix,
   forRange,
   gatherBindingCode,
   gatherBindingPatternTypeSuffix,
@@ -67,6 +68,7 @@ import {
   processProgram,
   processProgramAsync,
   processRangeExpression,
+  pushForBlockDeclaration,
   processTryBlock,
   processUnaryExpression,
   processUnaryNestedExpression,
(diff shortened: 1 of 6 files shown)


STRd6 (approved):
Thanks!

PR #2185, opened 2026-08-10
Null-safe pipe `?|>`

### LINKED ISSUE #897: Non-null pipe `?|>`
@STRd6 suggests the ability to short-circuit the rest of a pipe when the left-hand side is null:

```js
x ?|> f
---
x != null ? f(x) : x
//or
x == null ? x : f(x)
```

edemaine:
## Summary

Adds null-safe pipes `?|>` and taps `?||>`, including guarded `return` support. Complicated example:

```js
data |>= process ?|> filter |> return
---
let ref;
if ((ref = process(data)) != null) {
  return (data = filter(ref));
} else {
  data = ref;
}
```

Fixes #897 (assuming we don't want truthy pipes). See https://github.com/DanielXMoore/Civet/issues/897#issuecomment-5054923797 for why I chose this notation.

## Syntax and behavior

- Adds null-safe pipe `?|>` with Unicode alias `?▷`.
- Adds null-safe tap pipe `?||>` with Unicode alias `?|▷`.
- Null-safe pipes guard the remaining syntactic pipeline when the current value is `null` or `undefined`:
  ```js
  value ?|> f |> g
  ---
  value != null ? g(f(value)) : value
  ```
  This won't call `f` or `g` unless `value?`.
- Parentheses delimit the guarded pipeline:
  ```js
  (value ?|> f) |> g
  ---
  g(value != null ? f(value) : value)
  ```
  This won't call `f` unless `value?`, but will always call `g`.
- Null-safe taps `?||>` run their tap only for non-nullish values, preserve the input value, and guard subsequent steps.
- Complex guarded inputs are evaluated once.
- Guarded pipelines compose with ordinary pipes, taps, `yield`, `throw`, and repeated guards.

## Returning pipelines

With some effort, this PR supports `|> return` inside null-safe pipelines:

```civet
value ?|> f |> return
```

This transpiles to an `if`, allowing the function to fall through when the value is nullish:

```js
if (value != null) {
  return f(value)
}
```

Returning pipelines now work with:

- Ordinary and tap steps before or after a guard.
- Repeated null-safe guards.
- Direct guarded returns.
- Complex inputs requiring temporary refs.
- Parenthesized pipelines.
- Pipe assignment (`|>=`).

For returning pipe assignments:

- Successful guarded paths assign the final value and return it.
- Nullish paths assign the guarded value and fall through.
- Complex assignment targets are evaluated once.
- Negative-index targets preserve and hoist their generated refs.

## Bug fixes

- Fix implicit-return handling for pipelines ending in `return` or `throw`.
  - Replaces positional AST inspection with an `exits` property meaning the active pipeline segment definitely exits.
  - Prevents invalid output such as `return return value` and `return throw value`.
  - Correctly distinguishes a definitely exiting guarded suffix from an outer conditional pipeline that may fall through.
- Fix `get() |> &.-1` evaluating `get()` twice.
  - Negative-index expansion now marks an ampersand function non-unrollable when it duplicates its parameter or placeholder.
  - Safe cases such as `value |> &.items.-1` remain unrolled.
- Fix stale and missing AST parent pointers introduced while rebuilding pipe and negative-index nodes.
- Preserve shared ref identities when deeply copying pipeline assignment targets.
- Ensure `?||>` is tokenized before its `?|>` prefix.
- Improve AST typing for ampersand functions, pipeline hoists, return expressions, and placeholder metadata.

## Implementation cleanup

- Rework optional-pipe processing into the existing pipeline loop instead of recursively processing suffixes.
- Track nested guarded suffixes as they are constructed.
- Use explicit AST node constructors and targeted parent updates.
- Centralize final pipeline-segment handling.
- Document the distinction between syntactic `pipelineReturns` lookahead and definite `exits` metadata.
- Resolve 9 type errors

## Documentation

- Document null-safe pipe semantics and the effect of parentheses.
- Document null-safe taps.
- Add all ASCII and Unicode pipe forms to the operator reference.
- Link Civet tap pipes to related constructs in Elixir and magrittr.

### DIFF
--- a/civet.dev/reference.md
+++ b/civet.dev/reference.md
@@ -942,14 +942,36 @@ fetch url |> await
 |> return
 </Playground>
 
+Null-safe pipes `?|>` skip the remaining steps of the current pipeline when
+the left-hand value is `null` or `undefined`:
+
+<Playground>
+user
+  ?|> .profile
+  ?|> normalizeProfile
+  |> renderProfile
+</Playground>
+
+Use parentheses to keep steps from being aborted by a null-safe pipe:
+
+<Playground>
+(user
+  ?|> .profile
+  ?|> normalizeProfile
+) |> renderProfile
+</Playground>
+
 Pipe assignment:
 
 <Playground>
 data |>= .content
 </Playground>
 
-Fat pipes `||>` pass the left-hand value to the next two steps in the pipeline
-(ignoring the output from the right-hand side):
+Fat pipes `||>` are tap or tee pipes:
+they run the right-hand side but ignore its result,
+preserving the same left-hand value for the next step.
+(Similar to [Elixir's `Kernel.tap/2`](https://hexdocs.pm/elixir/Kernel.html#tap/2)
+and [magrittr's `%T>%` tee pipe](https://magrittr.tidyverse.org/reference/tee.html).)
 
 <Playground>
 array
@@ -980,10 +1002,16 @@ document.createElement('div')
 ||> .appendChild document.createTextNode 'Civet'
 </Playground>
 
+<Playground>
+document.querySelector('#main')
+?||> .className = 'civet'
+||> .click()
+</Playground>
+
 Unicode forms:
 
 <Playground>
-data ▷= func1 |▷ func2 ▷ func3
+data ▷= func1 |▷ func2 ▷ func3 ?▷ func4
 </Playground>
 
 ### Await Operators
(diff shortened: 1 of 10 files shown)


edemaine:
I forgot to say: the main decision here is to make a failed `?|>` stop the rest of the pipeline. The analogy for me was `x?.y.z` which stops if `x` is null — but necessarily so. I'm less certain for pipes, but it does seem potentially useful, and parens can limit the impact... but parens are not so convenient. So this decision is definitely open for discussion.

After this, I'm also thinking of adding `??|>` by analogy to `??` which only does the pipeline when the LHS is null. This could be nice for e.g. early return. Something like this:

```js
data |> fetch
??||> return
|> process
|> return
---
let ref = fetch(data)
if(ref == null) return ref
return process(ref)
```

This example is maybe instructive... if I wanted to write it with `?|>` I'd have to use parens:

```js
(
  data |> fetch
  ?|> process
) |> return
```

Perhaps it would be better if we could write this?

```js
data |> fetch
?|> process
|> return
```

STRd6 (approved):
Nice addition!

edemaine:
I think I see now that shortcutting makes sense from a type perspective, at least generically. `x ?|> f |> g` just wouldn't pass the right type to `g` if we skipped `f`, unless `f` is an automorphism. Same argument as `x?.y?.z` really. So let's go with this, at least for now. Maybe we can invent syntax for leaving a shortcutting pipeline that doesn't involve parentheses...

PR #2187, opened 2026-08-17
Fix `export default` with nested implicit object literal (#2186)

### LINKED ISSUE #2186: `export default` with `key: value` in new lines
```coffee
export default
	mode: 'development'
	entry: './main.civet'
```

should compile to
```js
export default {
  mode: 'development',
  entry: './main.civet'
};
```

instead i get
```js
export default {
  mode: "development"({
    entry: "./main.civet",
  }),
};
```

STRd6:
Fixes #2186.

The `__` after `Default` greedily consumed the newline and indentation before the value, preventing `MaybeNestedPostfixedExpressionOrCommaLoop` from recognizing a `NestedImplicitObjectLiteral` spanning multiple lines. Split the rule so the nested-expression case uses `_?` (same-line whitespace only), leaving the newline+indent for the value to consume itself.

### DIFF
--- a/source/parser.hera
+++ b/source/parser.hera
@@ -6932,7 +6932,12 @@ ExportDeclaration
     ]
   # NOTE: This covers all forms that can be directly export defaulted in TS.
   # NOTE: Using Expression to allow If/Switch expressions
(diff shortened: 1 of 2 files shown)


edemaine (approved):
Looks good! I fixed the typing issue (expressions can't be `ts: true`) and added a test.

PR #2188, opened 2026-08-18
Merge multiple JSX `class` attributes like shorthand

edemaine:
Implements [Option 3 from the PR 2181 discussion](https://github.com/DanielXMoore/Civet/pull/2181#issuecomment-5241768585): repeated explicit JSX `class` attributes are merged using the same logic as dotted class shorthand.

- Repeated `class` attributes merge with other `class` attributes.
- Repeated `className` attributes merge with other `className` attributes.
- `class` and `className` remain separate.
- Dotted shorthand uses the first explicit class-like attribute name, falling back to the configured default otherwise.
- The first attribute remains the output anchor, preserving its position relative to spreads and other attributes.
- Static, dynamic, and conditional values retain their source order.

For example:

```jsx
<div class="a" class="b" />
```

becomes:

```jsx
<div class="a b" />
```

while:

```jsx
<div class="a" className="b" />
```

remains unchanged.

The refactor collects class sources by name and updates the first attribute node in place, avoiding index arithmetic and array splicing.

### DIFF
--- a/civet.dev/reference.md
+++ b/civet.dev/reference.md
@@ -3614,6 +3614,15 @@ Specify the `"civet react"` directive to use the `className` attribute instead:
 <div .foo>Civet
 </Playground>
 
+Multiple `class` (or `className`) attributes automatically merge into one,
+along with `.class` shorthand, automatically dropping falsey classes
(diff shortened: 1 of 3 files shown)


STRd6 (approved):
Nice!

PR #2197, opened 2026-08-21
LSP: Fix completions via better newline handling

edemaine:
Fix LSP member completion when a Civet expression ends immediately before a generated closing delimiter and the source file has a trailing newline.

For example (as reported by kutu on Discord):

```civet
new ResizeObserver (entries) ->
.observe document.b
```

With cursor after `b`, VS Code previously offered global completions such as `bigint`, `boolean`, and `break` instead of `document.body`.

The fix avoids emitting source-map entries for empty segments created when a generated newline-only token is split into lines. These segments contain no generated characters, so they should not introduce mapping positions.

## Root cause

Civet generates:

```typescript
new ResizeObserver(function(entries) {})
.observe(document.b)
```

The completion cursor is after `b` in the Civet source:

```text
.observe document.b|
                   ^ source column 19
```

The corresponding generated cursor should be before the compiler-inserted `)`:

```text
.observe(document.b|)
                   ^ generated column 19
```

This lets TypeScript see that the cursor is completing `document.b`, so it returns members of `Document`, including `body`.

The source-map generator processes output one token at a time. Near the end of this expression, it processes:

1. The generated `")"` token.
2. The generated `"\n"` token.

Both tokens are anchored at the end of the Civet line, source column 19.

Processing `")"` correctly creates this mapping:

| Generated position | Source position | Meaning |
| --- | --- | --- |
| Line 1, column 19 | Line 1, column 19 | Before the generated `)` |

The problem occurred while processing `"\n"`. `updateSourceMap` splits every output token at line breaks:

```civet
outLines := outputStr.split(EOL)
```

Splitting the newline-only token produces:

```javascript
["", ""]
```

These are zero-width segments on either side of the line break. The old code nevertheless emitted mappings for them. In particular, the first empty segment produced:

| Generated position | Source position | Meaning |
| --- | --- | --- |
| Line 1, column 20 | Line 1, column 19 | After the generated `)` |

There were therefore two generated positions for the same source position:

```text
generated 1:19 ─┐
                 ├── source 1:19
generated 1:20 ─┘
```

`forwardMap` resolves equally good source mappings in favor of the later generated mapping, so it chose generated column 20. TypeScript consequently received this cursor:

```text
.observe(document.b)|
                    ^ generated column 20
```

At that position, the call expression is already complete. The cursor is no longer inside `document.b`, so TypeScript returns scope-level completions instead of `Document` members.

This also explains why the problem depended on a trailing newline. Without the generated newline token, the redundant mapping at column 20 was never created, and the cursor remained at column 19.

## Fix

Skip empty segments after performing the necessary line-state updates:

```civet
// A newline-only token splits into empty segments on either side of the
// line break. The line state advances above, but neither segment has
// generated characters to map.
continue unless line#
```

The newline still advances the generated and source line state. It simply does not create mappings for positions that cover no generated characters.

After this change, the only applicable mapping is generated column 19, before the inserted `)`. The existing `forwardMap` behavior then produces the correct completion position without any LSP-specific cursor adjustment or mapping bias.

## Tests

Added coverage at three levels:

- Source-map generation does not emit mappings for empty newline segments.
- Forward mapping keeps the completion cursor before generated closing punctuation.
- Protocol-level and VS Code end-to-end tests verify that `document.body` is offered inside the chained `.observe` call.

Validation:

- Compiler: 4,620 passing, 21 pending
- LSP: 302 passing
- VS Code E2E: 58 passing

### DIFF
--- a/lsp/server/package.json
+++ b/lsp/server/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@danielx/civet-language-server",
   "description": "Civet Language Server (standalone, editor-agnostic)",
-  "version": "0.3.41",
+  "version": "0.3.42",
   "repository": {
     "type": "git",
     "url": "https://github.com/DanielXMoore/civet.git"
--- a/lsp/server/test/lsp-features.test.civet
+++ b/lsp/server/test/lsp-features.test.civet
@@ -1048,6 +1048,28 @@ describe "LSP features (shared project server)", ->
       resolved := await client.request "completionItem/resolve", target
       assert resolved, "expected resolved completion"
 
+  it "returns member completions in an unindented chained call argument", ->
+    uri := docUri path.join(projectRoot, "chained-call-completion.civet")
+    text := "new ResizeObserver (entries) ->\n.observe document.b\n"
+    client.notify "textDocument/didOpen", {
+      textDocument: { uri, languageId: "civet", version: 1, text }
+    }
+    await waitForDiag(client, (p) => p.uri is uri)
+
+    lines := text.split "\n"
+    useLine := lines.findIndex (line) => line.includes "document.b"
+    result := await client.request "textDocument/completion", {
+      textDocument: { uri }
+      position: {
+        line: useLine
+        character: lines[useLine].indexOf("document.b") + "document.b".length
+      }
+    }
+    list := result?.items ?? result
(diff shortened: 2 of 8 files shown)


STRd6 (approved):

kutu:
similar issue
```civet
class Waveform < HTMLElement
	@()
		super()
		@|
```
don't get completions for `classList` unless it's last line of a file

PR #2218, opened 2026-09-11
Empty slice `x[..]` compiles to `x.slice()`

edemaine:
For arrays and strings, `.slice(0)` is equivalent to `.slice()`. For `x[..]` and `x[...]`, we currently omit the former but I think we should omit the latter, as in this PR, because it seems closer to the source.

This could impact custom types, but that seems unlikely; we maintain the implicit `slice` contract.

On the other hand, CoffeeScript outputs `.slice(0)`, so we could decide this isn't worth breaking. What do you think?


STRd6 (approved):

STRd6:
We can add a coffeeScript compat flag if it ever comes up.

